PR #28, opened 2016-09-01
Use util.logging's logger and update logging documentation #26

### LINKED ISSUE #26: Use util.logging's logger and update logging documentation

m133225:
Fixes #26

- [X] Use default logger
- [x] Update logger documentation


build.gradle
```diff
@@ -7,7 +7,7 @@
  
 plugins {
     id "com.github.kt3k.coveralls" version "2.4.0"
-    id 'com.github.johnrengelman.shadow' version '1.2.3'
+    id "com.github.johnrengelman.shadow" version '1.2.3'
```
damithc: This reminds me, we need to document how to use gradle to download dependencies. I don't think eclipse does it automatically. You may want to check the gradle plugin for eclipse to see if it can do it.

m133225: Yes, have touched on it slightly in Managing Dependencies.md. However, we are planning to put the dependencies in the repository, right? Then the downloading of dependencies using gradle should only matter in Travis CI.

damithc: Yes, we can do that too, although putting libs in source control is bad form. If they can be downloaded with gradle (even outside eclipse), i'd rather do that.

PR #35, opened 2016-09-09
Increase height of StatusBarHeader #25

### LINKED ISSUE #25: Increase height of StatusBarHeader
Make it 3 lines instead of current 1 line? Some error messages have more than one line.

m133225:
Fixes #25


m133225:
@damithc Is this okay? Seems abit weird though.

damithc:
Yes it' ok. I guess it will not look so weird after we extract the command box as a separate component and put it above the status message area.

PR #51, opened 2016-09-14
Tweak GUI to emphasize active components #48

### LINKED ISSUE #48: Tweak GUI to emphasize active components
Mainly set the status area to white and modify margins.

m133225:
Fixes #48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ build/
 lib/*
 *.iml
 *.log
+*.log.*
 *.csv
 config.json
 *.xml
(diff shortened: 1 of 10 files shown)


damithc:
Now we can put the command box above the status message? BTW, shall we call it `Result Display`? That way, there's no confusion between it and the `Status Bar` at the bottom.

m133225:
screen shot 2016-09-14 at 7 47 47 pm

@damithc Done.

damithc:
Looks good. Go ahead.

m133225:
@damithc Does it look ok on your machine? Result display seems to be abit too short for my Windows, probably need some adjustments.

damithc:
Yes, a little bit cut off in my office comp.
image
Probably depends on resolution. May be make the height adjustable? Having a scroll bar is useful too, because we never know how long the result can be e.g. if it was an error message with a stack trace.

m133225:
@damithc Did a quick change on master to a non-editable text area to allow for scrolling.

PR #58, opened 2016-09-18
GUI tests: use setText instead of write #53

### LINKED ISSUE #53: GUI tests: use setText instead of write
I think we can speed up the tests if we set the entire command in one shot instead of typing one letter at a time.

damithc:
Fixes #53 

Speeds up tests about 2m 50s to 2m 10s.

### DIFF
--- a/src/test/java/guitests/guihandles/GuiHandle.java
+++ b/src/test/java/guitests/guihandles/GuiHandle.java
@@ -93,8 +93,8 @@ public class GuiHandle {
      */
     protected void typeTextField(String textFieldId, String newText) {
         guiRobot.clickOn(textFieldId);
-        guiRobot.push(KeyCode.SHORTCUT, KeyCode.A).eraseText(1)
-                .write(newText);
+        ((TextField)guiRobot.lookup(textFieldId).tryQuery().get()).setText(newText);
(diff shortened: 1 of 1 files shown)


damithc:
@yl-coder for your review and merging

yl-coder:
.setText doesn't really go thru the GUI layer and setText is just modify the underlying value of the GUI node. If we assume that JavaFX UIControl is working, i think it is fine.

damithc:
.write() is painfully slow. Let's use setText by default. We can use write in some cases, especially if some event handling happens after every keystroke (e.g. live search).

yl-coder:
I'm just thinking of extreme case. For example, If there is parent node consuming the key event. This method will pass the test cases, but the behaviour in reality may be different. 

We can speed up the write method by implementing our own.

damithc:
That kind of 'show stopper' bugs are usually noticed even without automated tests. We can make most of the tests use .setText() but a few tests use .write().
Another thing we can do is to use .setText() for the first n-2 chars and use write for the last 2 chars.

damithc:
Or use write for the first few letters and set text for the remainder. For now, let's merge this. We can add a refinement later.

yl-coder:
Right.

PR #68, opened 2016-09-19
Flatten doc structure

damithc:
I've flatten the doc structure a bit and added some missing parts.


damithc:
@m133225 for your review and merging. Go ahead and fix any problems you notice. 
A lot more to be done, but I'll do those as separate PRs.

damithc:
@m133225 I merged because this is needed for another PR. You can review later and fix any problems in master or as a separate PR.

PR #85, opened 2016-09-22
Refine logic and parser

damithc:
As per title


damithc:
Main changes:
* Parser no longer depends on model to make Parser more cohesive. As a result, IndexOutOfBound error is detected during executing commands, not during construction.
* Minimized using exceptions to handle user errors. User errors are not 'exceptional' situations.
* Moved filter construction code into the model. 
* Consolidated logic, parser, and commands into one component. 

@YijinL note above changes for your info.

PR #90, opened 2016-09-25
Complete dev guide

damithc:
Work in progress


damithc:
Merging this because it has some code tweaks. Dev guide is not complete, but I'll do it as a separate PR.

PR #113, opened 2016-09-27
Implement 'Last updates timestamp' feature #98

### LINKED ISSUE #98: Implement 'Last updates timestamp' feature
Footer at bottom left "Last updated: [not implemented yet]"

mauris:
Implementation takes advantage of the EventsCenter to check for updates to AddressBook

however when it starts up, the app is not aware of the address book's last updated (before the app starts up) date/time. 

Fixes #98

### DIFF
--- a/src/main/java/seedu/address/ui/StatusBarFooter.java
+++ b/src/main/java/seedu/address/ui/StatusBarFooter.java
@@ -89,7 +89,7 @@ public class StatusBarFooter extends UiPart {
     }
 
     @Subscribe
-    public void handleAddressBookChangedEvent(AddressBookChangedEvent mce) {
+    public void handleAddressBookChangedEvent(AddressBookChangedEvent abce) {
         setSyncStatus("Last Updated: " + (new Date()).toString());
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/seedu/address/ui/StatusBarFooter.java
```diff
@@ -81,4 +88,8 @@ public String getFxmlPath() {
         return FXML;
     }
 
+    @Subscribe
+    public void handleAddressBookChangedEvent(AddressBookChangedEvent mce) {
```
damithc: `mce` --> `abce`? (change it in the other place as well).

src/main/java/seedu/address/ui/StatusBarFooter.java
```diff
@@ -36,7 +43,7 @@ public static StatusBarFooter load(Stage stage, AnchorPane placeHolder, String s
     public void configure(String saveLocation) {
         addMainPane();
         addSyncStatus();
-        setSyncStatus("Last updated: [not implemented yet]");
+        setSyncStatus("Ready");
```
damithc: `Unknown` or `Not updated yet in this session`?

src/main/java/seedu/address/ui/StatusBarFooter.java
mauris: renamed instances of `mce` to `abce`.

src/main/java/seedu/address/ui/StatusBarFooter.java
mauris: chose `Not updated yet in this session`. We can also choose to read in the last updated from file system.

Issue #126, opened 2016-09-28
Log file recreated every time


damithc:
Problem 1. Log file is not used when running from IDE (log file has only 2 lines)
Problem 2. When running from jar, log file is recreated every time I run the app.

damithc:
@m133225 can fix this one?

m133225:
Does this mean that the old log messages are being overwritten?

damithc:
Yes.

m133225:
Can't seem to reproduce 1). Maybe deleting the possibly old config file will help.

damithc:
When running from IDE, this is all I see in the log file.
```
Sep 28, 2016 4:15:25 PM seedu.address.storage.XmlAddressBookStorage readAddressBook
INFO: AddressBook file .\src\test\data\XmlAddressBookStorageTest\NonExistentFile.xml not found
```

damithc:
But problem 2 is still there. Every time I start the app (whether it is from IDE or jar file), the file is wiped clean.

damithc:
Confirmed from home comp. If you cannot reproduce this, it could be a Windows-only thing.

damithc:
@lejolly can you help to check this on your Mac?

lejolly:
1. doesn't happen to me
2. does

PR #127, opened 2016-09-28
Set rolling file for logging #122 #126

### LINKED ISSUE #122: Set rolling file for logging
Currently writes to a log file without any limit which are the defaults for `FileHandler`.

Set to about 5 files of 5MB, and document this.

m133225:
Fixes #122 and #126 
Combining these 2 since they are somewhat related.


m133225:
Note that with the rolling file, the file names will be `addressbook.log<index>`  where `<index>` is the file count (1-5). This is the behaviour for `logging.util.Logger` which does not seem to be customisable

damithc:
Go ahead and merge. What was the problem before?

m133225:
It simply just wasn't set to append mode.

PR #130, opened 2016-09-29
Tweak Gradle preferences file

damithc:
Tweaked this file because Eclipse added it during the set up process, and gives an error if I remove it.

### DIFF
--- a/.settings/org.eclipse.buildship.core.prefs
+++ b/.settings/org.eclipse.buildship.core.prefs
@@ -1,6 +1,7 @@
 build.commands=org.eclipse.jdt.core.javabuilder
 connection.arguments=
 connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
+connection.gradle.user.home=null
 connection.java.home=null
 connection.jvm.arguments=
 connection.project.dir=


damithc:
@m133225 @okkhoy can you try this branch and see if your eclipse project still works?

okkhoy:
I can try it tomorrow morning, since I don;t have my laptop now :-/

m133225:
Reimported the project and it seems to work for me.

okkhoy:
works now as well. all tests pass too! same observation, eclipse deletes the added line.

damithc:
Ha ha, in some comps, it adds the line if its missing. In others, it deletes the line if added. :-p

damithc:
On my computer it doesn't delete the added line.
@m133225 I wonder if we can omit the eclipse project files and let eclipse generate them during the import process.

m133225:
@damithc I think that there are some required eclipse project files in order for it to be imported successfully. However, it seems to work well even if `.settings/org.eclipse.buildship.core.prefs` is removed before importing. The buildship plugin probably re-evaluates the `build.gradle` script.

PR #150, opened 2016-10-08
User can add email, address, tags in any order #138

### LINKED ISSUE #138: User can add email, address, tags in any order
... so that the user doesn't have to remember the attribute order in the command

ndt93:
Fixes #138


src/main/java/seedu/address/logic/parser/ArgumentsParser.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.util.*;
+
+/**
+ * Parses arguments string of the form: non-prefix-args <prefix>value <prefix>value ...
+ * An argument's value is assumed to not contain its prefix and any leading or trailing
+ * whitespaces will be discarded. Argument may be allowed to be repeated or not. If a
+ * non-repeatable argument is repeated, its last value will take precedence.
+ */
+public class ArgumentsParser {
```
damithc: Is `CommandLexer` or `CommandTokenizer` a better name?

src/main/java/seedu/address/logic/parser/ArgumentsParser.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public static class ParsedArguments {
+        private String nonPrefixArgument = "";
+        private Map<String, String> onceArguments = new HashMap<>();
+        private Map<String, List<String>> repeatableArguments = new HashMap<>();
+
+        /**
+         * Adds the result for an argument
+         * @param argument
```
damithc: We discourage empty @param/@return comments. Either omit or give a description.

src/main/java/seedu/address/logic/parser/ArgumentsParser.java
```diff
@@ -0,0 +1,184 @@
...
+        extractArgumentValues(argumentsString);
+        return this.parsedArguments;
+    }
+
+    private void extractArgumentParsingData(String argumentsString, Argument argument) {
+        int argumentStart = argumentsString.indexOf(argument.prefix);
+        while (argumentStart != -1) {
+            ArgumentParsingData argumentParsingData =
+                    new ArgumentParsingData(argument, argumentStart);
+            parsingData.add(argumentParsingData);
```
damithc: The line does a state change. Stateless methods are preferred. May be use the return valueof the method  (or add one more parameter) to avoid the state change?

src/main/java/seedu/address/logic/parser/ArgumentsParser.java
```diff
@@ -0,0 +1,184 @@
...
+            ArgumentParsingData nextArg = this.parsingData.get(i + 1);
+            value = argumentsString.substring(currentArg.startPos + currentArg.prefix.length(),
+                                                     nextArg.startPos);
+            this.parsedArguments.addArgument(currentArg, value.trim());
+        }
+
+        ArgumentParsingData lastArg = this.parsingData.get(this.parsingData.size() - 1);
+        value = argumentsString.substring(lastArg.startPos + lastArg.prefix.length());
+        this.parsedArguments.addArgument(lastArg, value.trim());
+    }
```
damithc: See if you can SLAP this method a bit more.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -87,36 +88,28 @@ public Command parseCommand(String userInput) {
...
-                    matcher.group("name"),
-                    matcher.group("phone"),
-                    matcher.group("email"),
-                    matcher.group("address"),
-                    getTagsFromArgs(matcher.group("tagArguments"))
+                    parsedArguments.getNonPrefixArgument().get(),
+                    parsedArguments.getOnceArgumentValue(phoneNumberArg).get(),
+                    parsedArguments.getOnceArgumentValue(emailArg).get(),
+                    parsedArguments.getOnceArgumentValue(addressArg).get(),
+                    getTagsFromParsedArgs(parsedArguments)
```
damithc: This line is slightly higher level than the previous 4 lines (each of which is calling a method from parsedArguments). 
`toSet(parsedArguments.getRepeatableArgumentValue(tagArgs))`?

src/main/java/seedu/address/logic/parser/ArgumentsParser.java
```diff
@@ -0,0 +1,184 @@
...
+            this.name = name;
+            this.prefix = prefix;
+            this.isRepeatable = false;
+        }
+
+        public Argument(String name, String prefix, boolean isRepeatable) {
+            this(name, prefix);
+            this.isRepeatable = isRepeatable;
+        }
+    }
```
damithc: Even better if you can do this:
Argument has two children: NonPrefixedArgument and PrefixedArgument (which has two children: Repeatable and NonRepeatable)

src/test/java/seedu/address/logic/parser/ArgumentsParserTest.java
```diff
@@ -0,0 +1,61 @@
...
+import seedu.address.logic.parser.ArgumentsParser.ParsedArguments;
+
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public class ArgumentsParserTest {
```
damithc: It's better if you can test this class in general, not referring to the add command. The SUT is not specific to the add command either.

Also, it's better if you can test this without involving the Parser class. But you'll have to add more tests to the ParserTest to verify that the add command works for different parameter orders.

ndt93:
@damithc I've made some changes according to the suggestions

src/main/java/seedu/address/logic/parser/CommandTokenizer.java
```diff
@@ -0,0 +1,313 @@
...
+
+        public Argument(String name) {
+            this.name = name;
+        }
+
+        public String getName() {
+            return this.name;
+        }
+
+        public abstract boolean isPrefixed();
```
damithc: See if you can get rid of this. It defeats the purpose of having a sub class for prefixed arguments. An alternative is to assume all arguments have a prefix but have a sub class that doesn't have prefix. In that case, be careful not to violate the Liskov Substitution Principle.

src/main/java/seedu/address/logic/parser/CommandTokenizer.java
```diff
@@ -0,0 +1,313 @@
...
+        @Override
+        public boolean isPrefixed() {
+            return true;
+        }
+
+        public abstract boolean isRepeatable();
+    }
+
+    /**
+     * Describes a repeatable prefixed argument. All values of this argument are accumulated
```
damithc: The 2nd sentence is not relevant to this class. In fact I'm beginning to wonder if we need this class.  The fact that an argument is repeatable should not be a concern to the argument itself.

src/main/java/seedu/address/logic/parser/CommandTokenizer.java
```diff
@@ -0,0 +1,313 @@
...
+        @Override
+        public boolean isRepeatable() {
+            return false;
+        }
+    }
+
+    /**
+     * Stores values for all arguments extracted from the arguments string
+     */
+    public static class ParsedArguments {
```
damithc: 1. The `Parsed` part is not relevant. This class should not care what happened to the argument before it was given to this class.
2. Shouldn't this class contain a `Map<String, List<Argument>>`? It should not care if an argument is repeatable.

ndt93: I think there are some misunderstandings of the logic of this class:

Currently, the `*Arguments` are passed into the the class to describe the string to be interpreted. The outputs are the string value(s) for each of those `*Argument`. So this acts more like the original RegEx logic than a Lexer/Tokenizer.

I suppose the logic you are thinking of is that this class *outputs* a set of `*Arguments` as tokens provided that the string passed in follows certain conventions such as each prefix is ended with a `/` for example.

ndt93: We can code it to behave in the second way too. So we can discuss this later.

damithc: Yes, seems I got two things mixed up a bit. See if this works:
```java
class ArgumentTokenizer{
  ArgumentTokenizer(Prefix... prefixes);
  tokenize(String argsString);
  Optional<String> getValue(Prefix prefix); //returns last value with given prefix
  Optional<List<String>> getValues(Prefix prefix); //returns all values with the given prefix
  Optional<String> getPreamble(); //returns text before the first valid prefix, if any
}

class Prefix{
  Prefix(String);
}
```

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,204 @@
...
+     * Creates an ArgumentTokenizer that can tokenize arguments string as described by prefixes
+     */
+    public ArgumentTokenizer(Prefix... prefixes) {
+        this.prefixes = Arrays.asList(prefixes);
+    }
+
+    /**
+     * @param argsString arguments string of the form: preamble <prefix>data <prefix>data ...
+     */
+    public void tokenize(String argsString) {
```
damithc: I haven't looked at your algo. one simple algo is to repeatedly split by prefix. e.g. given prefixes `/x` `/y` `/z`  and a string
`aaa /y yyyy /x xxx xxx xx /z zzz`,
split by `/x`,
```
p: aaa /y yyyy
/x: xxx xxx xx /z zzz
```
split by `/y`,

```
p: aaa 
/y: yyyy
/x: xxx xxx xx /z zzz
```
split by `/z`,

```
p: aaa 
/y: yyyy
/x: xxx xxx xx 
/z: zzz
```
Repeated arguments will complicate the matters though.

Also keep in mind that in future we might allow multiple prefixes for the same argument and want it to be case insensitive.

ndt93: The current algorithm is like this:

1. For each prefix, find all its positions
2. Sort the list of all all positions (for all prefixes)
3. The values of the arguments are then extracted between consecutive positions in the sorted list.

So in the end we also know where each prefix is in the original string. And we don't have to create many substring instances (in each splitting iteration). What do you think?

Multiple prefixes, case insensitive etc. can be easily extended by adding more fields to the `Prefix` class and modify step (1) to take all those into account.

ndt93: And since we more controls in step (1), we can even do things like ignoring prefixes between double quotes (so that user can input argument values that contain a prefix between double quotes).

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,204 @@
...
+
+        public PrefixExtended(Prefix prefix, int startPos) {
+            super(prefix);
+            this.startPos = startPos;
+        }
+
+        public int getStartPos() {
+            return this.startPos;
+        }
+    }
```
damithc: Use composition instead? e.g. `PrefixPosition` <>----`Prefix`

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
damithc: Sounds good.  Perhaps tweak the code like this to make it clear.
```java
List<PrefixPosition> positions = getAllPrefixPositions(...);
sort(positions);
extractArguments(positions);
```

damithc:
Let me know when ready for review.

ndt93:
@damithc Yup, it's ready for review. I'm just waiting for the CI to pass.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+                return true;
+            }
+
+            Prefix otherPrefix = (Prefix) obj;
+            return otherPrefix.getPrefix().equals(getPrefix());
+        }
+    }
+
+    /**
+     * Extra data needed for each prefix encountered during tokenization process
```
damithc: Update comment to explain what the class represents.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+        public int getStartPos() {
+            return this.startPos;
+        }
+
+        public Prefix getPrefix() {
+            return this.prefix;
+        }
+    }
+
+    private final List<Prefix> prefixes;
```
damithc: Can do without this? This is same as the keys of the Map, right?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Finds all positions in an arguments string at which a given `prefix` appears
+     */
+    private List<PrefixPosition> findPrefixPositions(String argsString, Prefix prefix) {
+        List<PrefixPosition> extendedPrefixes = new ArrayList<>();
+
+        int argumentStart = argsString.indexOf(prefix.getPrefix());
+        while (argumentStart != -1) {
+            PrefixPosition extendedPrefix = new PrefixPosition(prefix, argumentStart);
```
damithc: Variable name mismatch

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+            extractMiddleArguments(argsString, prefixPositions, i);
+        }
+
+        extractLastArgument(argsString, prefixPositions);
+    }
+
+    /**
+     * Extracts and stores the preamble part of the `argsString`
+     */
+    private void extractPreamble(String argsString, List<PrefixPosition> prefixPositions) {
```
damithc: Should this have a return value?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+        this.preamble = value;
+    }
+
+    /**
+     * Extracts and stores the value of an argument with specified prefix in the arguments string
+     * @param argsString: The arguments string
+     * @param prefixPositions: the list of all prefixes positions in the string
+     * @param prefixIndex: the index of the argument's prefix in `prefixPositions` list
+     */
+    private void extractMiddleArguments(String argsString, List<PrefixPosition> prefixPositions, int prefixIndex) {
```
damithc: `extractMiddleArgument` (i.e. singular)?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+        PrefixPosition targetPrefixPosition = prefixPositions.get(prefixIndex);
+        PrefixPosition nextPrefixPosition = prefixPositions.get(prefixIndex + 1);
+        Prefix prefix = targetPrefixPosition.getPrefix();
+
+        int valueStartPos = targetPrefixPosition.getStartPos() + prefix.getPrefix().length();
+        String value = argsString.substring(valueStartPos, nextPrefixPosition.getStartPos());
+
+        addPrefixValue(prefix, value.trim());
+    }
```
damithc: I feel these `extract___Argument` should return a value rather than change state.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
damithc: Perhaps give current prefix position and next prefix position as parameters, instead of passing the entire list?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private void extractLastArgument(String argumentsString, List<PrefixPosition> extendedPrefixes) {
```
damithc: parameter name mismatch

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * @param argsString arguments string of the form: preamble <prefix>data <prefix>data ...
+     */
+    public void tokenize(String argsString) {
+        List<PrefixPosition> positions = findAllPrefixPositions(argsString);
+        positions.sort((prefix1, prefix2) -> prefix1.getStartPos() - prefix2.getStartPos());
+        extractArguments(argsString, positions);
+    }
```
damithc: Shouldn't there be an explicit reset at the begining of this method? The tokenizer objects can be instantiated once and used multiple times.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
ndt93: This is the list of prefixes passed into the constructor, which will be reused with the tokenizer. Also, the Map may not contain all given prefixes.

ndt93: Well, unless the `tokenize` method also takes in a list of prefixes...

damithc: What I had in mind was to init the map with all given prefixes and empty value list for each.  That way we can use map.keys() any time we need the list of prefixes. But this is ok too.

ndt93:
@damithc I've some changes according to the review. You can take a look again.

ndt93:
@damithc I've updated the tests and the user docs. You can check them again.

damithc:
Looks good. I'll update the diagrams and merge later.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -0,0 +1,208 @@
...
+        // Sort by start position
+        prefixPositions.sort((prefix1, prefix2) -> prefix1.getStartPosition() - prefix2.getStartPosition());
+
+        // Insert a PrefixPosition to represent the preamble
+        PrefixPosition preambleMarker = new PrefixPosition(new Prefix(""), 0);
+        prefixPositions.add(0, preambleMarker);
+
+        // Add a dummy PrefixPosition to represent the end of the string
+        PrefixPosition endPositionMarker = new PrefixPosition(new Prefix(""), argsString.length());
+        prefixPositions.add(endPositionMarker);
```
damithc: I used two dummy PrefixPositions objects to fool the extraction algorithm so that we can use the same algo for extracting preamble, middle arguments, and last argument. I don't usually like to have that kind of 'clever tricks' in the code, but may be the resulting reduction is code justifies it this time. What do you think?

src/test/java/seedu/address/logic/parser/ArgumentTokenizerTest.java
```diff
@@ -0,0 +1,136 @@
+package seedu.address.logic.parser;
+
+import org.junit.Test;
+import seedu.address.logic.parser.ArgumentTokenizer.Prefix;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+
+public class ArgumentTokenizerTest {
```
damithc: Had to rewrite this class to decouple it from Parser.

damithc:
@ndt93 reminder to review.

ndt93:
@damithc Looks good to me.

PR #151, opened 2016-10-09
Set up check style #50

### LINKED ISSUE #50: Set up check style
- Properly setup checkstyle
- Plugins for <s>dependencies and</s> checkstyle on eclipse

m133225:
Fixes #50


src/main/java/seedu/address/commons/core/Messages.java
```diff
@@ -7,7 +7,8 @@
 
     public static final String MESSAGE_UNKNOWN_COMMAND = "Unknown command";
     public static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format! \n%1$s";
-    public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is invalid";
+    public static final String MESSAGE_INVALID_PERSON_DISPLAYED_INDEX = "The person index provided is " +
+                                                                        "invalid";
```
damithc: Which rule is triggering this? If the max length is 80, increase it to 110

src/test/java/seedu/address/testutil/TestUtil.java
```diff
@@ -38,18 +38,16 @@
...
             executable.run();
-        }
-        catch (Throwable actualException) {
-            if (!actualException.getClass().isAssignableFrom(expected)) {
-                String message = String.format("Expected thrown: %s, actual: %s", expected.getName(),
-                        actualException.getClass().getName());
-                throw new AssertionFailedError(message);
-            } else return;
+        } catch (Throwable actualException) {
+            if (actualException.getClass().isAssignableFrom(expected)) return;
```
damithc: One line form is ok, but braces should be compulsory.

src/test/java/seedu/address/testutil/TestUtil.java
```diff
@@ -58,37 +56,47 @@ public static void assertThrows(Class<? extends Throwable> expected, Runnable ex
...
+                new Person(new Name("Elle Meyer"), new Phone("9482224"), new Email("werner@gmail.com"),
+                           new Address("michegan ave"), new UniqueTagList()),
+                new Person(new Name("Fiona Kunz"), new Phone("9482427"), new Email("lydia@gmail.com"),
+                           new Address("little tokyo"), new UniqueTagList()),
+                new Person(new Name("George Best"), new Phone("9482442"), new Email("anna@google.com"),
+                           new Address("4th street"), new UniqueTagList()),
+                new Person(new Name("Hoon Meier"), new Phone("8482424"), new Email("stefan@mail.com"),
+                           new Address("little india"), new UniqueTagList()),
+                new Person(new Name("Ida Mueller"), new Phone("8482131"), new Email("hans@google.com"),
+                           new Address("chicago ave"), new UniqueTagList())
```
damithc: In rare cases like this, it should be ok to make an exception and use longer lines. I hope checkstyle has a way to define exceptions to rules.

m133225:
Are we also adding the checkstyle tasks to Travis? i.e. enforcing that the PR abides the coding standard.

damithc:
Yes, we can do that, as long as the dev has a reasonable way to auto-detect violations before pushing. We don't want a case where every PR has to go through an extra round of tweaks after it has been pushed to github. Also, it should be easy to detect what's the offending line and what's the offense when travis fails.

src/test/java/seedu/address/testutil/TestUtil.java
m133225: I don't think this is configurable. If we enforce braces, we also have to enforce linebreaks, since there does not seem to be a way to detect if it is a 1-line statement, aside from a binary setting which also disables the braces check.

Will disallow single-line statements for now.

damithc: Come to think of it, it's ok if there are no braces in the single line case. But we should enforce braces if it's two lines. Is that possible? If we go for it, we should update our style guide accordingly.

m133225: That was the way it was initially in this PR. Will revert it back to that.

damithc:
Update documentation about the use of checkstyle? how to turn off rules, how to install plugin, how to check before pushing etc.

PR #153, opened 2016-10-12
Add tests for StringUtil::containsWordIgnoreCase

damithc:
Refined the test cases to show the usage of test case design heuristics


damithc:
@YijinL can you review this one? Not sure if the original method had a different intention. I refined based on what I thought it is supposed to be doing and ended up with a simpler algorithm.

YijinL:
The original code was not written by me, but this implementation is correct. Merging.

PR #159, opened 2016-10-16
Remove non-static initializers

damithc:
PMD discourages use of non-static initializer blocks as they are 'rarely used and confusing to the reader'


damithc:
@YijinL what do you think?

damithc:
On the other hand, we can keep it just to illustrate this 'rarely used code construct' but add a comment to say its use is discouraged.

YijinL:
That sounds good. Everything the initialiser block does can be done in the no-argument constructor anyway.

PR #162, opened 2016-10-19
Increase encapsulation of UniquePersonList #87

### LINKED ISSUE #87: Increase encapsulation of UniquePersonList
It's a bit odd to have a `getInternalList()` method. On the surface it looks like a violation of encapsulation. To be investigated further.

YijinL:
Fixes #87.

Also done for `UniqueTagList`.

Closed up quite a few other holes (eg. unecessary exposure of internals) for safety.

### DIFF
--- a/src/main/java/seedu/address/model/person/UniquePersonList.java
+++ b/src/main/java/seedu/address/model/person/UniquePersonList.java
@@ -2,6 +2,7 @@ package seedu.address.model.person;
 
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
+import seedu.address.commons.core.UnmodifiableObservableList;
 import seedu.address.commons.util.CollectionUtil;
 import seedu.address.commons.exceptions.DuplicateDataException;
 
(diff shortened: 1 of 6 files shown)


YijinL:
I have no idea what's up with the codacy thing though

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -36,6 +37,11 @@ protected DuplicatePersonException() {
 
 
     /**
+     * Constructs empty PersonList.
+     */
+    public UniquePersonList() {}
```
damithc: Can be omitted. Codacy is likely to complain about this because empty constructors are auto-generated by the compiler.

damithc:
Regarding codacy errors, click on the 'Details' link to find out the problems.
1. Remove 'no parameter' constructor in UniquePersonList. If we want to keep the constructor, we can ask codacy to ignore that case.
2. Move inner classes to the bottom of the class. Apparently, that's the convention.

PR #178, opened 2016-11-29
build.gradle: add `run` and `runShadow` tasks #177

### LINKED ISSUE #177: build.gradle: add `run` and `runShadow` tasks
The gradle [application plugin](https://docs.gradle.org/current/userguide/application_plugin.html) provides a `run` task to allow developers to conveniently launch the application from the command line. (Very convenient for me in practice)

The shadow plugin (already used in this project) will automatically [integrate with the application plugin](http://imperceptiblethoughts.com/shadow/#integrating_with_application_plugin) to provide the `runShadow` task which does the same thing but with the built fat JAR.

The PR can be adapted from https://github.com/CS2103AUG2016-T11-C4/main/pull/3 (trying to upstream my changes so they will not bitrot :-P)

pyokagan:
Fixes #177 

The `run` and `runShadow` tasks provide a convenient way for developers
to run the built application from its *.class files or the built fat JAR
respectively.

This is accomplished using the application plugin[1] which implements
the 'run' task.

The shadow plugin, which we already use, will automatically detect the
`application` plugin and will integrate with it[2], providing the
`runShadow` task. It also knows that `mainClassName` should be used
as the "Main-Class" of the JAR manifest, so we don't need to specify
that anymore.

[1] https://docs.gradle.org/current/userguide/application_plugin.html
[2] http://imperceptiblethoughts.com/shadow/#integrating_with_application_plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,11 @@
 plugins {
     id "com.github.kt3k.coveralls" version "2.4.0"
     id "com.github.johnrengelman.shadow" version '1.2.3'
+    id 'application'
 }
+
+// Specifies the entry point of the application
+mainClassName = 'seedu.address.MainApp'
 
 allprojects {
     apply plugin: 'idea'
@@ -82,10 +86,6 @@ allprojects {
 
     shadowJar {
         archiveName = "addressbook.jar"
-
-        manifest {
-            attributes "Main-Class": "seedu.address.MainApp"
-        }
 
         destinationDir = file("${buildDir}/jar/")
     }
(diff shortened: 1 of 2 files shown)


m133225:
Looks good.

pyokagan:
@damithc Can I press the "Merge Pull Request" button once the merge is approved by you? I took a look in [TEAMMATE's process](https://github.com/TEAMMATES/teammates/blob/master/docs/process.md) but couldn't find any info on who presses the button.

Side note, but I noticed that the "merge commit" and "rebase and merge" options are enabled even though this project seems to use only squashes. Perhaps those options could be disabled to minimize human error?

damithc:
Yes, the dev can merge after I have approved the merge.
Yes, we can limit to that option only. We are still in the 'let's try the new merge features and see how it goes' phase. We tried the 'squash' option and seems it works fine. Should be finalized and documented soon. Meanwhile, GitHub keeps coming up with new features :-p

Issue #184, opened 2016-11-30
Simplify construction of UiParts


pyokagan:
Currently there is quite a lot of [boilerplate](https://github.com/se-edu/addressbook-level4/blob/61b1f6fa5d259677efcd8df61b73b7569b7d5715/src/main/java/seedu/address/ui/CommandBox.java#L32-L64) involved if a developer wants to implement a new `UiPart`:

```java
    public static CommandBox load(Stage primaryStage, AnchorPane commandBoxPlaceholder, Logic logic) {
        CommandBox commandBox = UiPartLoader.loadUiPart(primaryStage, commandBoxPlaceholder, new CommandBox());
        commandBox.configure(logic);
        commandBox.addToPlaceholder();
        return commandBox;
    }

    public void configure(Logic logic) {
        this.logic = logic;
        registerAsAnEventHandler(this);
    }

    private void addToPlaceholder() {
        SplitPane.setResizableWithParent(placeHolderPane, false);
        placeHolderPane.getChildren().add(commandTextField);
        FxViewUtil.applyAnchorBoundaryParameters(commandPane, 0.0, 0.0, 0.0, 0.0);
        FxViewUtil.applyAnchorBoundaryParameters(commandTextField, 0.0, 0.0, 0.0, 0.0);
    }

    @Override
    public void setNode(Node node) {
        commandPane = (AnchorPane) node;
    }

    @Override
    public String getFxmlPath() {
        return FXML;
    }

    @Override
    public void setPlaceholder(AnchorPane pane) {
        this.placeHolderPane = pane;
}
```

In short, the developer needs to override:

* `setNode()`
* `setFxmlPath()`
* potentially `setPlaceholder()`

and think of what is the sane thing to do for each of these methods. Additionally, to make the API pleasant to use, the developer also needs to implement a static `load()` method (and in that method, we need to decide which of the three overloaded `UiPartLoader::loadUiPart()` methods we should call)

The reason why is because the current implementation (though `UiPartLoader`) [puts the responsibility on creating new controller instances on the `FXMLLoader`](https://github.com/se-edu/addressbook-level4/blob/master/src/main/java/seedu/address/ui/UiPartLoader.java#L27), but the `FXMLLoader` does not allow us to pass arguments to the constructed controller. This leads to a lot of boilerplate code just to work around this limitation.

[...]

PR #185, opened 2016-12-01
UiParts should not store a reference to the primaryStage #183

### LINKED ISSUE #183: UiParts should not store a reference to the primaryStage
Currently, all UiParts [store a reference to their primaryStage](https://github.com/se-edu/addressbook-level4/blob/61b1f6fa5d259677efcd8df61b73b7569b7d5715/src/main/java/seedu/address/ui/UiPart.java#L21), which means that they can potentially modify the primaryStage at their own fancy. This makes it hard to reason about which UiPart is controlling the primaryStage.

Furthermore, only `MainWindow` requires access the primaryStage. All other UiParts do not, but they are still forced to depend on the primaryStage (violates ISP?)

Perhaps the `primaryStage` field should be pushed down from `UiPart` to `MainWindow`. That way, all other UiParts lose access to the primaryStage, which reduces unneeded coupling.

cc @yl-coder 
See also: https://github.com/CS2103AUG2016-T11-C4/main/pull/38/commits/49bbe31a469438ae7e819244db6bf051154a77d1

pyokagan:
Fixes #183

### DIFF
--- a/src/main/java/seedu/address/commons/util/FxViewUtil.java
+++ b/src/main/java/seedu/address/commons/util/FxViewUtil.java
@@ -20,7 +20,7 @@ public class FxViewUtil {
      * Sets the given image as the icon for the given stage.
      * @param iconSource e.g. {@code "/images/help_icon.png"}
      */
-    public static void setIcon(Stage stage, String iconSource) {
+    public static void setStageIcon(Stage stage, String iconSource) {
         stage.getIcons().setAll(AppUtil.getImage(iconSource));
     }
 
--- a/src/main/java/seedu/address/commons/util/FxViewUtil.java
+++ b/src/main/java/seedu/address/commons/util/FxViewUtil.java
@@ -2,6 +2,7 @@ package seedu.address.commons.util;
 
 import javafx.scene.Node;
 import javafx.scene.layout.AnchorPane;
+import javafx.stage.Stage;
 
 /**
  * Contains utility methods for JavaFX views
@@ -14,4 +15,13 @@ public class FxViewUtil {
         AnchorPane.setRightAnchor(node, right);
         AnchorPane.setTopAnchor(node, top);
     }
+
+    /**
+     * Sets the given image as the icon for the given stage.
+     * @param iconSource e.g. {@code "/images/help_icon.png"}
+     */
+    public static void setIcon(Stage stage, String iconSource) {
+        stage.getIcons().setAll(AppUtil.getImage(iconSource));
+    }
+
 }
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -48,7 +48,7 @@ public class HelpWindow extends UiPart {
         //Null passed as the parent stage to make it non-modal.
         dialogStage = createDialogStage(TITLE, null, scene);
         dialogStage.setMaximized(true); //TODO: set a more appropriate initial size
-        setIcon(dialogStage, ICON);
+        FxViewUtil.setIcon(dialogStage, ICON);
 
         WebView browser = new WebView();
         browser.getEngine().load(USERGUIDE_URL);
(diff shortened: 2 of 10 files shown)


damithc:
@yl-coder @lejolly can one of you review this please?

src/main/java/seedu/address/commons/util/FxViewUtil.java
```diff
@@ -14,4 +15,13 @@ public static void applyAnchorBoundaryParameters(Node node, double left, double
         AnchorPane.setRightAnchor(node, right);
         AnchorPane.setTopAnchor(node, top);
     }
+
+    /**
+     * Sets the given image as the icon for the given stage.
+     * @param iconSource e.g. {@code "/images/help_icon.png"}
+     */
+    public static void setStageIcon(Stage stage, String iconSource) {
+        stage.getIcons().add(AppUtil.getImage(iconSource));
```
yl-coder: Add a clear() before add(). Otherwise, change it to addStageIcon.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -180,6 +186,14 @@ public PersonListPanel getPersonListPanel() {
         return this.personListPanel;
     }
 
+    /**
+     * Sets the given image as the icon for the primary stage.
+     * @param iconSource e.g. {@code "/images/help_icon.png"}
+     */
+    private void setIcon(String iconSource) {
+        primaryStage.getIcons().add(AppUtil.getImage(iconSource));
+    }
```
yl-coder: Could use the method setStageIcon() defined in FxViewUtil for this.

src/main/java/seedu/address/ui/HelpWindow.java
```diff
@@ -26,9 +26,9 @@
 
     private Stage dialogStage;
 
-    public static HelpWindow load(Stage primaryStage) {
+    public static HelpWindow load() {
         logger.fine("Showing help page about the application.");
-        HelpWindow helpWindow = UiPartLoader.loadUiPart(primaryStage, new HelpWindow());
+        HelpWindow helpWindow = UiPartLoader.loadUiPart(null, new HelpWindow());
```
yl-coder: Possible improvement by using Optional.Empty() instead of null.

src/main/java/seedu/address/commons/util/FxViewUtil.java
pyokagan: Good catch. I'll go with adding a clear() before add(), since the list of icons are supposed to be of different sizes of the same image, so I envision that API users are more likely to set all the icons at one shot rather than adding them one by one at different points.

src/main/java/seedu/address/ui/MainWindow.java
pyokagan: Yep, makes sense.

src/main/java/seedu/address/ui/HelpWindow.java
pyokagan: Yeah, this is just a slight workaround to deal with how all the `loadUiPart` methods are overloaded.

Ideally, I would like #184 to be implemented though, which would mean that `UiPartLoader` will be completely removed in the near future anyway.

pyokagan:
@yl-coder v2 ready for review

src/main/java/seedu/address/ui/HelpWindow.java
lejolly: sure, but before that gets done, why not add/rename the methods in `UiPartLoader` so that:
1. you don't have to pass a null
2. to better describe the differences between what goes on within each method

pyokagan: Hrmph. Okay then.

pyokagan:
@yl-coder v3 ready for review

src/main/java/seedu/address/commons/util/FxViewUtil.java
```diff
@@ -14,4 +15,14 @@ public static void applyAnchorBoundaryParameters(Node node, double left, double
...
         AnchorPane.setTopAnchor(node, top);
     }
+
+    /**
+     * Sets the given image as the icon for the given stage.
+     * @param iconSource e.g. {@code "/images/help_icon.png"}
+     */
+    public static void setStageIcon(Stage stage, String iconSource) {
+        stage.getIcons().clear();
+        stage.getIcons().add(AppUtil.getImage(iconSource));
```
pyokagan: Ah, this can actually be expressed more succinctly as:
```java
stage.getIcons().setAll(AppUtil.getImage(iconSource));
```

pyokagan:
@yl-coder v4 ready for review

yl-coder:
Nice work! thanks @pyokagan

PR #187, opened 2016-12-01
Construct BrowserPanel in FXML #186

### LINKED ISSUE #186: Construct BrowserPanel in FXML
The DeveloperGuide states that:
```markdown
The `UI` component uses JavaFx UI framework. The layout of these UI parts are defined in matching `.fxml` files
 that are in the `src/main/resources/view` folder.<br>
 For example, the layout of the [`MainWindow`](../src/main/java/seedu/address/ui/MainWindow.java) is specified in
 [`MainWindow.fxml`](../src/main/resources/view/MainWindow.fxml)
```

While this is mostly true, there is one oddball UiPart that does not have a corresponding `.fxml` file -- [BrowserPanel](https://github.com/se-edu/addressbook-level4/blob/61b1f6fa5d259677efcd8df61b73b7569b7d5715/src/main/java/seedu/address/ui/BrowserPanel.java).

For consistency with the rest of the UiParts, let's construct the `BrowserPanel` using FXML rather than using code.

pyokagan:
Fixes #186 

The DeveloperGuide states that the layout of individual UiParts are
specified in matching `.fxml` files. e.g. The layout of
`MainWindow.java` is specified in `MainWindow.fxml`.

While this is true for almost all UiParts, there is one oddball UiPart
that does not have a corresponding `.fxml` file -- BrowserPanel.

For consistency with the rest of the UiParts, let's construct the
BrowserPanel using FXML rather than using code.


damithc:
@yl-coder for your review please.

src/main/java/seedu/address/ui/BrowserPanel.java
```diff
@@ -42,8 +39,7 @@ public String getFxmlPath() {
      */
     public static BrowserPanel load(AnchorPane placeholder){
         logger.info("Initializing browser");
-        BrowserPanel browserPanel = new BrowserPanel();
-        browserPanel.browser = new WebView();
+        BrowserPanel browserPanel = UiPartLoader.loadUiPart(null, placeholder, new BrowserPanel());
```
yl-coder: Should this be 
UiPartLoader.loadUiPart(placeholder, new BrowserPanel());?

src/main/resources/view/BrowserPanel.fxml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.layout.StackPane?>
+<?import javafx.scene.web.WebView?>
+
+<StackPane xmlns:fx="http://javafx.com/fxml/1" fx:controller="seedu.address.ui.BrowserPanel">
+    <WebView fx:id="browser"/>
+</StackPane>
```
yl-coder: Any particular reason on why you use StackPane instead of AnchorPane?

src/main/java/seedu/address/ui/BrowserPanel.java
pyokagan: Ah you're right, this PR needs to be rebased.

Which makes me wonder why Travis did not catch this. Turns out it does not rebuild PRs if master changes, which means I need to remember to explicitly trigger rebuilds in the future before merging any PR to master ><

src/main/resources/view/BrowserPanel.fxml
pyokagan: Ah right, that's because in one of the previous iterations of the PR (when the order in which I wanted to do these UI PRs were not so clear), I was toying with the idea of removing all the:
```java
FxViewUtil.applyAnchorBoundaryParameters(..., 0.0, 0.0, 0.0, 0.0);
```
that litters the UI code by replacing them with `StackPanes`. This is just a remnant of that.

In the context of this PR (since I did not touch the `applyAnchorBoundaryParameters` calls), it should be an `AnchorPane`. Will change, thanks.

pyokagan: With that said, why are `AnchorPane`s used? :-P

pyokagan:
@yl-coder Sorry, I need to close and make a new PR since I deleted my original fork where I made the PR from in order to make my new fork fork off from se-edu rather than nus-cs2103-AY1617S1.

src/main/java/seedu/address/ui/BrowserPanel.java
damithc: I wish there is a way to know at least the Travis result is out of date. @m133225 is that possible?

src/main/resources/view/BrowserPanel.fxml
yl-coder: First, you want the content to fill the entire placeholder. Second u only want to hold one item. Although, Stackpane provides both requirements. imo, the stackpane should only be used when u need to stack multiple items.

pyokagan: I'll continue the discussion in #208 . Sorry for the PR closing noise.

PR #189, opened 2016-12-01
Construct StatusBarFooter in FXML fully #188

### LINKED ISSUE #188: Construct StatusBarFooter in FXML fully
[StatusBarFooter](https://github.com/se-edu/addressbook-level4/blob/61b1f6fa5d259677efcd8df61b73b7569b7d5715/src/main/java/seedu/address/ui/StatusBarFooter.java#L61-L75) currently constructs and sets up its StatusBar controls using code.

The StatusBar controls can actually be initialized directly in the FXML file. We can avoid placeholder elements and make the code simpler by just doing that.

Now, one may think: "This means Scene Builder cannot edit this FXML file any more!", since the `StatusBar` control comes from `controlsfx`, which is not supported by Scene Builder by default. However,

1. Scene Builder can actually be configured to support custom controls by [following this Stack Overflow answer](http://stackoverflow.com/a/30078204)

2. Having too many placeholder nodes all over the place, and constructing the actual scene graph nodes in code, kind of defeats the purpose of FXML which is to construct the scene graph declaratively without code.

pyokagan:
Fixes #188 

StatusBarFooter currently constructs and sets up its StatusBar controls
using pure code.

The StatusBar control is actually a real control that can be initialized
directly in the FXML file. We can avoid placeholder elements and make
things simpler by just doing that.

Now, one may think: "This means Scene Builder cannot edit this FXML file
any more!", since the StatusBar control comes from controlsfx, which is
not supported by Scene Builder by default. However,

1. Scene Builder can actually be configured[1] to support the custom
   controls from controlsfx, allowing it to edit the FXML file.

2. Having too many placeholder nodes, and constructing the actual scene
   graph in code, kind of defeats the advantage of FXML which is to
   construct the scene graph declaratively without code.

[1] http://stackoverflow.com/a/30078204

Coverage decrease is due to the overall line count of the code base decreasing.

### DIFF
--- a/src/main/java/seedu/address/ui/StatusBarFooter.java
+++ b/src/main/java/seedu/address/ui/StatusBarFooter.java
@@ -19,16 +19,13 @@ import java.util.logging.Logger;
  */
 public class StatusBarFooter extends UiPart {
     private static final Logger logger = LogsCenter.getLogger(StatusBarFooter.class);
+
+    @FXML
     private StatusBar syncStatus;
+    @FXML
     private StatusBar saveLocationStatus;
 
     private GridPane mainPane;
-
-    @FXML
-    private AnchorPane saveLocStatusBarPane;
-
-    @FXML
-    private AnchorPane syncStatusBarPane;
 
     private AnchorPane placeHolder;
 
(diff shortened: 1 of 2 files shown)


damithc:
@yl-coder another one to review

damithc:
@yl-coder can review?

yl-coder:
@damithc yes, I will look into it during weekend.

damithc:
@yl-coder reminder

yl-coder:
Looks good!

Issue #191, opened 2016-12-07
checkstyle: NewlineAtEndOfFile errors with vanilla Git config on Windows


pyokagan:
Checkstyle errors on Windows only when running `gradle checkstyleMain`. Most likely a line endings configuration problem:
```
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\MainApp.java:0: File does not end with a newline. [NewlineAtEndOfF
ile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\ComponentManager.java:0: File does not end with a new
line. [NewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\Config.java:0: File does not end with a newline. [New
lineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\EventsCenter.java:0: File does not end with a newline
. [NewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\GuiSettings.java:0: File does not end with a newline.
 [NewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\LogsCenter.java:0: File does not end with a newline.
[NewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\Messages.java:0: File does not end with a newline. [N
ewlineAtEndOfFile]
[ant:checkstyle] [ERROR] C:\Users\pyokagan\pyk\addressbook-level4\src\main\java\
seedu\address\commons\core\UnmodifiableObservableList.java:0: File does not end
with a newline. [NewlineAtEndOfFile]
```

cc @m133225

m133225:
Is this on the master branch? Can't seem to reproduce this.

pyokagan:
> Does this happen on the master branch? Can't seem to reproduce this.

Yeah, master branch. Do you have `core.autocrlf` switched on in Git?

[...]

PR #193, opened 2016-12-07
Set Up AppVeyor CI #192

### LINKED ISSUE #192: Set Up AppVeyor CI
Jim is a Windows user, so he would be able to sleep more peacefully at night if he knows that this project is being tested regularly on the operating system he uses.

[AppVeyor](https://www.appveyor.com/) is a Windows CI service that is free for open source projects.

pyokagan:
Fixes #192 

AppVeyor[1] is a continuous integration service used to build and test
projects on a Windows VM. Adding AppVeyor support to this project will
allow us to easily ensure that our tests do not break on Windows.

Add an appveyor.yml file that will teach AppVeyor how to build this
project and run its tests.

[1] https://www.appveyor.com/

Work in progress, will ping when it is ready.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,24 @@
+init:
+    - git config --global core.autocrlf true
+
+# Call on gradle to build and run tests
+# --no-daemon: Prevent the daemon from launching to prevent file-in-use errors
+#     when we cache the ~/.gradle directory
+
+build_script:
+    - gradlew.bat --no-daemon assemble
+
+test_script:
+    - gradlew.bat --no-daemon headless checkstyleMain checkstyleTest allTests
+
+environment:
+    matrix:
+        - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+        - JAVA_HOME: C:\Program Files (x86)\Java\jdk1.8.0
+
+# Files/folders to preserve between builds to speed them up
+cache:
+    - C:\Users\appveyor\.gradle
+
+matrix:
+    fast_finish: true  # Finish build once any job fails


pyokagan:
Here's the badge for this PR:

[Build status](https://ci.appveyor.com/project/pyokagan/cs2103aug2016-t11-c4-main/branch/192-appveyor)

Some points to discuss (if we agree to integrate AppVeyor):

1. We have the option of testing with 32-bit or 64-bit Java, or both. This should only matter if the project starts using native libraries, but which option should we go with?

2. Given that Travis already runs checkstyle for us, should we get Appveyor to run checkstyle as well? I'm inclined to say yes, because checkstyle does not add a significant cost to build times anyway.

cc @damithc @m133225

pyokagan:
This PR currently explicitly switches on `core.autocrlf`. I think I will solve #191 first so we can keep the `appveyor.yml` configuration as simple as possible.

damithc:
Nice work @pyokagan 👍 
Any comments @m133225 ?

damithc:
To answer the questions,
1. The more common one which I guess is 64-bit?
2. I'm fine to running checkstyle on both CIs

pyokagan:
@damithc 
> The more common one which I guess is 64-bit?

Yeah I guess we should just go with one. Testing both 32-bit and 64-bit would double the build times, and I don't think we will have any architecture-specific bugs since we don't use any native libs.

If we ever start using native libs then we should switch to testing both architectures.

> I'm fine to running checkstyle on both CIs

OK. I also just realised that running checkstyle on Windows would help us find issues like #191.

pyokagan:
Not ready for review yet, #194 should be merged first :-P

damithc:
Seems @m133225 gave the 👍 
So you can go ahead and merge whenever you are ready @pyokagan

pyokagan:
@damithc Thanks. This PR still needs a bit of cleanup though, I'll ping again when it's really ready.

pyokagan:
Superseded by #218

Issue #173, opened 2016-11-27
Add background color to tags


damithc:
Similar to how GitHub shows labels.
All tags can be shown with the same color

pyokagan:
Fixed in #200

PR #200, opened 2016-12-08
Add background color to tags #173

edmundmok:
For #173 

Preview: 
image

### DIFF
--- a/src/main/java/seedu/address/ui/PersonCard.java
+++ b/src/main/java/seedu/address/ui/PersonCard.java
@@ -3,6 +3,7 @@ package seedu.address.ui;
 import javafx.fxml.FXML;
 import javafx.scene.Node;
 import javafx.scene.control.Label;
+import javafx.scene.layout.FlowPane;
 import javafx.scene.layout.HBox;
 import seedu.address.model.person.ReadOnlyPerson;
 
@@ -23,7 +24,7 @@ public class PersonCard extends UiPart {
     @FXML
     private Label email;
     @FXML
-    private Label tags;
+    private FlowPane tags;
 
     private ReadOnlyPerson person;
     private int displayedIndex;
@@ -43,7 +44,7 @@ public class PersonCard extends UiPart {
         phone.setText(person.getPhone().value);
         address.setText(person.getAddress().value);
         email.setText(person.getEmail().value);
-        tags.setText(person.tagsString());
+        initTags();
     }
 
     public HBox getLayout() {
@@ -58,5 +59,9 @@ public class PersonCard extends UiPart {
     @Override
     public String getFxmlPath() {
         return FXML;
+    }
+
+    private void initTags() {
+        person.getTags().forEach(tag -> tags.getChildren().add(new Label(tag.tagName)));
     }
 }
--- a/src/main/resources/view/PersonListCard.fxml
+++ b/src/main/resources/view/PersonListCard.fxml
@@ -1,18 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<?import javafx.geometry.Insets?>
-<?import javafx.scene.control.*?>
-<?import javafx.scene.layout.*?>
 <?import java.net.URL?>
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.layout.ColumnConstraints?>
+<?import javafx.scene.layout.FlowPane?>
+<?import javafx.scene.layout.GridPane?>
+<?import javafx.scene.layout.HBox?>
+<?import javafx.scene.layout.RowConstraints?>
+<?import javafx.scene.layout.VBox?>
+
 <HBox id="cardPane" fx:id="cardPane" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1">
     <children>
-        <GridPane HBox.hgrow="ALWAYS" >
+        <GridPane HBox.hgrow="ALWAYS">
             <columnConstraints>
                 <ColumnConstraints hgrow="SOMETIMES" minWidth="10.0" prefWidth="150.0" />
-                <ColumnConstraints hgrow="SOMETIMES" maxWidth="100.0" minWidth="10.0" prefWidth="100.0" />
             </columnConstraints>
             <children>
-                <VBox alignment="CENTER_LEFT" maxHeight="150.0" minHeight="105.0" prefHeight="115.0" GridPane.columnIndex="0">
+                <VBox alignment="CENTER_LEFT" minHeight="105.0" GridPane.columnIndex="0">
                     <stylesheets>
                         <URL value="@DarkTheme.css" />
                         <URL value="@Extensions.css" />
@@ -28,9 +33,9 @@
                                     <Label fx:id="id" styleClass="cell_big_label"></Label>
                                     <Label fx:id="name" text="\$first" styleClass="cell_big_label"/>
                                 </HBox>
-                                <Label fx:id="tags" styleClass="cell_small_label" text="\$tags" />
                             </children>
                         </HBox>
+                        <FlowPane fx:id="tags" />
                         <Label fx:id="phone" styleClass="cell_small_label" text="\$phone" />
                         <Label fx:id="address" styleClass="cell_small_label" text="\$address" />
                         <Label fx:id="email" styleClass="cell_small_label" text="\$email" />
--- a/src/test/java/guitests/guihandles/PersonCardHandle.java
+++ b/src/test/java/guitests/guihandles/PersonCardHandle.java
@@ -1,9 +1,15 @@
 package guitests.guihandles;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 import guitests.GuiRobot;
 import javafx.scene.Node;
+import javafx.scene.control.Labeled;
+import javafx.scene.layout.Region;
 import javafx.stage.Stage;
 import seedu.address.model.person.ReadOnlyPerson;
+import seedu.address.model.tag.UniqueTagList;
 
 /**
  * Provides a handle to a person card in the person list panel.
@@ -13,6 +19,7 @@ public class PersonCardHandle extends GuiHandle {
     private static final String ADDRESS_FIELD_ID = "#address";
     private static final String PHONE_FIELD_ID = "#phone";
     private static final String EMAIL_FIELD_ID = "#email";
+    private static final String TAGS_FIELD_ID = "#tags";
 
     private Node node;
 
@@ -41,10 +48,36 @@ public class PersonCardHandle extends GuiHandle {
         return getTextFromLabel(EMAIL_FIELD_ID);
     }
 
+    public List<String> getTags() {
+        return getTags(getTagsContainer());
+    }
+
+    private List<String> getTags(Region tagsContainer) {
+        return tagsContainer
+                .getChildrenUnmodifiable()
+                .stream()
+                .map(node -> ((Labeled) node).getText())
+                .collect(Collectors.toList());
+    }
+
+    private List<String> getTags(UniqueTagList tags) {
+        return tags
+                .asObservableList()
+                .stream()
+                .map(tag -> tag.tagName)
+                .collect(Collectors.toList());
+    }
+
+    private Region getTagsContainer() {
+        return guiRobot.from(node).lookup(TAGS_FIELD_ID).query();
+    }
+
     public boolean isSamePerson(ReadOnlyPerson person) {
-        return getFullName().equals(person.getName().fullName) && getPhone().equals(person.getPhone().value)
+        return getFullName().equals(person.getName().fullName)
+                && getPhone().equals(person.getPhone().value)
                 && getEmail().equals(person.getEmail().value)
-                && getAddress().equals(person.getAddress().value);
+                && getAddress().equals(person.getAddress().value)
+                && getTags().equals(getTags(person.getTags()));
     }
 
     @Override
@@ -52,7 +85,10 @@ public class PersonCardHandle extends GuiHandle {
         if (obj instanceof PersonCardHandle) {
             PersonCardHandle handle = (PersonCardHandle) obj;
             return getFullName().equals(handle.getFullName())
-                    && getAddress().equals(handle.getAddress()); //TODO: compare the rest
+                    && getPhone().equals(handle.getPhone())
+                    && getEmail().equals(handle.getEmail())
+                    && getAddress().equals(handle.getAddress())
+                    && getTags().equals(handle.getTags());
         }
         return super.equals(obj);
     }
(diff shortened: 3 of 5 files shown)


src/main/resources/view/PersonListCard.fxml
```diff
@@ -1,18 +1,23 @@
...
     <children>
-        <GridPane HBox.hgrow="ALWAYS" >
+        <GridPane HBox.hgrow="ALWAYS">
             <columnConstraints>
                 <ColumnConstraints hgrow="SOMETIMES" minWidth="10.0" prefWidth="150.0" />
-                <ColumnConstraints hgrow="SOMETIMES" maxWidth="100.0" minWidth="10.0" prefWidth="100.0" />
             </columnConstraints>
             <children>
-                <VBox alignment="CENTER_LEFT" maxHeight="150.0" minHeight="105.0" prefHeight="115.0" GridPane.columnIndex="0">
+                <VBox alignment="CENTER_LEFT" maxHeight="1.7976931348623157E308" minHeight="105.0" GridPane.columnIndex="0">
```
pyokagan: Huh? `1.7976931348623157E308` ? :-D

src/main/resources/view/PersonListCard.fxml
```diff
@@ -1,18 +1,23 @@
...
+<?import javafx.scene.layout.RowConstraints?>
+<?import javafx.scene.layout.VBox?>
+
+<HBox id="cardPane" fx:id="cardPane" xmlns="http://javafx.com/javafx/8.0.60" xmlns:fx="http://javafx.com/fxml/1">
     <children>
-        <GridPane HBox.hgrow="ALWAYS" >
+        <GridPane HBox.hgrow="ALWAYS">
             <columnConstraints>
                 <ColumnConstraints hgrow="SOMETIMES" minWidth="10.0" prefWidth="150.0" />
-                <ColumnConstraints hgrow="SOMETIMES" maxWidth="100.0" minWidth="10.0" prefWidth="100.0" />
```
pyokagan: This change is fine -- multiple `ColumnConstraints` does not make any sense.

src/main/resources/view/PersonListCard.fxml
```diff
@@ -22,21 +27,24 @@
...
                                 <HBox>
-                                    <Label fx:id="id" styleClass="cell_big_label"></Label>
-                                    <Label fx:id="name" text="\$first" styleClass="cell_big_label"/>
+                                    <Label fx:id="id" styleClass="cell_big_label" />
+                                    <Label fx:id="name" styleClass="cell_big_label" text="\$first" />
                                 </HBox>
-                                <Label fx:id="tags" styleClass="cell_small_label" text="\$tags" />
                             </children>
                         </HBox>
+                  <FlowPane fx:id="tags" />
```
pyokagan: The indentation for this should be consistent with the rest.

src/main/resources/view/PersonListCard.fxml
```diff
@@ -22,21 +27,24 @@
...
                         <Label fx:id="phone" styleClass="cell_small_label" text="\$phone" />
                         <Label fx:id="address" styleClass="cell_small_label" text="\$address" />
                         <Label fx:id="email" styleClass="cell_small_label" text="\$email" />
                     </children>
                 </VBox>
             </children>
+         <rowConstraints>
+            <RowConstraints />
+         </rowConstraints>
```
pyokagan: Probably not needed.

src/main/resources/view/PersonListCard.fxml
```diff
@@ -22,21 +27,24 @@
                     </padding>
 
                     <children>
-                        <HBox spacing="5" alignment="CENTER_LEFT">
+                        <HBox alignment="CENTER_LEFT" spacing="5">
```
pyokagan: Yeah, no point making a change just to swap the order of attributes :-P

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -59,4 +60,7 @@ public void setNode(Node node) {
     public String getFxmlPath() {
         return FXML;
     }
+    private void initTags() {
+        person.getTags().forEach(tag -> tags.getChildren().add(new Label(tag.tagName)));
+    }
```
pyokagan: OK. I do wonder what happens when [`Person::setTags()`](https://github.com/se-edu/addressbook-level4/blob/7ff3ad82caaf32e71def3d26f133c1152cad0ac6/src/main/java/seedu/address/model/person/Person.java#L68) is called though :-D I guess the UI wouldn't update in that case. Making `Person` mutable seems to be problematic, but I don't think that needs to be fixed in this PR until we get around to doing the edit command.

src/main/resources/view/PersonListCard.fxml
edmundmok: There was a gridpane in the original card, which I removed since it was quite pointless (not sure about this, maybe something was planned there?). Leaving it there would lead to wasted space on the right of the card when the number of tags becomes quite large.

src/main/resources/view/PersonListCard.fxml
edmundmok: MAX_VALUE for max height of the card so that it grows vertically when the number of tags is quite large.

src/main/resources/view/PersonListCard.fxml
edmundmok: By the way, these changes were made using SceneBuilder... not sure why it changes these stuff, I can change it back if needed.

src/main/resources/view/PersonListCard.fxml
edmundmok: I'll change this manually.

src/main/resources/view/PersonListCard.fxml
pyokagan: Oh, why not just remove the `maxHeight` then?

edmundmok:
Changed the card handle to check the tags as well by converting them into a string like how  `tagsString()` does, and checking if they are the same.

pyokagan:
@edmundmok Sorry for the delay. I have to delay for another one day though, so sorry about that ><.

edmundmok:
@pyokagan No problem!

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +41,26 @@ public String getEmail() {
         return getTextFromLabel(EMAIL_FIELD_ID);
     }
 
+    public String getTags() {
+        return getTextFromTags(node);
```
pyokagan: This seems a little too inflexible. What happens, for example, if I want to iterate over each tag of the person card? Maybe this should return a `List<String>` instead.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +41,26 @@ public String getEmail() {
...
+    }
+
     public boolean isSamePerson(ReadOnlyPerson person) {
         return getFullName().equals(person.getName().fullName) && getPhone().equals(person.getPhone().value)
                 && getEmail().equals(person.getEmail().value)
-                && getAddress().equals(person.getAddress().value);
+                && getAddress().equals(person.getAddress().value)
+                && getTags().equals(person.tagsString());
```
pyokagan: I'm a bit concerned that we are comparing the *string representation* of the tags. This may make the tests too sensitive. (See "Overspecified software" in http://xunitpatterns.com/Fragile%20Test.html)

We are not concerned with the string formatting of the tags, but just whether we actually inserted each tag as a label node in the scene graph. So we should compare the tags as a list of strings or something.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -68,6 +71,29 @@ public void pressEnter() {
         guiRobot.type(KeyCode.ENTER).sleep(500);
     }
 
+    protected String getTextFromTags(Node parentNode) {
```
pyokagan: Since this is in the `GuiHandle` base class, *all* derived classes will get this method.

However, this method seems to require a scene graph structure that is a bit too specific to how the scene graph of the person card is implemented. This should probably be a private method of `PersonCardHandle` instead.

Also, this method requires a certain scene graph structure, so that should probably be documented in its javadoc.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -68,6 +71,29 @@ public void pressEnter() {
...
+
+        final String tagsFieldId = "#tags";
```
pyokagan: This should be a constant in `PersonCardHandle`.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -68,6 +71,29 @@ public void pressEnter() {
...
+            .forEach(tag -> {
+                String tagAsString = TestUtil.getTagAsString(((Label) tag).getText());
+                buffer.append(tagAsString).append(separator);
+            });
+
+        if (buffer.length() == 0) {
+            return "";
+        } else {
+            return buffer.substring(0, buffer.length() - separator.length());
+        }
```
pyokagan: All of these can probably be more succinctly written as (untested)
```java
        Region tagsContainer = guiRobot.from(parentNode).lookup("#tags").query();
        return tagsContainer.getChildrenUnmodifiable().stream()
            .map(node -> ((Labeled) node).getText())
            .map(TestUtil::getTagAsString)
            .collect(Collectors.joining(", "));
```

You should really consider returning a `List<String>` though.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -68,6 +71,29 @@ public void pressEnter() {
         guiRobot.type(KeyCode.ENTER).sleep(500);
     }
 
+    protected String getTextFromTags(Node parentNode) {
+
+        final String tagsFieldId = "#tags";
+        final StringBuffer buffer = new StringBuffer();
+        final String separator = ", ";
+
+        FlowPane tagsFlowPane = (FlowPane) guiRobot.from(parentNode).lookup(tagsFieldId).tryQuery().get();
```
pyokagan: Requiring a `FlowPane` is a bit too specific. The most low level class that we require is `Region`, which provides `getChildrenUnmodifiable()`.

src/main/resources/view/PersonListCard.fxml
```diff
@@ -25,12 +30,12 @@
                         <HBox spacing="5" alignment="CENTER_LEFT">
                             <children>
                                 <HBox>
-                                    <Label fx:id="id" styleClass="cell_big_label"></Label>
-                                    <Label fx:id="name" text="\$first" styleClass="cell_big_label"/>
+                                    <Label fx:id="id" styleClass="cell_big_label" />
+                                    <Label fx:id="name" styleClass="cell_big_label" text="\$first" />
```
pyokagan: Unnecessary change: switching the order of attributes.

damithc:
@edmundmok reminder to resume this PR

src/test/java/guitests/guihandles/PersonCardHandle.java
edmundmok: Noted!

edmundmok:
Sorry for the delay. Have made some changes directly into the commits and squashed them.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +48,40 @@ public String getEmail() {
...
+    }
+
     public boolean isSamePerson(ReadOnlyPerson person) {
         return getFullName().equals(person.getName().fullName) && getPhone().equals(person.getPhone().value)
                 && getEmail().equals(person.getEmail().value)
-                && getAddress().equals(person.getAddress().value);
+                && getAddress().equals(person.getAddress().value)
+                && getTags().equals(person.getTags().asObservableList()
+                        .stream()
+                        .map(tag -> tag.toString())
```
pyokagan: `tag.toString()` adds that annoying `[` `]` brackets, which is why you need to use `getTagAsString()` above.

If you just compare `tag.tagName` directly, then you will not need to implement and use `getTagAsString()` at all.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +48,40 @@ public String getEmail() {
...
+    }
+
+    private List<String> getTagsAsList(Region tagsContainer) {
+        return tagsContainer.getChildrenUnmodifiable().stream()
+                .map(node -> TestUtil.getTagAsString(((Labeled) node).getText()))
+                .collect(Collectors.toList());
+    }
+
+    private Region getTagsContainer() {
+        return (Region) guiRobot.from(node).lookup(TAGS_FIELD_ID).query();
```
pyokagan: No need the `(Region)` cast.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +48,40 @@ public String getEmail() {
         return getTextFromLabel(EMAIL_FIELD_ID);
     }
 
+    public List<String> getTags() {
+        Region tagsContainer = getTagsContainer();
+        return getTagsAsList(tagsContainer);
```
pyokagan: Can just call `getTagsAsList(getTagsContainer())` directly.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +48,40 @@ public String getEmail() {
...
+    }
+
+    private List<String> getTagsAsList(Region tagsContainer) {
```
pyokagan: Consider making this an overload of `getTags()`, e.g. `getTags(Region tagsContainer)`.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -41,18 +48,40 @@ public String getEmail() {
...
+
     public boolean isSamePerson(ReadOnlyPerson person) {
         return getFullName().equals(person.getName().fullName) && getPhone().equals(person.getPhone().value)
                 && getEmail().equals(person.getEmail().value)
-                && getAddress().equals(person.getAddress().value);
+                && getAddress().equals(person.getAddress().value)
+                && getTags().equals(person.getTags().asObservableList()
+                        .stream()
+                        .map(tag -> tag.toString())
+                        .collect(Collectors.toList()));
```
pyokagan: Maybe move all these stream processing stuff into a separate method. It can be an overload of `getTags()`, e.g. `getTags(UniqueTagList tags)`.

pyokagan:
@edmundmok Re: code coverage. I think you can just remove `ReadOnlyPerson::tagsString()`.

damithc:
Good work @edmundmok 👍

edmundmok:
Thanks @pyokagan for the helpful comments and guidance! I've learnt a lot from this.

PR #202, opened 2016-12-08
Users can delete multiple persons in one shot

### LINKED ISSUE #140: User can delete multiple persons in one shot
... so that user can do batch clean up of outdated contacts

cheec:
Fixes #140 
Implement multiple deleting 
* Users can now delete multiple persons in a single command
* Users must seperate multiple indices with at least a single whitespace
* Users can also specify a range indicator '-' for range deleting
* Overlapping ranges are also handled appropriately
* Indices specified in ranges do not need to be in ascending order
* Indices cannot exceed list size

Examples: 
* `delete 5 1 7 10-16` deletes persons 1, 5, 7, 10 to 16
* `delete 1 1 1 1 1` deletes person 1
* `delete 5-1 1-5 1 2 3 4 5` deletes persons 1 to 5
* `delete 1-3 2-5 1-10` deletes persons 1 to 10
* `delete 6-1` delete persons 1 to 6

This PR contains major changes pertinent to:
* `Parser` - add new helper methods to parse arguments above
* `DeleteCommand` - implement above functionality
* `DeleteCommandTest` - add some GUI tests
* `LogicManagerTest` - delete command tests added/updated
* `StringUtilTest` - added tests for new helper method `StringUtil::toIndexedListString` 
* `ListUtil` - create utility class holding methods relevant to lists
* `CollectionUtil` - fixed and refactored the otherwise not functioning `CollectionUtil::assertNoNullElements`
    * I'm guessing no one found this out because of the lack of `CollectionUtil` tests, time to open a new issue I think ;^)
* Docs - update UG to match new functionality 

ss

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level4/202)


damithc:
Good start @cheec 
Let us know when ready for review.

cheec:
Should be ready for review now

damithc:
@YijinL can review this one?

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -13,38 +19,69 @@
...
         } catch (PersonNotFoundException pnfe) {
             assert false : "The target person cannot be missing";
         }
 
-        return new CommandResult(String.format(MESSAGE_DELETE_PERSON_SUCCESS, personToDelete));
+        return personsToDelete;
+    }
+
+    /** Returns true if all indices are valid, that is, they are within bounds of {@code targetList}. */
+    private boolean isValidIndicesOf(UnmodifiableObservableList<ReadOnlyPerson> targetList) {
```
YijinL: Param type can be more general? No logical requirement to tie implementation to the `UnmodifiableObservableList` or `ReadOnlyPerson` classes.
```java
areValidIndicesOf(List<? extends ReadOnlyPerson> targetList) {
```

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -153,27 +277,53 @@ private Command prepareSelect(String args) {
...
-     * @param args full command args string
-     * @return the prepared command
+     * Returns an {@code Optional.empty()} otherwise.
      */
-    private Command prepareFind(String args) {
-        final Matcher matcher = KEYWORDS_ARGS_FORMAT.matcher(args.trim());
-        if (!matcher.matches()) {
-            return new IncorrectCommand(String.format(MESSAGE_INVALID_COMMAND_FORMAT,
-                    FindCommand.MESSAGE_USAGE));
+    private Optional<List<Integer>> parseIndex(List<String> tokens) {
```
YijinL: Should be named `parseIndices`.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -153,27 +277,53 @@ private Command prepareSelect(String args) {
...
+            this.end = Math.max(start, end);
+        }
+
+        private int getStart() {
+            return this.start;
+        }
+
+        private int getEnd() {
+            return this.end;
+        }
```
YijinL: Final primitive fields can be safely made public, the below example is shorter and has less boilerplate code.

Also, if this represents a 'range of *indices*' then you should work some index-related guarantees into the constructor, like `start and end > 0`. 
Not a big issue since this is a private class, but good practice to hard-code constraints that people expect from the name of the class.
```java
private class IndexRange {
    public final int START;
    public final int END;
    private IndexRange(int a, int b) { // since you don't need it to be asc here
        assert start > 0 && end > 0;
        this.START = Math.min(a, b);
        this.END = Math.max(a, b);
    }
}
```

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,31 +1,48 @@
...
-        targetIndex = currentList.length / 2;
-        assertDeleteSuccess(targetIndex, currentList);
+        deleteRawArgs = "" + currentList.length / 2;
+        assertDeleteSuccess(deleteRawArgs, currentList, currentList.length / 2);
+        currentList = TestUtil.removePersonFromList(currentList, currentList.length / 2);
+
+        //duplicates should be ignored
+        deleteRawArgs = "1 1 1 1 1 1-1";
+        assertDeleteSuccess(deleteRawArgs, currentList, 1);
+        currentList = TestUtil.removePersonFromList(currentList, 1);
```
YijinL: Not true unit test; can split into following methods:

- `delete_fromFront()` covers continuous sections (can be single) starting from the front
- `delete_fromBack()` same idea
- `delete_fromMiddle()` same idea
- `delete_ignoresDuplicates()` ...

etc...
Also follow the coding style guide for test names for consistency.

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -38,17 +55,27 @@ public void delete() {
      * @param targetIndexOneIndexed e.g. index 1 to delete the first person in the list,
      * @param currentList A copy of the current list of persons (before deletion).
      */
-    private void assertDeleteSuccess(int targetIndexOneIndexed, final TestPerson[] currentList) {
-        TestPerson personToDelete = currentList[targetIndexOneIndexed - 1]; // -1 as array uses zero indexing
-        TestPerson[] expectedRemainder = TestUtil.removePersonFromList(currentList, targetIndexOneIndexed);
+    private void assertDeleteSuccess(String targetIndicesArgs, final TestPerson[] currentList,
+            Integer... expectedIndicesDeleted) {
```
YijinL: Just curious why you put `final` for the `currentList` param. Also why not the others? 

My personal preference is to only finalise method parameters if I reassign *any* of the other parameters in the method body. 

Otherwise it feels like the cost in visual clutter and consistency (some methods have final params, some don't) is not worth the gain in compile-time safety.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -507,4 +596,38 @@ Person generatePersonWithName(String name) throws Exception {
...
+         * Returns the expected feedback from deleting.
+         * Persons are also supplied by {@code testPersons}.
+         */
+        String removePersonsFrom(AddressBook expectedAB, List<Person> testPersons, Integer... targetIndices)
+                throws Exception {
+            List<Person> toDelete = Stream.of(targetIndices)
+                    .distinct() // Removing duplicate indices.
+                    .sorted() // Sorting as targetIndices may be unordered.
+                    .map(i -> testPersons.get(i - 1))
+                    .collect(Collectors.toList());
```
YijinL: These 5 lines appear very often in the entire changeset. Strongly consider abstracting it out as a public static utility method in the most appropriate class.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -298,18 +329,76 @@ public void execute_deleteIndexNotFound_errorMessageShown() throws Exception {
...
+                expectedAB.getPersonList());
+
+        // deleting with duplicate indices
+        expectedIndicesDeleted = new Integer[] {3}; // dupes are ignored.
+        expectedFeedback = deleteHelper.removePersonsFrom(expectedAB, testPersons, expectedIndicesDeleted);
+        assertCommandBehavior("delete 3 3 3 3 3 3",
+                expectedFeedback,
+                expectedAB,
+                expectedAB.getPersonList());
+    }
```
YijinL: See my comments for `DeleteCommandTest`.
Names can be formatted `execDelete_<parition case>_removesCorrectly()`.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
YijinL: @damithc there are a lot of `is*` method names across the repos that sound very awkward because they are forcing the [coding guidelines (see no. 10)](https://docs.google.com/document/pub?id=1iAESIXM0zSxEa5OY7dFURam_SgLiSMhPQtU0drQagrs&embedded=true). Awkward sounding = loss of readability.

The guidelines do mention other alternative prefixes like 'can', 'has', 'should' but I think most people miss them. I will come up with an extended guideline on boolean method naming in an issue and tag everyone for discussion.

src/main/java/seedu/address/commons/util/CollectionUtil.java
```diff
@@ -28,7 +24,7 @@ public static boolean isAnyNull(Object... items) {
      */
     public static void assertNoNullElements(Collection<?> items) {
         assert items != null;
-        assert !isAnyNull(items);
+        assert items.stream().allMatch(i -> i != null);
```
pyokagan: Tangential to this PR, but we have both `isAnyNull()` (which operates on an array), and `assertNoNullElements()` (which operates on a collection). This leads to inconsistent assertion styles throughout the code base where we have e.g.

(from [`Person.java`](https://github.com/se-edu/addressbook-level4/blob/e68f4529ea42ac5929472b76fbb4a7bf9e8cbbf2/src/main/java/seedu/address/model/person/Person.java#L25) )
```java
assert !CollectionUtil.isAnyNull(name, phone, email, address, tags);
```

and (from [`UniqueTagList`](https://github.com/se-edu/addressbook-level4/blob/e68f4529ea42ac5929472b76fbb4a7bf9e8cbbf2/src/main/java/seedu/address/model/tag/UniqueTagList.java#L44) )
```java
CollectionUtil.assertNoNullElements(tags);
```

Personally I prefer how `isAnyNull` is used -- the use of the `assert` keyword (rather than putting it in the method body) makes it clear that it is an assertion that will only be run if assertions are enabled.

So, the end result may be clearer and more consistent if we:

1. Introduce an `isAnyNull(Collection<?> items)` overload
2. Replace `assertNoNullElements` with `assert !isAnyNull(items)`

src/test/java/guitests/DeleteCommandTest.java
cheec: That `final` was left in from the original code :P 
I'll make appropriate changes.

pyokagan:
@cheec 
>  Codacy: remove unused import
> DeleteCommand: add assertions

These seem to be commits that fix up errors in previous commits, and thus could be squashed into them.

src/main/java/seedu/address/commons/util/CollectionUtil.java
cheec: @pyokagan makes a point :-D

damithc: Yes, it would be good to use `assert` consistently. `CollectionUtil.assertNoNullElements(tags)` could be a legacy method from level 3 in which we didn't want to use `assert`s

src/main/java/seedu/address/logic/commands/DeleteCommand.java
damithc: Good idea @YijinL 
Yes, the more important thing is that enhances the readability (i.e. reads more 'natural')

cheec: Noted, will change accordingly

cheec:
Made changes, please review again @YijinL 
I created a new utility class `ListUtil` but tests are not done for it. Seeing that this PR is already getting quite large and `CollectionUtil`'s tests are yet to implemented, perhaps `ListUtil` and `CollectionUtil` tests can be implemented together in a different issue.

damithc:
@cheec good job on this PR so far. This is a significant feature; so let's keep at it until it is merged.

YijinL:
@damithc unrelated but this branch was rebased after my review, so I couldn't use the `view changes since last review` feature. Since we are going to squash this anyway do you think it would be better to leave the original commits and allow merge commits?

YijinL:
@cheec looks pretty good
Next time if you rebase please add a brief overview of new changes as a PR comment, because it feels inefficient to look through all changes again :P
Or don't rebase and we can use github's 'see changes since last review' feature.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -13,38 +19,63 @@
...
 
+        return new CommandResult(String.format(MESSAGE_DELETE_PERSON_SUCCESS, deletedPersons.size(),
+                StringUtil.toIndexedListString(deletedPersons)));
+    }
+
+    /*
+     * =========================================================================
+     *                             Helper Methods
+     * =========================================================================
+     */
```
damithc: Not sure if we really need this banner here.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -13,38 +19,63 @@
...
+
+    /** Deletes persons from {@code targetList} then returns the list of deleted persons. */
+    private List<ReadOnlyPerson> deletePersonsFrom(UnmodifiableObservableList<ReadOnlyPerson> targetList) {
+        List<ReadOnlyPerson> personsToDelete = ListUtil.sublistFromIndices(targetList, -1, targetIndices);
```
damithc: The method signature is easier to understand if the targetIndices is passed as a parameter?

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -13,38 +19,63 @@
...
+            }
         } catch (PersonNotFoundException pnfe) {
             assert false : "The target person cannot be missing";
         }
+        return personsToDelete;
+    }
 
-        return new CommandResult(String.format(MESSAGE_DELETE_PERSON_SUCCESS, personToDelete));
+    /** Returns true if all indices are valid, that is, they are within bounds of {@code targetList}. */
+    private boolean hasValidIndicesFor(List<? extends ReadOnlyPerson> targetList) {
```
damithc: How about this?
```java
/** Returns true if all indices are within bounds of {@code targetList}. */
    private boolean areIndicesWithinBounds(List<? extends ReadOnlyPerson> targetList, List<Integer> targetIndices) { ...}
```

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +163,113 @@ private Command prepareSelect(String args) {
...
+     * Parses arguments in the context of the find person command.
+     *
+     * @param args full command args string
+     * @return the prepared command
+     */
+    private Command prepareFind(String args) {
+        final Matcher matcher = KEYWORDS_ARGS_FORMAT.matcher(args.trim());
+        if (!matcher.matches()) {
+            return new IncorrectCommand(String.format(MESSAGE_INVALID_COMMAND_FORMAT,
+                    FindCommand.MESSAGE_USAGE));
```
damithc: This indentation makes it easier to read. The brackets appear extreme left and extreme right of the content that should be within those brackets.
```java
return new IncorrectCommand(String.format(MESSAGE_INVALID_COMMAND_FORMAT,
                                                 FindCommand.MESSAGE_USAGE));
```

Another option (line break should be as high in the content tree as possible):
```java
return new IncorrectCommand(
        String.format(MESSAGE_INVALID_COMMAND_FORMAT, FindCommand.MESSAGE_USAGE));
```

damithc: Just realized this was not your code. The point is still valid. :-p

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +163,113 @@ private Command prepareSelect(String args) {
...
+     *
+     * These tokens must also contain either: <br>
+     * <ul>
+     *   <li> a positive unsigned integer i.e. a non-ranged index </li>
+     *   <li> or positive unsigned integers surrounding a {@code PERSON_INDEX_RANGE_INDICATOR}
+     *          i.e. a ranged index</li>
+     * </ul>
+     *
+     * Hence 'mixed' as {@code tokens} may contain both ranged and non-ranged indices. <br>
+     * The returned indices may be unordered and have duplicates.
```
damithc: `may have duplicates`?

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +163,113 @@ private Command prepareSelect(String args) {
...
+     * Returns an {@code Optional.empty()} otherwise.
+     */
+    private Optional<List<Integer>> parseMixedIndices(List<String> tokens) {
```
damithc: I wonder if we can omit the `Mixed`.  If we don't specify a specific index type, 'anything goes' is kind of implied?

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +163,113 @@ private Command prepareSelect(String args) {
...
+        if (!startIndex.isPresent() || !endIndex.isPresent()) {
+            return Optional.empty();
+        }
+
+        return Optional.of(new IndexRange(startIndex.get(), endIndex.get()));
+    }
+
+    /**
+     * Returns a list of indices extracted from {@code tokens} if each token contains positive
+     * unsigned integers surrounding a {@code PERSON_INDEX_RANGE_INDICATOR}.
```
damithc: This sentence sound like is any of the tokens is not a positive integer, you return nothing. Is that correct?

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,146 @@
...
+        TestPerson[] expectedRemainder = getDeletedRemainder(currentList, expectedPersonsDeleted);
+        assertDeleteSuccess("5 1-3 7", expectedRemainder, expectedPersonsDeleted);
+    }
 
-        //delete the last in the list
-        currentList = TestUtil.removePersonFromList(currentList, targetIndex);
-        targetIndex = currentList.length;
-        assertDeleteSuccess(targetIndex, currentList);
+    @Test
+    public void delete_ignoresDuplicates() {
```
damithc: `delete_duplicateIndices_duplicatesIgnored`

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -296,20 +326,142 @@ public void execute_deleteIndexNotFound_errorMessageShown() throws Exception {
...
+        List<Person> testPersons = helper.generatePersonList(5);
+        AddressBook expectedAB = helper.generateAddressBook(testPersons);
+        helper.addToModel(model, testPersons);
+
+        String expectedFeedback = deleteHelper.deleteAtIndices(expectedAB, testPersons, 3, 4, 5);
+        assertCommandBehavior("delete       3          4      5    ",
+                expectedFeedback,
+                expectedAB,
+                expectedAB.getPersonList());
+    }
```
damithc: As the DeleteCommand is unit tested separately, there is no need to have lot of test cases here? If LogicManager can execute one DeleteCommand correctly, we can assume it can execute all delete commands correctly (as long as the DeleteCommand is working correctly).

damithc:
>@damithc unrelated but this branch was rebased after my review, so I couldn't use the `view changes since last review` feature. Since we are going to squash this anyway do you think it would be better to leave the original commits and allow merge commits?

I too have that concern about rebasing mid-PR. @pyokagan your thoughts?

src/main/java/seedu/address/logic/commands/DeleteCommand.java
YijinL: Can potentially shift this to `ListUtil` too

src/test/java/seedu/address/logic/LogicManagerTest.java
YijinL: Maybe can add to the header comment for `LogicManagerTest` what we don't need to check?
There's all this overlap because we aren't using mocks/stubs at this level, and it may not be clear how much we should test at higher levels.

pyokagan:
@damithc says:
> I too have that concern about rebasing mid-PR. @pyokagan your thoughts?

My opinion as a reviewer:

@YijinL says:
> unrelated but this branch was rebased after my review, so I couldn't use the view changes since last review feature.

Hrm, seems like yet another huge failing with Github's review funcionality :-/

FWIW what I do is that for every review iteration I do, I save a copy of the PR using my `git-fetch-pr` script. (I've dumped it [here](https://gist.github.com/pyokagan/a7905f12c8248e20a4cf83b5ace5748a)). I can then view the diff between iterations using something like:
```shell
git diff pr/56-55-implement-more-rigorous-validity-check-file-path-v1 pr/56-55-implement-more-rigorous-validity-check-file-path-v2
```
This works with rebases, which is why I'm puzzled why Github's review system does not support this.

An alternative is to use an external system like https://reviewable.io/ . It can be used to view diffs between revisions without actually needing access to the repo, so it can be used on a personal basis.

> Since we are going to squash this anyway do you think it would be better to leave the original commits and allow merge commits?

Hrm, I think in this case @cheec used rebase to squash any changes into the PR directly into their respective commits and not to update the branch with changes from master (?), so probably the discussion should be about "should we allow contributors to rewrite their PR branches"?

I think yes, because this allows contributors to submit properly-organized PRs, where each commit is a single logical change with a commit message that explains the context and rationale for that change. With this, reviewers can review each commit one by one, allowing much finer and effective reviews compared to staring at a gigantic diff.

As https://twitter.com/iamdevloper/status/397664295875805184 says:

> 10 lines of code = 10 issues.
> 500 lines of code = "looks fine."
> Code reviews.

Proper commits will solve that :-)

Some examples where being able to rewrite PRs benefited me: https://github.com/CS2103AUG2016-T11-C4/main/pull/19 https://github.com/CS2103AUG2016-T11-C4/main/pull/2 https://github.com/CS2103AUG2016-T11-C4/main/pull/120 https://github.com/CS2103AUG2016-T11-C4/main/pull/128 Try looking through these PRs commit-by-commit compared to just looking at the "Files Changed" tab :-)

(Some may notice that these PRs are big because they contain "cleanup" commits, which serve to code base nicer to implement the new feature on top of. Some may argue that these cleanup commits should be in separate PRs. *However*, doing this becomes difficult when the commits which implement those new features strongly *depend* on prior cleanups, which is why keeping them in the same PR, while separating them into separate commits are useful. Note that cleanup commits are usually inevitable in when implementing new big features, so this seems to be another plus for allowing PRs to be rewritten.)

By not allowing PRs to be rewritten, big PRs will naturally accumulate a load of "fixup" commits which pollute the PR. These commits add a lot of noise and make it impossible to view the PR as individual logical changes for review.

Furthermore, if we use regular merges to master rather than "squash and merge", this will give Git much more fine-grained information for `git-blame` and `git-bisect`. With "squash and merge", when we bisect to the commit which introduced the bug, we may still end up with the 600+ lines diff which we would need to then sieve through for the line of code which introduced the bug.

Even if we continued to use "squash and merge", Github combines the commit messages of all the commits in the PR to make the squashed commit's commit message. So, having logical commits with proper commit messages would still bring an improvement to the repo's commit log, allowing developers (or outsiders) to understand better at a glance what has changed with each commit.

So, if I were to choose between "allowing PRs to be rewritten and using additional tooling to make up for Github's problems", and "not allowing PRs to be rewritten and sacrificing code review quality" (which affects every single large PR now and in the future which I have to review), I'd prefer allowing PRs to be rewritten.

Thanks.

src/main/java/seedu/address/logic/parser/Parser.java
pyokagan: I think this is an example of multiple logical changes being smooshed together into one commit, thus making the diff hard to read.

At the very least, `Clean up imports` and `Shift all helper methods to the bottom for readability` are both clear-cut separate logical changes and should be split into their own commits. (I haven't looked at the other changes yet).

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +163,113 @@ private Command prepareSelect(String args) {
...
+            final Optional<IndexRange> indexRange = parseRangedIndex(token);
+            if (!indexRange.isPresent()) {
+                return Optional.empty();
+            }
+
+            // Adding every index from start to end (inclusive) to 'indices'.
+            IntStream.rangeClosed(indexRange.get().start, indexRange.get().end).forEach(indices::add);
+        }
+
+        return Optional.of(indices);
```
pyokagan: For fun, I checked out how difficult it would be to convert this code to use the streams API.

I came up with something like this:
```diff
diff --git a/src/main/java/seedu/address/commons/util/ListUtil.java b/src/main/java/seedu/address/commons/util/ListUtil.java
index fcef339..4ece580 100644
--- a/src/main/java/seedu/address/commons/util/ListUtil.java
+++ b/src/main/java/seedu/address/commons/util/ListUtil.java
@@ -1,8 +1,11 @@
 package seedu.address.commons.util;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -51,8 +54,45 @@ public class ListUtil {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Returns a collector that transforms a List<Optional<T>> into a Optional<List<T>>
+     */
+    public static final <T> Collector<Optional<? extends T>, ?, Optional<List<T>>> collectOptional() {
+        return Collector.of(
+            () -> new CollectOptional<T>(),
+            CollectOptional<T>::add,
+            CollectOptional<T>::combine,
+            CollectOptional<T>::finish);
+    }
+
     /** Returns true if every index in {@code indices} is within bounds of {@list} after applying the {@code offset} */
     private static <T> boolean areValidIndicesOf(List<T> list, int offset, List<Integer> indices) {
         return Collections.min(indices) + offset >= 0 && Collections.max(indices) + offset < list.size();
     }
+
+    private static class CollectOptional<E> {
+        boolean empty;
+        List<E> items = new ArrayList<>();
+
+        void add(Optional<? extends E> item) {
+            if (item.isPresent()) {
+                items.add(item.get());
+            } else {
+                empty = true;
+            }
+        }
+
+        CollectOptional<E> combine(CollectOptional<? extends E> other) {
+            if (other.empty) {
+                empty = true;
+            } else {
+                items.addAll(other.items);
+            }
+            return this;
+        }
+
+        Optional<List<E>> finish() {
+            return empty ? Optional.empty() : Optional.of(items);
+        }
+    }
 }
diff --git a/src/main/java/seedu/address/logic/parser/Parser.java b/src/main/java/seedu/address/logic/parser/Parser.java
index 73c6ede..cc43522 100644
--- a/src/main/java/seedu/address/logic/parser/Parser.java
+++ b/src/main/java/seedu/address/logic/parser/Parser.java
@@ -18,6 +18,7 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import seedu.address.commons.exceptions.IllegalValueException;
+import seedu.address.commons.util.ListUtil;
 import seedu.address.commons.util.StringUtil;
 import seedu.address.logic.commands.AddCommand;
 import seedu.address.logic.commands.ClearCommand;
@@ -249,18 +250,17 @@ public class Parser {
      * Returns an {@code Optional.empty()} otherwise.
      */
     private Optional<List<Integer>> parseRangedIndices(List<String> tokens) {
-        final List<Integer> indices = new ArrayList<>();
-        for (final String token : tokens) {
-            final Optional<IndexRange> indexRange = parseRangedIndex(token);
-            if (!indexRange.isPresent()) {
-                return Optional.empty();
-            }
-
-            // Adding every index from start to end (inclusive) to 'indices'.
-            IntStream.rangeClosed(indexRange.get().start, indexRange.get().end).forEach(indices::add);
-        }
+        return tokens.stream()
+            .map(this::parseRangedIndex) // Step 1
+            .collect(ListUtil.collectOptional()) // Step 2
+            .map(Parser::flattenIndexRanges); // Step 3
+    }
 
-        return Optional.of(indices);
+    private static List<Integer> flattenIndexRanges(List<IndexRange> indexRanges) {
+        return indexRanges.stream()
+                .flatMapToInt(IndexRange::toIntStream)
+                .boxed()
+                .collect(Collectors.toList());
     }
 
     /**
@@ -318,6 +318,10 @@ public class Parser {
             this.start = Math.min(start, end);
             this.end = Math.max(start, end);
         }
+
+        IntStream toIntStream() {
+            return IntStream.rangeClosed(start, end);
+        }
     }
 
 }
```
The benefit in this case of using the streams API is it makes it clear that `parseRangedIndices()` does three operations. The bad is that we need a lot of additional utility methods.

Step 2 (Convert the `List<Optional<T>>` into a `Optional<List<T>>`) seems like it would be a common operation if we are moving towards using `Optionals` for everything, so it might be good to have that as a utility function.

This is just for fun, no changes requested in this comment.

cheec:
:O so should I rebase mid-PR or just add fixup commits? @damithc @YijinL @pyokagan

cheec:
Or could I do the latter but organise the fixup commits for easier squashing later on?

pyokagan:
@cheec Follow what your assigned reviewer @YijinL says.

YijinL:
Hey @cheec for this PR you can rebase but try to leave commits up to the last review untouched.

damithc:
@cheec reminder

cheec:
@damithc sorry I've been busy, will work on it soon

src/main/java/seedu/address/logic/commands/DeleteCommand.java
cheec: @damithc isn't that the case? or am i missing something?

src/main/java/seedu/address/commons/util/ListUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Returns a sublist of items from {@code list} specified by {@code indices} with an {@code offset}.<br>
+     * <ul>
+     *   <li>{@code indices} come in an {@code Integer} array.</li>
+     *   <li>{@code offset} is applied before item extraction.</li>
+     *   <li>Original list order is maintained in the extracted list.</li>
+     *   <li>Duplicate indices are ignored.</li>
+     * </ul>
+     */
+    public static <T> List<T> sublistFromIndices(List<T> list, int offset, Integer... indices) {
```
cheec: i was going for Java's own .sublist(), i guess i'll make it `sublist(...)` then

src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -148,5 +151,48 @@ public void getDetails_nullGiven_assertionError() {
...
+        assertEquals(StringUtil.toIndexedListString(items),
+                "1. " + items.get(0).toString() + "\n"
+                + "2. " + items.get(1).toString() + "\n"
+                + "3. " + items.get(2).toString());
+    }
+
+    @Test
+    public void toIndexListString_emptyList_emptyResult() {
+        List<Object> emptyList = new ArrayList<>();
+        String emptyString = "";
```
cheec: @pyokagan  ;^)

src/test/java/seedu/address/logic/LogicManagerTest.java
cheec: > Maybe can add to the header comment for LogicManagerTest what we don't need to check?

@YijinL is this directed to me or to Prof @damithc?

src/main/java/seedu/address/commons/util/ListUtil.java
damithc: Yes, `subList(...)` is fine too.

src/test/java/seedu/address/logic/LogicManagerTest.java
damithc: Yes @YijinL  (and @cheec ) an explanatory comment is a good idea.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
damithc: Sorry @cheec , the comment is meant for the previous line `63`. 
`deletePersonsFrom(UnmodifiableObservableList<ReadOnlyPerson> targetList)` feels 'incomplete' because it doesn't specify which person to delete.

cheec: oh okay, noted

cheec:
@YijinL btw master branch has quite a few changes, some of them are relevant to this branch e.g. the conflict and `CollectionUtil` tests. Is it ok if I merge them into this branch and apply my new commits after them?

damithc:
@cheec yes, you should merge master

cheec:
Ready for review again @YijinL

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -72,10 +71,4 @@ public CommandResult execute() {
         return personsToDelete;
     }
```
YijinL: I strongly recommend adding and using a `deletePersons` method at the model level instead.

This preserves abstraction levels and you can rely on the `Collection.removeAll` optimisations. (right now we use an array list, so removing n items from an m size list takes O(mn) time, potentially quadratic)

damithc:
@pyokagan sorry I missed your detailed response on 'merge vs rebase'. We'll probably need a two-tiered approach given that some of contributors (i'm thinking of all our OSS projects, not just SE-EDU) are very new to Git etc. 
1. tier one: novice contributors, smaller issues, commit at will, review whole diff, merge preferred over rebase, squash at merge to get rid of all the ugliness of commits
2. tier two: more experienced devs, bigger issues, break PR into logical commits, review by commits, rebase preferred

Do you have any experience with reviewable.io? If it is good and not invasive (i.e. some members can use it while others don't), we can encourage reviewers to use it.

pyokagan:
@damithc 
> Do you have any experience with reviewable.io?

No, I don't. Just heard good things about it (Well, GitHub's review interface is a pretty low bar ;-) ). I could try it out on some PRs.

damithc:
@pyokagan 
>  I could try it out on some PRs.

I'll do the same. I've enabled it for addressbook levels 1-4

cheec:
@YijinL Ready for review

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -120,6 +120,14 @@ public boolean removePerson(ReadOnlyPerson key) throws UniquePersonList.PersonNo
         }
     }
 
+    public boolean removePersons(Collection<ReadOnlyPerson> keys) throws UniquePersonList.PersonNotFoundException {
```
damithc: Needs header comment. Also, need to clarify what happens if some persons exist in the list but others don't. One option is to remove those found andreturn an Optional list indicating those that could not be found.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -120,6 +120,14 @@ public boolean removePerson(ReadOnlyPerson key) throws UniquePersonList.PersonNo
...
+        if (persons.removeAll(keys)) {
+            return true;
+        } else {
+            throw new UniquePersonList.PersonNotFoundException();
+        }
```
damithc: It's odd to return true for success but throw an exception for failure (i.e. never returns false).

cheec: yes, I found it odd too, I adapted the implementation from removePerson() and talked about it [here](https://github.com/se-edu/addressbook-level4/pull/202#issuecomment-269491100)

src/test/java/seedu/address/commons/util/IntegerUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertApplyOffsetSuccess(2, 1, 5, 0);
+        assertApplyOffsetSuccess(5, -1, 4, 3);
+
+        // negative offset
+        assertApplyOffsetSuccess(-3, -10, 0, 2);
+        assertApplyOffsetSuccess(-1, 0, 0, 0);
+
+        // zero offset
+        assertApplyOffsetSuccess(0, -1, -6, -8);
+        assertApplyOffsetSuccess(0, 1, 0 , -1);
```
damithc: I think you should specify the expected output too. 
e.g. `assertApplyOffsetSuccess(1, new int[]{1,0,-1}, new int[]{2,1,0}`

Calculating the expected output in the utility method has the risk of that algorithm being equally flawed (i.e. a bug in test cancelling out bug in SUT)

src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -148,5 +152,46 @@ public void getDetails_nullGiven_assertionError() {
...
+    @Test
+    public void toIndexedListString_validList_correctResult() {
+        List<Object> items = new ArrayList<>();
+        // testing different object types
+        items.add("bob bob bob bobbobob");
+        items.add(new Integer(1));
+        items.add(new int[] {1, 2});
+        String expectedString = "1. " + items.get(0).toString() + "\n"
+                + "2. " + items.get(1).toString() + "\n"
+                + "3. " + items.get(2).toString();
```
damithc: Use minimal logic when specifying expected values. Give the full string literal instead? That way it is easy to see what the expected value is.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -288,6 +308,16 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
...
+
+        // no indices around range indicator
+        assertCommandBehavior("delete -", expectedMessage);
+
+        // range indicator only partially surrounded
+        assertCommandBehavior("delete 5-", expectedMessage);
+        assertCommandBehavior("delete -5", expectedMessage);
+
+        // excessive range indicators
+        assertCommandBehavior("delete - - --- -", expectedMessage);
```
damithc: These should be tested at command level. LogicManager treats all invalid delete commands the same way. i.e. all these cases are in the same equivalence class, as far as LogicManager is concerned.

src/test/java/seedu/address/logic/LogicManagerTest.java
damithc: @cheec now that we have unit tests for DeleteCommand, don't add more test cases here. If at all, remove duplicated test cases from here. But include a comment in the test method `\\ other variations are tested in DeleteCommandTest`

src/main/java/seedu/address/commons/util/CollectionUtil.java
```diff
@@ -2,33 +2,27 @@
...
+     * Returns true if {@code items} itself or any element of {@code items} is null.
      */
     public static boolean isAnyNull(Object... items) {
-        for (Object item : items) {
-            if (item == null) {
-                return true;
-            }
-        }
-        return false;
+        return items == null || Stream.of(items).anyMatch(Objects::isNull);
```
cheec: yes it's possible, if you pass in a null `Object[]` reference. E.g. see [here](https://github.com/se-edu/addressbook-level4/blob/22623403b72e49a360683c25384d07b767f154c7/src/main/java/seedu/address/model/tag/UniqueTagList.java#L33) (if tags == null) and [here](http://ideone.com/BZOHU4)

src/test/java/seedu/address/logic/LogicManagerTest.java
cheec: @damithc, noted

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,134 @@
...
+        String expectedFeedback = StringUtil.toIndexedListString(expectedPersonsDeleted);
+
         //confirm the result message is correct
-        assertResultMessage(String.format(MESSAGE_DELETE_PERSON_SUCCESS, personToDelete));
+        assertResultMessage(
+                String.format(MESSAGE_DELETE_PERSON_SUCCESS, expectedPersonsDeleted.size(), expectedFeedback));
+    }
+
+    /** Returns a copy of {@code currentList} without {@code personsToDelete}. */
+    private TestPerson[] getDeletedRemainder(TestPerson[] currentList, List<TestPerson> personsToDelete) {
```
cheec: @damithc I was thinking it'll be misleading as `deletePersons` might imply that `currentList`'s data is mutated, but in actual fact it's not. `getDeletedRemainder` simply returns a new list holding the leftover persons not specified by `personsToDelete` from `currentList`.

YijinL:
@cheec the boolean return is likely a remnant of one of the old versions where exceptions were not used. It can be a void function.

src/test/java/guitests/DeleteCommandTest.java
damithc: I see. In that case, 
`getRemainder(TestPerson[] currentList, List<TestPerson> personsToExclude)`?

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,134 @@
...
+        assertDeleteSuccess("4-5 4-6 5-6 4-4 5-5 6-6", expectedRemainder, expectedPersonsDeleted);
+    }
+
+    @Test
+    public void delete_excessiveWhitespaces() {
+        TestPerson[] currentList = td.getTypicalPersons();
+        List<TestPerson> expectedPersonsDeleted = ListUtil.subList(Arrays.asList(currentList), 1, 2);
+        TestPerson[] expectedRemainder = getDeletedRemainder(currentList, expectedPersonsDeleted);
+        assertDeleteSuccess("         2      3      ", expectedRemainder, expectedPersonsDeleted);
+    }
```
cheec: @damithc  with this refactoring, should I then squash all the delete testcases into one test method like [here](https://github.com/se-edu/addressbook-level4/blob/c9281baefb6952d3eb8c9259372705d34ce1fbe2/src/test/java/seedu/address/commons/util/CollectionUtilTest.java#L16-L44)? Seeing that if I don't they would all just be one liners. e.g.

```java
    @Test
    public void delete_excessiveWhitespaces() {
        assertDeleteSuccess("         2      3      ", 1, 2);
    }
```

src/test/java/guitests/DeleteCommandTest.java
cheec: sounds good!

src/test/java/guitests/DeleteCommandTest.java
cheec: hmm nevermind, I think it's okay as above

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,134 @@
...
+
+    @Test
+    public void delete_invalidIndex() {
+        TestPerson[] currentList = td.getTypicalPersons();
         commandBox.runCommand("delete " + currentList.length + 1);
         assertResultMessage("The person index provided is invalid");
-
     }
```
cheec: @damithc `4-2` is already tested here I believe.

cheec:
Ready for review

docs/UserGuide.md
```diff
@@ -147,8 +168,8 @@ There is no need to save manually.
 
 * **Clear** : `clear`
 
-* **Delete** : `delete INDEX` <br>
-   e.g. `delete 3`
+* **Delete** : `delete INDICES [INDICES]...` <br>
```
damithc: Indices is plural, so once is enough

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,137 @@
...
+
+    @Test
+    public void delete_overlappingRanges() {
+        assertDeleteSuccess("4-5 4-6 5-6 4-4 5-5 6-6", 3, 4, 5);
     }
 
+    @Test
+    public void delete_excessiveWhitespaces() {
+        assertDeleteSuccess("         2      3      ", 1, 2);
+    }
```
damithc: It is fine to combine some of these tests into one method if there are no dependencies between them.

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,137 @@
...
+        commandBox.runCommand("delete " + deleteArgs);
+        assertResultMessage(expectedMessage);
     }
 
+    /** Returns a copy of {@code currentList} but without {@code personsToDelete}. */
+    private TestPerson[] getRemainder(TestPerson[] currentList, List<TestPerson> personsToDelete) {
+        return TestUtil.removePersonsFromList(
+                currentList, personsToDelete.toArray(new TestPerson[personsToDelete.size()]));
+    }
 }
```
damithc: Sorry, I didn't notice this is a GUI test. You should minimize GUI tests because they are slow to run. GUI behavior does not depend on how you use indices in the command. One successful deletion and one unsuccessful one should do at GUI level? 

Other cases should be tested at lower levels. As we don't have a unit test for DeleteCommand class, the lowest level test we have is `LogicManagerTest` level. So need to move some of these test cases there. Previously I was asking you to cut down on LogicManagerTest cases because I mistook this class for a lower level unit test. Sorry about that.

src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -148,5 +153,46 @@ public void getDetails_nullGiven_assertionError() {
...
+
+    @Test
+    public void toIndexedListString_validList_correctResult() {
+        Collection<Object> items = new ArrayList<>();
+        // testing different object types
+        items.add("bob bob bob bobbobob");
+        items.add(new Integer(1));
+        items.add(Arrays.asList(5, 6, 7));
+        String expectedString = "1. bob bob bob bobbobob\n"
+                + "2. 1\n"
```
damithc: Indent to align with the first line so that it looks like the final output.

src/main/java/seedu/address/commons/util/IntegerUtil.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods related to Integers
+ */
+public class IntegerUtil {
+
+    /** Applies an integer {@code offset} to each integer element of {@code source}. */
```
cheec: I was trying to be consistent with the method name, but that works too

src/test/java/guitests/DeleteCommandTest.java
cheec: noted, no worries!

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -55,6 +56,24 @@ public boolean remove(ReadOnlyPerson toRemove) throws PersonNotFoundException {
         return personFoundAndDeleted;
     }
 
+    /**
+     * Removes all persons specified by {@code personsToRemove} from this list.<br>
+     *
+     * Returns an {@code Optional.empty()} if all persons in {@code personsToRemove} exist in this list.<br>
+     *
+     * Otherwise, returns an {@code Optional} of persons that do not exist in this list.
```
cheec: Thought it would aid readability - will remove

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -120,6 +120,17 @@ public boolean removePerson(ReadOnlyPerson key) throws UniquePersonList.PersonNo
         }
     }
 
+    /**
+     * Removes all persons specified by {@code keys} from this {@code AddressBook}'s {@code UniquePersonList}.<br>
+     *
+     * Returns an {@code Optional.empty()} if all persons in {@code keys} exist in this list.<br>
+     *
+     * Otherwise, returns an {@code Optional} of persons that do not exist in this list.
+     */
```
cheec: Is it ok if I still return the Optionals instead of an exception? I think it would provide more information as to who's missing. But yes i'll ensure removals don't happen if any person is missing. @damithc

src/test/java/guitests/DeleteCommandTest.java
cheec: > As we don't have a unit test for DeleteCommand class
Should I make one? @damithc

cheec: actually nevermind, probably out of scope of this PR

src/main/java/seedu/address/model/AddressBook.java
damithc: If you try to do that, the method might end up doing two things.  i.e. delete, query
Missing person info can be packaged into the exception?

src/main/java/seedu/address/commons/util/IntegerUtil.java
damithc: Change both?

src/main/java/seedu/address/model/AddressBook.java
cheec: I see, ok

src/main/java/seedu/address/commons/util/IntegerUtil.java
cheec: How about just `offset(..)`?

damithc: @cheec 

> How about just .offset(..)?

We prefer method names to be verb phrases.

You can also keep it as it is if you think `apply` implies `add` and not `deduct`. You can still delete the `an integer` in `Applies an integer` because that part is redundant.

cheec: offset can be a verb too if im not wrong.

> You can also keep it as it is if you think apply implies add and not deduct. You can still delete the an integer in Applies an integer because that part is redundant.

Im ok with both, i just thought .offset(..) looked cleaner :P

damithc: @cheec 

> offset can be a verb too if im not wrong.

Yes, but when it can be one of two things, it's ambiguous (i.e. it looks like a write op as well as a read op).  Can keep the current one.

cheec: ah ok

src/main/java/seedu/address/model/AddressBook.java
cheec: > If you try to do that, the method might end up doing two things. i.e. delete, query

Sorry, just to clarify, should i still return the actual missing persons in some form of way e.g. an exception? Or should the method simply flag when a removal is not possible?

damithc: I prefer returning void and including missing person info in the exception.

cheec:
@damithc ready for review

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -288,28 +309,163 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
...
+
+        List<Person> testPersons = helper.generatePersonList(5);
+        AddressBook expectedAB = helper.generateAddressBook(testPersons);
+        helper.addToModel(model, testPersons);
+
+        String expectedFeedback = deleteHelper.deleteAtIndices(expectedAB, testPersons, 2, 3, 4);
+        assertCommandBehavior("delete       3          4      5    ",
+                expectedFeedback,
+                expectedAB,
+                expectedAB.getPersonList());
```
damithc: Instead of these two statements, it's cleaner and more readable to see something like this:
```java
assertDeleteSuccess("delete       3          4      5    ", 2, 3, 4);
```
It clearly says `assert that "delete       3          4      5    " results in the deletion of items 2, 3, 4` without any other noise.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -288,28 +309,163 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
...
+        assertCommandBehavior("delete 1-3",
+                expectedFeedback,
+                expectedAB,
+                expectedAB.getPersonList());
 
-        assertCommandBehavior("delete 2",
-                String.format(DeleteCommand.MESSAGE_DELETE_PERSON_SUCCESS, threePersons.get(1)),
+        // deleting range in descending order
+        expectedFeedback = deleteHelper.deleteAtIndices(expectedAB, testPersons, 1, 2, 3);
+        assertCommandBehavior("delete 4-2",
```
cheec: it's an intended feature since the start of this PR, no reason why ranges should be only ascending in my opinion

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -507,4 +663,29 @@ Person generatePersonWithName(String name) throws Exception {
...
+     * Utility class for deleting
+     */
+    private class DeleteHelper {
+
+        /**
+         * Deletes persons specified by {@code targetIndices} from {@code expectedAB} and {@code testPersons}.<br>
+         * Returns the expected feedback from deleting.<br>
+         * Persons are supplied by {@code testPersons}.
+         */
+        String deleteAtIndices(AddressBook expectedAB, List<Person> testPersons, Integer... targetIndices)
```
cheec: noted

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,40 +162,123 @@ private Command prepareSelect(String args) {
...
+        public final int start;
+        public final int end;
+
+        private IndexRange(int start, int end) {
+            assert start > 0 && end > 0;
+            // Forcing ascending order.
+            this.start = Math.min(start, end);
+            this.end = Math.max(start, end);
+        }
     }
```
cheec: sure, there are actually quite a few issues i uncovered from this PR, ill add them to the tracker once this is merged.

cheec:
@damithc ready for review

docs/UserGuide.md
```diff
@@ -83,14 +83,16 @@ Examples:
...
+Deletes the specified persons from the address book. Irreversible.<br>
+Format: `delete INDICES`
 
-> Deletes the person at the specified `INDEX`. <br>
-> The index refers to the index number shown in the most recent listing.<br>
-> The index **must be a positive integer** 1, 2, 3, ...
+> * Deletes the persons at the specified `INDICES`.<br>
+> * `INDICES` can only be integers or integer ranges separated by spaces.<br>
+> * Integer ranges are specified using `-`. <br>
+> * Only positive integers are accepted.<br>
```
damithc: Resolved.

docs/UserGuide.md
```diff
@@ -83,14 +83,16 @@ Examples:
...
+> * If any of the above is not satisfied, the command will not execute at all.
```
damithc: Resolved.

src/main/java/seedu/address/logic/parser/Parser.java
damithc: Resolved.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,12 +113,20 @@ private void syncMasterTagListWith(UniquePersonList persons) {
...
+     * @throws PersonNotFoundException if any person can't be found in the list.
+     */
+    public void removePersons(ReadOnlyPerson... keys) throws PersonNotFoundException {
+        removePersons(Arrays.asList(keys));
+    }
+
+    /**
+     * Removes all persons specified by {@code keys} from this list.<br>
+     * @throws PersonNotFoundException if any person can't be found in the list.
+     */
```
damithc: Resolved

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -42,17 +44,19 @@ public void add(Person toAdd) throws DuplicatePersonException {
     }
 
     /**
-     * Removes the equivalent person from the list.
-     *
-     * @throws PersonNotFoundException if no such person could be found in the list.
+     * Removes all persons specified by {@code personsToRemove} from this list.<br>
+     * @throws PersonNotFoundException if any person can't be found in this list.
```
damithc: `@throws PersonNotFoundException without deleting any persons if any of the {@code personsToRemove} can't be found in this list.`

src/test/java/seedu/address/logic/LogicManagerTest.java
damithc: It's OK to have that feature.  Is it mentioned in the documentation? Perhaps one the given examples can be a descending range.

src/main/java/seedu/address/model/person/UniquePersonList.java
damithc: Carried forward to next review.

src/test/java/seedu/address/logic/LogicManagerTest.java
damithc: Resolved.

src/test/java/seedu/address/logic/LogicManagerTest.java
cheec: yep it's already mentioned

src/test/java/seedu/address/logic/LogicManagerTest.java
damithc: Resolved.

src/test/java/guitests/DeleteCommandTest.java
```diff
@@ -1,54 +1,63 @@
...
 
-        //confirm the list now contains all previous persons except the deleted person
+        //confirm the list now contains all previous persons except the deleted
         assertTrue(personListPanel.isListMatching(expectedRemainder));
+    }
 
-        //confirm the result message is correct
-        assertResultMessage(String.format(MESSAGE_DELETE_PERSON_SUCCESS, personToDelete));
+    /** Runs the delete command specified by {@code deleteArgs} and confirms result matches {@code expectedMessage}. */
+    private void assertDeleteCommandResult(String deleteArgs, String expectedMessage) {
```
damithc: Is this used only for testing delete failures? In that case, name it `asserDeleteFailure` to match the previous method?

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -288,28 +309,108 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
...
+        assertCommandBehavior("delete -5", expectedMessage);
+
+        // excessive range indicators
+        assertCommandBehavior("delete - - --- -", expectedMessage);
     }
 
     @Test
     public void execute_deleteIndexNotFound_errorMessageShown() throws Exception {
-        assertIndexNotFoundBehaviorForCommand("delete");
+        assertIndexNotFoundBehaviorForCommand("delete 5", 3);
```
damithc: As here we have multiple cases inside one method, add a comment to explain the rationale for each specific case?

src/test/java/guitests/DeleteCommandTest.java
cheec: It's used in assertDeleteSuccess too

cheec:
@damithc ready for review again

cheec:
ready for review @damithc

src/test/java/guitests/DeleteCommandTest.java
damithc: > But shouldn't method names only describe exactly what it's doing regardless of where it's used and the context it's used in?

Yes. I meant 'does this method tests delete failures only'.

I didn't notice it's being used in the other method. Can keep.

damithc:
over to you @pyokagan :-)

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,40 +162,123 @@ private Command prepareSelect(String args) {
...
-        String index = matcher.group("targetIndex");
-        if (!StringUtil.isUnsignedInteger(index)) {
+        final Optional<Integer> startIndex = parseIndex(matcher.group("rangeStart"));
+        final Optional<Integer> endIndex = parseIndex(matcher.group("rangeEnd"));
+        if (!startIndex.isPresent() || !endIndex.isPresent()) {
             return Optional.empty();
         }
-        return Optional.of(Integer.parseInt(index));
 
+        return Optional.of(new IndexRange(startIndex.get(), endIndex.get()));
```
cheec: is there a reason why the limit for token.split(PERSON_INDEX_RANGE_INDICATOR, -1); is -1 and not 0 or even just .split(range_indicator)? just curious

cheec: upon your first review, i too thought of doing what you suggested, wrapping the emulation within parseRangeIndex() itself, however, in my mind it contradicted with what you also said - that a method should only have a single responsibility. That's why i left the emulation job to some other place, in this case it was parseIndices().

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,40 +162,123 @@ private Command prepareSelect(String args) {
...
+        // keywords delimited by whitespace
+        final String[] keywords = matcher.group("keywords").split("\\s+");
+        final Set<String> keywordSet = new HashSet<>(Arrays.asList(keywords));
+        return new FindCommand(keywordSet);
+    }
+
+    /*
+     * =========================================================================
+     *                             Helper Methods
+     * =========================================================================
```
cheec: boils down to personal pref i guess, i can take it off

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,40 +162,123 @@ private Command prepareSelect(String args) {
...
+     * Returns a {@code Collection} of indices extracted from {@code tokens} if each token contains either:<br>
+     * - a positive unsigned integer i.e. a non-ranged index<br>
+     * - or positive unsigned integers surrounding a {@code PERSON_INDEX_RANGE_INDICATOR}
+     *          i.e. a ranged index<br>
+     *
      * Returns an {@code Optional.empty()} otherwise.
      */
-    private Optional<Integer> parseIndex(String command) {
-        final Matcher matcher = PERSON_INDEX_ARGS_FORMAT.matcher(command.trim());
+    private Optional<Collection<Integer>> parseIndices(Collection<String> tokens) {
```
cheec: > A method must only do one thing and one thing well, so these two things need to be split up.

agreed :D:D:D but then it kinda depends how loose/broad you define "one thing" doesn't it?

pyokagan: @cheec 
> it kinda depends how loose/broad you define "one thing" doesn't it?

Yes, it depends on the *one* responsibility given to the method. In this case, there is one definite responsibility given to a parser function which your design and naming does not adhere to.

Now, here are the rules of the Parser:

1. A parser is a function that transform a string into an object of type `T`.

2. The parser is named according to this type `T`. E.g., for `parseIndex()`, this type `T` is `Integer`. For `parseRangedIndex()`, this type `T` is `IndexRange`. (Note: some methods in `Parser.java` are currently named inconsistently, e.g. `prepareAdd()` could probably be better named as `parseAddCommand()`)

3. **The one responsibility of a parser function**: Each parser has an associated format which is accepts as input. For example, `parseIndex()` parses the "index" format as input and returns an `Integer`. The `parseRangedIndex()` method parses the "ranged index" format as input and returns an `IndexRange`.

If you take a look at all the methods currently in `Parser.java`, you will see they all follow the above rules. E.g. `prepareAdd()` parses the "add command" format as input and returns an `AddCommand`.

You should now see why `parseIndices()` violates the above rules. It claims it parses "indices", yet it also parses the "ranged index" format.

So, it's fine if you want the "ranged index" format to not be a superset of the "index" format. However, you *must* then introduce a `parseRangedOrNormalIndex()` function which accepts both a ranged index or normal index, and returns the correct type `T`. You can then introduce an additional method `parseRangedOrNormalIndices()` which maps the `parseRangedOrNormalIndex()` over a list and returns a `List<T>`.

cheec: noice i see, thanks for explaining Paul

pyokagan:
githubpls

GitHub why

cheec:
@pyokagan ready for review by the way, oops

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +155,70 @@ private Command prepareSelect(String args) {
...
+
+            indexRanges.add(indexRange.get());
+        }
+        return Optional.of(indexRanges);
+    }
+
+    /**
+     * Returns an {@code IndexRange} extracted from {@code token} if it contains either positive
+     * unsigned integers surrounding a {@code PERSON_INDEX_RANGE_INDICATOR}, or positive
+     * unsigned integers, i.e. 1-2 or 10.
```
pyokagan: Doesn't `unsigned` imply `positive`?

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -140,11 +155,70 @@ private Command prepareSelect(String args) {
...
+     *
+     * If the latter is encountered, an {@code IndexRange} with the same start and end will be used
+     * to represent it. e.g. 5 -> 5-5.
+     *
+     * Returns an {@code Optional.empty()} otherwise.
+     */
+    private Optional<IndexRange> parseRangedIndex(String token) {
+        final String[] components = token.split(PERSON_INDEX_RANGE_INDICATOR, -1);
```
pyokagan: Maybe add a comment to explain what the `-1` does.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -153,27 +227,50 @@ private Command prepareSelect(String args) {
...
+            this.end = Math.max(start, end);
         }
 
-        // keywords delimited by whitespace
-        final String[] keywords = matcher.group("keywords").split("\\s+");
-        final Set<String> keywordSet = new HashSet<>(Arrays.asList(keywords));
-        return new FindCommand(keywordSet);
+        /**
+         * Returns all intermediate values of this index range's start and end (inclusive) in a {@code List}.<br>
+         * E.g. for an {@code IndexRange}, {@code [2, 6]}.<br>
```
pyokagan: Hrm, not very obvious that `[2, 6]` refers to the start and end of the `IndexRange`, given that you use the same syntax in the next line to represent a list.

Maybe just use plain prose, something like:
`for an IndexRange with start 2 and end 6`.

Do we really need to `{@code}` all class names?

src/main/java/seedu/address/logic/parser/Parser.java
cheec: adapted from existing comments but yes it does

src/main/java/seedu/address/logic/parser/Parser.java
cheec: >Do we really need to {@code} all class names?

I was following Oracle's Javadoc style guidelines. It's also quite useful when devs mouseover a method and it highlights in the Javadoc tooltip that this is a class etc.

cheec:
@pyokagan ready for review

cheec:
This PR is continued [here](https://github.com/se-edu/addressbook-level4/pull/248).

damithc:
Good work @cheec The new PR should be able carry this PR through the last stretch.

PR #208, opened 2016-12-10
Construct BrowserPanel in FXML #186

### LINKED ISSUE #186: Construct BrowserPanel in FXML

pyokagan:
Fixes #186 

The DeveloperGuide states that the layout of individual UiParts are
specified in matching `.fxml` files. e.g. The layout of
`MainWindow.java` is specified in `MainWindow.fxml`.

While this is true for almost all UiParts, there is one oddball UiPart
that does not have a corresponding `.fxml` file -- BrowserPanel.

For consistency with the rest of the UiParts, let's construct the
BrowserPanel using FXML rather than using code.

(This PR was previously #187, but I'm recreating it because the original repo got deleted)


src/main/resources/view/BrowserPanel.fxml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.layout.AnchorPane?>
+<?import javafx.scene.web.WebView?>
+
+<AnchorPane xmlns:fx="http://javafx.com/fxml/1" fx:controller="seedu.address.ui.BrowserPanel">
+    <WebView fx:id="browser"/>
```
yl-coder: You can apply the anchorpane boundary alignment here actually.

pyokagan: Yeah, by doing something like this:
```diff
diff --git a/src/main/java/seedu/address/ui/BrowserPanel.java b/src/main/java/seedu/address/ui/BrowserPanel.java
index 2ac2a56..710a34d 100644
--- a/src/main/java/seedu/address/ui/BrowserPanel.java
+++ b/src/main/java/seedu/address/ui/BrowserPanel.java
@@ -42,7 +42,6 @@ public class BrowserPanel extends UiPart {
         BrowserPanel browserPanel = UiPartLoader.loadUiPart(placeholder, new BrowserPanel());
         placeholder.setOnKeyPressed(Event::consume); // To prevent triggering events for typing inside the
                                                      // loaded Web page.
-        FxViewUtil.applyAnchorBoundaryParameters(browserPanel.browser, 0.0, 0.0, 0.0, 0.0);
         placeholder.getChildren().add(browserPanel.browser);
         return browserPanel;
     }
diff --git a/src/main/resources/view/BrowserPanel.fxml b/src/main/resources/view/BrowserPanel.fxml
index 21c41d4..025267a 100644
--- a/src/main/resources/view/BrowserPanel.fxml
+++ b/src/main/resources/view/BrowserPanel.fxml
@@ -4,5 +4,5 @@
 <?import javafx.scene.web.WebView?>
 
 <AnchorPane xmlns:fx="http://javafx.com/fxml/1" fx:controller="seedu.address.ui.BrowserPanel">
-    <WebView fx:id="browser"/>
+    <WebView fx:id="browser"  AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="0.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0"/>
 </AnchorPane>
```
which would make `BrowserPanel.fxml` look more complicated. Also, all the other `UiParts` use `applyAnchorBoundaryParameters()`, so for this PR it may be better not to deviate from the norm for now.

imo, I'd much rather we use one of the other panes (e.g. `StackPane`, maybe even `HBox` or `VBox`) on the basis that they do not require us to add all these anchor alignment boilerplate everywhere.

AFAIK, I think `AnchorPane` also stacks its children as well?

yl-coder: AnchorPane doesn't really stack its children. Unless u set all its children to have the same alignment.

PR #210, opened 2016-12-11
Support F1 accelerator even when the focus is not in the list #149

### LINKED ISSUE #149: Support F1 accelerator even when the focus is not in the list
Currently, it works only if the focus is in the person list

chao1995:
Fixes #149 

This issue may be caused by a JavaFX bug, as suggested here https://bugs.openjdk.java.net/browse/JDK-8131666.

It's because TextField and TextAreas in CommandBox and ResultDisplay will consume F1 keyevents. As a result, the accelerators will not be fired up.

The discussion in the post shows that the bug will only be fixed in later version of JDK (though it's openJDK but I guess it should be similar case for JDK itself.)

Thus in this PR, what I implemented is just a workaround.

- Explicitly bind F1 keyevents to open help window for CommandBox TextField and ResultDisplay TextArea using `addEventFilter`
- Remove `personListPanel.clickOnListView()` in `HelpWindowTest` because now the focus doesn't have to be in the person list for F1 accelerator to work.

Ready for review

### DIFF
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -1,10 +1,13 @@
 package seedu.address.ui;
 
+import javafx.event.ActionEvent;
 import javafx.fxml.FXML;
 import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.MenuItem;
+import javafx.scene.control.TextInputControl;
 import javafx.scene.input.KeyCombination;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.VBox;
 import javafx.stage.Stage;
@@ -104,7 +107,37 @@ public class MainWindow extends UiPart {
     }
 
     private void setAccelerators() {
-        helpMenuItem.setAccelerator(KeyCombination.valueOf("F1"));
+        setAccelerator(helpMenuItem, KeyCombination.valueOf("F1"));
+    }
+
+    /**
+     * Sets the accelerator of a MenuItem.
+     * @param keyCombination the KeyCombination value of the accelerator
+     */
+    private void setAccelerator(MenuItem menuItem, KeyCombination keyCombination) {
+        menuItem.setAccelerator(keyCombination);
+
+        /*
+         * TODO: the code below can be removed once the bug reported here
+         * https://bugs.openjdk.java.net/browse/JDK-8131666
+         * is fixed in later version of SDK.
+         *
+         * According to the bug report, TextInputControl (TextField, TextArea) will
+         * consume function-key events. Because CommandBox contains a TextField, and
+         * ResultDisplay contains a TextArea, thus some accelerators (e.g F1) will
+         * not work when the focus is in them because the key event is consumed by
+         * the TextInputControl(s).
+         *
+         * For now, we add following event filter to capture such key events and open
+         * help window purposely so to support accelerators even when focus is
+         * in CommandBox or ResultDisplay.
+         */
+        rootLayout.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+            if (event.getTarget() instanceof TextInputControl && keyCombination.match(event)) {
+                menuItem.getOnAction().handle(new ActionEvent());
+                event.consume();
+            }
+        });
     }
 
     void fillInnerParts() {
--- a/src/test/java/guitests/HelpWindowTest.java
+++ b/src/test/java/guitests/HelpWindowTest.java
@@ -4,12 +4,13 @@ import guitests.guihandles.HelpWindowHandle;
 import org.junit.Test;
 
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 public class HelpWindowTest extends AddressBookGuiTest {
 
     @Test
     public void openHelpWindow() {
-
+        //use accelerator
         commandBox.clickOnTextField();
         assertHelpWindowOpen(mainMenu.openHelpWindowUsingAccelerator());
 
@@ -19,14 +20,23 @@ public class HelpWindowTest extends AddressBookGuiTest {
         personListPanel.clickOnListView();
         assertHelpWindowOpen(mainMenu.openHelpWindowUsingAccelerator());
 
+        browserPanel.clickOnWebView();
+        assertHelpWindowNotOpen(mainMenu.openHelpWindowUsingAccelerator());
+
+        //use menu button
         assertHelpWindowOpen(mainMenu.openHelpWindowUsingMenu());
 
+        //use command
         assertHelpWindowOpen(commandBox.runHelpCommand());
-
     }
 
     private void assertHelpWindowOpen(HelpWindowHandle helpWindowHandle) {
         assertTrue(helpWindowHandle.isWindowOpen());
         helpWindowHandle.closeWindow();
     }
+
+    private void assertHelpWindowNotOpen(HelpWindowHandle helpWindowHandle) {
+        assertFalse(helpWindowHandle.isWindowOpen());
+    }
+
 }
--- a/src/test/java/guitests/HelpWindowTest.java
+++ b/src/test/java/guitests/HelpWindowTest.java
@@ -10,8 +10,13 @@ public class HelpWindowTest extends AddressBookGuiTest {
     @Test
     public void openHelpWindow() {
 
-        personListPanel.clickOnListView();
+        commandBox.clickOnTextField();
+        assertHelpWindowOpen(mainMenu.openHelpWindowUsingAccelerator());
 
+        resultDisplay.clickOnTextArea();
+        assertHelpWindowOpen(mainMenu.openHelpWindowUsingAccelerator());
+
+        personListPanel.clickOnListView();
         assertHelpWindowOpen(mainMenu.openHelpWindowUsingAccelerator());
 
         assertHelpWindowOpen(mainMenu.openHelpWindowUsingMenu());
--- a/src/test/java/guitests/guihandles/BrowserPanelHandle.java
+++ b/src/test/java/guitests/guihandles/BrowserPanelHandle.java
@@ -1,0 +1,25 @@
+package guitests.guihandles;
+
+import guitests.GuiRobot;
+import javafx.stage.Stage;
+import seedu.address.TestApp;
+
+/**
+ * A handler for the BrowserPanel of the UI
+ */
+public class BrowserPanelHandle extends GuiHandle {
+
+    private static final String BROWSER_ID = "#browser";
+
+    public BrowserPanelHandle(GuiRobot guiRobot, Stage primaryStage) {
+        super(guiRobot, primaryStage, TestApp.APP_TITLE);
+    }
+
+    /**
+     * Clicks on the WebView.
+     */
+    public void clickOnWebView() {
+        guiRobot.clickOn(BROWSER_ID);
+    }
+
+}
--- a/src/test/java/guitests/guihandles/CommandBoxHandle.java
+++ b/src/test/java/guitests/guihandles/CommandBoxHandle.java
@@ -14,6 +14,13 @@ public class CommandBoxHandle extends GuiHandle {
         super(guiRobot, primaryStage, stageTitle);
     }
 
+    /**
+     * Clicks on the TextField.
+     */
+    public void clickOnTextField() {
+        guiRobot.clickOn(COMMAND_INPUT_FIELD_ID);
+    }
+
     public void enterCommand(String command) {
         setTextField(COMMAND_INPUT_FIELD_ID, command);
     }
--- a/src/test/java/guitests/guihandles/HelpWindowHandle.java
+++ b/src/test/java/guitests/guihandles/HelpWindowHandle.java
@@ -17,7 +17,7 @@ public class HelpWindowHandle extends GuiHandle {
     }
 
     public boolean isWindowOpen() {
-        return getNode(HELP_WINDOW_ROOT_FIELD_ID) != null;
+        return guiRobot.lookup(HELP_WINDOW_ROOT_FIELD_ID).tryQuery().isPresent();
     }
 
     public void closeWindow() {
--- a/src/test/java/guitests/guihandles/ResultDisplayHandle.java
+++ b/src/test/java/guitests/guihandles/ResultDisplayHandle.java
@@ -16,6 +16,13 @@ public class ResultDisplayHandle extends GuiHandle {
         super(guiRobot, primaryStage, TestApp.APP_TITLE);
     }
 
+    /**
+     * Clicks on the TextArea.
+     */
+    public void clickOnTextArea() {
+        guiRobot.clickOn(RESULT_DISPLAY_ID);
+    }
+
     public String getText() {
         return getResultDisplay().getText();
     }
--- a/src/test/java/guitests/AddressBookGuiTest.java
+++ b/src/test/java/guitests/AddressBookGuiTest.java
@@ -44,6 +44,7 @@ public abstract class AddressBookGuiTest {
     protected PersonListPanelHandle personListPanel;
     protected ResultDisplayHandle resultDisplay;
     protected CommandBoxHandle commandBox;
+    protected BrowserPanelHandle browserPanel;
     private Stage stage;
 
     @BeforeClass
@@ -64,6 +65,7 @@ public abstract class AddressBookGuiTest {
             personListPanel = mainGui.getPersonListPanel();
             resultDisplay = mainGui.getResultDisplay();
             commandBox = mainGui.getCommandBox();
+            browserPanel = mainGui.getBrowserPanel();
             this.stage = stage;
         });
         EventsCenter.clearSubscribers();
--- a/src/test/java/guitests/guihandles/MainGuiHandle.java
+++ b/src/test/java/guitests/guihandles/MainGuiHandle.java
@@ -29,6 +29,10 @@ public class MainGuiHandle extends GuiHandle {
         return new MainMenuHandle(guiRobot, primaryStage);
     }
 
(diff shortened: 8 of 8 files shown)


src/main/java/seedu/address/ui/CommandBox.java
```diff
@@ -40,6 +42,18 @@ public void configure(Logic logic) {
         registerAsAnEventHandler(this);
     }
 
+    /**
+     * Because JavaFX TextField consumes F1 keyevents, below is
+     * a workaround to support responding to F1 accelerators (for Menu Help).
+     */
+    public void configureF1Accelerator(Runnable callback) {
+        commandTextField.addEventFilter(KeyEvent.KEY_PRESSED, (evt) -> {
```
pyokagan: Maybe just use `event` instead of `evt`? We try to avoid abbreviations when they do not add much value.

Also, I think we don't need the brackets around `(evt)`.

src/main/java/seedu/address/ui/CommandBox.java
```diff
@@ -40,6 +42,18 @@ public void configure(Logic logic) {
...
+            if (evt.getCode() == KeyCode.F1) {
+                callback.run();
```
pyokagan: Do we need to `evt.consume()`?

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -113,6 +113,11 @@ void fillInnerParts() {
         resultDisplay = ResultDisplay.load(getResultDisplayPlaceholder());
         statusBarFooter = StatusBarFooter.load(getStatusbarPlaceholder(), config.getAddressBookFilePath());
         commandBox = CommandBox.load(getCommandBoxPlaceholder(), logic);
+
+        // To support F1 accelerator, resultDisplay and commandBox require extra configuration
+        // because they have UI elements that will consume F1 keyevents.
```
pyokagan: Maybe put a link to https://bugs.openjdk.java.net/browse/JDK-8131666 so that any reader who chances upon the code can check if this workaround is still needed or not.

Maybe prepend a `FIXME:` to this comment as well to make it clear that this is a workaround which should be fixed in the future.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -113,6 +113,11 @@ void fillInnerParts() {
...
+        resultDisplay.configureF1Accelerator(() -> handleHelp());
+        commandBox.configureF1Accelerator(() -> handleHelp());
```
pyokagan: This should probably be in `setAccelerators()`.

Also, there's quite a bit of repeated code between `CommandBox` and `ResultDisplay`.

First of all, should this filter just apply to the command box and result display? What about when we focus on the web browser pane? I think pressing F1 then should also trigger the help window.

So, to avoid repeating code and to make sure that this workaround always works, I do wonder if we should just set the event filter on the root layout, like this:
```diff
diff --git a/src/main/java/seedu/address/ui/MainWindow.java b/src/main/java/seedu/address/ui/MainWindow.java
index ba00e17..dc29b62 100644
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -4,7 +4,9 @@ import javafx.fxml.FXML;
 import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.MenuItem;
+import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyCombination;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.VBox;
 import javafx.stage.Stage;
@@ -105,6 +107,12 @@ public class MainWindow extends UiPart {
 
     private void setAccelerators() {
         helpMenuItem.setAccelerator(KeyCombination.valueOf("F1"));
+        rootLayout.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+            if (event.getCode() == KeyCode.F1) {
+                handleHelp();
+                event.consume();
+            }
+        });
     }
 
     void fillInnerParts() {
```

src/test/java/guitests/HelpWindowTest.java
```diff
@@ -10,8 +10,6 @@
     @Test
     public void openHelpWindow() {
 
-        personListPanel.clickOnListView();
-
```
pyokagan: With this change, I can see that when the person list panel is *not* in focus, the `F1` accelerator will work.

However, I cannot tell if this PR will break the case where the person list panel *is* in focus, because that test has been removed with this change.

So I think that rather than modifying this test case, we add a new test case instead.

chao1995:
@pyokagan Thanks for the review. Because I found in `BrowserPanel.java` this line:
```java
placeholder.setOnKeyPressed(Event::consume); // To prevent triggering events for typing inside the loaded Web page.
```
So I didn't add the event filter to Browser Panel. Do you think we should add to it as well?

src/main/java/seedu/address/ui/MainWindow.java
chao1995: @pyokagan I agree. The code is not DRY. Thanks for the suggestion!

src/main/java/seedu/address/ui/MainWindow.java
chao1995: @pyokagan Fixed as suggested. But instead of FIXME, codestyle check suggests using TODO.

damithc:
Regarding the browser panel, the current thinking is that the web page might have its own key bindings and we didn't want to interfere with those. So it's ok if F1 doesn't work when the focus is in the browser.

pyokagan:
@damithc says:
> Regarding the browser panel, the current thinking is that the web page might have its own key bindings and we didn't want to interfere with those. 

Ah okay. I'll just note that pressing F1 in Google Chrome will also open the help window :-). It doesn't work in firefox though. Hmm..

@chao1995 I still do think that attaching the event filter to the root node is the way to go though, as it keeps the workaround code small and without duplication.

https://bugs.openjdk.java.net/browse/JDK-8089884 states that this bug only affects `TextInputControls`, so maybe we can just restrict the workaround to them. Something like this?
```diff
diff --git a/src/main/java/seedu/address/ui/MainWindow.java b/src/main/java/seedu/address/ui/MainWindow.java
index ba00e17..71c0a6c 100644
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -4,7 +4,10 @@ import javafx.fxml.FXML;
 import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.MenuItem;
+import javafx.scene.control.TextInputControl;
+import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyCombination;
+import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.AnchorPane;
 import javafx.scene.layout.VBox;
 import javafx.stage.Stage;
@@ -105,6 +108,12 @@ public class MainWindow extends UiPart {
 
     private void setAccelerators() {
         helpMenuItem.setAccelerator(KeyCombination.valueOf("F1"));
+        rootLayout.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+            if (event.getTarget() instanceof TextInputControl && event.getCode() == KeyCode.F1) {
+                handleHelp();
+                event.consume();
+            }
+        });
     }
 
     void fillInnerParts() {
```

chao1995:
@pyokagan I agree that keeping the affected code small is a good idea, especially since this is a temporary workaround only. I have updated the commit according to your suggested code snippet. Thanks!

src/test/java/guitests/guihandles/CommandBoxHandle.java
```diff
@@ -14,6 +17,15 @@ public CommandBoxHandle(GuiRobot guiRobot, Stage primaryStage, String stageTitle
         super(guiRobot, primaryStage, stageTitle);
     }
 
+    /**
+     * Clicks on the TextField.
```
pyokagan:
```
Clicks on the command box's TextField.
```

src/test/java/guitests/guihandles/ResultDisplayHandle.java
```diff
@@ -16,6 +18,15 @@ public ResultDisplayHandle(GuiRobot guiRobot, Stage primaryStage) {
         super(guiRobot, primaryStage, TestApp.APP_TITLE);
     }
 
+    /**
+     * Clicks on the TextArea.
```
pyokagan:
```
Clicks on the result display's TextArea.
```

src/test/java/guitests/guihandles/ResultDisplayHandle.java
```diff
@@ -16,6 +18,15 @@ public ResultDisplayHandle(GuiRobot guiRobot, Stage primaryStage) {
...
+     */
+    public void clickOnTextField() {
```
pyokagan: Shouldn't this be `clickOnTextArea()`?

src/test/java/guitests/guihandles/CommandBoxHandle.java
```diff
@@ -14,6 +17,15 @@ public CommandBoxHandle(GuiRobot guiRobot, Stage primaryStage, String stageTitle
...
+     */
+    public void clickOnTextField() {
+        TextField textField = (TextField) getNode(COMMAND_INPUT_FIELD_ID);
+        Point2D point = TestUtil.getScreenMidPoint(textField);
+        guiRobot.clickOn(point.getX(), point.getY());
```
pyokagan: I'm aware that this is adapted from `PersonListPanelHandle#clickOnListView()`.

Won't
```java
guiRobot.clickOn(COMMAND_INPUT_FIELD_ID);
```
accomplish the same thing?

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -105,6 +108,28 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+         * ResultDisplay contains a TextArea, thus the F1 accerelator will not be
+         * fired up when the focus is in them because the keyevent is consumed by
+         * the TextInputControl(s).
+         *
+         * For now, we add following event filter to capture F1 keyevents and open
+         * help windowin purposely so to support F1 accelerator even when focus is
+         * in CommandBox or ResultDisplay.
+         */
+        rootLayout.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+            if (event.getTarget() instanceof TextInputControl && event.getCode() == KeyCode.F1) {
```
pyokagan: OK.

I was hoping you may have an idea on how to avoid the duplication of `KeyCode.F1` though. We may wish to add other accelerators in the future with this workaround still active, and we will have to repeat the code for each accelerator.

chao1995:
@pyokagan Hi Paul, thank you very much for the review again. I fixed the issues in the test files. Also I tried to refactor the duplicate code into a separate method, you may have some suggestions for this approach? Thanks.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +106,39 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+
+    /**
+     * This methods sets an accelerator for a menuItem. And it accpets a handler as
+     * a callback to support accelerators even when the focus is in CommandBox or
+     * ResultDisplay due to a JavaFX bug. See comments below for details.
+     *
+     * TODO: this method can be removed and setAccelerators() method should be changed
+     * accordingly once this bug reported here:
+     * https://bugs.openjdk.java.net/browse/JDK-8131666
+     * is fixed in later version of SDK.
```
pyokagan: This is giving too many specifics into how the method is implemented, which should not be the concern of developers using this method.

Something like
```
Sets the accelerator of a MenuItem.
```
should do.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +106,39 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+     */
+    private void setAccelerator(MenuItem menuItem, String keyName, Runnable handler) {
+        menuItem.setAccelerator(KeyCombination.valueOf(keyName));
+
+        /**
+         * According to the bug report, TextInputControl (TextField, TextArea) will
+         * consume function-key events. Because CommandBox contains a TextField, and
+         * ResultDisplay contains a TextArea, thus the F1 accerelator will not be
```
pyokagan:
```
s/accerelator/accelerator
```

There are other spelling errors in the PR, so I'll leave you to do a spell check with Eclipse :-)

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +106,39 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+         * fired up when the focus is in them because the keyevent is consumed by
+         * the TextInputControl(s).
+         *
+         * For now, we add following event filter to capture F1 keyevents and open
+         * help windowin purposely so to support F1 accelerator even when focus is
+         * in CommandBox or ResultDisplay.
+         */
+        rootLayout.addEventFilter(KeyEvent.KEY_PRESSED, event -> {
+            if (event.getTarget() instanceof TextInputControl && event.getCode().getName().equals(keyName)) {
```
pyokagan: Hrm, this will not handle the case where the `KeyCombinaton` for the accelerator is `Ctrl+P` or something.

Maybe we can use [`KeyCombination#match(KeyEvent)`](http://docs.oracle.com/javase/8/javafx/api/javafx/scene/input/KeyCodeCombination.html#match-javafx.scene.input.KeyEvent-)

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +106,39 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+                handler.run();
```
pyokagan: This requires us to repeat the `MenuItem`'s action handler (`handleHelp` in this case), in both the FXML file and the code.

Maybe we can just call the `MenuItem's` action handler directly? Something like:
```java
menuItem.getOnAction().handle(new ActionEvent());
```

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +106,39 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+     *
+     * TODO: this method can be removed and setAccelerators() method should be changed
+     * accordingly once this bug reported here:
+     * https://bugs.openjdk.java.net/browse/JDK-8131666
+     * is fixed in later version of SDK.
+     */
+    private void setAccelerator(MenuItem menuItem, String keyName, Runnable handler) {
+        menuItem.setAccelerator(KeyCombination.valueOf(keyName));
+
+        /**
```
pyokagan: Not a javadoc, just `/*` will do.

src/main/java/seedu/address/ui/MainWindow.java
chao1995: @pyokagan Thanks! I need this one so much!

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +107,38 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
     }
 
     private void setAccelerators() {
-        helpMenuItem.setAccelerator(KeyCombination.valueOf("F1"));
+        setAccelerator(helpMenuItem, KeyCombination.valueOf("F1"));
+    }
+
+    /*
```
pyokagan: This is a javadoc, so should use `/**` here.

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -104,7 +107,38 @@ private void configure(Stage primaryStage, String appTitle, String addressBookNa
...
+     * Sets the accelerator of a MenuItem.
+     * @param menuItem a menuItem
```
pyokagan: No need for this line since it's obvious from looking at the method signature.

chao1995:
@pyokagan @damithc fixed.

chao1995:
@pyokagan Done. Please help review. Thanks!

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -0,0 +1,25 @@
...
+import guitests.GuiRobot;
+import javafx.stage.Stage;
+import seedu.address.TestApp;
+
+/**
+ * A handler for the BrowserPanel of the UI
+ */
+public class BrowserPanelHandle extends GuiHandle {
+
+    public static final String BROWSER_ID = "#browser";
```
pyokagan: Should be `private`.

src/test/java/guitests/guihandles/HelpWindowHandle.java
```diff
@@ -17,7 +17,11 @@ public HelpWindowHandle(GuiRobot guiRobot, Stage primaryStage) {
     }
 
     public boolean isWindowOpen() {
-        return getNode(HELP_WINDOW_ROOT_FIELD_ID) != null;
+        try {
+            return getNode(HELP_WINDOW_ROOT_FIELD_ID) != null;
+        } catch (IllegalStateException ise) {
+            return false;
+        }
```
pyokagan: Huh.

The first thing I wonder is why we are checking for `null` return from `getNode()`.

Looking at [`GuiHandle#getNode()`](https://github.com/se-edu/addressbook-level4/blob/38a52f5ad5e1bb97574f38950aa3b1246cc3b469/src/test/java/guitests/guihandles/GuiHandle.java#L53), we have:
```java
return guiRobot.lookup(query).tryQuery().get();
```
So a return of `null` should be impossible because `Optional` can't store nulls.

I think these lines of `isWindowOpen()` can be replaced by:
```java
guiRobot.lookup(HELP_WINDOW_ROOT_FIELD_ID).tryQuery().isPresent()
```

chao1995:
@pyokagan Thanks! Fixed as suggested.

PR #211, opened 2016-12-14
Dummy PR to test review requests

damithc:
Please any review requests from this PR

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -61,7 +61,7 @@ public class MainApp extends Application {
 
         uiManager = new UiManager(logicManager, config, userPrefs);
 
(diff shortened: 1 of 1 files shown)


pyokagan:
@damithc fwiw I did get the following email:
> @damithc requested your review on: se-edu/addressbook-level4#211 Dummy PR to test review requests.

damithc:
Thanks for the info @pyokagan 
The problem I'm facing is sometimes I cannot seem to be able to request reviews. 
The last time I tried to add a reviewer to this dummy PR, it worked. I removed the request soon after, but I guess the alert was already sent out by that time. The problem seems to be an intermittent one, so I will be adding/removing review requests to this PR at different times. If you receive more review requests from this PR, you can safely ignore them.

PR #214, opened 2016-12-17
Allow updating addressBookPath and addressBookName at runtime. #213

### LINKED ISSUE #213: Allow updating addressBookPath and addressBookName at runtime.
As per title.

chao1995:
Fixes #213 

~Replace `ConfigUtil` with `ConfigStorage` whose functionality is similar to `JsonUserPrefsStorage` for `UserPrefs`.~

~`ConfigStorage` knows what is the current file path of the config, and provides methods for read and save the config, thus it supports updating config at runtime by calling `configStorage.saveConfig(config);` before app exits. (see `MainApp.java`)~

Changes:
- Move addressBookFilePath and addressBookName from` Config` to `UserPrefs`
- Fix JSON test data.
- Fix tests. `ConfigTest`, `ConfigUtilTest` and `JsonUserPrefsStorageTest`
- Add `setAddressBookPath(String addressBookPath)` to `Storage` interface.
- Implement the new interface method in `StorageManager`.
- Fix the initialization steps (e.g. setup `UserPrefs`, use `UserPrefs` to get addressBookFilePath instead of `Config`) in `MainApp` and `TestApp` and `MainWindow`

Ready for review.

```
Users are not able to update addressBookName and addressBookPath,
which are stored in Config, because data in Config are not meant to be
changed at runtime.

To allow modifying addressBookName and addressBookPath at runtime,
let's move them from Config to UserPrefs.
```

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -26,8 +26,12 @@ import seedu.address.model.ModelManager;
 import seedu.address.model.ReadOnlyAddressBook;
 import seedu.address.model.UserPrefs;
 import seedu.address.model.util.SampleDataUtil;
+import seedu.address.storage.AddressBookStorage;
+import seedu.address.storage.JsonUserPrefsStorage;
 import seedu.address.storage.Storage;
 import seedu.address.storage.StorageManager;
+import seedu.address.storage.UserPrefsStorage;
+import seedu.address.storage.XmlAddressBookStorage;
 import seedu.address.ui.Ui;
 import seedu.address.ui.UiManager;
 
@@ -53,7 +57,10 @@ public class MainApp extends Application {
         super.init();
 
         config = initConfig(getApplicationParameter("config"));
-        storage = new StorageManager(config.getAddressBookFilePath(), config.getUserPrefsFilePath());
+
+        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(config.getAddressBookFilePath());
+        UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
+        storage = new StorageManager(addressBookStorage, userPrefsStorage);
 
         userPrefs = initPrefs(config);
 
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -58,11 +58,10 @@ public class MainApp extends Application {
 
         config = initConfig(getApplicationParameter("config"));
 
-        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(config.getAddressBookFilePath());
         UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
+        userPrefs = initPrefs(userPrefsStorage);
+        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(userPrefs.getAddressBookFilePath());
         storage = new StorageManager(addressBookStorage, userPrefsStorage);
-
-        userPrefs = initPrefs(config);
 
         initLogging(config);
 
@@ -135,10 +134,8 @@ public class MainApp extends Application {
         return initializedConfig;
     }
 
-    protected UserPrefs initPrefs(Config config) {
-        assert config != null;
-
-        String prefsFilePath = config.getUserPrefsFilePath();
+    protected UserPrefs initPrefs(UserPrefsStorage storage) {
+        String prefsFilePath = storage.getUserPrefsFilePath();
         logger.info("Using prefs file : " + prefsFilePath);
 
         UserPrefs initializedPrefs;
--- a/src/main/java/seedu/address/commons/core/Config.java
+++ b/src/main/java/seedu/address/commons/core/Config.java
@@ -14,9 +14,6 @@ public class Config {
     private String appTitle = "Address App";
     private Level logLevel = Level.INFO;
     private String userPrefsFilePath = "preferences.json";
-    private String addressBookFilePath = "data/addressbook.xml";
-    private String addressBookName = "MyAddressBook";
-
 
     public String getAppTitle() {
         return appTitle;
@@ -42,23 +39,6 @@ public class Config {
         this.userPrefsFilePath = userPrefsFilePath;
     }
 
-    public String getAddressBookFilePath() {
-        return addressBookFilePath;
-    }
-
-    public void setAddressBookFilePath(String addressBookFilePath) {
-        this.addressBookFilePath = addressBookFilePath;
-    }
-
-    public String getAddressBookName() {
-        return addressBookName;
-    }
-
-    public void setAddressBookName(String addressBookName) {
-        this.addressBookName = addressBookName;
-    }
-
-
     @Override
     public boolean equals(Object other) {
         if (other == this) {
@@ -72,14 +52,12 @@ public class Config {
 
         return Objects.equals(appTitle, o.appTitle)
                 && Objects.equals(logLevel, o.logLevel)
-                && Objects.equals(userPrefsFilePath, o.userPrefsFilePath)
-                && Objects.equals(addressBookFilePath, o.addressBookFilePath)
-                && Objects.equals(addressBookName, o.addressBookName);
+                && Objects.equals(userPrefsFilePath, o.userPrefsFilePath);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(appTitle, logLevel, userPrefsFilePath, addressBookFilePath, addressBookName);
+        return Objects.hash(appTitle, logLevel, userPrefsFilePath);
     }
 
     @Override
@@ -88,8 +66,6 @@ public class Config {
         sb.append("App title : " + appTitle);
         sb.append("\nCurrent log level : " + logLevel);
         sb.append("\nPreference file Location : " + userPrefsFilePath);
-        sb.append("\nLocal data file location : " + addressBookFilePath);
-        sb.append("\nAddressBook name : " + addressBookName);
         return sb.toString();
     }
 
--- a/src/main/java/seedu/address/model/UserPrefs.java
+++ b/src/main/java/seedu/address/model/UserPrefs.java
@@ -10,6 +10,8 @@ import seedu.address.commons.core.GuiSettings;
 public class UserPrefs {
 
     public GuiSettings guiSettings;
+    private String addressBookFilePath = "data/addressbook.xml";
+    private String addressBookName = "MyAddressBook";
 
     public GuiSettings getGuiSettings() {
         return guiSettings == null ? new GuiSettings() : guiSettings;
@@ -27,6 +29,22 @@ public class UserPrefs {
         guiSettings = new GuiSettings(width, height, x, y);
     }
 
+    public String getAddressBookFilePath() {
+        return addressBookFilePath;
+    }
+
+    public void setAddressBookFilePath(String addressBookFilePath) {
+        this.addressBookFilePath = addressBookFilePath;
+    }
+
+    public String getAddressBookName() {
+        return addressBookName;
+    }
+
+    public void setAddressBookName(String addressBookName) {
+        this.addressBookName = addressBookName;
+    }
+
     @Override
     public boolean equals(Object other) {
         if (other == this) {
@@ -38,17 +56,23 @@ public class UserPrefs {
 
         UserPrefs o = (UserPrefs) other;
 
-        return Objects.equals(guiSettings, o.guiSettings);
+        return Objects.equals(guiSettings, o.guiSettings)
+                && Objects.equals(addressBookFilePath, o.addressBookFilePath)
+                && Objects.equals(addressBookName, o.addressBookName);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(guiSettings);
+        return Objects.hash(guiSettings, addressBookFilePath, addressBookName);
     }
 
     @Override
     public String toString() {
-        return guiSettings.toString();
+        StringBuilder sb = new StringBuilder();
+        sb.append("Gui Settings : " + guiSettings.toString());
+        sb.append("\nLocal data file location : " + addressBookFilePath);
+        sb.append("\nAddressBook name : " + addressBookName);
+        return sb.toString();
     }
 
 }
--- a/src/main/java/seedu/address/storage/StorageManager.java
+++ b/src/main/java/seedu/address/storage/StorageManager.java
@@ -30,10 +30,6 @@ public class StorageManager extends ComponentManager implements Storage {
         this.userPrefsStorage = userPrefsStorage;
     }
 
-    public StorageManager(String addressBookFilePath, String userPrefsFilePath) {
-        this(new XmlAddressBookStorage(addressBookFilePath), new JsonUserPrefsStorage(userPrefsFilePath));
-    }
-
     // ================ UserPrefs methods ==============================
 
     @Override
--- a/src/main/java/seedu/address/storage/StorageManager.java
+++ b/src/main/java/seedu/address/storage/StorageManager.java
@@ -37,6 +37,11 @@ public class StorageManager extends ComponentManager implements Storage {
     // ================ UserPrefs methods ==============================
 
     @Override
+    public String getUserPrefsFilePath() {
+        return userPrefsStorage.getUserPrefsFilePath();
+    }
+
+    @Override
     public Optional<UserPrefs> readUserPrefs() throws DataConversionException, IOException {
         return userPrefsStorage.readUserPrefs();
     }
--- a/src/test/java/seedu/address/storage/JsonUserPrefsStorageTest.java
+++ b/src/test/java/seedu/address/storage/JsonUserPrefsStorageTest.java
@@ -61,8 +61,7 @@ public class JsonUserPrefsStorageTest {
 
     @Test
     public void readUserPrefs_fileInOrder_successfullyRead() throws DataConversionException {
-        UserPrefs expected = new UserPrefs();
-        expected.setGuiSettings(1000, 500, 300, 100);
+        UserPrefs expected = getTypicalUserPrefs();
         UserPrefs actual = readUserPrefs("TypicalUserPref.json").get();
         assertEquals(expected, actual);
     }
@@ -75,11 +74,18 @@ public class JsonUserPrefsStorageTest {
 
     @Test
     public void readUserPrefs_extraValuesInFile_extraValuesIgnored() throws DataConversionException {
-        UserPrefs expected = new UserPrefs();
-        expected.setGuiSettings(1000, 500, 300, 100);
+        UserPrefs expected = getTypicalUserPrefs();
         UserPrefs actual = readUserPrefs("ExtraValuesUserPref.json").get();
 
         assertEquals(expected, actual);
+    }
+
+    private UserPrefs getTypicalUserPrefs() {
+        UserPrefs userPrefs = new UserPrefs();
+        userPrefs.setGuiSettings(1000, 500, 300, 100);
+        userPrefs.setAddressBookFilePath("addressbook.xml");
+        userPrefs.setAddressBookName("TypicalAddressBookName");
+        return userPrefs;
     }
 
     @Test
--- a/src/main/java/seedu/address/storage/JsonUserPrefsStorage.java
+++ b/src/main/java/seedu/address/storage/JsonUserPrefsStorage.java
@@ -19,6 +19,11 @@ public class JsonUserPrefsStorage implements UserPrefsStorage {
     }
 
     @Override
+    public String getUserPrefsFilePath() {
+        return filePath;
+    }
+
+    @Override
     public Optional<UserPrefs> readUserPrefs() throws DataConversionException, IOException {
         return readUserPrefs(filePath);
     }
--- a/src/main/java/seedu/address/storage/UserPrefsStorage.java
+++ b/src/main/java/seedu/address/storage/UserPrefsStorage.java
@@ -12,6 +12,11 @@ import seedu.address.model.UserPrefs;
 public interface UserPrefsStorage {
 
     /**
+     * Returns the file path of the UserPrefs data file.
+     */
+    String getUserPrefsFilePath();
+
+    /**
      * Returns UserPrefs data from storage.
      *   Returns {@code Optional.empty()} if storage file is not found.
      * @throws DataConversionException if the data in storage is not in the expected format.
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -36,6 +36,7 @@ public class MainWindow extends UiPart<Region> {
     private BrowserPanel browserPanel;
     private PersonListPanel personListPanel;
     private Config config;
+    private UserPrefs prefs;
 
     @FXML
     private AnchorPane browserPlaceholder;
@@ -62,6 +63,7 @@ public class MainWindow extends UiPart<Region> {
         this.primaryStage = primaryStage;
         this.logic = logic;
         this.config = config;
+        this.prefs = prefs;
 
         // Configure the UI
         setTitle(config.getAppTitle());
@@ -116,7 +118,7 @@ public class MainWindow extends UiPart<Region> {
         browserPanel = new BrowserPanel(browserPlaceholder);
         personListPanel = new PersonListPanel(getPersonListPlaceholder(), logic.getFilteredPersonList());
         new ResultDisplay(getResultDisplayPlaceholder());
-        new StatusBarFooter(getStatusbarPlaceholder(), config.getAddressBookFilePath());
+        new StatusBarFooter(getStatusbarPlaceholder(), prefs.getAddressBookFilePath());
         new CommandBox(getCommandBoxPlaceholder(), logic);
     }
 
--- a/src/test/data/ConfigUtilTest/ExtraValuesConfig.json
+++ b/src/test/data/ConfigUtilTest/ExtraValuesConfig.json
@@ -2,7 +2,5 @@
   "appTitle" : "Typical App Title",
   "logLevel" : "INFO",
   "userPrefsFilePath" : "C:\\preferences.json",
-  "addressBookFilePath" : "addressbook.xml",
-  "addressBookName" : "TypicalAddressBookName",
   "extra" : "extra value"
 }
--- a/src/test/data/ConfigUtilTest/TypicalConfig.json
+++ b/src/test/data/ConfigUtilTest/TypicalConfig.json
@@ -1,7 +1,5 @@
 {
   "appTitle" : "Typical App Title",
   "logLevel" : "INFO",
-  "userPrefsFilePath" : "C:\\preferences.json",
-  "addressBookFilePath" : "addressbook.xml",
-  "addressBookName" : "TypicalAddressBookName"
+  "userPrefsFilePath" : "C:\\preferences.json"
 }
--- a/src/test/data/JsonUserPrefsStorageTest/ExtraValuesUserPref.json
+++ b/src/test/data/JsonUserPrefsStorageTest/ExtraValuesUserPref.json
@@ -8,5 +8,7 @@
       "y" : 100,
       "z" : 99
     }
-  }
+  },
+  "addressBookFilePath" : "addressbook.xml",
+  "addressBookName" : "TypicalAddressBookName"
 }
--- a/src/test/data/JsonUserPrefsStorageTest/TypicalUserPref.json
+++ b/src/test/data/JsonUserPrefsStorageTest/TypicalUserPref.json
@@ -6,5 +6,7 @@
       "x" : 300,
       "y" : 100
     }
-  }
+  },
+  "addressBookFilePath" : "addressbook.xml",
+  "addressBookName" : "TypicalAddressBookName"
 }
--- a/src/test/java/seedu/address/TestApp.java
+++ b/src/test/java/seedu/address/TestApp.java
@@ -8,6 +8,7 @@ import seedu.address.commons.core.Config;
 import seedu.address.commons.core.GuiSettings;
 import seedu.address.model.ReadOnlyAddressBook;
 import seedu.address.model.UserPrefs;
+import seedu.address.storage.UserPrefsStorage;
 import seedu.address.storage.XmlSerializableAddressBook;
 import seedu.address.testutil.TestUtil;
 
@@ -45,18 +46,18 @@ public class TestApp extends MainApp {
     protected Config initConfig(String configFilePath) {
         Config config = super.initConfig(configFilePath);
         config.setAppTitle(APP_TITLE);
-        config.setAddressBookFilePath(saveFileLocation);
         config.setUserPrefsFilePath(DEFAULT_PREF_FILE_LOCATION_FOR_TESTING);
-        config.setAddressBookName(ADDRESS_BOOK_NAME);
         return config;
     }
 
     @Override
-    protected UserPrefs initPrefs(Config config) {
-        UserPrefs userPrefs = super.initPrefs(config);
+    protected UserPrefs initPrefs(UserPrefsStorage storage) {
+        UserPrefs userPrefs = super.initPrefs(storage);
         double x = Screen.getPrimary().getVisualBounds().getMinX();
         double y = Screen.getPrimary().getVisualBounds().getMinY();
         userPrefs.updateLastUsedGuiSetting(new GuiSettings(600.0, 600.0, (int) x, (int) y));
+        userPrefs.setAddressBookFilePath(saveFileLocation);
+        userPrefs.setAddressBookName(ADDRESS_BOOK_NAME);
         return userPrefs;
     }
 
--- a/src/test/java/seedu/address/commons/core/ConfigTest.java
+++ b/src/test/java/seedu/address/commons/core/ConfigTest.java
@@ -16,9 +16,7 @@ public class ConfigTest {
     public void toString_defaultObject_stringReturned() {
         String defaultConfigAsString = "App title : Address App\n" +
                 "Current log level : INFO\n" +
-                "Preference file Location : preferences.json\n" +
-                "Local data file location : data/addressbook.xml\n" +
-                "AddressBook name : MyAddressBook";
+                "Preference file Location : preferences.json";
 
         assertEquals(defaultConfigAsString, new Config().toString());
     }
--- a/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
@@ -76,8 +76,6 @@ public class ConfigUtilTest {
         config.setAppTitle("Typical App Title");
         config.setLogLevel(Level.INFO);
         config.setUserPrefsFilePath("C:\\preferences.json");
-        config.setAddressBookFilePath("addressbook.xml");
-        config.setAddressBookName("TypicalAddressBookName");
         return config;
     }
 
--- a/src/test/java/seedu/address/logic/LogicManagerTest.java
+++ b/src/test/java/seedu/address/logic/LogicManagerTest.java
@@ -47,7 +47,9 @@ import seedu.address.model.person.Phone;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.tag.Tag;
 import seedu.address.model.tag.UniqueTagList;
+import seedu.address.storage.JsonUserPrefsStorage;
 import seedu.address.storage.StorageManager;
+import seedu.address.storage.XmlAddressBookStorage;
 
 
 public class LogicManagerTest {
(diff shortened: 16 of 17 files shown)


damithc:
@chao1995 is this needed for your other PR?

chao1995:
@damithc Yes. `SaveAsCommand` needs to change
```xml
"addressBookFilePath" : "./data/myaddressbook.xml",
```
in `config.json` so that upon next start we can use the new location by default.

chao1995:
@damithc Currently there is no testing for ExitCommand, so the line I added in stop() method in `MainApp.java` decreases the coverage by 0.09%.

If I want to add an ExitCommandTest, should I add it in this PR or another PR? Thanks.

pyokagan:
@chao1995 At this point, why not

1. Move `Config` into the model component
2. Introduce `ConfigStorage` in the storage component?

This would make it consistent with the architecture design, rather than letting `Config` and `ConfigStorage` float around in `commons`.

damithc:
If we allow data file to be configurable by the user, then it should be part of `UserPref` rather than `Config`?

pyokagan:
@damithc I think `Config` and `UserPrefs` should be merged together :-) No point having two sources of configuration data.

damithc:
I agree with @pyokagan. We can avoid that issue altogether we move the file path to `UserPref`

chao1995:
@damithc @pyokagan OK! I agree too.

damithc:
@chao1995 You can ignore the coverage drop. It's not easy to test for the exit command because exit calls `System.exit()`.

chao1995:
@damithc Noted.

chao1995:
@damithc @pyokagan 

When I attempted to move addressBookFilePath from `Config` to `UserPrefs`, one issue occured and I need your suggestions.

Previously `StorageManager` required both addressBookFilePath and userPrefsFilePath to get instantiated. It implied there is no dependency between these two paths. However now beacuse we want to move addressBookFilePath to `UserPrefs`, we need to `readUserPrefs()` and get an `UserPrefs` before we can get addressBookFilePath. So my code looks like this:
```java
 public StorageManager(String userPrefsFilePath) {
        UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(userPrefsFilePath);
        UserPref initializedPrefs;
        try {
            Optional<UserPrefs> prefsOptional = userPrefsStorage.readUserPrefs();
            initializedPrefs = prefsOptional.orElse(new UserPrefs());
        } catch (DataConversionException e) {
            initializedPrefs = new UserPrefs();
        } catch (IOException e) {
            initializedPrefs = new UserPrefs();
        }
        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(initializedPrefs.getAddressBookFilePath());
        this(addressBookStorage, userPrefsStorage);
}
```

However, in `MainApp` once we instantiate `StorageManager`, we use it to read `UserPrefs`, and we do the above code again
```java
protected UserPrefs initPrefs(Config config) {
        assert config != null;

        String prefsFilePath = config.getUserPrefsFilePath();

        UserPrefs initializedPrefs;
        try {
            Optional<UserPrefs> prefsOptional = storage.readUserPrefs();
            initializedPrefs = prefsOptional.orElse(new UserPrefs());
        } catch (DataConversionException e) {
            initializedPrefs = new UserPrefs();
        } catch (IOException e) {
            initializedPrefs = new UserPrefs();
        }
        ...
        return initializedPrefs;
}
```

Any suggestions for not doing this? Thanks.

pyokagan:
@damithc says:
> We can avoid that issue altogether we move the file path to UserPref

Actually, my suggestion is to:

1. Move `Config` into the model component
2. Introduce a `ConfigStorage` in the storage component
3. Move all of the fields from `UserPrefs` into `Config`.
4. Get rid of the `UserPrefs` class entirely :-)

pyokagan:
@chao1995 Maybe construct the `JsonUserPrefsStorage` in `MainApp.java` and then pass it into the `StorageManager's` constructor.

damithc:
@pyokagan Config is meant to represent system config (we can rename it as such). These are the things a programmer will manipulate to alter primitive system behaviors and also, these are the things needed from the very beginning, even before the full app is fleshed out. e.g. logging level. They are also read-only and can only be modified by editing the file (or supplying a different file as a startup parameter). UserPrefs are related to the end user preferences that can be changed by the end user (or deduced by user behavior) and therefore can be considered as part of the model. So I would like to keep the two separate.

chao1995:
@damithc But there is one problem if we move addressBookPath to `UserPrefs`,

If we want to get the addressBookPath,

You need an `UserPrefs` to get an addressBookPath.
You need a `StorageManger` to get a `UserPref`.
You need an addressBookPath to get a `StorageManager` (But we need the addressBookPath)

Where to break the dependency?

damithc:
@chao1995 Another option is to allow a 'partially initialized' StorageManager to remain for a short while. 
e.g.
1. Create an empty StorageManager
2. Set userPref path of StorageManager
2. Read UserPrefs using StorageManager 
3. Find the data location from the loaded UserPrefs
4. set address book location in StorageManager

Would that work? @pyokagan your thoughts?

pyokagan:
@damithc says:
> Would that work? @pyokagan your thoughts?

Yeah, that will work. I think this is the best we can do without radically changing the architecture.

pyokagan:
@damithc says:
> They are also read-only and can only be modified by editing the file (or supplying a different file as a startup parameter).

Oh okay. I agree that moving the address book path config field to `UserPrefs` is what we should do in this case.

chao1995:
@damithc @pyokagan Ready for review.

Changes:
- Move addressBookFilePath and addressBookName from` Config` to `UserPrefs`
- Fix JSON test data.
- Fix tests. `ConfigTest` and `JsonUserPrefsStorageTest`
- Add `setAddressBookPath(String addressBookPath)` to `Storage`, `AddressBookStorage` interface.
- Implement the interfaces in `StorageManager` and `XmlAddressBookStorage`.
- Fix the initialization steps (e.g. setup `UserPrefs`) in `MainApp` and `TestApp` and `HelpWindow`

src/main/java/seedu/address/commons/core/Config.java
```diff
@@ -14,8 +14,6 @@
     private String appTitle = "Address App";
     private Level logLevel = Level.INFO;
     private String userPrefsFilePath = "preferences.json";
-    private String addressBookFilePath = "data/addressbook.xml";
-    private String addressBookName = "MyAddressBook";
```
damithc: Just curious: is this variable used for anything?

chao1995: @damithc I think addressBookName has not yet been used anywhere in the xml or code.

damithc: That's what I thought. Once this PR is done, we can create an issue to do something about it and leave it to a first time contributor.

src/main/java/seedu/address/storage/StorageManager.java
```diff
@@ -29,6 +29,10 @@ public StorageManager(AddressBookStorage addressBookStorage, UserPrefsStorage us
         this.userPrefsStorage = userPrefsStorage;
     }
 
+    public StorageManager(String userPrefsFilePath) {
```
pyokagan: Having a constructor with just a `userPrefsFilePath` seems inconsistent. Why not have a constructor that accepts a `addressBookFilePath` instead?

The answer is we cannot, since we are designing this constructor with `MainApp` in mind, which indicates this constructor is too tied to implementation details.

Instead, maybe we should use the `StorageManager(String addressBookFilePath, String userPrefsFilePath)` instead, and adapt it to allow `null`s as `addressBookFilePath` and `userPrefsFilePath`.

src/main/java/seedu/address/storage/StorageManager.java
```diff
@@ -29,6 +29,10 @@ public StorageManager(AddressBookStorage addressBookStorage, UserPrefsStorage us
...
+        this(null, new JsonUserPrefsStorage(userPrefsFilePath));
+    }
+
     public StorageManager(String addressBookFilePath, String userPrefsFilePath) {
         this(new XmlAddressBookStorage(addressBookFilePath), new JsonUserPrefsStorage(userPrefsFilePath));
```
pyokagan: With the introduction of `setAddressBookFilePath()` below, this method is not DRY anymore, since both this constructor and `setAddressBookFilePath()` references the address book storage we construct internally (`XmlAddressBookStorage`)

I recommend:

1. Introducing `setUserPrefsFilePath()` to be consistent with `setAddressBookFilePath()`.
2. Then the constructor can just do:
```java
this(null, null);
setAddressBookStorage(addressBookFilePath);
setUserPrefsFilePath(userPrefsFilePath);
```

pyokagan:
@chao1995 Please also update the commit message to indicate the rationale for these code changes. (I didn't follow discussion that closely, and the issue and PR did not say anything about *why* this change is a good thing).

src/main/java/seedu/address/model/UserPrefs.java
```diff
@@ -38,17 +56,23 @@ public boolean equals(Object other) {
...
     }
 
     @Override
     public String toString() {
-        return guiSettings.toString();
+        StringBuilder sb = new StringBuilder();
+        sb.append("Gui Settings : " + guiSettings.toString());
+        sb.append("\nLocal data file location : " + addressBookFilePath);
+        sb.append("\nAddressBook name : " + addressBookName);
+        return sb.toString();
```
pyokagan: @damithc What do you think of adopting a standard `toString()` format? Maybe using Guava's [`MoreObjects.toStringHelper()`](https://github.com/google/guava/wiki/CommonObjectUtilitiesExplained#tostring)

Also, I'm not too fond of object string representations returning multiple lines since they are kind of hard to read in log files. e.g., the above string representation will look like:
```
INFO: Initializing with address book: seedu.address.storage.XmlSerializableAddressBook@766ccbf and user prefs Gui Settings : Width : 684.0
Height : 1005.0
Position : java.awt.Point[x=1236,y=25]
Local data file location : data/addressbook.xml
AddressBook name : MyAddressBook
Dec 18, 2016 11:32:40 PM seedu.address.MainApp start
```
in the log file.

damithc: Yes, that's a good idea @pyokagan 
Please create an issue for it.

src/main/java/seedu/address/storage/StorageManager.java
chao1995: I think `StorageManager(String addressBookFilePath, String userPrefsFilePath)` is a legacy method in some sense, since we cannot get an addressBookFilePath before we can get a `StorageManager`, except for making up one.

In my current implementation, another issue is that it doesn't enforce the users to call `setAddressBookFilePath` before calling `readAddressBook`, which will then fail if the addressBookFilePath have not yet been set. So I would say this is not a good API yet. Allowing `null` has the same problem.

So how about we don't use the `set...FilePath()` methods, and instead we create a `factory` method for `StorageManager`, which takes in a `userPrefsFilePath`.

Then in the factory method we
1. construct a `StorageManager` using something like `this(null, null)`
2. set `userPrefsFilePath` using the required parameter `userPrefsFilePath`
3. read `UserPrefs` from the path
4. read `addressBookFilePath` from the `UserPrefs`
5. set `addressBookFilePath`

Finally we can return a fully initialised `StorageManager`

Using the factory method, `StorageManager` will be protected from partial initialisation.

chao1995:
@pyokagan Thanks for the notice. Will update the commit message in the later fixing commits.

src/main/java/seedu/address/storage/StorageManager.java
pyokagan: > So how about we take away the set...FilePath() methods,

One question, how can the address book file path be updated?

> 1. construct a `StorageManager` using something like `this(null, null)`
> 2. set `userPrefsFilePath` using the required parameter `userPrefsFilePath`
> 3. read `UserPrefs` from the path
> 4. read `addressBookFilePath` from the `UserPrefs`
> 5. set `addressBookFilePath`

Sounds good, although I think this should be in `MainApp.java` for consistency with `initModelManager()`, `initConfig()` etc.

Why not try it out? :-) (Of course, keep a copy of this version in case things go south)

chao1995: > how can the address book file path be updated?

We can directly update the storage as this is instance variable in `StorageManager` anaways.

Will try to implement this and see how it looks like later. :)

pyokagan:
@chao1995 Any updates?

pyokagan:
@chao1995 So any updates? :-P

pyokagan:
Closing due to inactivity.

CanIHasReview:
# v1

@chao1995 submitted v1 for review.

* [v1 1/2] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/24c9cab85717caff624d86084acbff5d25be7c73)
* [v1 2/2] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/f81f6a2e868d5ecf11203d1f821e0ac42d8bcdef)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/f40ee42cb2e29a99c9c6b8757b64b0b0f4284d67...f81f6a2e868d5ecf11203d1f821e0ac42d8bcdef))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/1/head:BRANCHNAME
```

</details>

chao1995:
@yamgent Would you like to peer review this PR before seniors review it?

Long-time ago... we discussed several approaches to solve the problem below, but in the new version (v1), I used a different approach, summarised as follows (copied from second commit's message):

The problem is if we move addressBookFilePath to `UserPrefs`, there will be a initialization dependency cycle as follows:

1. to initialize a `StorageManager`, we need an addressBookFilePath
   (or an `AddressBookStorage`, which also needs an addressBookFilePath)
2. to get an addressBookFilePath, we now need an initialized `UserPrefs`
3. to get an `UserPrefs`, we need to read it from StorageManager

That means, we cannot even initialize a StorageManager in `MainApp`!

To break the cycle in `MainApp`, instead of reading the `UserPrefs` storage from
`StorageManager`, we construct a `JsonUserPrefsStorage` and read `UserPrefs`
from it. Then to initialize `StorageManager`, we can pass in a `UserPrefsStorage` and an `AddressBookStorage` instead of file paths.

This is a better practice which follows Law of Demeter since what
`StorageManager` needs is the Storage not the file paths. `StorageManager`
should not construct the Storage objects itself from file paths, but
asks for them. Actually, the constructor (taking two file paths)
should be removed.

chao1995:
Will open a new PR.

yamgent:
> This is a better practice which follows Law of Demeter since what
> StorageManager needs is the Storage not the file paths. StorageManager
> should not construct the Storage objects itself from file paths, but
> asks for them. Actually, the constructor (taking two file paths)
> should be removed.

This seems kind of strange though. Is there a way to ask `StorageManager` to initialize `AddressBookStorage` by itself? `MainApp` should probably be able to call a certain method in the `Storage` API to initialize (and subsequently reinitialize when renaming) the `AddressBookStorage` inside `StorageManager`, and shouldn't have to manually make a storage and pass it in.

Also the second commit is too huge, it is very difficult for me to review it all at one go without seperated context. :P

CanIHasReview:
# v2

@chao1995 submitted v2 for review.

* [v2 1/4] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/e7a2becbaecdceaaf916492898b36d04b94091a3)
* [v2 2/4] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/8502b4a32342ce0694a46cac03c780c4c2539e7c)
* [v2 3/4] [Fix Json](https://github.com/se-edu/addressbook-level4/pull/214/commits/dde064471bd595bde35a6cca5c5efffd5414bc6d)
* [v2 4/4] [Fix test](https://github.com/se-edu/addressbook-level4/pull/214/commits/b14db3f1c5c21910e79badf79d8843d65f8337c6)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...b14db3f1c5c21910e79badf79d8843d65f8337c6)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/14c8b19b4faa50de99b57da16a0c9000020aee1b~1...14c8b19b4faa50de99b57da16a0c9000020aee1b))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/2/head:BRANCHNAME
```

</details>

chao1995:
@yamgent Split the huge commit into 3 smaller ones. The [2/4] commit is moving stuff around and make sure the MainApp can run. The later two commits are just fix json and tests respectively and can be squashed after code review.

src/test/java/seedu/address/storage/JsonUserPrefsStorageTest.java
```diff
@@ -75,13 +74,20 @@ public void readUserPrefs_valuesMissingFromFile_defaultValuesUsed() throws DataC
...
     public void readUserPrefs_extraValuesInFile_extraValuesIgnored() throws DataConversionException {
-        UserPrefs expected = new UserPrefs();
-        expected.setGuiSettings(1000, 500, 300, 100);
+        UserPrefs expected = getTypicalUserPrefs();
         UserPrefs actual = readUserPrefs("ExtraValuesUserPref.json").get();
 
         assertEquals(expected, actual);
     }
 
+    private UserPrefs getTypicalUserPrefs() {
```
yamgent: Probably should be placed below. Right now it is in the middle of tests.

chao1995:
@yamgent Thanks for the suggestion! I find by breaking down [2/4] to two commits, there is no need to explain the dependency cycle style because it doesn't appear anymore, thus shortening the commit message by a lot. Thanks again!

src/test/java/seedu/address/storage/JsonUserPrefsStorageTest.java
chao1995: The guideline for ordering of methods is not uniformed. Some insist public-private etc ordering, some suggest it is better to organise related methods together. So I put it there because it is used in tests right above it.

CanIHasReview:
# v3

@chao1995 submitted v3 for review.

* [v3 1/5] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/e7a2becbaecdceaaf916492898b36d04b94091a3)
* [v3 2/5] [MainApp: Refactor construction of StorageManager](https://github.com/se-edu/addressbook-level4/pull/214/commits/1b5fd35ff39dde4171480f21824b398c1519d256)
* [v3 3/5] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/8ee10799d1f55ccb40ea3695f969cd870e919fab)
* [v3 4/5] [Fix Json](https://github.com/se-edu/addressbook-level4/pull/214/commits/87b03d6cf6edb80cb237ed6b69db165d4d3535ce)
* [v3 5/5] [Fix test](https://github.com/se-edu/addressbook-level4/pull/214/commits/35067cca84723b005229f1bc49ee8424e531dca1)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...35067cca84723b005229f1bc49ee8424e531dca1)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/f46ed467027c52665ce56e5eeea82e5e5c7c3fa7~1...f46ed467027c52665ce56e5eeea82e5e5c7c3fa7))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/3/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/storage/JsonUserPrefsStorageTest.java
yamgent: Ok we will leave it here then.

CanIHasReview:
# v4

@chao1995 submitted v4 for review.

* [v4 1/5] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/230bb7108a371d7baad0580c7e23f90bfacf05c0)
* [v4 2/5] [MainApp: Refactor construction of StorageManager](https://github.com/se-edu/addressbook-level4/pull/214/commits/beb235d2f3dc792a066f856c530725592ebe3c40)
* [v4 3/5] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/5be8e246881d411340514782269e162d7f46da7f)
* [v4 4/5] [Fix Json](https://github.com/se-edu/addressbook-level4/pull/214/commits/5f810e88ea17e8c3eef314a725f5b1a2fdca4a12)
* [v4 5/5] [Fix test](https://github.com/se-edu/addressbook-level4/pull/214/commits/f3c742b97da71086651ffd4c0e94619da343bdac)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...f3c742b97da71086651ffd4c0e94619da343bdac)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/ecafc36c3647928475199a7483a0308365571736~1...ecafc36c3647928475199a7483a0308365571736))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/4/head:BRANCHNAME
```

</details>

chao1995:
@damithc Prof, I wrote the rationale in commit [v4 3/5], which moves stuff from `Config` to `UserPref`, and explained why we wanted to so. The rationale is as follows:

> addressBookFilePath and addressBookName are stored in Config, which is
meant to represent system configuration and should be read-only.<br>
That means we should not update addressBookFilePath and addressBookName
at runtime. However, this creates problems if we want to allow users to
change the address book file path, or choose a different name for the
address book.<br>
Thus these two pieces of info should be kept in UserPrefs instead,
because it is meant for such user preference settings, and can be
easily updated at runtime.<br>
Let's move addressBookFilePath and addressBookName from Config to
UserPrefs.

damithc:
> @damithc Prof, I wrote the rationale in commit [v4 3/5], which moves stuff from Config to UserPref, and explained why we wanted to so.

1. This rationale does not match with the PR/issue title exactly. Update PR/issue title if necessary.
2. We still need a merge commit message which is supposed to explain the entire PR. So the merge commit message is unlikely to be exactly the same as that of one of the commits in the PR. That's why I asked you to propose a merge commit message in the PR description. The idea is, the person merging the PR can simply use that in the merge commit.

BTW, why do you have temporary commits in the PR?

chao1995:
> This rationale does not match with the PR/issue title exactly. Update PR/issue title if necessary.

Will append `addressBookName` to PR/issue title.

> We still need a merge commit message which is supposed to explain the entire PR

will add a proposed merge commit message to PR description.

> why do you have temporary commits in the PR?

The third commit originally contains three commits (3rd, 4th and 5th), which results in a commit diff of 10+ files and it's hard to review. So I just separate the commits to production, test data (json) and testing. It's just for review purposes.

damithc:
> The third commit originally contains three commits (3rd, 4th and 5th), which results in a commit diff of 10+ files and it's hard to review. So I just separate the commits to production, test data (json) and testing. It's just for review purposes.

It's not nice to have a commit organization just for reviewing. Is there no way to split it up permanently so that the commit is small enough to review but still contain a logical step forward (without breaking the build)?

chao1995:
> Is there no way to split it up permanently so that the commit is small enough to review but still contain a logical step forward

The smaller break-down commits are just testing code. Since testing code should be together with the production code, they should just be in one commit.

I can't really think of other logical ways to break it down, other than moving one thing at a time from Config to UserPref, or move to UserPref first followed by cleanup Config, but they seem like not involving less file changes, which is why I break down the commit for in the first place.

yamgent:
@chao1995 I suggest that you can just squash them. I agree with you that splitting the third commit is difficult, because it just so happens that moving the two variables from one class to another causes a lot of changes in different areas of the application. So if we split them up further it is likely that some commits are going to break the build one way or another anyway.

It is possible to split it such that one commit implements the variables on the new class, while the second commit removes the variables on the old class. However, this is likely to confuse the reviewers (defeats the purpose of splitting the commit in the first place), and the decrease in the diff would be negligible to warrant such a split.

damithc:
> @chao1995 I suggest that you can just squash them.

Yes, if there is no way to split them permanently, it is cleaner to keep it as one commit although a bit harder to review.

CanIHasReview:
# v5

@chao1995 submitted v5 for review.

* [v5 1/3] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/230bb7108a371d7baad0580c7e23f90bfacf05c0)
* [v5 2/3] [MainApp: Refactor construction of StorageManager](https://github.com/se-edu/addressbook-level4/pull/214/commits/beb235d2f3dc792a066f856c530725592ebe3c40)
* [v5 3/3] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/4eefcd41f24f01c33d175df16c626b09aae2a0e9)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...4eefcd41f24f01c33d175df16c626b09aae2a0e9)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/036bd35321c93f61f3ae9d51f4f6a0472a0ed7bf~1...036bd35321c93f61f3ae9d51f4f6a0472a0ed7bf))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/5/head:BRANCHNAME
```

</details>

CanIHasReview:
# v6

@chao1995 submitted v6 for review.

* [v6 1/3] [Add getUserPrefsFilePath\(\) to UserPrefsStorage interface](https://github.com/se-edu/addressbook-level4/pull/214/commits/230bb7108a371d7baad0580c7e23f90bfacf05c0)
* [v6 2/3] [MainApp: Refactor construction of StorageManager](https://github.com/se-edu/addressbook-level4/pull/214/commits/769882eaf49c52c5c805742e587901561c55f4d2)
* [v6 3/3] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/74061c589b04201bac7cfadbc1970e82bf0a8585)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...74061c589b04201bac7cfadbc1970e82bf0a8585)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/567bd88ebc60854426167c55c58fcfc4f8a947c6~1...567bd88ebc60854426167c55c58fcfc4f8a947c6))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/6/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/MainApp.java
```diff
@@ -58,11 +58,13 @@ public void init() throws Exception {
...
 
-        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(config.getAddressBookFilePath());
         UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
-        storage = new StorageManager(addressBookStorage, userPrefsStorage);
 
-        userPrefs = initPrefs(config);
+        userPrefs = initPrefs(userPrefsStorage);
+
+        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(userPrefs.getAddressBookFilePath());
+
```
pyokagan: I would like to strongly discourage you from adding too many blank lines. The [end result](https://github.com/chao1995/addressbook-level4/blob/74061c589b04201bac7cfadbc1970e82bf0a8585/src/main/java/seedu/address/MainApp.java#L59-L77) looks extremely weird:

```java
        logger.info("=============================[ Initializing AddressBook ]===========================");
        super.init();

        config = initConfig(getApplicationParameter("config"));

        UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());

        userPrefs = initPrefs(userPrefsStorage);

        AddressBookStorage addressBookStorage = new XmlAddressBookStorage(userPrefs.getAddressBookFilePath());

        storage = new StorageManager(addressBookStorage, userPrefsStorage);

        initLogging(config);

        model = initModelManager(storage, userPrefs);

        logic = new LogicManager(model, storage);

        ui = new UiManager(logic, config, userPrefs);

        initEventsCenter();
```
That's a lot of wasted vertical space.

src/main/java/seedu/address/model/UserPrefs.java
pyokagan: Interesting that this change did not trip any tests, which suggests this method is not being tested?

With that said, I don't really thing we need to test `toString()` either (since it's mainly supposed to be a debugging tool). Just showing that I'm reading the diff :-P

chao1995:
> Huh? I was under the impression that Law of Demeter is about the invocation of
> methods of objects, which from the rest of the commit message I don't think it
> is about.

Yes, invocation of methods of objects using more than one dot is a symptom of violating Law of Demeter. The root cause of this is not asking for direct dependency, which is the case for the String parameters. They are not direct dependencies.

As for why this constructor is that bad, in the third commit, I worte

>The problem of this constructor is that it's not asking for direct
dependency. This makes it hard to write unit tests for StorageManager
as it's impossible to isolate it from XmlAddressBookStorage using
this constructor. A real XmlAddressBookStorage is always created,
you just can't stub it out. This coupling has another problem, imagine
we want to add another JsonAddressBookStorage, then we'll have a hard
time deciding which type of AddressBookStorage to create in this
constructor.

CanIHasReview:
# v7

@chao1995 submitted v7 for review.

* [v7 1/4] [UserPrefsStorage: introduce getUserPrefsFilePath\(\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/15c572e6b0c258ec24f603fa8150525109ec35f4)
* [v7 2/4] [MainApp: refactor construction of StorageManager](https://github.com/se-edu/addressbook-level4/pull/214/commits/9e9bb2770d8edc7a6771bac480ed034f57632c0c)
* [v7 3/4] [StorageManager: remove constructor StorageManager\(String\, String\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/d09bd10a9b654a7b06234e8c700be6f0a61e5d52)
* [v7 4/4] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/f2a2270b69308c7d9bfed15dca375fbeb22e1af1)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...f2a2270b69308c7d9bfed15dca375fbeb22e1af1)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/compare/c72fbcaf3a3d524e23da5b69eb5f708432e70d34~1...c72fbcaf3a3d524e23da5b69eb5f708432e70d34))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/7/head:BRANCHNAME
```

</details>

pyokagan:
> @damithc Could you help me

Thanks in advance :-P

damithc:
> @damithc Could you help me verify if this is correct? 
> FWIW, this StackOverflow answer, as I'm understanding it, actually says that forcing consumers to do the injection of dependencies actually breaks this law, interestingly. (But SO answers aren't always right.)

'Law of Demeter' (which says "don't talk to strangers") allows calling methods of the objects created by yourself (objects you create yourself are considered 'friends'). So this justification using LoD is not correct. A better justification: creating dependencies inside the constructor reduces restability.

chao1995:
[v7 2/4]
> FWIW, this StackOverflow answer, as I'm understanding it, actually says that forcing consumers to do the injection of dependencies actually breaks this law, interestingly. (But SO answers aren't always right.)

The answer actually mentions that if we are using factories/DI framework, we can avoid breaking LoD.

[v7 3/4]

> Why can't we just use the StorageManager(AddressBookStorage, UserPrefsStorage) constructor?

Yes we can.

> I guess we could write some code to select
> XmlAddressBookStorage/JsonAddressBookStorage based on, e.g. the file extension,
> but why would that be hard?

It's actually just about testability. Having a constructor does real work is not a good idea. Plus with this method StorageManager(AddressBookStorage, UserPrefsStorage), the problem is trivially resolved.


> This reasoning seems quite similar to [v7 2/4], so I guess this can be squashed in.

ok. makes sense to me!

CanIHasReview:
# v8

@chao1995 submitted v8 for review.

* [v8 1/3] [UserPrefsStorage: introduce getUserPrefsFilePath\(\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/a7432e6f55f513bae8de4cad878ae2e1434f6a93)
* [v8 2/3] [StorageManager: remove constructor StorageManager\(String\, String\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/2f9c72a11ef12677dda2fd2d0c5260a1974b7550)
* [v8 3/3] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/2560b6bf2c70c5e3330a2441d6e5e7f23acdf9a7)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...2560b6bf2c70c5e3330a2441d6e5e7f23acdf9a7)) ([:chart_with_upwards_trend: Interdiff between v7 and v8](https://github.com/se-edu/addressbook-level4/compare/02d355d893d50e43a257ae5c725c580ea4fd5c73~1...02d355d893d50e43a257ae5c725c580ea4fd5c73))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/8/head:BRANCHNAME
```

</details>

chao1995:
> The constructor StorageManager(String, String) allows the callers to 
> avoid having to create XmlAddressBookStorage and JsonUserPrefsStorage
> by themselves.

Actually I think the thing that hurts testability is creating new object inside constructor. So I actually changed my previous sentence to

> The constructor StorageManager(String, String) creates two objects,
namely XmlAddressBookStorage and JsonUserPrefsStorage inside itself.

CanIHasReview:
# v9

@chao1995 submitted v9 for review.

* [v9 1/3] [UserPrefsStorage: introduce getUserPrefsFilePath\(\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/a7432e6f55f513bae8de4cad878ae2e1434f6a93)
* [v9 2/3] [StorageManager: remove constructor StorageManager\(String\, String\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/7ff7c378c63e6da76f0565b09606309e67a69fe0)
* [v9 3/3] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/18e04a6334f70b763d03f8dd908ec8f321e7be38)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...18e04a6334f70b763d03f8dd908ec8f321e7be38)) ([:chart_with_upwards_trend: Interdiff between v8 and v9](https://github.com/se-edu/addressbook-level4/compare/8b92c6743580e08ce807861a50469fdcfe38ae63~1...8b92c6743580e08ce807861a50469fdcfe38ae63))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/9/head:BRANCHNAME
```

</details>

CanIHasReview:
# v10

@chao1995 submitted v10 for review.

* [v10 1/3] [UserPrefsStorage: introduce getUserPrefsFilePath\(\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/1e370962a6c6939e0fc815688f96a542270cd5a6)
* [v10 2/3] [StorageManager: remove constructor StorageManager\(String\, String\)](https://github.com/se-edu/addressbook-level4/pull/214/commits/535ed3a77aac64ccde8a5ad6e10752b322ef50ec)
* [v10 3/3] [Move addressBookFilePath and addressBookName from Config to UserPrefs](https://github.com/se-edu/addressbook-level4/pull/214/commits/3a4617fc799ffcd33d5b50d46fa13a629f03fb1f)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/1f784b06cd8055123a5aa5ad06da7a1a0382bffc...3a4617fc799ffcd33d5b50d46fa13a629f03fb1f)) ([:chart_with_upwards_trend: Interdiff between v9 and v10](https://github.com/se-edu/addressbook-level4/compare/55b5c1a73912962ced33257edc19aae455a0508b~1...55b5c1a73912962ced33257edc19aae455a0508b))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/214/10/head:BRANCHNAME
```

</details>

chao1995:
V10: change "Storage" to "storage" in [v10 2/3] commit message.

damithc:
@chao1995 can you update the proposed merge commit message to follow our usual format?

chao1995:
@damithc Prof, updated the merge commit message.

PR #225, opened 2016-12-20
Make `GuiHandle::getNode()` return a `<T extends Node> T` #219

### LINKED ISSUE #219: Make `GuiHandle::getNode()` return a `<T extends Node> T`
It's slightly undesirable that [`GuiHandle::getNode()`](https://github.com/se-edu/addressbook-level4/blob/311ab3c0ccd4d6e49d81234e6d8fe42646e3c834/src/test/java/guitests/guihandles/GuiHandle.java#L53) returns a `Node`, since it means that we need to add down-casts in order to get the subclass of `Node` that we want.

The PR should also remove the casts at callsites that were previously due to `getNode()` returning a `Node`.

MightyCupcakes:
Resolves #219

### DIFF
--- a/src/test/java/guitests/guihandles/GuiHandle.java
+++ b/src/test/java/guitests/guihandles/GuiHandle.java
@@ -52,7 +52,7 @@ public class GuiHandle {
 
     @SuppressWarnings("unchecked")
     protected <T extends Node> T getNode(String query) {
-        return (T) guiRobot.lookup(query).tryQuery().get();
+        return (T) guiRobot.lookup(query).query();
     }
 
     protected String getTextFieldText(String filedName) {
--- a/src/test/java/guitests/guihandles/GuiHandle.java
+++ b/src/test/java/guitests/guihandles/GuiHandle.java
@@ -52,7 +52,7 @@ public class GuiHandle {
 
     @SuppressWarnings("unchecked")
     protected <T extends Node> T getNode(String query) {
-        return (T) guiRobot.lookup(query).query();
+        return guiRobot.lookup(query).query();
     }
 
     protected String getTextFieldText(String filedName) {
--- a/src/test/java/guitests/guihandles/GuiHandle.java
+++ b/src/test/java/guitests/guihandles/GuiHandle.java
@@ -50,8 +50,9 @@ public class GuiHandle {
         logger.info("Finishing focus " + stageTitle);
     }
 
-    protected Node getNode(String query) {
-        return guiRobot.lookup(query).tryQuery().get();
+    @SuppressWarnings("unchecked")
+    protected <T extends Node> T getNode(String query) {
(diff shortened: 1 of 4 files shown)


src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -50,17 +50,20 @@ public void focusOnWindow(String stageTitle) {
         logger.info("Finishing focus " + stageTitle);
     }
 
-    protected Node getNode(String query) {
-        return guiRobot.lookup(query).tryQuery().get();
+    @SuppressWarnings("unchecked")
+    protected <T extends Node> T getNode(String query) {
+        return (T) guiRobot.lookup(query).tryQuery().get();
```
pyokagan: Shouldn't `guiRobot.lookup(query).query()` be sufficient?

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -50,17 +50,20 @@ public void focusOnWindow(String stageTitle) {
         logger.info("Finishing focus " + stageTitle);
     }
 
-    protected Node getNode(String query) {
-        return guiRobot.lookup(query).tryQuery().get();
+    @SuppressWarnings("unchecked")
+    protected <T extends Node> T getNode(String query) {
+        return (T) guiRobot.lookup(query).query();
```
pyokagan: I think you can remove the cast as well.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -50,17 +50,20 @@ public void focusOnWindow(String stageTitle) {
         logger.info("Finishing focus " + stageTitle);
     }
 
-    protected Node getNode(String query) {
-        return guiRobot.lookup(query).tryQuery().get();
+    @SuppressWarnings("unchecked")
```
pyokagan: This can be removed.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -73,7 +76,8 @@ protected String getTextFromLabel(String fieldId, Node parentNode) {
     }
 
     protected String getTextFromLabel(String fieldId) {
-        return ((Label) guiRobot.lookup(fieldId).tryQuery().get()).getText();
+        Label label = getNode(fieldId);
+        return label.getText();
```
pyokagan: Re: code coverage. Turns out this method is not used at all. I think we can do some cleanup by removing this method.

pyokagan:
@MightyCupcakes The name of the PR needs to match the issue title.

damithc:
Thanks for the contribution @MightyCupcakes 👍 .
You may also create a PR to add yourself to the list of contributors if you haven't done so already.

PR #226, opened 2016-12-20
Show some sample data when the app is run for the first time #60

### LINKED ISSUE #60: Show some sample data when the app is run for the first time
We can check for the file existence and add some sample data if the file doesn't exist.

edmundmok:
Fixes #60 

If no save file is detected, sample data is loaded into memory.
File is not created with the sample data.

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level4/226)

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -15,6 +15,7 @@ import seedu.address.commons.util.StringUtil;
 import seedu.address.logic.Logic;
 import seedu.address.logic.LogicManager;
 import seedu.address.model.*;
+import seedu.address.model.util.SampleDataUtil;
 import seedu.address.storage.Storage;
 import seedu.address.storage.StorageManager;
 import seedu.address.ui.Ui;
@@ -73,9 +74,9 @@ public class MainApp extends Application {
         try {
             addressBookOptional = storage.readAddressBook();
             if (!addressBookOptional.isPresent()) {
-                logger.info("Data file not found. Will be starting with an empty AddressBook");
+                logger.info("Data file not found. Will be starting with a sample AddressBook");
             }
-            initialData = addressBookOptional.orElse(new AddressBook());
+            initialData = addressBookOptional.orElseGet(SampleDataUtil::getSampleAddressBook);
         } catch (DataConversionException e) {
             logger.warning("Data file not in the correct format. Will be starting with an empty AddressBook");
             initialData = new AddressBook();
--- a/src/main/java/seedu/address/model/util/SampleDataUtil.java
+++ b/src/main/java/seedu/address/model/util/SampleDataUtil.java
@@ -1,0 +1,53 @@
+package seedu.address.model.util;
+
+import seedu.address.commons.exceptions.IllegalValueException;
+import seedu.address.model.AddressBook;
+import seedu.address.model.ReadOnlyAddressBook;
+import seedu.address.model.person.Address;
+import seedu.address.model.person.Email;
+import seedu.address.model.person.Name;
+import seedu.address.model.person.Person;
+import seedu.address.model.person.Phone;
+import seedu.address.model.person.UniquePersonList.DuplicatePersonException;
+import seedu.address.model.tag.UniqueTagList;
+
+public class SampleDataUtil {
+    public static Person[] getSamplePersons() {
+        try {
+            return new Person[] {
+                new Person(new Name("Alex Yeoh"), new Phone("87438807"), new Email("alexyeoh@gmail.com"),
+                    new Address("Blk 30 Geylang Street 29, #06-40"),
+                    new UniqueTagList("friends")),
+                new Person(new Name("Bernice Yu"), new Phone("99272758"), new Email("berniceyu@gmail.com"),
+                    new Address("Blk 30 Lorong 3 Serangoon Gardens, #07-18"),
+                    new UniqueTagList("colleagues", "friends")),
+                new Person(new Name("Charlotte Oliveiro"), new Phone("93210283"), new Email("charlotte@yahoo.com"),
+                    new Address("Blk 11 Ang Mo Kio Street 74, #11-04"),
+                    new UniqueTagList("neighbours")),
+                new Person(new Name("David Li"), new Phone("91031282"), new Email("lidavid@google.com"),
+                    new Address("Blk 436 Serangoon Gardens Street 26, #16-43"),
+                    new UniqueTagList("family")),
+                new Person(new Name("Irfan Ibrahim"), new Phone("92492021"), new Email("irfan@outlook.com"),
+                    new Address("Blk 47 Tampines Street 20, #17-35"),
+                    new UniqueTagList("classmates")),
+                new Person(new Name("Roy Balakrishnan"), new Phone("92624417"), new Email("royb@gmail.com"),
+                    new Address("Blk 45 Aljunied Street 85, #11-31"),
+                    new UniqueTagList("colleagues"))
+            };
+        } catch (IllegalValueException e) {
+            throw new AssertionError("sample data cannot be invalid", e);
+        }
+    }
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        try {
+            AddressBook sampleAB = new AddressBook();
+            for (Person samplePerson : getSamplePersons()) {
+                sampleAB.addPerson(samplePerson);
+            }
+            return sampleAB;
+        } catch (DuplicatePersonException e) {
+            throw new AssertionError("sample data cannot contain duplicate persons", e);
+        }
+    }
+}
--- a/src/test/java/guitests/SampleDataTest.java
+++ b/src/test/java/guitests/SampleDataTest.java
@@ -1,0 +1,30 @@
+package guitests;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+import seedu.address.model.AddressBook;
+import seedu.address.model.person.Person;
+import seedu.address.model.util.SampleDataUtil;
+import seedu.address.testutil.TestUtil;
+
+public class SampleDataTest extends AddressBookGuiTest {
+    @Override
+    protected AddressBook getInitialData() {
+        // return null to force test app to load data from file only
+        return null;
+    }
+
+    @Override
+    protected String getDataFileLocation() {
+        // return a non-existent file location to force test app to load sample data
+        return TestUtil.getFilePathInSandboxFolder("SomeFileThatDoesNotExist1234567890.xml");
+    }
+
+    @Test
+    public void addressBook_dataFileDoesNotExist_loadSampleData() throws Exception {
+        Person[] expectedList = SampleDataUtil.getSamplePersons();
+        assertTrue(personListPanel.isListMatching(expectedList));
+    }
+}
--- a/src/main/java/seedu/address/model/tag/UniqueTagList.java
+++ b/src/main/java/seedu/address/model/tag/UniqueTagList.java
@@ -5,6 +5,7 @@ import javafx.collections.ObservableList;
 import seedu.address.commons.core.UnmodifiableObservableList;
 import seedu.address.commons.util.CollectionUtil;
 import seedu.address.commons.exceptions.DuplicateDataException;
+import seedu.address.commons.exceptions.IllegalValueException;
 
 import java.util.*;
 
@@ -24,6 +25,18 @@ public class UniqueTagList implements Iterable<Tag> {
      * Constructs empty TagList.
      */
     public UniqueTagList() {}
+
+    /**
+     * Creates a UniqueTagList using given String tags.
+     * Enforces no nulls or duplicates.
+     */
+    public UniqueTagList(String... tags) throws DuplicateTagException, IllegalValueException {
+        final List<Tag> tagList = new ArrayList<Tag>();
+        for (String tag : tags) {
+            tagList.add(new Tag(tag));
+        }
+        setTags(tagList);
+    }
 
     /**
      * Creates a UniqueTagList using given tags.
--- a/src/test/java/seedu/address/testutil/TypicalTestPersons.java
+++ b/src/test/java/seedu/address/testutil/TypicalTestPersons.java
@@ -19,13 +19,13 @@ public class TypicalTestPersons {
                     .withTags("friends").build();
             benson = new PersonBuilder().withName("Benson Meier").withAddress("311, Clementi Ave 2, #02-25")
                     .withEmail("johnd@gmail.com").withPhone("98765432")
-                    .withTags("owesMoney", "friends").build();
+                    .withTags("colleagues", "friends").build();
             carl = new PersonBuilder().withName("Carl Kurz").withPhone("95352563")
-                    .withEmail("heinz@yahoo.com").withAddress("wall street").build();
+                    .withEmail("heinz@yahoo.com").withAddress("36, Toa Payoh Street 3, #04-13").build();
             daniel = new PersonBuilder().withName("Daniel Meier").withPhone("87652533")
-                    .withEmail("cornelia@google.com").withAddress("10th street").build();
+                    .withEmail("cornelia@google.com").withAddress("448, Bukit Merah Street 1, #03-01").build();
             elle = new PersonBuilder().withName("Elle Meyer").withPhone("9482224")
-                    .withEmail("werner@gmail.com").withAddress("michegan ave").build();
+                    .withEmail("werner@gmail.com").withAddress("332, Jurong West Street 14, #11-39").build();
             fiona = new PersonBuilder().withName("Fiona Kunz").withPhone("9482427")
                     .withEmail("lydia@gmail.com").withAddress("little tokyo").build();
             george = new PersonBuilder().withName("George Best").withPhone("9482442")
--- a/src/test/java/seedu/address/testutil/TypicalTestPersons.java
+++ b/src/test/java/seedu/address/testutil/TypicalTestPersons.java
@@ -19,13 +19,13 @@ public class TypicalTestPersons {
                     .withTags("friends").build();
             benson = new PersonBuilder().withName("Benson Meier").withAddress("311, Clementi Ave 2, #02-25")
                     .withEmail("johnd@gmail.com").withPhone("98765432")
-                    .withTags("colleagues", "friends").build();
+                    .withTags("owesMoney", "friends").build();
             carl = new PersonBuilder().withName("Carl Kurz").withPhone("95352563")
-                    .withEmail("heinz@yahoo.com").withAddress("36, Toa Payoh Street 3, #04-13").build();
+                    .withEmail("heinz@yahoo.com").withAddress("wall street").build();
             daniel = new PersonBuilder().withName("Daniel Meier").withPhone("87652533")
-                    .withEmail("cornelia@google.com").withAddress("448, Bukit Merah Street 1, #03-01").build();
+                    .withEmail("cornelia@google.com").withAddress("10th street").build();
             elle = new PersonBuilder().withName("Elle Meyer").withPhone("9482224")
-                    .withEmail("werner@gmail.com").withAddress("332, Jurong West Street 14, #11-39").build();
+                    .withEmail("werner@gmail.com").withAddress("michegan ave").build();
             fiona = new PersonBuilder().withName("Fiona Kunz").withPhone("9482427")
                     .withEmail("lydia@gmail.com").withAddress("little tokyo").build();
             george = new PersonBuilder().withName("George Best").withPhone("9482442")


damithc:
1. Is it better if we save sample data just like we do with normal data?
2. Now that creating Person objects is happening in the app, may be we can bring over the PersonBuilder class from the test code to production code?
@pyokagan any thoughts?

edmundmok:
Regarding 1, I think there would be no difference since if the user does enter a data mutating command a save file would then be created as per normal. Otherwise, if there is no change done to the sample data by the user then there is no real need to save the data. 

For 2, I considered using the `PersonBuilder` as a nested class but the current implementation of `Person` guarantees all fields to be present and not null. Wouldn't using a builder suggest that some options may be left out? (Not sure about this, I have no knowledge of the Builder pattern ...) Eventually I opted to use a private method that builds the Person in a similar manner.

Also want to check if it is okay to provide sample data upon incorrect data file format (DataConversionException) and problem reading from file (IOException).

pyokagan:
@edmundmok says:
> Wouldn't using a builder suggest that some options may be left out?

Yup. `PersonBuilder` as it is currently implemented may be appropriate for test code (since if we accidentally leave out any fields we will get an exception), but it is not suitable for the main code.

> Eventually I opted to use a private method that builds the Person in one shot.

Another option is to use a [Step builder](http://www.svlada.com/step-builder-pattern/), but it seems like a lot of code just to have *named arguments* (We can't even specify them out of order!) So really, I think for simplicity we should just go with plain ol' arguments, and ignore Codacy :-P

pyokagan:
@damithc says:
> Is it better if we save sample data just like we do with normal data?

I dunno, do you have any benefits in mind? Maybe it may be easier for non-Java devs to edit, or to edit using the application itself.

We will also need to teach our storage component how to load the sample data from our fat JAR file, I think.

What are the goals for the sample data, anyway? Do we aim to have a lot of entries? Maybe we could implement a random-person generator for this?

If not, maybe we could use our test persons data directly, thus not needing to have two sets of data?

damithc:
Objectives for sample data
1. to give new users some data to play around with from the very beginning. 
2. to show new users the GUI in its full glory (instead of a blank list) so that it makes a good first impression. That means sample data should be well chosen to showcase our features in the best possible way.

Given the above objectives, I guess we should not fall back to sample data if the data file is corrupted (because it is not a case of a 'new user').

I didn't mean we should put the sample data in a text file. Rather, once the data is injected to the GUI, they should be treated as any normal data so that the code doesn't need additional branches to deal with sample data. So we should not have rules like 'sample data are not saved'. However, as @edmundmok explained (i missed that point earlier), that's how it works now. The data will get saved if they are mutated, just like any other data. So we are good.

The `build()` method of the Builder class is supposed to do an integrity check. That is, builders should not produces invalid objects. In fact the `build()` method can call the normal many-parameter constructor that does the intergrity check anyway. Yes, that will cause a codacy warning but the warning will be limited to the builder rather than everywhere we create a person. We can configure codacy to ignore that specific instance too. 
Anyway, it's ok if we don't use the builder here.

pyokagan:
@damithc 
> The build() method of the Builder class is supposed to do an integrity check.

Right, but it converts an error (not all required fields specified) from a compile-time error into a runtime error.

... which shouldn't be a problem in practice I guess if we have proper test coverage, although I slightly prefer letting things be caught at compile-time because it means the IDE can help me more (immediately give me all the call-sites that need to be updated rather than me needing to explicitly run the tests or find out the call sites).

So from this perspective, that is why I prefer "many arguments" over the builder pattern, in the case where the majority of arguments are required.

edmundmok:
Sample data to be loaded will contain 5 people, while test data contains 7 people so that we can test that the sample data is actually being loaded and not running the default test data

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import seedu.address.model.person.Address;
+import seedu.address.model.person.Email;
+import seedu.address.model.person.Name;
+import seedu.address.model.person.Person;
+import seedu.address.model.person.Phone;
+import seedu.address.model.person.UniquePersonList;
+import seedu.address.model.tag.Tag;
+import seedu.address.model.tag.UniqueTagList;
+
+public class SampleDataUtil {
```
damithc: Currently, this class produces a person list and a tag list. Why not create an empty AddressBook and keep adding persons to it instead? Just curious.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,61 @@
...
+                new TestApp(this::getInitialData, getDataFileLocation()));
+        } else {
+            testApp = (TestApp) FxToolkit.setupApplication(() ->
+                new TestApp(TestUtil.getFilePathInSandboxFolder("somefilethatdoesnotexist1234567890.xml")));
+        }
+
+        FxToolkit.showStage();
+        while (!stage.isShowing());
+        mainGui.focusOnMainApp();
+    }
```
damithc: Sure this big chunk of code duplication cannot be avoided? May be tweak the TestApp code to facilitate loading the app without a data file (which should trigger the sample data load automatically)?

edmundmok: Hi Prof, I have tried to change the code again to prevent duplication.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: I think I just wanted the AddressBook class to be the central place to get all AddressBooks from. Perhaps it would be better to put it in the SampleDataUtil as the sample addressbook is a sample data.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import seedu.address.model.person.Phone;
+import seedu.address.model.tag.Tag;
+import seedu.address.model.tag.UniqueTagList;
+
+public class SampleDataUtil {
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        AddressBook sampleAB = new AddressBook();
+        try {
+            addSamplePersons(sampleAB);
```
damithc: Can be inlined?

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,61 @@
...
+        TestPerson[] expectedList = {ttp.alice, ttp.benson, ttp.carl, ttp.daniel, ttp.elle};
+        assertTrue(personListPanel.isListMatching(expectedList));
+    }
+
+    @Test
+    public void addressBook_dataFileExists_loadSaveFileData() throws Exception {
+        setupTestForExistingDataFile();
+        TestPerson[] expectedList = new TypicalTestPersons().getTypicalPersons();
+        assertTrue(personListPanel.isListMatching(expectedList));
+    }
```
damithc: May be no need for this test case? If this functionality is not working, many of the other GUI tests will fail anyway. If you don't have to support this test case, this test class can become simpler.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private static void addSamplePersons(AddressBook sampleAB) throws IllegalValueException {
+        sampleAB.addPerson(createPerson("Alice Pauline", "85355255", "alice@gmail.com",
+                    "123, Jurong West Ave 6, #08-111", "friends"));
+        sampleAB.addPerson(createPerson("Benson Meier", "98765432", "johnd@gmail.com",
+                    "311, Clementi Ave 2, #02-25", "owesMoney", "friends"));
+        sampleAB.addPerson(createPerson("Carl Kurz", "95352563", "heinz@yahoo.com", "wall street"));
+        sampleAB.addPerson(createPerson("Daniel Meier", "87652533", "cornelia@google.com", "10th street"));
+        sampleAB.addPerson(createPerson("Elle Meyer", "9482224", "werner@gmail.com", "michegan ave"));
```
damithc: These three guys don't have realistic addresses.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,49 @@
...
+            assert false : "not possible";
+        }
+        return sampleAB;
+    }
+
+    private static void addSamplePersons(AddressBook sampleAB) throws IllegalValueException {
+        sampleAB.addPerson(createPerson("Alice Pauline", "85355255", "alice@gmail.com",
+                    "123, Jurong West Ave 6, #08-111", "friends"));
+        sampleAB.addPerson(createPerson("Benson Meier", "98765432", "johnd@gmail.com",
+                    "311, Clementi Ave 2, #02-25", "owesMoney", "friends"));
```
damithc: replace owesMoney with a more typical tag? e.g. `colleagues`?

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Noted.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
edmundmok: Yes, that is true. Added this test case for completeness only.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,49 @@
...
+        sampleAB.addPerson(createPerson("Carl Kurz", "95352563", "heinz@yahoo.com", "wall street"));
+        sampleAB.addPerson(createPerson("Daniel Meier", "87652533", "cornelia@google.com", "10th street"));
+        sampleAB.addPerson(createPerson("Elle Meyer", "9482224", "werner@gmail.com", "michegan ave"));
+    }
+
+    private static Person createPerson(String name, String phone, String email, String address, String ... tags)
+            throws IllegalValueException {
```
edmundmok: Noted.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Updated.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Updated.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Test;
+
+import seedu.address.model.AddressBook;
+import seedu.address.testutil.TestPerson;
+import seedu.address.testutil.TestUtil;
+import seedu.address.testutil.TypicalTestPersons;
+
+public class AddressBookSampleDataGuiTest extends AddressBookGuiTest {
+
+    private static final String NONEXISTENT_FILENAME = "somefilethatdoesnotexist1234567890.xml";
```
damithc: Let's use UpperCamelCase for file names.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    protected AddressBook getInitialData() {
+        return null;
+    }
+
+    @Override
+    protected String getDataFileLocation() {
+        return TestUtil.getFilePathInSandboxFolder(NONEXISTENT_FILENAME);
```
damithc: This constant can be inlined without any loss of readability (as the file name is self-explanatory and used only here).

damithc: You can add a comment to explain why we supply a non-existent file.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Test
+    public void addressBook_dataFileDoesNotExist_loadSampleData() throws Exception {
+        TypicalTestPersons ttp = new TypicalTestPersons();
+        TestPerson[] expectedList = {ttp.alice, ttp.benson, ttp.carl, ttp.daniel, ttp.elle};
```
damithc: Is this correct? These are not sample persons.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
edmundmok: Noted.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
edmundmok: Noted.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
edmundmok: I have modified the typical test persons to be equivalent to the sample persons (with the more realistic addresses). Or should I leave the typical test persons alone?

edmundmok: The sample persons are just the first 5 of the typical test persons, so that I could just reuse the existing test persons.

damithc: @edmundmok Test data and sample data have different purposes. In test persons we might have weird data to flush out bugs while sample data is there to look nice. So  I'm not sure sharing the two is a good idea. The coupling here is even stranger because they are actually different objects that happen to have same values at this moment. So, it's better to use person objects from `SampleDataUtil` here.

edmundmok: Got it, will update.

edmundmok:
@damithc Updated sample data to be separate from test persons.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+import seedu.address.commons.util.SampleDataUtil;
+import seedu.address.model.AddressBook;
+import seedu.address.model.person.Person;
+import seedu.address.testutil.TestUtil;
+
+public class AddressBookSampleDataGuiTest extends AddressBookGuiTest {
```
damithc: The class is in `guitests` package, so no need to have `GuiTest` in class name. `SampleDataTest`?

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Override
+    protected AddressBook getInitialData() {
+        return null;
+    }
+
+    @Override
+    protected String getDataFileLocation() {
+        // override file location to provide non-existent file to test address book loading with sample data
```
damithc: `override file location` is implied by lines 19,20. 
`// return a non-existent file to force test app to load sample data`?

src/test/java/guitests/AddressBookSampleDataGuiTest.java
```diff
@@ -0,0 +1,30 @@
...
+import seedu.address.commons.util.SampleDataUtil;
+import seedu.address.model.AddressBook;
+import seedu.address.model.person.Person;
+import seedu.address.testutil.TestUtil;
+
+public class AddressBookSampleDataGuiTest extends AddressBookGuiTest {
+
+    @Override
+    protected AddressBook getInitialData() {
+        return null;
```
damithc: `// return null to force test app to load data from file`?

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,66 @@
...
+            Phone personPhone = new Phone(phone);
+            Email personEmail = new Email(email);
+            Address personAddress = new Address(address);
+            UniqueTagList personTags = new UniqueTagList();
+            for (String tag: tags) {
+                personTags.add(new Tag(tag));
+            }
+            return new Person(personName, personPhone, personEmail, personAddress, personTags);
+        } catch (IllegalValueException e) {
+            assert false : "sample data cannot be invalid";
```
pyokagan: throw an `AssertionError` here? That way you don't have to `return null`.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,66 @@
...
+        };
+        return samplePersons;
+    }
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        AddressBook sampleAB = new AddressBook();
+
+        try {
+            Person[] samplePersons = getSamplePersons();
+            for (Person samplePerson : samplePersons) {
```
pyokagan: I think you can just call `getSamplePersons()` here. No need for the `samplePersons` variable.

src/main/java/seedu/address/MainApp.java
```diff
@@ -73,9 +74,9 @@ private Model initModelManager(Storage storage, UserPrefs userPrefs) {
         try {
             addressBookOptional = storage.readAddressBook();
             if (!addressBookOptional.isPresent()) {
-                logger.info("Data file not found. Will be starting with an empty AddressBook");
+                logger.info("Data file not found. Will be starting with a sample AddressBook");
             }
-            initialData = addressBookOptional.orElse(new AddressBook());
+            initialData = addressBookOptional.orElse(SampleDataUtil.getSampleAddressBook());
```
pyokagan: This calls `SampleDataUtil.getSampleAddressBook()` at all times, even if `addressBookOptional` is present, which is slightly wasteful. Maybe use [`Optional::orElseGet()`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseGet-java.util.function.Supplier-) instead?

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Alright.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Good point.

src/main/java/seedu/address/MainApp.java
edmundmok: Wow, never knew that. Good catch.

src/test/java/guitests/AddressBookSampleDataGuiTest.java
edmundmok: Noted.

edmundmok:
Updated the PR.
@damithc @pyokagan

src/main/java/seedu/address/MainApp.java
```diff
@@ -73,9 +74,9 @@ private Model initModelManager(Storage storage, UserPrefs userPrefs) {
         try {
             addressBookOptional = storage.readAddressBook();
             if (!addressBookOptional.isPresent()) {
-                logger.info("Data file not found. Will be starting with an empty AddressBook");
+                logger.info("Data file not found. Will be starting with a sample AddressBook");
             }
-            initialData = addressBookOptional.orElse(new AddressBook());
+            initialData = addressBookOptional.orElseGet(() -> SampleDataUtil.getSampleAddressBook());
```
pyokagan: Does `SampleDataUtil::getSampleAddressBook` work?

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } catch (DuplicatePersonException e) {
+            assert false : "sample data cannot contain duplicate persons";
+        }
+        return sampleAB;
+    }
+
+    private static Person createPerson(String name, String phone, String email, String address, String ... tags)
+           throws AssertionError {
+
```
pyokagan: Don't need this blank line?

src/test/java/guitests/SampleDataTest.java
```diff
@@ -0,0 +1,31 @@
...
+        // return a non-existent file location to force test app to load sample data
+        return TestUtil.getFilePathInSandboxFolder("SomeFileThatDoesNotExist1234567890.xml");
+    }
+
+    @Test
+    public void addressBook_dataFileDoesNotExist_loadSampleData() throws Exception {
+        Person[] expectedList = SampleDataUtil.getSamplePersons();
+        assertTrue(personListPanel.isListMatching(expectedList));
+    }
+}
```
pyokagan: If you have a blank line just after the `public class ... {` at the top, then there should be a blank line here as well.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,66 @@
...
+                        "Blk 47 Tampines Street 20, #17-35", "classmates"),
+                createPerson("Roy Balakrishnan", "92624417", "royb@gmail.com",
+                        "Blk 45 Aljunied Street 85, #11-31", "colleagues")
+        };
+        return samplePersons;
+    }
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        AddressBook sampleAB = new AddressBook();
+
```
pyokagan: Unneeded blank line? (Not symmetric with the `return sampleAB;` below)

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,66 @@
...
+                sampleAB.addPerson(samplePerson);
+            }
+        } catch (DuplicatePersonException e) {
+            assert false : "sample data cannot contain duplicate persons";
+        }
+        return sampleAB;
+    }
+
+    private static Person createPerson(String name, String phone, String email, String address, String ... tags)
+           throws AssertionError {
```
pyokagan: Interestingly this seems to only be indented by 11 spaces? Maybe make this 12 spaces?

src/main/java/seedu/address/MainApp.java
edmundmok: Yes, looks neater too.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Noted.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Noted.

src/test/java/guitests/SampleDataTest.java
edmundmok: Will remove the above blank line.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Sorry about that, updated to 12.

edmundmok:
@pyokagan Updated. Thanks!

src/test/java/guitests/SampleDataTest.java
pyokagan: @edmundmok 
> Will remove the above blank line.

There's still a blank line?

edmundmok: Oh wow, just updated and removed it. Sorry about that.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,64 @@
...
+            for (Person samplePerson : getSamplePersons()) {
+                sampleAB.addPerson(samplePerson);
+            }
+        } catch (DuplicatePersonException e) {
+            assert false : "sample data cannot contain duplicate persons";
+        }
+        return sampleAB;
+    }
+
+    private static Person createPerson(String name, String phone, String email, String address, String ... tags)
```
pyokagan: For our code style, the `...` attaches to the type e.g. `String... tags`

edmundmok: Noted.

edmundmok:
@pyokagan Fixed the `String...` issue. Thanks!

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,64 @@
...
+            for (Person samplePerson : getSamplePersons()) {
+                sampleAB.addPerson(samplePerson);
+            }
+        } catch (DuplicatePersonException e) {
+            assert false : "sample data cannot contain duplicate persons";
+        }
+        return sampleAB;
+    }
+
+    private static Person createPerson(String name, String phone, String email, String address, String... tags)
```
pyokagan: Wait... isn't this dangerous? What happens if we add a new field after `address` but before `tags`, but forget to properly update the sample data of a person above? Then one tag of that person will accidentally become the value of the new field.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,64 @@
...
+            throws AssertionError {
+        try {
+            Name personName = new Name(name);
+            Phone personPhone = new Phone(phone);
+            Email personEmail = new Email(email);
+            Address personAddress = new Address(address);
+            UniqueTagList personTags = new UniqueTagList();
+            for (String tag: tags) {
```
pyokagan: Oh, this should be `String tag : tags` (space between the `:` )

src/main/java/seedu/address/commons/util/SampleDataUtil.java
```diff
@@ -0,0 +1,64 @@
+package seedu.address.commons.util;
```
pyokagan: This is in the `commons` package, but it depends on the `model` package, which violates our architectural design. This class should probably be in the `model` package instead.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
pyokagan: > Wait... isn't this dangerous?

While having a `createPerson()` helper method would make the `getSamplePersons()` code look slightly neater, I think it will open the door to trouble if we go about adding/reordering person fields in the future.

So, I think that it may actually be better if we go the straightforward way and just implement `getSamplePersons()` using code like:
```java
        try {
            return new Person[] {
                new Person(new Name("Alex Yeoh"), new Phone("87438807"), new Email("alexyeoh@gmail.com"),
                        new Address("Blk 30 Geylang Street 29, #06-40"),
                        new UniqueTagList("friends")),
                new Person(new Name("Bernice Yu"), new Phone("99272758"), new Email("berniceyu@gmail.com"),
                        new Address("Blk 30 Lorong 3 Serangoon Gardens, #07-18"),
                        new UniqueTagList("colleagues", "friends")),
...
            };
        } catch (IllegalValueException e) {
            throw new AssertionError("cannot be invalid", e);
        }
```
The presence of all the `new Name()`, `new Phone()` etc. makes the code look slightly messier. BUT, it makes the code self-documenting (we immediately know which strings are the name, phone, email etc.) It is also much more type safe -- the compiler will immediately tell us if we accidentally try to put a name into a phone field, for example.

We'll also need a new `UniqueTagList` convenience constructor, of course. Something like:
```java
    public UniqueTagList(String... tags) throws IllegalValueException, DuplicateTagException {
        final List<Tag> tagList = new ArrayList<>();
        for (String tag : tags) {
            tagList.add(new Tag(tag));
        }
        setTags(tagList);
    }
```

edmundmok: Hmm, that's true about the new field and reordering etc. Will update it to the recommended approach.

src/main/java/seedu/address/commons/util/SampleDataUtil.java
edmundmok: Noted.

edmundmok:
@pyokagan Updated as recommended. Thanks!

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -0,0 +1,55 @@
...
+                    new UniqueTagList("family")),
+                new Person(new Name("Irfan Ibrahim"), new Phone("92492021"), new Email("irfan@outlook.com"),
+                    new Address("Blk 47 Tampines Street 20, #17-35"),
+                    new UniqueTagList("classmates")),
+                new Person(new Name("Roy Balakrishnan"), new Phone("92624417"), new Email("royb@gmail.com"),
+                    new Address("Blk 45 Aljunied Street 85, #11-31"),
+                    new UniqueTagList("colleagues"))
+            };
+        } catch (IllegalValueException e) {
+            throw new AssertionError("sample data cannot be invalid");
```
pyokagan: Set the `IllegalValueException` as the cause of the `AssertionError`, i.e.
```java
throw new AssertionError("sample data cannot be invalid", e);
```
to make it easier for devs to diagnose the problem if it ever happens.

edmundmok: Sounds good.

edmundmok:
@pyokagan Updated. Thanks for the review.

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        AddressBook sampleAB = new AddressBook();
+        try {
+            for (Person samplePerson : getSamplePersons()) {
+                sampleAB.addPerson(samplePerson);
+            }
+        } catch (DuplicatePersonException e) {
+            assert false : "sample data cannot contain duplicate persons";
```
damithc: Any reason why we are using `assert` here but `AssertionError` in the method above?

pyokagan: @damithc 
> Any reason why we are using assert here but AssertionError in the method above?

The usage above is so that we will not need to return a bogus value.

I personally would prefer `AssertionError` to be used here as well since we are handling an exception, but it's not required since we don't need to return any value here, although I guess it may be bad if we are swallowing exceptions if assertions are disabled. (But then again, assertions would only be disabled in production, and we would have caught this assertion error during testing, so it may not be a problem in practice. In summary, I don't feel strongly about this.)

damithc: @pyokagan I meant like this:

```diff
diff --git a/src/main/java/seedu/address/model/util/SampleDataUtil.java b/src/main/java/seedu/address/model/util/SampleDataUtil.java
index acd7257..83c4ecd 100644
--- a/src/main/java/seedu/address/model/util/SampleDataUtil.java
+++ b/src/main/java/seedu/address/model/util/SampleDataUtil.java
@@ -47,7 +47,7 @@ public class SampleDataUtil {
                 sampleAB.addPerson(samplePerson);
             }
         } catch (DuplicatePersonException e) {
-            assert false : "sample data cannot contain duplicate persons";
+            throw new AssertionError("sample data cannot have duplicate persons", e);
         }
         return sampleAB;
     }
```

That shouldn't require a bogus return value right?

pyokagan: @damithc 
> I meant like this:

Yeah, I understood you in your first message. `The usage above ` in my message was referring to `getSamplePersons()`, sorry.

pyokagan: > In summary, I don't feel strongly about this.

Actually, now that I thought it through over dinner, I think I'll feel strongly about this :-P If we are responding to an exception that is already thrown, then we should definitely throw an `AssertionError` regardless of whether assertions are enabled or not.

damithc: > If we are responding to an exception that is already thrown, then we should definitely throw an AssertionError regardless of whether assertions are enabled or not.

I agree. @edmundmok one more change to do :-)

edmundmok: Sorry for the late reply! So in general, if we caught an impossible exception like in this scenario, it would be more appropriate to handle it with an assertionError instead of just an assert? Just to confirm as there are a few other instances in the code where simply an assert is used.

edmundmok:
@damithc @pyokagan Updated as recommended. Thanks!

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static ReadOnlyAddressBook getSampleAddressBook() {
+        try {
+            AddressBook sampleAB = new AddressBook();
+            for (Person samplePerson : getSamplePersons()) {
+                sampleAB.addPerson(samplePerson);
+            }
+            return sampleAB;
+        } catch (DuplicatePersonException e) {
+            throw new AssertionError("sample data cannot contain duplicate persons", e);
```
edmundmok: @damithc @pyokagan Reasking this in case the other reply on the old commit gets hidden: So in general, if we caught an impossible exception like in this scenario, it would be more appropriate to handle it with an assertionError instead of just an assert? Just to confirm as there are a few other instances in the rest of the code where simply an assert is used.

edmundmok: @damithc @pyokagan Reasking this in case the other reply on the old commit gets hidden: So in general, if we caught an impossible exception like in this scenario, it would be more appropriate to handle it with an assertionError instead of just an assert? Just to confirm as there are a few other instances in the rest of the code where simply an assert is used.

damithc: @edmundmok I'm ok with that convention.

pyokagan: @edmundmok 
> So in general, if we caught an impossible exception like in this scenario, it would be more appropriate to handle it with an assertionError instead of just an assert?

Yes. `AssertionError` will always be thrown, while `assert` will only work when it is enabled. So `assert` should only be used for "extra checks". In this case, we are already performing the "duplicate person" check, so there is no point using `assert` since we are not gaining any performance or anything. In fact, it is worse as if `assert` is disabled, then we are essentially ignoring this exception which is clearly signaling that something is wrong.

pyokagan:
@damithc This PR should be ready now.

damithc:
Excellent work @edmundmok 👍 
Minor nit for future reference: The PR title should be a copy-paste of issue title, including the issue number.

PR #238, opened 2016-12-27
 Check for line endings in Travis CI #195 

### LINKED ISSUE #195: Check for line endings in Travis CI
Once #194 is merged, we can then teach the CI to check that all line endings in the repo are LF, to make sure that other line endings will not sneak into the code base.

cc @damithc , @m133225

pyokagan:
Fixes #195 

In our repository, we want:

* All text files to use only LF ('\n') as their line endings. Not CRLF
  ('\r\n')

* All text files to end with a newline (Newline at EOF).

So, ideally, we should have checks that automatically check for these
requirements in our repository.

However, we can't impose these checks on all developers, namely Windows
developers. They could be using either LF or CRLF (e.g. if the developer
activated Git's core.autocrlf configuration). Even if we did, we can't
guarantee that developers do not bypass these checks. As such, we have
to implement these checks on the Travis CI level instead.

The next question is how do we implement these checks. Checkstyle is
already helping us with that, but it is not completely effective as it
does not check *all* text files -- it only checks the source files which
are used for compilation.  This means that it does not check things like
resource files (CSS and FXML files), test data and documentation.

While we could probably teach it to check all files, we meet another
problem; Checkstyle does not know what a text file is. At best, we can
only give it a filter of file extensions to check.

On the other hand, Git knows the difference between text files and
binary files. As such, let's implement some shell scripts that will use
Git's facilities to implement our line ending checks on all text files.
These checks will be run on Travis CI, thus guaranteeing that master
will not have any more line ending problems.

These checks are implemented in POSIX shell scripts so that Travis CI
would not need to download any additional dependencies in order to run
them. Furthermore, developers running Linux or Macs can also run them
locally without downloading additional dependencies as well. Windows
devs would be unable to run them, but then again these line ending
checks are not useful to them because as mentioned above, Windows devs
may be using CRLF line endings anyway.

While our checkstyle configuration already checks for trailing
whitespace in our java files, it does not cover all text files in our
repository. Notably: our documentation files and other configuration
files such as `build.gradle`.

While we could probably coax checkstyle and gradle to check for these
files, it is much easier to just hook onto our existing "Travis CI
checks" architecture to implement this check.

Trailing whitespace should be considered an error, however we treat
trailing whitespace in markdown (`*.md`) files as warnings instead to
accomodate developers who prefer to edit documentation using the GitHub
web interface which does not provide facilities to easily detect and
remove trailing whitespace.

---
This change is [Reviewable](https://reviewable.io/reviews/se-edu/addressbook-level4/238)


pyokagan:
v1 Ready for review.

cc @m133225

damithc:
@pyokagan go ahead and merge. @m133225 can take a look later when he resurfaces after holidays.
Should this (i.e. your explanation above) be documented somewhere, may be in dev guide?

pyokagan:
@damithc 
> Should this (i.e. your explanation above) be documented somewhere, may be in dev guide?

Hrm, the explanation is about the thought process which led to the implementation in this PR. I think that is irrelevant to the dev guide, which documents current implementation details?

But yes, I could probably add some notes about the implementation inside `UsingTravis.md` or something.

pyokagan:
@pyokagan
> The author considered implementing "trailing whitespace" checks as
> well, but noticed that many of the current "trailing whitespace" errors
> are in the documentation files,

Hmm... this isn't actually a very good reason for not implementing the "trailing whitespace" checks. There are still other files not covered by checkstyle that would benefit from it, such as the CSS and FXML files...

So maybe I could implement the "trailing whitespace" checks, but just make any errors in the `*.md` files warnings or something...

Will submit an updated version.

damithc:
@pyokagan 

>Hrm, the explanation is about the thought process which led to the implementation in this PR. I think that is irrelevant to the dev guide, which documents current implementation details?

True, the dev guide is mostly about HOW the current design works. It's still good to have at least a pointer to WHY (the rationale) too. It's a matter of time before someone else comes along and says "this is stupid, we should get rid of it (something like that)" and then the justification can come in handy. :-p

pyokagan:
@damithc How about this?

---

Review status: 0 of 26 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/se-edu/addressbook-level4/238#-:-K_3sy-_xNunSymlgplk:bhvp0vt)*

damithc:
Reviewed 15 of 15 files at r1, 4 of 4 files at r2, 8 of 8 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[docs/UsingTravis.md, line 73 at r5](https://reviewable.io:443/reviews/se-edu/addressbook-level4/238#-K_44NLcX3VQ0_QbQVIq:-K_44NLcX3VQ0_QbQVIr:b4upvp8) ([raw file](https://github.com/se-edu/addressbook-level4/blob/9c29f8ed2c67b326b8bbeec88abe95b28bebbd57/docs/UsingTravis.md#L73)):*
> ```Markdown
> The `run-checks.sh` script will automatically pick up and run files named as such.
> 
> Check scripts should print out errors in the following format:
> ```

Dummy comment, please ignore.

---

*[docs/UsingTravis.md, line 82 at r5](https://reviewable.io:443/reviews/se-edu/addressbook-level4/238#-K_43ZRIn6cCjeCrMJUM:-K_43ZRJ2b5QlZGrzrR1:bgmxtv5) ([raw file](https://github.com/se-edu/addressbook-level4/blob/9c29f8ed2c67b326b8bbeec88abe95b28bebbd57/docs/UsingTravis.md#L82)):*
> ```Markdown
> and `MESSAGE` is the message explaining the error.
> 
> Check scripts must exit with a non-zero exit code if any errors occur.
> ```

:+1: yup, this should do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/se-edu/addressbook-level4/238)*

damithc:
:lgtm:

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/se-edu/addressbook-level4/238#-:-K_45Rv5DU6UbP1Cj-sa:bnfp4nl)*

pyokagan:
@damithc I just noticed that GitHub assigned you as the author for the squash commit d782898510a12b3b03b07848cb7a647f3076e241 :O Is that a GitHub bug?

damithc:
@pyokagan 

> I just noticed that GitHub assigned you as the author for the squash commit d782898 :O Is that a GitHub bug?

Possibly. Previously they used to mention all authors. e.g. 15d91fd909ad422aaa0cc13039cbb297c199b448

If this continues, it could affect mentionbot's effectiveness.

pyokagan:
@damithc Oh sorry, I did not notice that you merged it regularly rather than using the "squash and merge" option. In that case, that's the correct behavior. Sorry for the noise.

damithc:
I forgot about it myself :-p 
Yes, now I remember that I did a normal merge because the branch was clean and I wanted to preserve the individual commits. That's better than a squash, right? Or do you think a fast-forward is even better?

pyokagan:
@damithc 
> That's better than a squash, right?

Yes, it's better. It makes `git-blame` and `git-bisect` more fine-grained.

> Or do you think a fast-forward is even better?

Well, I think not, because it destroys the structure of topic branches (e.g. you cannot see where a PR's commits start and end in the commit history if you draw it as a graph).

PR #209, opened 2016-12-11
User can edit a person's details #141

Zhiyuan-Amos:
Fixes #141 

Ready for review.

* [1/6] [create constants for keys (to be used in HashMap for EditCommand) and make Person mutable.](https://github.com/se-edu/addressbook-level4/pull/209/commits/2df9e1b9427c95c334680f19c553c9755be14bf7)
* [2/6] [create EditCommand & update Parser and ArgumentTokenizer.](https://github.com/se-edu/addressbook-level4/pull/209/commits/24daab843f7baee1378612e049d2076038d61e13)
* [3/6] [update Model package (ModelManager, Model, AddressBook, UniquePersonList) to allowing editing.](https://github.com/se-edu/addressbook-level4/pull/209/commits/1270bdb4998f36e39bb1670b12384b6d61db1a55)
* [4/6] [create NoArgumentException](https://github.com/se-edu/addressbook-level4/pull/209/commits/029cd8779cb0dca7874c3c711f0ca70078925a4c)
* [5/6] [update User Guide to include edit.](https://github.com/se-edu/addressbook-level4/pull/209/commits/30a5c309ae36ad47b1e3dcf78b0da768ad6f3217)
* [6/6] [create EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/209/commits/aa14f2354368ddd9ba44e4fd2eaef99b1d3b336d)

Functionalities:
1. User can remove existing tags by typing "t/" without any tags behind it.
2. User can edit existing fields. Syntax of edit is similar to that of add (refer to UG), but you do not have to necessarily edit all the fields.

Issue #169, opened 2016-11-01
Discussion: Prepare Methods in Parser class in violation of SRP?


mauris:
While reviewing code, I find that the `Parser` class in Level 4 has prepare methods for each command. It feels like this is a violation of SRP and does not follow open-close principle. Adding new commands would require addition of such methods. 

If we were to refactor this, would following the Factory pattern - by having one concrete factory for each concrete command - be better?

damithc:
Yes, that's a good observation @mauris 
#137 is supposed to fix that problem.

mauris:
Right, couldn't find the corresponding issue. Thank you Prof @damithc!

PR #247, opened 2017-01-14
Split Parser into command-specific classes #137

### LINKED ISSUE #137: Split Parser into command-specific classes
To make it more compliant with the open-closed principle

MightyCupcakes:
Closes #137

This PR will split the individual parsers for each commands into their command-specific classes. 

With this change, a command parser registry will be added to contain the list of all registered commands. The registry is essentially a Map mapping `String` which are command words to a `CommandParser` which are the command specific classes.


CanIHasReview:
# v1

@MightyCupcakes submitted v1 for review.

* [v1 1/13] [Implement the abstract command parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/4db60af78fe204b38d8f74f843d39b4b8e381045)
* [v1 2/13] [Implement the generic command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/a46981f3ad3bbdda57106f6cf3aae5f8d1d494af)
* [v1 3/13] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/8da4f5c94b12c8c8720261163891f6dc961ea1d1)
* [v1 4/13] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/c33aa437e937f3edb20c81513f040f33d5a07321)
* [v1 5/13] [Split the parser for add command into AddCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/08cd392472feb2fbf56185e3c90d50db64cfd06d)
* [v1 6/13] [Split the parser for delete command into DeleteCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/3b890598f2705d4f749259a8536071ea113d49b4)
* [v1 7/13] [Split the parser for find command into FindCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/dd535e676170f8d2d887c1c249d1b29b2186d0f1)
* [v1 8/13] [Split the parser for select command into SelectCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/54b4b679cb4eb5856fe56e7e58309ffb5f510bd9)
* [v1 9/13] [Modify Parser class to register all existing commands ](https://github.com/se-edu/addressbook-level4/pull/247/commits/dba8db22b1ef69d4b44fccdf287d289bc1e448f1)
* [v1 10/13] [Modify Parser.java to use the parser registry ](https://github.com/se-edu/addressbook-level4/pull/247/commits/ab17fac91e6f40abcddf4936d6a4541ef0daf0c1)
* [v1 11/13] [Remove unused methods in Parser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f4aba0d59366f16bbbcea0b40a75d38adba1db5e)
* [v1 12/13] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/9b903f37fc53885f5f53de6186d593f4f766850f)
* [v1 13/13] [Implement test cases ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f08933e30a8dfdbad57e3ae0245083a6ee1008aa)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/32643f6b7435056071181de2bd34055bb3b05cc4...f08933e30a8dfdbad57e3ae0245083a6ee1008aa))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/1/head:BRANCHNAME
```

</details>

damithc:
@chao1995 @yamgent can peer review this PR? Use [this PR](https://github.com/se-edu/addressbook-level4/pull/209) as an example on how to review a multi-step PR incrementally.

pyokagan:
@MightyCupcakes Next time wait for all the checks to pass first before submitting a new iteration.

pyokagan:
@MightyCupcakes Also the commit message body needs to be wrapped at 72 characters. See https://github.com/oss-generic/process/blob/master/docs/FormatsAndConventions.md#commit-message

src/main/java/seedu/address/logic/parser/CommandParser.java
```diff
@@ -0,0 +1,54 @@
...
+            Pattern.compile("(?<keywords>\\S+(?:\\s+\\S+)*)"); // one or more keywords separated by whitespace
+
+    /*
+     * Sets the command word for this parser to the given command word.
+     * This is mainly used for parsers that parses more than
+     * one command.
+     */
+    public void setCommandWord(String commandWord) {
+        // NO-OP
+    }
```
yamgent: Sorry, I didn't really understand the purpose of this function, namely:

1. Why would `setCommandWord()` be called? Will the parser itself not know its own command?
1. Is no-op a reasonable default behaviour? If not, would it be possible to abstract it?
1. How will parsers that have more than one command override this command, since it only accepts one `commandWord`?

src/main/java/seedu/address/logic/parser/CommandParser.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Parses user input arguments in the context of the respective command classes
+ */
+public abstract class CommandParser {
+
+    public static final Prefix PREFIX_PHONE = new Prefix("p/");
+    public static final Prefix PREFIX_EMAIL = new Prefix("e/");
+    public static final Prefix PREFIX_ADDRESS = new Prefix("a/");
+    public static final Prefix PREFIX_TAG = new Prefix("t/");
```
yamgent: Seems to be an odd place to put these (makes `CommandParser` not "reusable" for non-addressbook applications).

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
```diff
@@ -0,0 +1,69 @@
...
+import seedu.address.logic.commands.HelpCommand;
+import seedu.address.logic.commands.IncorrectCommand;
+import seedu.address.logic.commands.ListCommand;
+
+/**
+ * The generic command parser class for all commands that do not require any
+ * arguments.
+ *
+ */
+public class GenericCommandParser extends CommandParser {
```
yamgent: Maybe this could be better named (e.g. `NoArgumentsCommandParser`?), as `Generic` doesn't really convey the meaning that the parser does not accept any arguments (had to read the comment to understand what it meant).

src/main/java/seedu/address/logic/parser/CommandParser.java
MightyCupcakes: This is mainly used for the NoArgumentsCommandParser where it is repsonsible for handling multiple commands. Making it abstract is not ideal as it will introduce too much boilerplate code in other command parser classes.

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
MightyCupcakes: Good point

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+    *
+    * Registers all associated command word strings with the provided
+    * command parser class. One command parser can be associated with
+    * multiple command words. However one command word can only be associated with one command parser.
+    *
+    * @return this ParserRegistry object for daisy chain
+    */
+    public ParserRegistry registerCommand(Class<? extends CommandParser> parser, Iterable<String> commandWords) {
```
yamgent: You could use varags (`String... commandWords`) to save the trouble of having to maintain two different versions of the same method.

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private static final Logger logger = LogsCenter.getLogger(GenericCommandParser.class);
+
+    private String commandWord = "";
+
+    static {
+        Builder<String, Class<? extends Command>> builder = new Builder<>();
+
+        builder
+            .put(ListCommand.COMMAND_WORD, ListCommand.class)
```
chao1995: Here the builder is not DRY as we already register these NoArgument commands in Parser once, through registry. We need this builder due to the fact that this parser needs to handle multiple commands, but we want to split the parser in the first place? So I suggest we don't use another map here, just create concrete subclasses of GenericCommandParser (same as I suggested in the abstract parser class)

yamgent:
After some thoughts, I feel that maybe we can simplify the design a bit.

Perhaps the `NoArgumentsCommandParser` class can accept a command as a constructor parameter, and then we can assign different command classes through the constructor. So it will end up be something like:

```
.registerCommand(new NoArgumentsCommandParser(ListCommand.class), ListCommand.COMMAND_WORD)
.registerCommand(new NoArgumentsCommandParser(ClearCommand.class), ClearCommand.COMMAND_WORD)
```

By doing so, we would eliminate the need for having an `ImmutableMap` inside `NoArgumentsCommandParser`, while still avoiding boilerplate codes.

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+     * @return this ParserRegistry object for daisy chain
+     */
+    public ParserRegistry registerCommand(Class<? extends CommandParser> parser, String commandWord) {
+        commandRegistry.put(commandWord, parser);
+
+        return this;
+    }
+
+    /**
+    *
```
chao1995: Need one more space indentation for the comment.

src/main/java/seedu/address/logic/parser/CommandParser.java
chao1995: As for now, only AddCommandParser uses these prefixes in ArgumentTokenizer. Maybe we can move these prefixes to ArgumentTokenizer?

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
chao1995: Here the builder is not DRY as we already register these NoArgument commands in Parser once, through registry. We need this builder due to the fact that this parser needs to handle multiple commands, but we want to split the parser in the first place? So I suggest we don't use another map here, just create concrete subclasses of GenericCommandParser (same as I suggested in the abstract parser class)

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
```diff
@@ -0,0 +1,69 @@
...
+    public void setCommandWord(String commandWord) {
+        this.commandWord = commandWord;
+    }
+
+    @Override
+    public Command prepareCommand(String args) {
+
+        Class<? extends Command> commandClass = COMMANDS.get(commandWord);
+
+        if (commandClass == null) {
```
chao1995: Will commandClass ever be null? If it's null, the parser registry should not return a GenericCommandParser for this commandWord right?

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
chao1995: Here the builder is not DRY as we already register these NoArgument commands in Parser once, through registry. We need this builder due to the fact that this parser needs to handle multiple commands, but we want to split the parser in the first place? So I suggest we don't use another map here, just create concrete subclasses of GenericCommandParser (same as I suggested in the abstract parser class)

src/main/java/seedu/address/logic/parser/GenericCommandParser.java
chao1995: Will commandClass ever be null? If it's null, the parser registry should not return a GenericCommandParser for this commandWord right?

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @return this ParserRegistry object for daisy chain
+     */
+    public ParserRegistry registerCommand(Class<? extends CommandParser> parser, String commandWord) {
+        commandRegistry.put(commandWord, parser);
+
+        return this;
+    }
+
+    /**
```
chao1995: Need one more space indentation for the comment.

src/main/java/seedu/address/logic/parser/ParserRegistry.java
chao1995: Need one more space indentation for the comment.

src/main/java/seedu/address/logic/parser/CommandParser.java
MightyCupcakes: If we do that we can get rid of NoArgumentsCommandParser and have all of them inherit from CommandParser directly.

src/main/java/seedu/address/logic/parser/CommandParser.java
chao1995: As for now, only AddCommandParser uses these prefixes in ArgumentTokenizer. Maybe we can move these prefixes to ArgumentTokenizer?

src/main/java/seedu/address/logic/parser/ParserRegistry.java
chao1995: Need one more space indentation for the comment.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -31,6 +31,25 @@
...
+    private ParserRegistry commandParserRegistry;
+
+    public Parser() {
+        commandParserRegistry = new ParserRegistry();
+        registerCommandParsers();
+    }
+
+    private void registerCommandParsers() {
+        commandParserRegistry
+            .registerCommand(AddCommandParser.class, AddCommand.COMMAND_WORD)
```
chao1995: Just to be consistent with GenericCommandParser builder, we can put the COMMNAD_WORD first, e.g.
```
commandParserRegistry.registerCommand(AddCommand.COMMAND_WORD, AddCommandParser.class);
```

src/main/java/seedu/address/logic/parser/CommandParser.java
chao1995: Same thought as @yamgent, this setCommandWord method conflicts with the fact we are already using command registry (which at core is a map) to create parsers for command words. So I would suggest removing this method and create the concrete parser classes for commands like list and help and just make them inherit from NoArgumentsCommandParser.

chao1995: I'm OK with that if the subclasses don't have command codes to share.

src/main/java/seedu/address/logic/parser/CommandParser.java
MightyCupcakes: ArgumentTokenizer is a general class that accepts a string with the set Prefixes and processes them. Individual prefixes should not be defined in the class at all. In the latest edit command PR, the edit command will also make use of these prefixes so it is not accurate to say only AddCommandParser uses these prefixes. In fact these prefixes can be used in many other commands too.

MightyCupcakes:
@yamgent 

That would not work as the registerParser method expects a class object itself, not an instance of a class

pyokagan:
@chao1995 @yamgent Just FYI, if you all have no more comments for this PR iteration, you can replace the `s.ToReview` label with the `s.Ongoing` label to signal to @MightyCupcakes that you all have no more comments.

Likewise, @CanIHasReview will automatically set the `s.ToReview` label to signal to y'all that the next iteration is ready for review.

chao1995:
@yamgent I have no more comments.

yamgent:
@MightyCupcakes okay sure, just a suggestion haha

CanIHasReview:
# v2

@MightyCupcakes submitted v2 for review.

* [v2 1/12] [Implement the abstract command parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/6778b02b355b576fe29aaf6028e43469b6d37350)
* [v2 2/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/21806c79f2d3ca4ec68690af43148c0e365ba05a)
* [v2 3/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/33777536e8dcd53631f513501a48c3d6f9257b6a)
* [v2 4/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/3a1653196b52cc1088d419c7cd1c95a7727abc8a)
* [v2 5/12] [Split the parser for add command into AddCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/1ad213c3ce5d4d8c4724f95ee36a14d6fedd6b29)
* [v2 6/12] [Split the parser for delete command into DeleteCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/aa63c67f954c6727ee10066c6eb574a31fd53401)
* [v2 7/12] [Split the parser for find command into FindCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/cd323e39c330b3762e384ee9d438fc053c69b7d6)
* [v2 8/12] [Split the parser for select command into SelectCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/d2336b991c5107f1996dfaafb4905c311ae218e5)
* [v2 9/12] [Split the parser for edit command into EditCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/b545bf67ce8a8cb92cd9912db5579f3bc7c9b1e2)
* [v2 10/12] [Modify Parser.java to use ParserRegistry and removed unused methods ](https://github.com/se-edu/addressbook-level4/pull/247/commits/93882126f723a971aec3e6cae0a2f627b8756dad)
* [v2 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/fefa566142b5735d718d2bc9172a03663ca1cbd7)
* [v2 12/12] [Implement test cases ](https://github.com/se-edu/addressbook-level4/pull/247/commits/7e1faad10fc802c3bc6d4b63e674984dfdffc02b)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/97c34f2b96ec34b3afd93876f3eda12455759c91...7e1faad10fc802c3bc6d4b63e674984dfdffc02b)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/commit/d2866375e891cd3ac58f26d09324755131d6b692))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/2/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/CommandParser.java
```diff
@@ -0,0 +1,48 @@
...
+
+import seedu.address.commons.util.StringUtil;
+import seedu.address.logic.commands.Command;
+
+/**
+ * Parses user input arguments in the context of the respective command classes
+ */
+public abstract class CommandParser {
+
+    protected static final Pattern PERSON_INDEX_ARGS_FORMAT = Pattern.compile("(?<targetIndex>.+)");
```
yamgent: Can remove the word `PERSON` (so that it becomes `INDEX_ARGS_FORMAT`).

src/main/java/seedu/address/logic/parser/CommandParser.java
```diff
@@ -0,0 +1,48 @@
...
+        return Optional.of(Integer.parseInt(index));
+
+    }
+
+    protected Set<String> toSet(Optional<List<String>> tagsOptional) {
+        List<String> tags = tagsOptional.orElse(Collections.emptyList());
+        return new HashSet<>(tags);
+    }
+
+    public abstract Command prepareCommand(String args);
```
yamgent: A header comment can be added for this method.

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+public class ParserRegistry {
+
+    protected final Map<String, Class<? extends CommandParser>> commandRegistry;
+
+    public ParserRegistry() {
+        commandRegistry = Maps.newHashMap();
+    }
+
+    /**
+     *
```
yamgent: The blank line at the top/bottom of the javadocs can be removed.

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+import seedu.address.model.tag.Tag;
+import seedu.address.model.tag.UniqueTagList;
+
+/**
+ * Parses input arguments in the context of the Edit command
+ */
+public class EditCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
```
yamgent: I feel that this method is rather long (there are a lot of different level of details in this method). Perhaps the SLAP can be further improved here.

MightyCupcakes: This is outside the scope of this PR as it is parser method for the edit command

yamgent: Ok, thanks for the heads up!

We can raise it as an issue after this PR is merged.

yamgent:
@chao1995 I am done with my reviews for the current iteration. Feel free to change to `s.Ongoing` after you are done. :)

pyokagan:
@MightyCupcakes The commit messages in this PR are quite empty. Please fill them up.

I expect to see some explanation on why you think the approach you took in this PR is the best one. Have you (or @yamgent or @chao1995 ) considered any alternatives? Also, a brief description on why the current Parser code is so bad such that it needs to be replaced would be appreciated.

Also, #137 talks about the open-closed principle and a related issue ( #169 ) mentions SRP. I expect to see those terms being used as well, and an explanation of how the approach in this PR adheres to those.

Thanks.

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,57 @@
...
+     * Registers all command words string with the provided
+     * command parser class. One command parser can be associated with
+     * multiple command words. However one command word can only be associated with one command parser.
+     *
+     * @return this ParserRegistry object for daisy chain
+     */
+    public ParserRegistry registerCommand(Class<? extends CommandParser> parser, String... commandWords) {
```
chao1995: As we have `protected final Map<String, Class<? extends CommandParser>> commandRegistry` which maps String to CommandParser, I think this `registerCommand` method can take commandWords first, then parser, which is more consistent at least to the developer.

MightyCupcakes: We are using Vaargs for registerCommand, therefore the String parameter must be the last parameter.

We can however split it up into two overloaded methods - one accepting String and another Iterable<String>; that way we can use your suggestion - but we now have to maintain two methods instead of one.

CanIHasReview:
# v3

@MightyCupcakes submitted v3 for review.

* [v3 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/d2d9a39db5ff6c21c0cd0892dcd2498bbdb0dbb2)
* [v3 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/2e1b19d6931588215e01de8c3c7a5c537a0306fe)
* [v3 3/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/3d1c74c95cda80d7661c856bf3e4bee058470019)
* [v3 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/51bec5bb3c70c9cc0e964ef126346f81308ee6f0)
* [v3 5/12] [Split the parser for add command into AddCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f8dc94696e1aadeaf47851b439a87647317cc404)
* [v3 6/12] [Split the parser for delete command into DeleteCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/afd4aaeca6b9614caa2c93ba37ceb96ad76c191b)
* [v3 7/12] [Split the parser for find command into FindCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/882ad3283792beaff8a76a1da8d1ea3e026c804c)
* [v3 8/12] [Split the parser for select command into SelectCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/c94d882304f514408d3b9c2f4021423d5fa0ac75)
* [v3 9/12] [Split the parser for edit command into EditCommandParser.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/79f8752062661b77a57dd187c81cb20ba0e0ad0d)
* [v3 10/12] [Modify Parser.java to use ParserRegistry and removed unused methods ](https://github.com/se-edu/addressbook-level4/pull/247/commits/32b728669a9c87e98e01081863f8e607b866d7a0)
* [v3 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/d16b322a7721062053a6591f5f0929f12d4b77ed)
* [v3 12/12] [Implement test cases ](https://github.com/se-edu/addressbook-level4/pull/247/commits/1a4e78757e2674ee2ea3dedf0c0134b3c31b36a3)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/0dd249a002f77bde21542f679f19f3d4bfa0444c...1a4e78757e2674ee2ea3dedf0c0134b3c31b36a3)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/commit/e504709239b91eeb6146af56beb7e48828687790))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/3/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ParserRegistry.java
chao1995: Oh, thanks for bearing me with my mistake. However, I want to know is there any specific case where we want to register more than one command words to a command parser? Since  in our current design each command should have its own parser.

MightyCupcakes: For now there is no case where you need to register more than one command word per parser. But in the future, there may be cases where you may want to associate multiple command words to a parser (like command aliases)

damithc: >  in the future, there may be cases where you may want to associate multiple command words to a parser (like command aliases)

Yes, this is the case. It's good that we can support aliases easily in a future version. But as the feature is not supported yet, may be we should not try to accommodate it yet? Ideally, the current design should be easy to extend in future directions but should not contain more code than it needs for the current functionality. see [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)

MightyCupcakes: Hmm, either way support aliases will be quite simple so if this change makes the code more readable and understandable for future developers, I have no objections modifying it and scrap the support for multiple command words for now in favour for consistency.

CanIHasReview:
# v4

@MightyCupcakes submitted v4 for review.

* [v4 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/d2d9a39db5ff6c21c0cd0892dcd2498bbdb0dbb2)
* [v4 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/b898bb3eac7704ff8f0a1ba523c02daff0a6d787)
* [v4 3/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/89332d7c02dc4c1f97246507d5a2f1891ad5a6ea)
* [v4 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/dd410ee2a5ae497bc3ae5de8a4b044c6beb6dc68)
* [v4 5/12] [Split the parser for add command into AddCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/613c85598aa8dcca030cd33ba192770bd58f0e44)
* [v4 6/12] [Split the parser for delete command into DeleteCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/4575435c3b2ec0e0acbcac574c3cd1f3e350b4e3)
* [v4 7/12] [Split the parser for find command into FindCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/cb6cd7f65d4cfae626cea53e862bb78c015a1df3)
* [v4 8/12] [Split the parser for select command into SelectCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/733793f10ee00d0af45a136787dbc23ae5dcc58e)
* [v4 9/12] [Split the parser for edit command into EditCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/3f2b7c5c95d9e09572756005c57d107723cefffc)
* [v4 10/12] [Modify Parser\.java to use ParserRegistry and removed unused methods ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f938664ea2af388f1ba097fcad465d3e0364464c)
* [v4 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/e9c081086cf19f859f4897606236d6d3a694a331)
* [v4 12/12] [Implement test cases ](https://github.com/se-edu/addressbook-level4/pull/247/commits/2ab07e75a803c3bb5f4ead4017b16d4456c9ae4f)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/0dd249a002f77bde21542f679f19f3d4bfa0444c...2ab07e75a803c3bb5f4ead4017b16d4456c9ae4f)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/commit/23c7f1226bea12258c0ae2e8020d38b6b7d9069f))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/4/head:BRANCHNAME
```

</details>

MightyCupcakes:
The latest iteration changes the parser registry register command API to follow the internal structure of the Map.

pyokagan:
@MightyCupcakes 

[v4 3/12], [v4 5/12] to [v4 9/12] seem to be code movement commits, moving code from `Parser.java` into their new classes.

In that case, I'd expect the code that is moved to be deleted from `Parser.java` in their respective commits as well, rather than deleting all of them at once in [v4 10/12]. This is to make it clear that code is being moved. Otherwise, you'd get cases like https://github.com/se-edu/addressbook-level4/pull/247#discussion_r96339260 where reviewers are left wondering where the code came from.

CanIHasReview:
# v5

@MightyCupcakes submitted v5 for review.

* [v5 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/55f78232b0431d35c678616cbed3eaf44ae34ba0)
* [v5 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/8ed22833032b214c6e36e218440cc7113d484398)
* [v5 3/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/d9a6bc8c9d92f3a9a9538c0364f70bf73b9decf8)
* [v5 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/c102b2c7a7ecb7b1da1df6ef0a1caeb0134544fb)
* [v5 5/12] [Split the parser for add command into AddCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/d143868d087b62dbcdad0c50c877309ac3272d29)
* [v5 6/12] [Split the parser for edit command into EditCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f1f5f515730075be421ded5cda69b58b769602cc)
* [v5 7/12] [Split the parser for delete command into DeleteCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/8c22314588ce487e4b02925bc0165b52e23d9c32)
* [v5 8/12] [Split the parser for select command into SelectCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/5bf6e0d6088437ebb75e946bcce23191ef4ce1e1)
* [v5 9/12] [Split the parser for find command into FindCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/ea548630ab3da7a74dae1092e2f8012f41f50a03)
* [v5 10/12] [Modify Parser\.java to use the Parser registry for command parsing](https://github.com/se-edu/addressbook-level4/pull/247/commits/d425366536dfed51b1410301ef9deff2a9d785e8)
* [v5 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/9e69faf47e59246e0cd2eb889e2a32075ba1cbd3)
* [v5 12/12] [Implement test cases ](https://github.com/se-edu/addressbook-level4/pull/247/commits/4e918f3327563ec3f94a21474bde145ace1b10fa)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...4e918f3327563ec3f94a21474bde145ace1b10fa)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/commit/db720f53ab88b06916fe028e7eebd0f9238fe5ac))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/5/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class ParserRegistryTest {
+
+    @Test
+    public void parserRegistry_emptyRegistry_incorrectCommand() {
+        TestParserRegistry registry = new TestParserRegistry();
+        registry.setCommandRegistry(Maps.newHashMap());
+
+        CommandParser parser = registry.getParserFromCommandWord("add");
+
+        assertTrue(parser instanceof IncorrectCommandParser);
```
yamgent: The two lines can be refactored into a method, so that you can just call this method for _all_ your test cases:

```
assertCorrectParserWithCommand("add", registry, IncorrectCommandParser.class);
```

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    public void parserRegistry_commandNotRegistered_incorrectCommand() {
+        TestParserRegistry registry = new TestParserRegistry();
```
yamgent: Since the empty registry is used for most of the test cases, maybe consider adding a static variable `emptyRegistry`, and initialize it in `setUp()`? So that you don't have to repeat this across multiple test cases.

```
@Before
public void setUp() throws Exception {
    emptyRegistry = new TestParserRegister();
}
```

yamgent:
@chao1995 Do you have any new comments to add?

CanIHasReview:
# v6

@MightyCupcakes submitted v6 for review.

* [v6 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/55f78232b0431d35c678616cbed3eaf44ae34ba0)
* [v6 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/8ed22833032b214c6e36e218440cc7113d484398)
* [v6 3/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/d9a6bc8c9d92f3a9a9538c0364f70bf73b9decf8)
* [v6 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/c102b2c7a7ecb7b1da1df6ef0a1caeb0134544fb)
* [v6 5/12] [Split the parser for add command into AddCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/d143868d087b62dbcdad0c50c877309ac3272d29)
* [v6 6/12] [Split the parser for edit command into EditCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f1f5f515730075be421ded5cda69b58b769602cc)
* [v6 7/12] [Split the parser for delete command into DeleteCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/8c22314588ce487e4b02925bc0165b52e23d9c32)
* [v6 8/12] [Split the parser for select command into SelectCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/5bf6e0d6088437ebb75e946bcce23191ef4ce1e1)
* [v6 9/12] [Split the parser for find command into FindCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/ea548630ab3da7a74dae1092e2f8012f41f50a03)
* [v6 10/12] [Modify Parser\.java to use the Parser registry for command parsing](https://github.com/se-edu/addressbook-level4/pull/247/commits/d425366536dfed51b1410301ef9deff2a9d785e8)
* [v6 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/9e69faf47e59246e0cd2eb889e2a32075ba1cbd3)
* [v6 12/12] [Implement unit test cases for parser registry ](https://github.com/se-edu/addressbook-level4/pull/247/commits/43c1a7dee40ca84d4dba887f546f2e3456813266)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...43c1a7dee40ca84d4dba887f546f2e3456813266)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/commit/5f1ef65313a8bf53d8448bf72fdc80d79bc36486))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/6/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,88 @@
...
+            Class<?> parser) {
+        CommandParser parserInRegistry = registry.getParserFromCommandWord(commandWord);
+
+        assertTrue(parser.isInstance(parserInRegistry));
+    }
+
+    /**
+     * A command parser class purely for testing. Does not do anything.
+     */
+    public static class TestCommandParser extends CommandParser {
```
damithc: It's nice if we can come up with a standard prefix for this sort of objects. `Test` is used as a standard suffix for test classes. While here we use it as a prefix, it's less confusing if it is an entirely different term.

src/main/java/seedu/address/logic/parser/CommandParser.java
```diff
@@ -0,0 +1,52 @@
...
+            return Optional.empty();
+        }
+        return Optional.of(Integer.parseInt(index));
+
+    }
+
+    protected Set<String> toSet(Optional<List<String>> tagsOptional) {
+        List<String> tags = tagsOptional.orElse(Collections.emptyList());
+        return new HashSet<>(tags);
+    }
```
MightyCupcakes: I agree. These can possibly be static methods within Parser.java. Additionally, the patterns can also be moved to parser so that the CommandParser abstract class is nothing more than an abstract prepareCommand method

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
MightyCupcakes: I was following other classes used solely in tests like `TestPerson`, `TestApp`

I was under the impression that any class (or file) prefixed with `Test` (for example `TestPerson`) are classes used solely in test cases and any class suffixed with `Test` (for example `StorageManagerTest`) will contain the test cases. I feel that this is a rather nice system that is already in place.

src/main/java/seedu/address/logic/parser/ClearCommandParser.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Parses input arguments in the context of the Clear command
+ */
+public class ClearCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
+        return new ClearCommand();
+    }
+}
```
pyokagan: @damithc 
> Classes like this makes me wonder if we should name these classes ClearCommandFactory etc.

Yes, parsers do have quite a lot of similarities with the factory pattern. However, they are much more *specific*, as they do have more constraints rather than just object creation, and so would benefit from a more specific name. Specifically:

* Parsers always *one* argument, a String, as input.

* Failure to parse the input string is treated as an user error which must be handled by the program, rather than a programmer error.

Classes named `Factory` do not necessarily obey the above constraints, which is why a more specific name `Parser` would be better.

> That is because in many cases there is no parsing involved, but object construction is always the end goal.

Yet we also can't ignore other parsers such as `add` and `find` which *do* contain some parsing logic.

Besides, what's to say we won't implement additional parsing logic in `ClearCommandParser` in the future? For example, it would be nice if I got an error when I type `clear foo` instead of just `clear`. Currently the parser just ignores `foo` silently.

So we shouldn't only be naming classes based on what they currently do, but more importantly we should name them based on what we *intend* for them to do. In this case, parsing, rather than just being a factory.

> If we adopt that approach, Parser parses the command to find the command word, picks the correct Factory class to delegate the command construction, and pass the remaining info to the Factory class. The downside is some of the Factory classes still need to do some 'parsing' of parameters although they are no longer called 'Parsers'

Yes, which strongly suggests that they *should* be called 'Parsers'.

> Perhaps the parsing parts can be extracted into utility classes e.g. IndexParser?

Yes that should be done.

But let's go back to your naming argument: if we end up calling `DeleteCommandParser` `DeleteCommandFactory`, for example, all `DeleteCommandFactory` would receive is `delete 1` with `delete` stripped off, aka just `1`. It then will need to decide to pass that `1` to the `IndexParser` for parsing.

HOWEVER, from the perspective of the caller (`Parser`), the caller does not know that `DeleteCommandFactory` actually used `IndexParser`. From the perspective of the caller, it would seem that `DeleteCommandFactory` actually did the parsing itself. And that's a good thing, because it means our abstraction isn't leaking.

In short, the name `Parser` is a good abstraction, so let's use it.

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,51 @@
...
+            return new IncorrectCommandParser();
+        }
+
+        Class<? extends CommandParser> parser = commandRegistry.get(commandWord);
+
+        try {
+            return parser.newInstance();
+
+        } catch (Exception e) {
+            return new IncorrectCommandParser();
```
pyokagan: Yes, the above `catch (Exception e)` means that *all* exceptions get caught, even programmer errors, which isn't a good thing.

src/main/java/seedu/address/logic/parser/ClearCommandParser.java
damithc: Agreed. Let's keep the `*Parser` names.

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
damithc: Yes, the current convention is adequate, so there's no need to change. It's just that sometimes it takes some effort to remember which is which. I can't think of a better term though; `Stub` or `Mock` don't work.

CanIHasReview:
# v7

@MightyCupcakes submitted v7 for review.

* [v7 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/28f202955f88c7f7b47797360c94d5bd10b79173)
* [v7 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/d2d747f71d5ab0212af2ea88f8cf45409c56c8c6)
* [v7 3/12] [Split all commands that requires no arguments to their classes](https://github.com/se-edu/addressbook-level4/pull/247/commits/d50ee94b0ed27a7630c9df9e0502ca64dce571ea)
* [v7 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/0df6dce397ec1a9fc7fddf741ac155261c122ff2)
* [v7 5/12] [Split the parser for add command into AddCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/dd9714269ae1f7eece69bbc0ac45b1bc8f2b5b87)
* [v7 6/12] [Split the parser for edit command into EditCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/7868d5701fc621393a9046e859ba415700b8efae)
* [v7 7/12] [Split the parser for delete command into DeleteCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/fb83d13ce6c204d354c6a1ef0c576fb6f723a27a)
* [v7 8/12] [Split the parser for select command into SelectCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/e501bb26e7a69d7989a6500ac50b9c97cc9c937a)
* [v7 9/12] [Split the parser for find command into FindCommandParser\.java ](https://github.com/se-edu/addressbook-level4/pull/247/commits/172b4ee7270ffcefd10a7ca7fbafe8f157241528)
* [v7 10/12] [Modify Parser\.java to use ParserRegistry](https://github.com/se-edu/addressbook-level4/pull/247/commits/e6f26684de2cdbd0233d73f1e1a5e94ecadb695f)
* [v7 11/12] [Implement TestParserRegistry for testing purposes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/71a1dd279a10ea07f88a0c1caec9aa060cf1ffec)
* [v7 12/12] [Implement unit test cases for parser registry ](https://github.com/se-edu/addressbook-level4/pull/247/commits/f9c0e7e10ef6f911f2df7f181557c09a399328b8)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...f9c0e7e10ef6f911f2df7f181557c09a399328b8)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/commit/63035064f926d3081cd73e8081397995201a3513))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/7/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,51 @@
...
+    protected final Map<String, Class<? extends CommandParser>> commandRegistry;
+
+    public ParserRegistry() {
+        commandRegistry = Maps.newHashMap();
+    }
+
+    /**
+     * Registers the given command word string with the provided
+     * command parser class. One command parser can be associated with
+     * multiple command words. However one command word can only be associated with one command parser.
```
damithc: Perhaps we should document what happens if the same command word is registered multiple times?

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,51 @@
+package seedu.address.logic.parser;
+
+import java.util.Map;
+
+import com.google.common.collect.Maps;
+
+/**
+ * A container to store all registered command parser classes that are
+ * responsible for parsing command strings provided by the user
```
damithc: This class has no business knowing about `parsing command strings provided by the user`.  As far as it knows, it acts as a registry of CommandParser objects.

The principle is, not to mention the usage contexts of a class/method when describing the class/method (as usages are incoming dependencies which the class/method is not aware of at all). If at all, usage context should only be mentioned as example usages, in order to make the description easier to understand.

src/main/java/seedu/address/logic/parser/ClearCommandParser.java
```diff
@@ -0,0 +1,15 @@
+package seedu.address.logic.parser;
```
damithc: Commit message:

>The NoArgCommandParser is deleted due to its unneeded complexity
it introduces to the code base. Furthermore, the commands might
see more modifications to its functionality later in development
and the class might be rendered obselete

Sound like you are describing something that has no trace in current commit history.

src/main/java/seedu/address/logic/parser/IncorrectCommandParser.java
```diff
@@ -0,0 +1,18 @@
+package seedu.address.logic.parser;
+
+import seedu.address.commons.core.Messages;
+import seedu.address.logic.commands.Command;
+import seedu.address.logic.commands.IncorrectCommand;
+
+/**
+ * Returns IncorrectCommand with an Unknown command message.
```
damithc: >  with an Unknown command message.

Can also sound like we don't know the returned message. Rephrase to make clearer?

src/main/java/seedu/address/logic/parser/AddCommandParser.java
```diff
@@ -0,0 +1,36 @@
+package seedu.address.logic.parser;
```
damithc: Commit message:
`Extract AddCommandParser from Parser class`?

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Parses input arguments and creates a new EditCommand object
+ */
+public class EditCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
+        assert args != null;
+        ArgumentTokenizer argsTokenizer = new ArgumentTokenizer(Parser.PREFIX_PHONE, Parser.PREFIX_EMAIL,
```
damithc: I wonder if `static` importing these PREFIX constants can make the code more concise. Applies to some other *CommandParser classes as well.

src/main/java/seedu/address/logic/parser/DeleteCommandParser.java
```diff
@@ -0,0 +1,27 @@
...
+import seedu.address.logic.commands.IncorrectCommand;
+
+/**
+ * Parses input arguments and creates a new DeleteCommand object
+ */
+public class DeleteCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
+        Optional<Integer> index = Parser.parseIndex(args);
```
damithc: This dependency back to `Parser` creates a bidirectional dependency, right? Can avoid be extracting the method to a separate utility class?

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -1,129 +1,109 @@
...
 
 /**
  * Parses user input.
  */
 public class Parser {
 
+    public static final Prefix PREFIX_PHONE = new Prefix("p/");
+    public static final Prefix PREFIX_EMAIL = new Prefix("e/");
+    public static final Prefix PREFIX_ADDRESS = new Prefix("a/");
+    public static final Prefix PREFIX_TAG = new Prefix("t/");
```
damithc: These too will cause bidirectional dependencies between `Parser` and `*CommandParser` classes. Not sure how to resolve it though. Perhaps extract 'language definition info' to a separate class? @pyokagan @MightyCupcakes any thoughts?

src/test/java/seedu/address/testutil/TestParserRegistry.java
```diff
@@ -0,0 +1,22 @@
+package seedu.address.testutil;
```
damithc: This commit can be squashed onto the next one?

src/test/java/seedu/address/testutil/TestParserRegistry.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.Map;
+
+import seedu.address.logic.parser.CommandParser;
+import seedu.address.logic.parser.ParserRegistry;
+
+/**
+ * A parser registry class solely used for testing
```
damithc: How does it differ from the normal ParserRegistry. i.e. why do we even need it?

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Before
+    public void setUp() {
+        registry = new TestParserRegistry();
+        registerTestParsers(registry);
+    }
+
+    @Test
+    public void parserRegistry_emptyRegistry_incorrectCommand() {
+        registry.setCommandRegistry(Maps.newHashMap());
```
damithc: Why not use a new ParserRegistry object here? That way, you can get rid of setCommandRegistry method.

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,96 @@
...
+        registry
+            .registerCommand("add", AddCommandParser.class)
+            .registerCommand("find", FindCommandParser.class)
+            .registerCommand("edit", EditCommandParser.class)
+            .registerCommand("delete", DeleteCommandParser.class)
+            .registerCommand("invalid", CommandParser.class);
+    }
+
+    private void assertParserWithCommand(String commandWord, ParserRegistry registry,
+            Class<?> parser) {
```
damithc: No need for line break?

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private void registerTestParsers(ParserRegistry registry) {
+        registry
+            .registerCommand("add", AddCommandParser.class)
+            .registerCommand("find", FindCommandParser.class)
+            .registerCommand("edit", EditCommandParser.class)
+            .registerCommand("delete", DeleteCommandParser.class)
+            .registerCommand("invalid", CommandParser.class);
+    }
+
+    private void assertParserWithCommand(String commandWord, ParserRegistry registry,
```
damithc: This method name needs a bit of head scratching to understand. `assertCommandParserIsInRegistry`?

src/test/java/seedu/address/logic/parser/ParserRegistryTest.java
damithc: `commandParserClass`?

src/main/java/seedu/address/logic/parser/ParserRegistry.java
```diff
@@ -0,0 +1,51 @@
...
+        commandRegistry.put(commandWord, parser);
+
+        return this;
+    }
+
+    /**
+     * Returns a CommandParser registered with the given command word. If no such CommandParser
+     * exists, an IncorrectCommandParser is returned instead.
+     */
+    public CommandParser getParserFromCommandWord(String commandWord) {
```
damithc: `getCommandParserForCommandWord`? I'm not sure about using `Parser` as a short name for `*CommandParser` as already have a separate class call `Parser`.
On a related note, I'm wondering if we can simply use `put` and `get` for the two methods in this class.

yamgent:
Now that I looked at it, I realize that one can easily confuse `CommandParser` with `Parser`. It is difficult to tell at first glance the difference in responsibilities of the two classes.

Following up with the factory suggestion, perhaps `Parser` might be a better candidate to be named as a "`ParserFactory`"?

src/main/java/seedu/address/logic/parser/Parser.java
MightyCupcakes: To get rid of the bidirectional dependency, one way is to create a utility class solely devoted to defining and holding constants values such as these. The classes can be prefixed with `Lib` as in Library like `LibParser`. This is how Minecraft internal code structure defines its constants to be used in multiple classes in server and client code without creating troublesome dependency across logical units.

But one problem is that we don't really have much constants to justify creating a new seperate utility class. Furthermore, these constants are only really used by two classes (only add and edit command uses it right now) so I am not sure if this is applicable to our case.

damithc:
> Following up with the factory suggestion, perhaps Parser might be a better candidate to be named as a ParserFactory?

I think you are right in that we can have a `ParserFactory` class (or a `CommandParserFactory`) that takes care of creating different `CommandParser` classes.  However, it should be a case of `Parser` using `CommandParserFactory` (i.e. extract out `CommandParser` construction to a new class) rather than renaming `Parser`. The reason is, we should present a high-level `Parser` API to clients who wants to parse something. Presenting those clients with a `ParserFactory` breaks the abstraction and exposes clients to how the parsing is implemented.

MightyCupcakes:
I am not sure about naming `Parser` to `ParserFactory`.

The other components like `Logic` have no business knowing about the internal workings of `Parser`. All they need to know is that they provide `Parser` with a command string, and `Parser` will return the appropriate `Command` object.

src/main/java/seedu/address/logic/parser/Parser.java
damithc: > But one problem is that we don't really have much constants to justify creating a new seperate utility class. Furthermore, these constants are only really used by two classes (only add and edit command uses it right now) so I am not sure if this is applicable to our case.

Personally, I don't consider the number of members and number of clients of a class as important criteria when defining classes. OOP favors small classes anyway. For me, the important criteria is getting the abstraction right (i.e. put things where they belong, and more importantly, not put things where they don't belong). So in this case, we can consider putting information related to the syntax of the CLI language in one class e.g. `CliSyntax`. `Parser` and `*CommandParser` classes need to depend on that class. The downside is that when adding new commands, there is one more place to update. Perhaps we can put high-level syntax info in `CliSyntax` but hold command-specific syntax info in the `*CommandParser` class itself. That way, plug-and-play type new commands can be added without modifying `CliSyntax` class but new commands that introduce a more invasive change to CLI syntax (e.g. adding a new attribute to persons) may need a change in `CliSyntax` as well as other existing command parsers e.g. `AddCommandParser`.

CanIHasReview:
# v8

@MightyCupcakes submitted v8 for review.

* [v8 1/12] [Implement the abstract command parser class](https://github.com/se-edu/addressbook-level4/pull/247/commits/6dbd935e05396b5acf7150c9bd7171f1ddd20cb7)
* [v8 2/12] [Implement the parser registry class](https://github.com/se-edu/addressbook-level4/pull/247/commits/b5bffe8d54b216c75305df050ac5d82f5d176a6b)
* [v8 3/12] [Split all commands that requires no arguments to their classes ](https://github.com/se-edu/addressbook-level4/pull/247/commits/a8c13fbfe9607a99f71f010937a450f0dda66be8)
* [v8 4/12] [Implement the incorrect command parser ](https://github.com/se-edu/addressbook-level4/pull/247/commits/93616efcda0c556b7bb7aac471e0478d380f688e)
* [v8 5/12] [Add seperate utility classes for parsing related tasks](https://github.com/se-edu/addressbook-level4/pull/247/commits/4f89faca8aa4be137837f8c06834c871f3977b9e)
* [v8 6/12] [Extract AddCommandParser from Parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/9bf25de7c23ab71a519d5edaf5007bb22ed969c9)
* [v8 7/12] [Extract EditCommandParser from Parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/114ff2fd158ce5cc38a1540343caff741b84302a)
* [v8 8/12] [Extract DeleteCommandParser from Parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/5541074e255acc0f9ab1564e7ef7a7144f48e3cb)
* [v8 9/12] [Extract SelectCommandParser from Parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/e5af1d427f8b6d5c4ed4d927a4a2e51265d4d3a5)
* [v8 10/12] [Extract FindCommandParser from Parser class ](https://github.com/se-edu/addressbook-level4/pull/247/commits/8a2d9858f30ab7d64773a0c966960b7951b74840)
* [v8 11/12] [Modify Parser\.java to use ParserRegistry](https://github.com/se-edu/addressbook-level4/pull/247/commits/a8f464e865b0ef7c784a231b5070b8c340756c5c)
* [v8 12/12] [Implement unit test cases for parser registry ](https://github.com/se-edu/addressbook-level4/pull/247/commits/90e79444c0506bb0018beed29b79ab6bbcc9b1f5)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/f74c40cb73cd505ba1d6540bd9361cab1369d830...90e79444c0506bb0018beed29b79ab6bbcc9b1f5)) ([:chart_with_upwards_trend: Interdiff between v7 and v8](https://github.com/se-edu/addressbook-level4/commit/3141de1df439781c5a2aacb4aab4f86309e50066))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/247/8/head:BRANCHNAME
```

</details>

MightyCupcakes:
Major changes in version 8:

* Moved utility parsing methods used by various `*Parser` classes to a seperate `ParserUtil` class located in `seedu.address.commons.util` package
* Moved constants and patterns used by `*Parser` classes and `Parser` class itself to a new container class `CliSyntax` located in the `seedu.address.logic.parser` package
* `PREFIX` constants are now `static` imports
* Removed `TestParserRegistry`
* Other minor changes suggested are implemented.

src/main/java/seedu/address/logic/parser/AddCommandParser.java
```diff
@@ -0,0 +1,41 @@
...
+import seedu.address.logic.commands.IncorrectCommand;
+
+/**
+ * Parses input arguments and creates a new AddCommand object
+ */
+public class AddCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
+        ArgumentTokenizer argsTokenizer = new ArgumentTokenizer(PREFIX_PHONE, PREFIX_EMAIL,
```
damithc: Line break after `(` so that all params appear in one line?

src/main/java/seedu/address/logic/parser/FindCommandParser.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Parses input arguments and creates a new FindCommand object
+ */
+public class FindCommandParser extends CommandParser {
+
+    @Override
+    public Command prepareCommand(String args) {
+        final Matcher matcher = KEYWORDS_ARGS_FORMAT.matcher(args.trim());
+        if (!matcher.matches()) {
+            return new IncorrectCommand(String.format(MESSAGE_INVALID_COMMAND_FORMAT,
```
damithc: Line break after `(`?. i.e. prefer to break at higher levels of the syntax tree.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -1,60 +1,46 @@
...
 
-    private static final Pattern KEYWORDS_ARGS_FORMAT =
-            Pattern.compile("(?<keywords>\\S+(?:\\s+\\S+)*)"); // one or more keywords separated by whitespace
-    private static final Prefix phoneNumberPrefix = new Prefix("p/");
-    private static final Prefix emailPrefix = new Prefix("e/");
-    private static final Prefix addressPrefix = new Prefix("a/");
-    private static final Prefix tagsPrefix = new Prefix("t/");
+    private void registerCommandParsers() {
+        commandParserRegistry
+            .registerCommand(AddCommand.COMMAND_WORD, AddCommandParser.class)
```
damithc: It's a bit odd that the `COMMAND_WORD` is in `AddCommand` class (the command word is not related to the command execution). If it is defined in `AddCommandParser` class, you don't even need this parameter here.

src/main/java/seedu/address/commons/util/ParserUtil.java
```diff
@@ -0,0 +1,44 @@
...
+        }
+        return Optional.of(Integer.parseInt(index));
+
+    }
+
+    /**
+     * Returns a new Set populated by all elements in the given list of String
+     * Returns an empty list if the provided list is empty
+     */
+    public static Set<String> toSet(Optional<List<String>> tagsOptional) {
```
damithc: This method doesn't sound like it's specific to Parsing. That is, if someone is looking for a method like this to reuse, that person is unlikely to look in ParserUtil class.

pyokagan:
@damithc
> Better to wait till @pyokagan has weighed in on this iteration before you proceed as he might have objections to the direction we are heading in this iteration.

Some general comments:

* Could `CommandParser` be a functional interface? This would allow us to possibly take advantage of the power of lambdas in the future. (Maybe for test cases)

* Remember that every single commit must be able to compile and pass the test suite. At the same time, each commit needs to fully carry out its stated responsibility. Some examples:

    * [v8 3/12]: I would expect that the current uses of `new ListCommand()`, `new ExitCommand()` etc. in `Parser.java` be updated to use the new `ListCommandParser`, `ExitCommandParser` etc.

    * [v8 5/12]: You should change all current users of `Parser::parseIndex()` to use the new `ParserUtil::parseIndex()`.

    * [v8 6/12] does not compile.

* What's [v8 4/12] used for? It's not immediately clear.

* [v8 2/12], [v8 12/12] can be squashed into [v8 11/12]. It will then be obvious that `ParserRegistry` is just replacing a `switch` block in `Parser.java`. Is that overkill? Please address that in the commit message. (I didn't fully read all commit messages, so it's possible the PR already addressed it)

* The commit message of [v8 2/12] references a lot of things that are not actually solved by the `ParserRegistry` itself. For example:

    > this class is responsible for the parsing of so many different commands with vastly different functionalities - this heavy responsibility is hardly one cohesive single responsilbility.

    This is solved already solved by splitting up the parsing code into different classes, not really by introducing a `ParserRegistry` class.

* We can probably use lambdas instead of using reflection in `ParserRegistry`.

pyokagan:
> [v8 2/12], [v8 12/12] can be squashed into [v8 11/12]. It will then be obvious that ParserRegistry is just replacing a switch block in Parser.java. Is that overkill?

To add on, issue #137 only requires the parsing code to be split into different classes, so it may be good if we just leave [v8 3/12] and [v8 5/12] to [v8 10/12] in this PR and move all the other commits (those concerning `ParserRegistry`) into another PR.

pyokagan:
@MightyCupcakes 

To add on again,

* Shouldn't `Parser` be demonstrating the open closed principle by allowing passing in a arbitrary `ParserRegistry` at construction time?

* `Parser` should be renamed. `Parser` is a shorter and simpler name than `CommandParser`, which implies that it is lower-level than a `CommandParser`. However, this is not the case. `Parser`, as implemented in this PR, is at a higher level since it calls on other sub-command parsers.

* Can't `Parser` inherit/implement `CommandParser` as well?

pyokagan:
> What's [v8 4/12] used for? It's not immediately clear.

If it's not immediately clear by looking at a commit how it fits in the whole PR, then it's likely an indication that it should be squashed into another commit to make it more cohesive.

MightyCupcakes:
> Shouldn't Parser be demonstrating the open closed principle by allowing passing in a arbitrary ParserRegistry at construction time?

No objections

>Parser should be renamed. Parser is a shorter and simpler name than CommandParser, which implies that it is lower-level than a CommandParser. However, this is not the case. Parser, as implemented in this PR, is at a higher level since it calls on other sub-command parsers.

I don't buy this logic at all. Simple class names imply generality.

> Can't Parser inherit/implement CommandParser as well?

Yes it can. But it won't make sense

pyokagan:
@MightyCupcakes 
> Simple class names imply generality.

Precisely. Lower level classes are usually more generic, since they are the building blocks used for higher level classes.

Want an example? Look at the implementation of `logic.commands`. `Command` is the base class, and then we have `AddCommand`, `DeleteCommand`, `FindCommand`...

> Yes it can. But it won't make sense

Because `Parser#prepareCommand()` does not take a String and return a Command?

damithc:
> Parser should be renamed. Parser is a shorter and simpler name than CommandParser, which implies that it is lower-level than a CommandParser. 

The current `Parser` is using composition rather than inheritance.

```
[Parser]<>-----[CommandParser]<|---------[AddCommandParser]
```

On terminology, I usually consider parent class _higher_ level than child class (they are at a higher level in the inheritance tree).  That means higher level classes have more general (and simpler) names.

pyokagan:
@damithc 
> I usually consider parent class higher level than child class (they are at a higher level in the inheritance tree).

Could we really use the same definition of "parent" and "child" in a inheritance relationship with the definition of "parent" and "child" in a composition relationship? After all, in inheritance the child depends on the parent (`child -> parent`, child is aware of parent, parent not aware of child), but in composition the parent depends on the child (`child <- parent`, parent is aware of child (otherwise it cannot construct the child), child is not aware of the parent). Assuming we don't have a bidirectional relationship, of course.

In other words, if we look at dependencies, we see that both the `Parser` and `AddCommandParser` both depend on `CommandParser`. Should `CommandParser` be considered the most highest level and get the name `Parser` instead?

Anyway, just to clarify: I'm looking from the perspective of well the name actually describes the responsibility of the class -- the name `Parser` does not give any indication of what the responsibility of the class is. Unless the class's reponsibility is so generic that it gets a generic name of course. Names should also highlight any patterns in behavior between classes, to aid API discoverability.

If we look at the `Parser` as it is currently implemented, we see that it:

1. Takes a *command string* (wonder what other classes do that)

2. Extracts the first word (hmm, what is the first word called? That would help us come up with a descriptive name for this class, since its responsibility involves looking at the first word)

3. Dispatches the rest of the input to the relevant command parser (without the first word)

4. Returns a *Command* (again, what other kind of classes do that?)

The only thing that is missing is the definition of what the first word is. Git, and most other command line programs, for example, has a name for something like that. It's called subcommands.

I vote for the name `SubCommandParser`, and it inherits from `CommandParser`.

damithc:
Let's continue this discussion at tomorrow's session where we can easily use diagrams as visual aids. Some thoughts:
1. One factor that might be complicating the naming of the current Parser class is that it plays multiple roles, one of which is to act as a Facade to the whole parsing process. It's the interface used by any clients who needs something to be parsed and those clients don't care what part of the parsing is done by the Parser class and what part is delegated to other classes. That is the role pulling towards a general name like `Parser` containing various `parse` methods (although we only have one parse method at the moment). Going forward, we can consider having a real Facade class instead of forcing a regular class to play the Facade. 
2. I'm OK with the name `SubCommandParser`. I suspect our situation is slightly different from Git in that for Git `git` is the main command (I suspect, not sure) and what follows it e.g. `branch` is the sub command. In our case, calling `add` a sub command is a bit odd as we never use `addresbook` as our main command. But if `sub command` is a common term for other CLI programs, no harm following the same.

pyokagan:
@damithc 
> It's the interface used by any clients who needs something to be parsed and those clients don't care what part of the parsing is done by the Parser class and what part is delegated to other classes.

If clients do not care about the internal details of a parser, they could just use the `CommandParser` interface directly? After all, it only exposes a "parse" method (somewhat misnamed I think as `prepareCommand()` in this PR). A `CommandParser` could internally delegate its parsing to other classes, however the caller would not know that at all.

> That is the role pulling towards a general name like Parser containing various parse methods (although we only have one parse method at the moment).

Okay, this puts things in a slightly better context. What other parse methods do you envision?

With that said, I'd rather we aim for as-simple-as-possible. I'm not sure if we really need a facade with our current application's requirements (YAGNI and all that)

> In our case, calling add a sub command is a bit odd as we never use addresbook as our main command.

Just for a bit more context: I'm striving for our classes to be small and reusable (SRP). While the top level `SubCommandParser` could be interpreted as parsing the "main command", we can nest multiple `SubCommandParsers` to create commands with sub-commands, sub-sub-commands etc.

pyokagan:
@damithc 
> Let's continue this discussion at tomorrow's session

Okay. For this session, I hope we can first discuss fixing the code to adhere to SRP in isolation, without considering OCP. We can then discuss how to implement OCP on top of that.

The reason is because I think this PR, by attempting to do both at the same time, kind of muddled up our discussions a bit.

pyokagan:
Summarizing my points of the discussion today:

* It is better to split this PR up into smaller PRs, each PR tackling a specific subproblem.

* Let's try to adopt the simplest solution to each problem for now. Simpler solutions are easy to reason and talk about. After these simple solutions get merged. we can then discuss about enhancing these solutions.

    * For SRP: The simplest solution is to just split the code into different classes. That's all.

    * For OCP: The simplest solution is likely just to apply Dependency Inversion -- introduce an interface. We should aim to have as minimal an interface as possible, as it makes it easier for clients to implement the interface. The simplest interface is a functional interface with only one method.

MightyCupcakes:
As @pyokagan mentioned, this PR will be closed in favour of having smaller PRs.

Closed and continued in #272

PR #248, opened 2017-01-14
Users can delete multiple persons in one shot

### LINKED ISSUE #140: User can delete multiple persons in one shot

cheec:
Fixes #140 
Implement multiple deleting 
* Users can now delete multiple persons in a single command
* Users must seperate multiple indices with at least a single whitespace
* Users can also specify a range indicator '-' for range deleting
* Overlapping ranges are also handled appropriately
* Indices specified in ranges do not need to be in ascending order
* Indices cannot exceed list size

Examples: 
* `delete 5 1 7 10-16` deletes persons 1, 5, 7, 10 to 16
* `delete 1 1 1 1 1` deletes person 1
* `delete 5-1 1-5 1 2 3 4 5` deletes persons 1 to 5
* `delete 1-3 2-5 1-10` deletes persons 1 to 10
* `delete 6-1` delete persons 1 to 6


cheec:
@pyokagan ready for review

CanIHasReview:
# v1

@cheec submitted v1 for review.

* [v1 1/9] [IntegerUtil: create utility class * API includes applyOffset()
* used for mapping one-indexed indices to zero-indexed indices
* tests are included in this commit as well
](https://github.com/se-edu/addressbook-level4/pull/248/commits/830b507cfbeed5ec1149e5b9689953d29a265663)
* [v1 2/9] [ListUtil: create utility class * API includes subList() and areIndicesWithinBounds()
* subList() used to extract items from a list at specified indices
* areIndicesWithinBounds() used to check index boundaries
](https://github.com/se-edu/addressbook-level4/pull/248/commits/7cf510b2b533f14b155b9bb37ba5a38a675106ca)
* [v1 3/9] [TestUtil: add mapIndexToObj() * created to substitute dependency on ListUtil::subList
](https://github.com/se-edu/addressbook-level4/pull/248/commits/6089a86a0015bdf27c58f009617ece26f548e77c)
* [v1 4/9] [StringUtil: add toIndexedListString * includes tests as well
](https://github.com/se-edu/addressbook-level4/pull/248/commits/892fab23b9b747e8a878ae0a673269cec1d3d331)
* [v1 5/9] [CollectionUtil: refactor assertNoNullElements -> isAnyNull(Collection) * previously, assertNoNullElements wasn't even working as intended at all,
this commit both fixes this issue and refactors it into an overloaded method.
* this commit also updates the tests from refactoring
](https://github.com/se-edu/addressbook-level4/pull/248/commits/c87fb63881281ad9c4cdd0a0499cd409a656491a)
* [v1 6/9] [Model: implement deletePersons * ModelManager::deletePersons
* AddressBook::removePersons
*UniquePersonList::removeAll
](https://github.com/se-edu/addressbook-level4/pull/248/commits/5ae62e87836f31e41d82a9d85758d00884322faf)
* [v1 7/9] [Parser/DeleteCommand: implement logic for deleting * morph parser to parse new arguments
* update logic of DeleteCommand created by parser
* DeleteCommandTest: update tests to match new functionality
* LogicManagerTest: update tests to match new functionality
](https://github.com/se-edu/addressbook-level4/pull/248/commits/f7562cfa8c23d99c14a683c7b6dd253a20b32ec6)
* [v1 8/9] [DG: update model API ](https://github.com/se-edu/addressbook-level4/pull/248/commits/a54ca27b337bebab18b094fe33522624a56d0995)
* [v1 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/3a5a2aa36083540bdef6f1e02c5c9537b6279ae5)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/32643f6b7435056071181de2bd34055bb3b05cc4...3a5a2aa36083540bdef6f1e02c5c9537b6279ae5))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/1/head:BRANCHNAME
```

</details>

pyokagan:
@cheec  You need to have a blank line between the commit subject and commit message body. Also the commit message body needs to be wrapped at 72 characters. See https://github.com/oss-generic/process/blob/master/docs/FormatsAndConventions.md#commit-message

pyokagan:
> Also the commit message body needs to be wrapped at 72 characters.

Oh, scratch that. I mixed this PR up with the other one.

CanIHasReview:
# v2

@cheec submitted v2 for review.

* [v2 1/9] [IntegerUtil: create utility class](https://github.com/se-edu/addressbook-level4/pull/248/commits/94cecb9adb2669830db7bcaaa0e73262111f83e9)
* [v2 2/9] [ListUtil: create utility class](https://github.com/se-edu/addressbook-level4/pull/248/commits/a1f2fe8ff875dc90218d1ac7ef20d0c13147dfca)
* [v2 3/9] [TestUtil: add mapIndexToObj()](https://github.com/se-edu/addressbook-level4/pull/248/commits/90b95ca03b3dd3d02c14d1cdf492eed3fa2f6c36)
* [v2 4/9] [StringUtil: add toIndexedListString](https://github.com/se-edu/addressbook-level4/pull/248/commits/b0e80cf78b396eb194862ef1f0f49c335257b0f1)
* [v2 5/9] [CollectionUtil: refactor assertNoNullElements -> isAnyNull(Collection)](https://github.com/se-edu/addressbook-level4/pull/248/commits/33bc31295c3967eb04af72868d2cbc7e6a204eb8)
* [v2 6/9] [Model: implement deletePersons](https://github.com/se-edu/addressbook-level4/pull/248/commits/78fef3a1799ffe9f0d31f650da50c12614129ce8)
* [v2 7/9] [Parser/DeleteCommand: implement logic for deleting](https://github.com/se-edu/addressbook-level4/pull/248/commits/5a0913f2613054f5d141ad002a1d31d8844069b3)
* [v2 8/9] [DG: update model API ](https://github.com/se-edu/addressbook-level4/pull/248/commits/36a5e3df952c8f3758f77b94813545a800ebc929)
* [v2 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/e39ed6a596b463e7629ff15e5e48c77149d78b84)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/bc2f54675f66a410eea7f3338f09b125dfc573aa...e39ed6a596b463e7629ff15e5e48c77149d78b84)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/commit/84174be599f17910c802f92a658be247ec950782))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/2/head:BRANCHNAME
```

</details>

cheec:
@pyokagan done, thanks

docs/UserGuide.md
```diff
@@ -147,9 +167,9 @@ There is no need to save manually.
 
 * **Clear** : `clear`
 
-* **Delete** : `delete INDEX` <br>
-   e.g. `delete 3`
-
+* **Delete** : `delete INDICES` <br>
+   e.g. `delete 3 5 6 1-2`
+
```
cheec: whitespace, will remove next iteration

cheec: whitespace, will remove next iteration

src/test/java/seedu/address/testutil/TestUtil.java
```diff
@@ -350,4 +351,9 @@ public static boolean compareCardAndPerson(PersonCardHandle card, ReadOnlyPerson
         return collect.toArray(new Tag[split.length]);
     }
 
+    /** Maps {@code indices} to its corresponding item in {@code items}. */
```
damithc: The description is not very clear.

src/main/java/seedu/address/commons/util/CollectionUtil.java
```diff
@@ -2,33 +2,27 @@
...
-        }
-        return false;
+        return items == null || Stream.of(items).anyMatch(Objects::isNull);
     }
 
-
-
     /**
-     * Throws an assertion error if the collection or any item in it is null.
+     * Returns true if {@code items} itself or any element of {@code items} is null.
```
damithc: This is not critical, but I feel the method is doing two things (`is collection null` and `does collection contains null`) which can lead to unintended results. e.g. user thinks he is passing a list containing a null but passes null by mistake, and yet gets `true` as the return value. Perhaps `items != null` should be a pre-condition?  @pyokagan what do you think?

src/main/java/seedu/address/model/Model.java
```diff
@@ -17,8 +18,18 @@
...
+    /**
+     * Deletes the given person.
+     * @see #deletePersons(Collection)
+     */
+    void deletePersons(ReadOnlyPerson... persons) throws UniquePersonList.PersonNotFoundException;
+
+    /**
+     * Deletes the given {@code persons}.
+     * @throws PersonNotFoundException without deleting any persons if any of the {@code persons}
+     * can't be found in this list.
```
damithc: Should be indented to indicate it is a continuation of the previous line?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,12 +112,17 @@ private void syncMasterTagListWith(UniquePersonList persons) {
...
     }
 
-    public boolean removePerson(ReadOnlyPerson key) throws UniquePersonList.PersonNotFoundException {
-        if (persons.remove(key)) {
-            return true;
-        } else {
-            throw new UniquePersonList.PersonNotFoundException();
-        }
+    /** @see #removePersons(Collection) */
+    public void removePersons(ReadOnlyPerson... keys) throws UniquePersonList.PersonNotFoundException {
```
damithc: We seem to use terms `delete` and `remove` both. Is that intended? If not, may be standardize on one? Can be left as a TODO to do as a separate PR.

damithc:
Reviewed [v2 1-6]. Will continue the rest later. As before, wait for @pyokagan's review before doing the next version.

src/main/java/seedu/address/commons/util/CollectionUtil.java
pyokagan: @damithc 
> Perhaps items != null should be a pre-condition?

I agree.

src/main/java/seedu/address/commons/util/StringUtil.java
```diff
@@ -55,4 +59,18 @@ public static String getDetails(Throwable t) {
     public static boolean isUnsignedInteger(String s) {
         return s != null && s.matches("^0*[1-9]\\d*$");
     }
+
+    /**
+     * Returns a string that has each item in {@code items} on its own 'one-indexed' new line.<br>
+     * E.g. [apple, bob, foo] becomes:<br>
```
pyokagan: Since you are interested in properly formatting your javadocs, then the `[apple, bob, foo]` should also probably be using `@code`.

Rather than inventing your own placeholder strings, you may wish to pick one of the [commonly used ones](https://en.wikipedia.org/wiki/Metasyntactic_variable#English)

src/main/java/seedu/address/commons/util/StringUtil.java
```diff
@@ -55,4 +59,18 @@ public static String getDetails(Throwable t) {
     public static boolean isUnsignedInteger(String s) {
         return s != null && s.matches("^0*[1-9]\\d*$");
     }
+
+    /**
+     * Returns a string that has each item in {@code items} on its own 'one-indexed' new line.<br>
```
pyokagan: Why is `'one-indexed'` quoted?

`new line` -> `line`.

src/main/java/seedu/address/commons/util/StringUtil.java
```diff
@@ -55,4 +59,18 @@ public static String getDetails(Throwable t) {
...
+     * E.g. [apple, bob, foo] becomes:<br>
+     *   1. apple\n<br>
```
pyokagan: Either write a literal `\n`, or use `<br>`, not both.

Shouldn't this be in a `<code>` block as well?

src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -148,5 +153,46 @@ public void getDetails_nullGiven_assertionError() {
...
+    }
+
+    @Test
+    public void toIndexedListString_validList_correctResult() {
+        Collection<Object> items = new ArrayList<>();
+        // testing different object types
+        items.add("bob bob bob bobbobob");
+        items.add(new Integer(1));
+        items.add(Arrays.asList(5, 6, 7));
+        String expectedString = "1. bob bob bob bobbobob\n"
```
pyokagan: So, how should the output be like if a string contains newlines inside?

And what if the string contains trailing whitespace?

src/main/java/seedu/address/commons/util/StringUtil.java
cheec: the \n is not for Javadocs but rather to show that this method appends a new line character to each line

src/test/java/seedu/address/commons/util/CollectionUtilTest.java
```diff
@@ -69,4 +113,5 @@ private void assertAreUnique(Object... objects) {
     private void assertNotUnique(Object... objects) {
         assertFalse(CollectionUtil.elementsAreUnique(Arrays.asList(objects)));
     }
+
```
pyokagan: Unnecessary change? Note that the top of the class doesn't have a blank line.

src/test/java/seedu/address/commons/util/StringUtilTest.java
cheec: the string will still be represented as is e.g. for a string "a\nb\nc\n"
1. a
b
c

2. ...
3. ...

src/main/java/seedu/address/commons/util/StringUtil.java
pyokagan: @cheec 
> the \n is not for Javadocs but rather to show that this method appends a new line character to each line except the last

Yes, I know your intent. But your javadoc does not make it clear whether the output is a *Java string literal* (by which `\n` will have a special meaning), or the displayed console output (by which `\n` does not have any special meaning).

src/test/java/seedu/address/commons/util/StringUtilTest.java
pyokagan: @cheec 
> the string will still be represented as

Oh sorry, the above message is a hint that you should test for that.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,12 +112,17 @@ private void syncMasterTagListWith(UniquePersonList persons) {
...
+        removePersons(Arrays.asList(keys));
+    }
+
+    /**
+     * Removes {@code keys} from this {@code AddressBook}'s {@code UniquePersonList}. If any person in {@code keys}
+     * is missing from the list, removal will be aborted with a {@code PersonNotFoundException}.
```
pyokagan: Use `@throws` here?

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -59,8 +61,13 @@ private void indicateAddressBookChanged() {
...
-    public synchronized void deletePerson(ReadOnlyPerson target) throws PersonNotFoundException {
-        addressBook.removePerson(target);
+    public synchronized void deletePersons(ReadOnlyPerson... persons) throws PersonNotFoundException {
+        deletePersons(Arrays.asList(persons));
+    }
+
+    @Override
+    public synchronized void deletePersons(Collection<ReadOnlyPerson> persons) throws PersonNotFoundException {
+        addressBook.removePersons(persons);
         indicateAddressBookChanged();
```
pyokagan: If I call `deletePersons()` with an empty list, should `indicateAddressBookChanged()` be called?

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -59,8 +61,13 @@ private void indicateAddressBookChanged() {
     }
 
     @Override
-    public synchronized void deletePerson(ReadOnlyPerson target) throws PersonNotFoundException {
-        addressBook.removePerson(target);
+    public synchronized void deletePersons(ReadOnlyPerson... persons) throws PersonNotFoundException {
+        deletePersons(Arrays.asList(persons));
```
pyokagan: Maybe this can be implemented as a default method in the `Model` interface.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -50,11 +52,28 @@ public boolean remove(ReadOnlyPerson toRemove) throws PersonNotFoundException {
         assert toRemove != null;
         final boolean personFoundAndDeleted = internalList.remove(toRemove);
         if (!personFoundAndDeleted) {
-            throw new PersonNotFoundException();
+            throw new PersonNotFoundException(toRemove.toString());
```
pyokagan: Interesting how the argument is called `indexedListStringOfPersons` and yet an indexed list string is not being passed in :-P

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -50,11 +52,28 @@ public boolean remove(ReadOnlyPerson toRemove) throws PersonNotFoundException {
...
         return personFoundAndDeleted;
     }
 
+    /**
+     * Removes all persons specified by {@code personsToRemove} from this list.<br>
+     * @throws PersonNotFoundException without deleting any persons if any of the {@code personsToRemove}
+     * can't be found in this list.
+     */
+    public void removeAll(Collection<ReadOnlyPerson> personsToRemove) throws PersonNotFoundException {
+        assert !CollectionUtil.isAnyNull(Arrays.asList(personsToRemove));
```
pyokagan: `assert personsToRemove != null`?

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -99,9 +118,13 @@ protected DuplicatePersonException() {
...
 
     /**
-     * Signals that an operation targeting a specified person in the list would fail because
-     * there is no such matching person in the list.
+     * Signals that an operation targeting specified persons in the list would fail because
+     * there is no such matching persons in the list.
      */
-    public static class PersonNotFoundException extends Exception {}
+    public static class PersonNotFoundException extends Exception {
+        public PersonNotFoundException(String indexedListStringOfPersons) {
```
pyokagan: Seems like a pretty arbitrary argument, since we can pass it any kind of string in here. We already *expect* that the string of persons being passed in here is formatted using `StringUtil.toIndexedListString()`, so why not just do it within this constructor to make sure that is true, rather than placing additional responsibilities on the caller?

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -99,9 +118,13 @@ protected DuplicatePersonException() {
     }
 
     /**
-     * Signals that an operation targeting a specified person in the list would fail because
-     * there is no such matching person in the list.
+     * Signals that an operation targeting specified persons in the list would fail because
+     * there is no such matching persons in the list.
      */
-    public static class PersonNotFoundException extends Exception {}
+    public static class PersonNotFoundException extends Exception {
```
pyokagan: Should this class be renamed to Person**s**NotFoundException?

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -50,11 +52,28 @@ public boolean remove(ReadOnlyPerson toRemove) throws PersonNotFoundException {
...
         }
         return personFoundAndDeleted;
     }
 
+    /**
+     * Removes all persons specified by {@code personsToRemove} from this list.<br>
+     * @throws PersonNotFoundException without deleting any persons if any of the {@code personsToRemove}
+     * can't be found in this list.
+     */
+    public void removeAll(Collection<ReadOnlyPerson> personsToRemove) throws PersonNotFoundException {
```
pyokagan: In the `AddressBook` and `Model` interfaces, you converted all the uses of `deletePerson(ReadOnlyPerson)` into `deletePersons(ReadOnlyPerson...)`.

So, I'm wondering why wasn't `remove(ReadOnlyPerson)` converted into a `removeAll(ReadOnlyPerson...)`.

src/main/java/seedu/address/model/person/UniquePersonList.java
cheec: I actually did that prior to this haha, will revert.

src/main/java/seedu/address/model/person/UniquePersonList.java
cheec: oops

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -288,28 +291,111 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
...
+        assertIndexNotFoundBehaviorForCommand("delete 5-3", 4);
     }
 
     @Test
-    public void execute_delete_removesCorrectPerson() throws Exception {
-        TestDataHelper helper = new TestDataHelper();
-        List<Person> threePersons = helper.generatePersonList(3);
+    public void execute_deleteNonRanged_removesCorrectly() throws Exception {
+        TestDataHelper tdh = new TestDataHelper();
+        tdh.addToModel(model, tdh.generatePersonList(10));
```
pyokagan: Actually, I don't think this test even uses the `model`? Same for all the tests below.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -507,4 +593,44 @@ Person generatePersonWithName(String name) throws Exception {
             );
         }
     }
+
+    /**
+     * Utility class for deleting
+     */
+    private class DeleteHelper {
```
pyokagan: Why introduce a full utility class that encapsulates its own state? Why not just use the `model` provided to all tests?

src/test/java/seedu/address/logic/LogicManagerTest.java
pyokagan: Make sure you add a test for running delete when only a subset of persons are shown (due to a filter).

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -507,4 +593,44 @@ Person generatePersonWithName(String name) throws Exception {
...
+            expectedRemainder.removeAll(expectedDeleted);
+
+            String expectedFeedback =
+                    String.format(DeleteCommand.MESSAGE_DELETE_PERSON_SUCCESS, expectedDeleted.size(),
+                                  StringUtil.toIndexedListString(expectedDeleted));
+
+            // set expectation of remainder after deleting
+            expectedAB.setPersons(expectedRemainder);
+
+            assertCommandBehavior(deleteArgs, expectedFeedback, expectedAB, expectedRemainder);
```
pyokagan: I'm thinking this is slightly wrong, since it operates on the `model`.

By the way, in the above tests, you use the code:
```java
        tdh.addToModel(model, tdh.generatePersonList(20));
        DeleteHelper dh = new DeleteHelper(tdh.generateAddressBook(20));
```
which means that the model and the DeleteHelper gets different persons. Since the tests pass, I'm guessing something is not being tested correctly.

cheec:
> I'm more interested in knowing the design decisions behind the code.
> 
> For example, if I recall you initially planned to have 3- to mean "delete from the 3rd person all the way to the last person. Why was that dropped? By the way, I'm not completely satisfied with the approach this commit takes, because it closes the door to us implementing such a feature in the future without needing to rewrite all this code.
> 
> Also, you decided to allow 3-4 and 4-3 to mean the same thing. Why is that so?

Short answer: time :P

mm I don't see anything wrong with 3-4, 4-3 being synonymous. If a user's scrolling through the list from the bottom up while deleting, it might be useful in those cases.

pyokagan:
I'll stop here for now.

cheec:
> Other than the annoying lone *, I don't really understand this argument. If
> you don't trust that ListUtil::subList will perform the job correctly, why
> not just write unit tests for ListUtil::subList to verify its correctness?
> 
> As it is right now, we are implementing the same logic in two different places,
> which is not good.

When I first made this, I commented that the PR was getting too big so i left ListUtil's tests out.
I can add them in though no worries

src/test/java/seedu/address/logic/LogicManagerTest.java
cheec: Actually, the implementation currently uses the same list of persons (by same i mean same details but diff objects). But I get your point, passing in the same reference would be somewhat more resistant to code changes if the internals of generatePersonList and generateAB were to differ in the future.

damithc:
@cheec you can go ahead and do the next version.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -13,38 +20,43 @@
...
     }
 
-
     @Override
     public CommandResult execute() {
-
         UnmodifiableObservableList<ReadOnlyPerson> lastShownList = model.getFilteredPersonList();
 
-        if (lastShownList.size() < targetIndex) {
+        IntegerUtil.applyOffset(targetIndices, -1); // converting to zero-indexed indices
```
cheec: @pyokagan I think it's fine as is? I don't quite see what's wrong with an additional comment. In my opinion it's already quite readable on it's own even without a comment.

src/main/java/seedu/address/model/AddressBook.java
cheec: I tried to retain the previous code naming conventions, I think clean-ups like these can be a separate PR.

src/main/java/seedu/address/model/ModelManager.java
cheec: I dont quite follow, both deletePersons extracted to a default method?

pyokagan: @cheec 
> both deletePersons extracted to a default method?

Just this method.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
pyokagan: @cheec 
> I think it's fine as is? I don't quite see what's wrong with an additional comment.

Our code base has lots of these "convert from one index to zero index" comments since we do lots of these conversions.  But these comments add a lot of bloat and make the code less succinct than it could have been.

So, these comments are problematic, which is why it is somewhat regrettable that we go through the trouble of introducing a new `applyOffset()` method but do not go the last mile of wrapping it up in a meaningful name which would remove the need for these comments.

> In my opinion it's already quite readable on it's own even without a comment.

These comments are needed since it's not always clear whether a `-1` is because we are performing this operation or because we are just doing some other math. A careful reader would need to spend extra time confirming if that is what we are indeed doing. A suitable name would make the intent more obvious, and save readers a lot of time.

But then again I won't push for this, because I have a suspicion that the right way to go about this is to actually make the `Model` interface accept one-indexes instead. This would rid the whole code base of the one-to-zero indexing conversions.

damithc: > But then again I won't push for this, because I have a suspicion that the right way to go about this is to actually make the Model interface accept one-indexes instead. This would rid the whole code base of the one-to-zero indexing conversions.

I think so too. In the meantime, a small wrapper method `oneIndexToZeroIndex` can get rid of the comments.

cheec: Ah I see. 
In my mind, there were all these conflicting principles. Mainly that a method name should be independent of the context it's used in to facilitate re-usability. That is why the name `oneIndexToZeroIndex` appeared so jarring to me when I first read it. But seeing that this method is used quite remotely for now, I guess a temporary rename is okay.

damithc: > a method name should be independent of the context it's used in to facilitate re-usability. 

Good point. It looks as if we are breaking this principle, but that's not the case. If the method is named `adjustIndexForMassDelete(int indexesOfpersonsToDelete)`, then the method has become dependent on the usage context. What we are doing here is to narrow the scope. It can still be used anywhere there is a need to convert from oneIndex to zeroIndex. While the more general method (the current one) makes it more widely reusable, it's too general to the extent that it needs an explanatory  comment every time it is being used. Hence the need to write a narrower-scoped method that is more self-explanatory.

pyokagan: @damithc 
> a small wrapper method oneIndexToZeroIndex can get rid of the comments.

I think we should go with a name that is as short as possible such as `oneToZeroIndex`, otherwise we might discourage devs from using it because the name is too long for an operation which seems trivial.

pyokagan: Just to make it clear:

> Hrm, this means that execute() can only be run once. Probably not what callers expect.

This is a bug that must be fixed.

cheec: What arguments should oneToZeroIndex have? Just the collection of indices alone?

In that case then do we still need IntegerUtil?

pyokagan: @cheec 
> What arguments should oneToZeroIndex accept? Just the collection of indices alone?

Yeah that would be nice. I'm thinking,
* `int oneToZeroIndex(int)`
* `List<Integer> oneToZeroIndex(List<Integer>)`
* `Set<Integer> oneToZeroIndex(Set<Integer>)`
* etc..

> In that case then do we still need IntegerUtil?

I dunno. Where else could we put these methods?

As an aside, I think the utility class may be better served being called `IndexUtil`. Even though we are using ints to store them, indexes aren't really integers since they don't span their whole range.

As another aside, I think we should allow the usage of
```java
import static seedu.address.commons.util.IndexUtil.oneToZeroIndex;
```
so that devs could call `oneToZeroIndex()` directly instead of `IndexUtil.oneToZeroIndex()`, which is extremely long. This is to encourage devs to use this method rather than trying to do the arithmetic themselves.

cheec: > import static seedu.address.commons.util.IndexUtil.oneToZeroIndex;

Could we do this in a separate PR? Allow this for all static utility methods or something

cheec:
@pyokagan, ready for review for now. Everything updated except LogicManagerTest. Im not too sure as to what approach i should go for. Sorry for the late work, I've been really busy the last couple days ><

CanIHasReview:
# v3

@cheec submitted v3 for review.

* [v3 1/9] [IndexUtil: implement oneToZeroIndex\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/f3c877dfb5cf7833428992079b0cd45055a97f1f)
* [v3 2/9] [ListUtil: implement subList\(\) and areIndicesWithinBounds\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/374101d30d4d978c4fc157a8eff029ab471aac7f)
* [v3 3/9] [TestUtil: implement removePersonsFromList\(Collection\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/4d5c6f30c8eca0b9005813804adce156fa63f88f)
* [v3 4/9] [StringUtil: implement toIndexedListString\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/372de35753dd6ed40745432398089be58e48e4cc)
* [v3 5/9] [CollectionUtil: refactor assertNoNullElements \-> isAnyNull\(Collection\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/e19b502fadbf29d3cefc0d59582dcad653276f89)
* [v3 6/9] [Model: implement deletePersons](https://github.com/se-edu/addressbook-level4/pull/248/commits/d2e8474a49e858dec2a34cd96be9862e72e7208d)
* [v3 7/9] [Parser/DeleteCommand: implement logic for deleting](https://github.com/se-edu/addressbook-level4/pull/248/commits/0bbce4dd71003e16579033aec773e7b00ce56100)
* [v3 8/9] [DG: update model API ](https://github.com/se-edu/addressbook-level4/pull/248/commits/b026a0471cc591336c7ce2f85409534eae50aab5)
* [v3 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/6453ee5237fc11f9bade5491e13b5a43826a8e5e)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...6453ee5237fc11f9bade5491e13b5a43826a8e5e)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/commit/033086d1647eb869cdc7089c6504a1254f953fa9))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/3/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/commons/util/IndexUtil.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods related to Integers
+ */
+public class IndexUtil {
+
+    /** Returns a new {@code List} with each {@code Integer} element of {@code source} subtracting {@code 1}. */
+    public static Collection<Integer> oneToZeroIndex(Collection<Integer> source) {
+        return source.stream().map(i -> i - 1).collect(Collectors.toList());
```
pyokagan: The `i - 1` can be abstracted into a lower level `oneToZeroIndex(int) -> int`

This lower level function can be used to perform per-element checks for us: e.g. ensure that all members of this collection are truly one-indices (`assert oneBasedIndex > 0`)

src/main/java/seedu/address/commons/util/IndexUtil.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods related to Integers
+ */
+public class IndexUtil {
+
+    /** Returns a new {@code List} with each {@code Integer} element of {@code source} subtracting {@code 1}. */
+    public static Collection<Integer> oneToZeroIndex(Collection<Integer> source) {
```
pyokagan: Maybe rename `source` to `oneBasedIndexCollection` ? Anyway, we should put the name `oneBasedIndex` in the argument to make it clear that the input is a one-based index thingy.

src/main/java/seedu/address/commons/util/IndexUtil.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods related to Integers
+ */
+public class IndexUtil {
+
+    /** Returns a new {@code List} with each {@code Integer} element of {@code source} subtracting {@code 1}. */
```
pyokagan: Description is too low level, we should be describing the higher-level concept instead.

Something like:
```
Transforms a collection of one based indices into a collection of zero based indices.
```

src/main/java/seedu/address/commons/util/IndexUtil.java
```diff
@@ -0,0 +1,16 @@
+package seedu.address.commons.util;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods related to Integers
```
pyokagan: More like related to indices.

src/main/java/seedu/address/commons/util/IndexUtil.java
pyokagan: Anyway, I thought a bit, and I think that returning a `List` as  `Collection` should be fine.

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void oneToZeroIndex_emptyCollection_emptyResult() throws Exception {
+        assertOneToZeroIndexSuccess(new Integer[0], new Integer[0]);
+    }
+
+    @Test
+    public void oneToZeroIndex_validElements_correctResult() throws Exception {
+        // positive elements
+        assertOneToZeroIndexSuccess(new Integer[] { 1, 5, 6 }, new Integer[] { 0, 4, 5 });
```
pyokagan: Hrm, would it be any less readable if we made this:
```java
assertOneToZeroIndexSuccess(Arrays.asList(1, 5, 6), Arrays.asList(0, 4, 5));
```

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,54 @@
...
+        assertOneToZeroIndexSuccess(new Integer[] { 10, 4, 3 }, new Integer[] { 9, 3, 2 });
+
+        // negative elements
+        assertOneToZeroIndexSuccess(new Integer[] { -10, -5, -2 }, new Integer[] { -11, -6, -3 });
+        assertOneToZeroIndexSuccess(new Integer[] { -3, -8, -1 }, new Integer[] { -4, -9, -2 });
+
+        // 0 elements
+        assertOneToZeroIndexSuccess(new Integer[] { 0, 0, 0 }, new Integer[] { -1, -1, -1 });
```
pyokagan: Eh we should be throwing an `AssertionError` or something for this case.

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Test
+    public void oneToZeroIndex_validElements_correctResult() throws Exception {
+        // positive elements
+        assertOneToZeroIndexSuccess(new Integer[] { 1, 5, 6 }, new Integer[] { 0, 4, 5 });
+        assertOneToZeroIndexSuccess(new Integer[] { 10, 4, 3 }, new Integer[] { 9, 3, 2 });
+
+        // negative elements
+        assertOneToZeroIndexSuccess(new Integer[] { -10, -5, -2 }, new Integer[] { -11, -6, -3 });
```
pyokagan: We should probably also be throwing `AssertionError` for these cases.

src/test/java/seedu/address/commons/util/IndexUtilTest.java
pyokagan: Try having a test case with repeated elements as well.

src/main/java/seedu/address/commons/util/ListUtil.java
```diff
@@ -0,0 +1,40 @@
...
+    public static <T> List<T> subList(List<T> list, Integer... indices) {
+        return subList(list, Arrays.asList(indices));
+    }
+
+    /**
+     * Returns a sublist of items from {@code list} specified by {@code indices}.<br>
+     * - Original list order is maintained in the extracted list.<br>
+     * - Duplicate indices are ignored.<br>
+     */
+    public static <T> List<T> subList(List<T> list, Collection<Integer> indices) {
```
pyokagan: I think we should make `subList` take a Set of indices instead, rather than silently ignoring duplicates. There are a few reasons why:

* The type signature of the method should describe its functionality as succinctly as possible. This allows readers of code to quickly grasp the responsibility of the method by looking at its input and output types.

* From a logical perspective, it makes sense to say "I want a sublist with items 0, 1 and 2", but it does not make sense to say "I want a sublist with items 0 and 0" [1]. Such a query is totally nonsense.

* This would force callers to think long and hard about whether they should be passing a collection of indices with duplicates in the first place. Passing a collection with duplicates may actually indicate a bug in the program (e.g. in test code, having test data with duplicate indices)

[1] An alternative interpretation of `subList()` could actually be to directly map each index to its respective item of the list, respecting the ordering and any duplicates. Taking a `Set` instead of a `Collection` would help readers avoid assuming this wrong interpretation.

src/main/java/seedu/address/commons/util/ListUtil.java
```diff
@@ -0,0 +1,40 @@
...
+ * Utility methods related to Lists
+ */
+public class ListUtil {
+
+    /**
+     * Returns a sublist of items from {@code list} specified by {@code indices}.<br>
+     * - Original list order is maintained in the extracted list.<br>
+     * - Duplicate indices are ignored.<br>
+     */
+    public static <T> List<T> subList(List<T> list, Integer... indices) {
```
pyokagan: I think we shouldn't have this overload in `ListUtil`: it allows developers to write nonsense code like:
```java
subList(someList, 0, 1, 2, 0);
```

All existing usages of this method are wrapper methods that already allow callers to pass a variable number of arguments anyway, and can even help to check for any accidental duplicate indices. So, we technically don't need this overload.

src/main/java/seedu/address/commons/util/ListUtil.java
```diff
@@ -0,0 +1,40 @@
...
+        assert areIndicesWithinBounds(list, indices);
+        return indices.stream()
+                .distinct()
+                .sorted()
+                .map(list::get)
+                .collect(Collectors.toList());
+    }
+
+    /** Returns true if every index in {@code indices} is within bounds of {@code list}. */
+    public static boolean areIndicesWithinBounds(List<?> list, Collection<Integer> indices) {
```
cheec: areIndicesWithinList sounds misleading though. some might think that indices are the actual elements of the list

src/main/java/seedu/address/commons/util/StringUtil.java
```diff
@@ -55,4 +59,20 @@ public static String getDetails(Throwable t) {
...
     }
+
+    /**
+     * Returns a string that has each item in {@code items} on its own one-indexed line.<br>
+     * E.g. {@code [spam, ham, eggs]} becomes:<br>
+     * <code>
+     *   1. spam<br>
+     *   2. ham<br>
+     *   3. eggs
+     * </code>
```
pyokagan: Would it be bad if we wrote this as `"1. spam\n2. ham\n3. eggs"` instead? To make it clear that it doesn't have a trailing newline.

cheec: i made it that way to properly reflect how it would look like in reality.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -1,50 +1,61 @@
...
+    public static final String MESSAGE_DELETE_PERSON_SUCCESS =
+            "Deleted %1$s Person(s):\n"
+            + "%2$s";
 
-    public final int targetIndex;
+    public final Collection<Integer> targetIndices;
 
-    public DeleteCommand(int targetIndex) {
-        this.targetIndex = targetIndex;
+    public DeleteCommand(Collection<Integer> targetIndices) {
```
pyokagan: I think this should take a `Set` as well, to force callers to think about whether they should allow duplicate `targetIndices`.

This way, whether to allow duplicate target indices or not becomes the responsibility of the `Parser`, as it should be. (since this is a command format thing)

src/main/java/seedu/address/commons/util/ListUtil.java
pyokagan: @cheec 
> are the actual elements

But elements != indices

src/main/java/seedu/address/commons/util/StringUtil.java
pyokagan: OK. Up to you.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -39,9 +39,9 @@ public CommandResult execute() {
         ReadOnlyPerson personToDelete = lastShownList.get(targetIndex - 1);
 
         try {
-            model.deletePerson(personToDelete);
-        } catch (PersonNotFoundException pnfe) {
-            assert false : "The target person cannot be missing";
+            model.deletePersons(personToDelete);
+        } catch (PersonsNotFoundException pnfe) {
+            assert false : "DeleteCommand: " + pnfe.getMessage();
```
pyokagan: No need for the `DeleteCommand:` prefix, the stack trace will show that the exception originated from `DeleteCommand`.

By the way, our new code policy requires that control invariants (this `assert` here) be converted into a `throw new AssertionError(pnfe)`.

pyokagan:
I'll stop here for now.

src/main/java/seedu/address/commons/util/ListUtil.java
pyokagan: Oh, how about `isListIndices`?

pyokagan: or `areIndicesOfList`.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
pyokagan: Hmph. Not sure why GitHub keeps showing this as "outdated".

CanIHasReview:
# v4

@cheec submitted v4 for review.

* [v4 1/9] [IndexUtil: implement oneToZeroIndex\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/a666f7d5e79fa4cb97dc8485a6f8c1fefed3a44d)
* [v4 2/9] [ListUtil: implement subList\(\) and areIndicesWithinBounds\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/32664e5f964540418a434eddef826258514cdd26)
* [v4 3/9] [TestUtil: implement removePersonsFromList\(Collection\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/bbb1afe4794b90f5cf1f66b5028213808c7e372b)
* [v4 4/9] [StringUtil: implement toIndexedListString\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/516cb02ec05bac7c1ce43f68f939f8bdc48ef35e)
* [v4 5/9] [CollectionUtil: refactor assertNoNullElements \-> isAnyNull\(Collection\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/ec46296e7adedee7f89194eeef46f90f530bdb47)
* [v4 6/9] [Model: implement deletePersons](https://github.com/se-edu/addressbook-level4/pull/248/commits/cfb769b81db9a53bb5f779b5aa71bb56f1a4d98e)
* [v4 7/9] [Parser/DeleteCommand: implement logic for deleting](https://github.com/se-edu/addressbook-level4/pull/248/commits/89b31896c3379cf4505ac967ca0adb024505e5bb)
* [v4 8/9] [DG: update model API ](https://github.com/se-edu/addressbook-level4/pull/248/commits/65bf40431e958bc9f162e96a9f71a23a5b6efef7)
* [v4 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/18ab88a281565dc4ce52bba045e5204cd508e651)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/da4dc8e78218b35ccf991108acf1c56ca8492f57...18ab88a281565dc4ce52bba045e5204cd508e651)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/commit/844b8236ded5fd022b3365e5f5c3e045c37120f7))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/4/head:BRANCHNAME
```

</details>

cheec:
@pyokagan ready for review

src/main/java/seedu/address/commons/util/ListUtil.java
cheec: oh oops, i didnt see this, i will add this next iteration

cheec: "But elements != indices"

mm yeah but i'm saying the name reads as thought the list contains 'indices' of some form

cheec: after all, Java throws IndexOutOf**Bounds**Exceptions. So I just followed along

pyokagan: @cheec 
> after all, Java throws IndexOutOfBoundsExceptions. So I just followed along

The circumstances are different. Note that `IndexOutOfBoundsException` is a `java.lang` exception (it is defined at the Java language level). In other words, it is not just meant to be used in the context of Java collections. As such, it needs to have an extremely generic name. 

On the other hand, `ListUtil` exists at the Java collections level, which is at a higher level than the lowest-level Java language level. As such, we can and should aim for more specific and descriptive names.

But this makes me wonder: why is a bounds-checking method implemented in `ListUtil`? At first glance, it may seem that it deserves to be lumped with other list utility methods because this specific implementation of `areIndicesWithinBounds()` just happens to take a `List<?>`

However, as I noted in the above message, "bounds-checking" is not only done with a List. It could also be done with a pair of start-end indices, or it could be done with an `IndexRange` (if that ever becomes a top-level class). It can also be done with Strings or Arrays as well. In other words, there is no single type which can represent an "index boundary".

There are also smaller issues like how the name of the method is called `areIndicesWithinBounds`, yet the "bounds" are the first argument and the "indices" the second argument, which is totally opposite from the name.

So, all these issues point to one conclusion: `ListUtil` is not a good place for a method called `areIndicesWithinBounds`. Instead, it should probably be in `IndexUtil` instead.

Starting from basics, a simple index bounds-checking method would look like:
```java
public static boolean areIndicesWithinBounds(Collection<Integer> indices, int start, int endExclusive) {
    return indices.stream().allMatch(i -> i >= start && i < endExclusive);
}
```

We can then build on top these basic method with other boundary types such as a `List`:
```java
public static boolean areIndicesWithinBounds(Collection<Integer> indices, List<?> bounds) {
    return areIndicesWithinBounds(indices, 0, bounds.size());
}
```

pyokagan:
@cheec Can you pull [v4 5/9] into its own PR? I'd like to merge it asap since it's a bugfix.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -237,23 +239,24 @@ private void assertIncorrectIndexFormatBehaviorForCommand(String commandWord, St
...
-     * targeting a single person in the shown list, using visible index.
-     * @param commandWord to test assuming it targets a single person in the last shown list
+     * Confirms the 'invalid argument index number behaviour' for the given {@code commandArgs}
+     * targeting persons in the shown list of size {@code listSize}, using visible index.
+     * @param commandArgs to test assuming it targets persons in the last shown list
      *                    based on visible index.
+     * @param listSize last shown list size
      */
-    private void assertIndexNotFoundBehaviorForCommand(String commandWord) throws Exception {
+    private void assertIndexNotFoundBehaviorForCommand(String commandArgs, int listSize) throws Exception {
```
pyokagan: @cheec Actually, I think the correct way to go around this is to have an `assertIndexNotFoundBehaviorForCommand(String commandArgs)` overload for tests that do not require a specific `listSize`.

That way, you don't have to touch all the current uses of that method.

pyokagan:
@cheec You can squash [v4 8/9] into [v4 6/9] as well.

pyokagan:
You can submit a new version now.

CanIHasReview:
# v5

@cheec submitted v5 for review.

* [v5 1/9] [TestUtil: implement asIntegerSet\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/ab0e889aa6723c08e5907b9458dffa1ce238488f)
* [v5 2/9] [IndexUtil: implement oneToZeroIndex\(\) and areIndicesWithinBounds\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/4cd0e95bf17252bbb63e48fd431b253ddf667669)
* [v5 3/9] [ListUtil: implement subList\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/a5c84295752b87a73eac3dc89761e466aa283b44)
* [v5 4/9] [TestUtil: implement removePersonsFromList\(Collection\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/04b9e7115dc861b22feed802f9d5098c15a2268a)
* [v5 5/9] [StringUtil: implement toIndexedListString\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/787d54b9a18c285d401692dbfed5ce74ab08a6bc)
* [v5 6/9] [Model: implement deletePersons](https://github.com/se-edu/addressbook-level4/pull/248/commits/35cb70d06436e89712d85198c9869f5b8aa335ed)
* [v5 7/9] [LogicManagerTest: refactor assertIndexNotFoundBehaviorForCommand\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/a4d1a406f3a88b4f35063ad41d356381a1aabc75)
* [v5 8/9] [Parser/DeleteCommand: implement logic for deleting](https://github.com/se-edu/addressbook-level4/pull/248/commits/54155be391b439b6abdd24caebaacb407c311d08)
* [v5 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/49256ba77a628c58504cf83d4da7f462d87ded0b)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/941d0afebe52dc6bd86b60bd4a11861f48273900...49256ba77a628c58504cf83d4da7f462d87ded0b)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/commit/753b737573b7ffc1507fb04c86114a2895731895))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/5/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/testutil/TestUtil.java
```diff
@@ -357,4 +359,9 @@ public static boolean compareCardAndPerson(PersonCardHandle card, ReadOnlyPerson
         return collect.toArray(new Tag[split.length]);
     }
 
+    /** Converts {@code items} to an {@code Integer Set}. */
+    public static Set<Integer> asIntegerSet(Integer... items) {
```
damithc: Seems like a generic `toSet<T>` can be used instead? Anyway, not strictly needed for this PR, so can be left out.

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,132 @@
...
+    public ExpectedException thrown = ExpectedException.none();
+
+    private void assertOneToZeroIndexSetSuccess(Set<Integer> testIntegers,
+                                                Set<Integer> expectedIntegers) {
+        assertEquals(expectedIntegers, IndexUtil.oneToZeroIndex(testIntegers));
+    }
+
+    private void assertOneToZeroIndexIntSuccess(int testInteger, int expectedInteger) {
+        assertEquals(expectedInteger, IndexUtil.oneToZeroIndex(testInteger));
+    }
```
damithc: The convention is to place lower-level utility methods below (not above) where they are used. That way, code can be read in a top down manner.

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @Test
+    public void oneToZeroIndexSet_containingNull_throwsNullPointerException() throws Exception {
+        thrown.expect(NullPointerException.class);
+        IndexUtil.oneToZeroIndex(asIntegerSet((Integer) null));
+    }
+
+    @Test
+    public void oneToZeroIndexSet_negativeIndices_assertionError() throws Exception {
+        thrown.expect(AssertionError.class);
+        IndexUtil.oneToZeroIndex(asIntegerSet(-10, -5, -2));
```
damithc: The boundary case for this is a list containing just one negative value `-1`

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    public void oneToZeroIndexSet_zeroValuedIndices_assertionError() throws Exception {
+        thrown.expect(AssertionError.class);
+        IndexUtil.oneToZeroIndex(asIntegerSet(0, 0, 0));
```
damithc: Just one `0` is enough?

src/test/java/seedu/address/commons/util/IndexUtilTest.java
```diff
@@ -0,0 +1,132 @@
...
+    @Test
+    public void oneToZeroIndexSet_emptySet_emptyResult() throws Exception {
+        assertOneToZeroIndexSetSuccess(Collections.emptySet(), Collections.emptySet());
+    }
+
+    @Test
+    public void oneToZeroIndexSet_validIndices_correctResult() throws Exception {
+        assertOneToZeroIndexSetSuccess(asIntegerSet(1, 5, 6), asIntegerSet(0, 4, 5));
+        assertOneToZeroIndexSetSuccess(asIntegerSet(10, 4, 3), asIntegerSet(9, 3, 2));
+        assertOneToZeroIndexSetSuccess(asIntegerSet(1, 1, 1), asIntegerSet(0, 0, 0));
```
damithc: Document the rational for each case and get rid of cases that don't cover new scenarios. These three cases seem to be testing the same thing.

src/test/java/seedu/address/commons/util/ListUtilTest.java
```diff
@@ -0,0 +1,78 @@
...
+    public void subList_nonEmptyListOutOfBoundsIndices_assertionError() {
+        List<Object> list = Arrays.asList("spam", 6, new ArrayList<>());
+        thrown.expect(AssertionError.class);
+        ListUtil.subList(list, asIntegerSet(list.size()));
+    }
+
+    @Test
+    public void subList_emptyListOutOfBoundsIndices_assertionError() {
+        thrown.expect(AssertionError.class);
+        ListUtil.subList(Collections.emptyList(), asIntegerSet(0, 3, 6));
```
damithc: `asIntegerSet(0)` should be enough? No need to use more test data than necessary.

src/test/java/seedu/address/commons/util/ListUtilTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void subList_validInputs_correctResult() {
+        // list containing nulls
+        List<Object> list = Arrays.asList(null, "spam", 5);
+        List<Object> expectedSubList = Arrays.asList((Object) null);
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(0)));
+
+        list = Arrays.asList("spam", 5, "eggs", "ham", new ArrayList<>());
+        expectedSubList = Arrays.asList("spam", "eggs", new ArrayList<>());
```
damithc: Use of `new ArrayList<>()` as an element in the list can be a bit confusing given that we are dealing with a few other lists in the code. may be use some other simpler object?

src/test/java/seedu/address/commons/util/ListUtilTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        // indices in sorted order
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(0, 2, list.size() - 1)));
+
+        // indices unsorted but original list order maintained
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(list.size() - 1, 2, 0)));
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(list.size() - 1, 0, 2)));
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(2, list.size() - 1, 0)));
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(2, 0, list.size() - 1)));
+        assertEquals(expectedSubList, ListUtil.subList(list, asIntegerSet(0, list.size() - 1, 2)));
```
damithc: `list.size() - 1` repeats a few times.
Perhaps define three constants `MIN, MID, MAX` first (for `0, 2, list.size() - 1`)?

src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -148,5 +153,66 @@ public void getDetails_nullGiven_assertionError() {
...
+    @Test
+    public void toIndexedListString_nullGiven_throwsNullPointerException() {
+        thrown.expect(NullPointerException.class);
+        StringUtil.toIndexedListString(null);
+    }
+
+    @Test
+    public void toIndexedListString_listContainsNull_throwsNullPointerException() {
+        Collection<Object> containingNull = new ArrayList<>();
+        containingNull.add(null);
```
damithc: Can these two lines be collapsed to one line using `Arrays.asList()`? That way, we can avoid the temporary lie in line 174 (i.e. the object doesn't contain any nulls at that point).

src/test/java/seedu/address/testutil/TestUtil.java
pyokagan: @damithc 
> Seems like a generic toSet<T> can be used instead?

Yes, in terms of naming, the method should definitely be called `toSet()`.

CanIHasReview:
# v6

@cheec submitted v6 for review.

* [v6 1/9] [TestUtil: implement toSet\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/73fbed84304235103c90e05b2f391f8415692e9d)
* [v6 2/9] [IndexUtil: implement oneToZeroIndex\(\) and areIndicesWithinBounds\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/5e45c1573e1af3d63b92736309f7771fa2ed80cf)
* [v6 3/9] [ListUtil: implement subList\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/a05908ae450e117768c5e9c3cb49a76081f2d790)
* [v6 4/9] [TestUtil: implement removePersonsFromList\(Collection\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/826802fdc189e69ee5d8cfa396ce824ea5d2d420)
* [v6 5/9] [StringUtil: implement toIndexedListString\(\) ](https://github.com/se-edu/addressbook-level4/pull/248/commits/72e39deb77725a6413b409c4f913c119a8ee3a72)
* [v6 6/9] [Model: implement deletePersons](https://github.com/se-edu/addressbook-level4/pull/248/commits/094106330849fdcc25c39ae4c53e63c9011efef8)
* [v6 7/9] [LogicManagerTest: refactor assertIndexNotFoundBehaviorForCommand\(\)](https://github.com/se-edu/addressbook-level4/pull/248/commits/001b2498fd2f11d88e7b36933147c9fb41cf9523)
* [v6 8/9] [Parser/DeleteCommand: implement logic for deleting](https://github.com/se-edu/addressbook-level4/pull/248/commits/a598e6085ab2405f24c231a1d0fde038f9d51b9f)
* [v6 9/9] [UG: update new delete functionality ](https://github.com/se-edu/addressbook-level4/pull/248/commits/6d028d180e59f6721b3b40b39837710cc6482200)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/2221d24f2e49a9ec8a9c0d45b9dcb03694d25b0f...6d028d180e59f6721b3b40b39837710cc6482200)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/8d16ae32007f63bd9167454982a5a9a465c6d7cf~1...8d16ae32007f63bd9167454982a5a9a465c6d7cf))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/248/6/head:BRANCHNAME
```

</details>

cheec:
Implemented suggestions by Prof @damithc

damithc:
@cheec Good job on this PR so far. Due to some other design cleanup that's going on, we need to put this on ice for a while. Hope you can resume this at a later time.

cheec:
@damithc, sure no worries. I'm glad I could help.

Zhiyuan-Amos:
What's the status of this PR? ^^

pyokagan:
@cheec Shall we continue with this PR?

@damithc Do you still want this feature?

cheec:
Hi Paul! Unfortunately, I don't think I have enough bandwidth to redo this PR as of now. Perhaps someone else can continue from where I left off.

pyokagan:
OK

damithc:
> @damithc Do you still want this feature?

No compelling need to add this feature for level 4. Will close issue.

PR #251, opened 2017-01-16
checkstyle: ensure overloaded methods are grouped together #229

### LINKED ISSUE #229: checkstyle: ensure overloaded methods are grouped together
This improves the readability of the code as it allows developers to see all possible variations of a method at one place.

chao1995:
Fixes #229 

Ready for review.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -209,7 +209,7 @@
 
 
     <!--
-    MODIFIERS CHECKS
+    ORDER CHECKS
     -->
 
     <module name="ModifierOrder">
@@ -220,6 +220,7 @@
         -->
     </module>
 
+    <module name="OverloadMethodsDeclarationOrder"/>
 
     <!--
     WHITESPACE CHECKS
--- a/src/main/java/seedu/address/commons/core/LogsCenter.java
+++ b/src/main/java/seedu/address/commons/core/LogsCenter.java
@@ -51,6 +51,14 @@ public class LogsCenter {
         return Logger.getLogger(name);
     }
 
+    /**
+     * Creates a Logger for the given class name.
+     */
+    public static <T> Logger getLogger(Class<T> clazz) {
+        if (clazz == null) return Logger.getLogger("");
+        return getLogger(clazz.getSimpleName());
+    }
+
     private static void addConsoleHandler(Logger logger) {
         if (consoleHandler == null) consoleHandler = createConsoleHandler();
         logger.addHandler(consoleHandler);
@@ -83,14 +91,6 @@ public class LogsCenter {
         ConsoleHandler consoleHandler = new ConsoleHandler();
         consoleHandler.setLevel(currentLogLevel);
         return consoleHandler;
-    }
-
-    /**
-     * Creates a Logger for the given class name.
-     */
-    public static <T> Logger getLogger(Class<T> clazz) {
-        if (clazz == null) return Logger.getLogger("");
-        return getLogger(clazz.getSimpleName());
     }
 
     /**
--- a/src/main/java/seedu/address/commons/core/UnmodifiableObservableList.java
+++ b/src/main/java/seedu/address/commons/core/UnmodifiableObservableList.java
@@ -38,13 +38,13 @@ public class UnmodifiableObservableList<E> implements ObservableList<E> {
     }
 
     @Override
-    public final void removeListener(ListChangeListener<? super E> listener) {
-        backingList.removeListener(listener);
+    public final void addListener(InvalidationListener listener) {
+        backingList.addListener(listener);
     }
 
     @Override
-    public final void addListener(InvalidationListener listener) {
-        backingList.addListener(listener);
+    public final void removeListener(ListChangeListener<? super E> listener) {
+        backingList.removeListener(listener);
     }
 
     @Override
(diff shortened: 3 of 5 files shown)


damithc:
@chao1995 you can request a peer review from one of the team member you think is suitable to review this.

damithc:
Also, you can add the label `toReview` :-)

yamgent:
I tested it and it seems to be working most of the time.

However, I was able to break this rule if there are other class/interface declarations in the middle of the two methods. Not sure whether we can fix this or not (and whether we should be concerned about it):

```diff
@@ -101,14 +101,18 @@ public class ModelManager extends ComponentManager implements Model {
    private void updateFilteredPersonList(Expression expression) {
        filteredPersons.setPredicate(expression::satisfies);
    }

    //========== Inner classes/interfaces used for filtering =================================================

    interface Expression {
        boolean satisfies(ReadOnlyPerson person);
        String toString();
    }

+   public void updateFilteredPersonList(String keyword, boolean someRandomBool) {
+       // do nothing
+   }
+
    private class PredicateExpression implements Expression {

        private final Qualifier qualifier;
```

damithc:
> However, I was able to break this rule if there are other class/interface declarations in the middle of the two methods. Not sure whether we can fix this or not:

I think we don't need to be worried about it. I sounds like a loophole in checkstyle which could be fixed in a future version.

damithc:
@m133225 for your inputs

config/checkstyle/checkstyle.xml
```diff
@@ -195,7 +195,7 @@
 
 
     <!--
-    MODIFIERS CHECKS
+    ORDER CHECKS
```
damithc: What's the reason for this change?

chao1995: This is because both "ModifierOrder" and "OverloadMethodsDeclarationOrder" have "Order" as suffixes.

damithc: I see. I wonder if these terms were in the original default checkstyle configuration that came with checkstyle (if there is such a thing) and in that case we may want to keep the same.

chao1995: Yes both "ModifierOrder" and "OverloadMethodsDeclarationOrder" are directly provided by checkstyle. So I think we are using them right (without modification to the names).

chao1995:
@yamgent Nice spot! Thanks!

As @damithc said, this is a potential loophole in checkstyle, though we can add ["InnerTypeLast"](http://checkstyle.sourceforge.net/config_design.html#InnerTypeLast) rule to help this specific case, so the violating overloaded method should not appear after the inner interface.

However, "InnerTypeLast" is not specified in Google Java code style, so I think for this PR's concern, we just don't add the "InnerTypeLast" first (maybe we can add in a separate PR if necessary), but add more info in the commit message documenting this potential loophole in checkstyle.

chao1995:
@m133225 @yamgent Thanks for the review! Updated commit message.

damithc:
@chao1995 can rebase branch while waiting for @m133225 's approval?

chao1995:
@damithc Rebased.

pyokagan:
Some comments on the commit message:

> Check overloaded methods are grouped together in checkstyle
> 

The rationale for this change is missing. Why do we need this extra checkstyle rule?

> Note that there is a potential loophole in checkstyle
> OverloadMethodsDeclarationOrder module which permits out-of-order
> overloaded methods exist when they are separated by a nested class

This reads a bit strangely. Maybe remove `exist`?

> or interface.
> 
> e.g. the overloaded updateFilteredPersonList in ModelManager.java
> will pass.

I think we can rephrase this more clearly as:
```
e.g. the following code in ModelManager.java still passes the check, even
though there is a nested interface between the two `updateFilteredPersonList()`
method definitions:
```
> 
> private void updateFilteredPersonList(Expression expression) {
>     filteredPersons.setPredicate(expression::satisfies);
> }
> 
> //========== Inner classes/interfaces ==================
> interface Expression {
>     boolean satisfies(ReadOnlyPerson person);
>     String toString();
> }
> 
> public void updateFilteredPersonList(String keyword, boolean someRandomBool) {
>     // do nothing
> }

Indent the code by 4 spaces to make it clear that it is a code block.

m133225:
@chao1995 Agree with the changes that @pyokagan has suggested. Do remember to update it!

CanIHasReview:
# v1

@chao1995 submitted v1 for review.

* [v1 1/1] [Check overloaded methods are grouped together in checkstyle](https://github.com/se-edu/addressbook-level4/pull/251/commits/dbf8a42ea961a54ac6efbbef8a3a1e9c5d9c62a6)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...dbf8a42ea961a54ac6efbbef8a3a1e9c5d9c62a6))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/251/1/head:BRANCHNAME
```

</details>

chao1995:
@m133225 Updated!

damithc:
@chao1995 do a rebase  while we wait for @pyokagan to have another look?

pyokagan:
Looks good overall.

Just a small nit that I think @chao1995 missed out from https://github.com/se-edu/addressbook-level4/pull/251#issuecomment-273757966 :

> Indent the code by 4 spaces to make it clear that it is a code block.

I've fixed that up on my end.

Also, the commit message wasn't wrapped at 72 characters. I fixed that up on my end too.

PR #144, opened 2016-10-05
Allow adding names with comma or hyphen #110

okkhoy:
Fixes #110

PR #265, opened 2017-01-21
Allow adding names with comma or hyphen #110

### LINKED ISSUE #110: Add doesn't allow adding names with comma or hyphen
The regex is too restrictive (`[\\p{Alnum} ]+`)

kychua:
Fixes #110 

based on #144

### DIFF
--- a/src/main/java/seedu/address/model/person/Name.java
+++ b/src/main/java/seedu/address/model/person/Name.java
@@ -20,11 +20,11 @@ public class Name {
      */
     public Name(String name) throws IllegalValueException {
         assert name != null;
-        String trimmedName = name.trim();
-        if (!isValidName(trimmedName)) {
+        String standardizedName = StringUtil.stripLineTerminators(name).trim();
+        if (!isValidName(standardizedName)) {
             throw new IllegalValueException(MESSAGE_NAME_CONSTRAINTS);
         }
-        this.fullName = trimmedName;
+        this.fullName = standardizedName;
     }
 
     /**
--- a/src/main/java/seedu/address/model/person/Name.java
+++ b/src/main/java/seedu/address/model/person/Name.java
@@ -1,6 +1,7 @@
 package seedu.address.model.person;
 
 import seedu.address.commons.exceptions.IllegalValueException;
+import seedu.address.commons.util.StringUtil;
 
 /**
  * Represents a Person's name in the address book.
@@ -8,14 +9,7 @@ import seedu.address.commons.exceptions.IllegalValueException;
  */
 public class Name {
 
-    public static final String MESSAGE_NAME_CONSTRAINTS =
-            "Person names should only contain alphanumeric characters and spaces, and it should not be blank";
-
-    /*
-     * The first character of the address must not be a whitespace,
-     * otherwise " " (a blank string) becomes a valid input.
-     */
-    public static final String NAME_VALIDATION_REGEX = "[\\p{Alnum}][\\p{Alnum} ]*";
+    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names can take any value and should not be blank";
 
     public final String fullName;
 
@@ -37,9 +31,8 @@ public class Name {
      * Returns true if a given string is a valid person name.
      */
     public static boolean isValidName(String test) {
-        return test.matches(NAME_VALIDATION_REGEX);
+        return !StringUtil.isBlank(test);
     }
-
 
     @Override
     public String toString() {
--- a/src/test/java/seedu/address/logic/LogicManagerTest.java
+++ b/src/test/java/seedu/address/logic/LogicManagerTest.java
@@ -50,6 +50,8 @@ import seedu.address.model.tag.UniqueTagList;
 import seedu.address.storage.JsonUserPrefsStorage;
 import seedu.address.storage.StorageManager;
 import seedu.address.storage.XmlAddressBookStorage;
+import seedu.address.testutil.ExpectedFailure;
+import seedu.address.testutil.ExpectedFailureRule;
 
 
 public class LogicManagerTest {
@@ -59,6 +61,9 @@ public class LogicManagerTest {
      */
     @Rule
     public TemporaryFolder saveFolder = new TemporaryFolder();
+
+    @Rule
+    public ExpectedFailureRule expectedFailure = new ExpectedFailureRule();
 
     private Model model;
     private Logic logic;
@@ -201,8 +206,6 @@ public class LogicManagerTest {
 
     @Test
     public void execute_add_invalidPersonData() {
-        assertCommandFailure("add []\\[;] p/12345 e/valid@e.mail a/valid, address",
-                Name.MESSAGE_NAME_CONSTRAINTS);
         assertCommandFailure("add Valid Name p/not_numbers e/valid@e.mail a/valid, address",
                 Phone.MESSAGE_PHONE_CONSTRAINTS);
         assertCommandFailure("add Valid Name p/12345 e/notAnEmail a/valid, address",
@@ -239,9 +242,24 @@ public class LogicManagerTest {
 
         // execute command and verify result
         assertCommandFailure(helper.generateAddCommand(toBeAdded),  AddCommand.MESSAGE_DUPLICATE_PERSON);
-
     }
 
+    @Test
+    @ExpectedFailure
+    public void execute_addDuplicateUnicode_notAllowed() throws Exception {
+        // setup expectations
+        TestDataHelper helper = new TestDataHelper();
+        Person personWithPrecomposedUnicodeName = helper.generatePersonWithName("Bj\u00F6rg");
+        Person personWithDecomposedUnicodeName = helper.generatePersonWithName("Bj\u006F\u0308rg");
+
+        // setup starting state
+        model.addPerson(personWithPrecomposedUnicodeName);
+
+        // TODO: unicode names that look the same but are written in different ways should be considered identical
+        assertCommandFailure(helper.generateAddCommand(personWithDecomposedUnicodeName),
+                AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+    }
 
     @Test
     public void execute_list_showsAllPersons() throws Exception {
@@ -395,6 +413,27 @@ public class LogicManagerTest {
     }
 
     @Test
+    @ExpectedFailure
+    public void execute_find_isNotCaseSensitiveForUnicode() throws Exception {
+        TestDataHelper helper = new TestDataHelper();
+        Person p1 = helper.generatePersonWithName("bla ﬂu\u00DFchen");
+        Person p2 = helper.generatePersonWithName("fluSSchen bla blah");
+
+        List<Person> personsWithUnicodeNames = helper.generatePersonList(p1, p2);
+        AddressBook expectedAB = helper.generateAddressBook(personsWithUnicodeNames);
+        List<Person> expectedList = personsWithUnicodeNames;
+
+        helper.addToModel(model, personsWithUnicodeNames);
+
+        // TODO: case insensitive comparisons do not work for unicode characters
+        // as String#equalsIgnoreCase(String) uses only simple case folding.
+        assertCommandSuccess("find FLUSSCHEN",
+                Command.getMessageForPersonListShownSummary(expectedList.size()),
+                expectedAB,
+                expectedList);
+    }
+
+    @Test
     public void execute_find_matchesIfAnyKeywordPresent() throws Exception {
         TestDataHelper helper = new TestDataHelper();
         Person pTarget1 = helper.generatePersonWithName("bla bla KEY bla");
@@ -413,6 +452,26 @@ public class LogicManagerTest {
                 expectedList);
     }
 
+    @Test
+    @ExpectedFailure
+    public void execute_find_matchesIfAnyUnicodeKeywordPresent() throws Exception {
+        TestDataHelper helper = new TestDataHelper();
+        Person target1 = helper.generatePersonWithName("bla bla Bj\u006F\u0308rg bla");
+        Person target2 = helper.generatePersonWithName("bla keY bla bsdfkj");
+        Person target3 = helper.generatePersonWithName("Bj\u00F6rg Bj\u00F6rg");
+        Person p1 = helper.generatePersonWithName("asdfds");
+
+        List<Person> fourUnicodePersons = helper.generatePersonList(target1, p1, target2, target3);
+        AddressBook expectedAB = helper.generateAddressBook(fourUnicodePersons);
+        List<Person> expectedList = helper.generatePersonList(target1, target2, target3);
+        helper.addToModel(model, fourUnicodePersons);
+
+        // TODO: unicode names that look the same but are written in different ways should be considered identical
+        assertCommandSuccess("find key Bj\u00F6rg",
+                Command.getMessageForPersonListShownSummary(expectedList.size()),
+                expectedAB,
+                expectedList);
+    }
 
     /**
      * A utility class to generate test data.
--- a/src/main/java/seedu/address/commons/util/StringUtil.java
+++ b/src/main/java/seedu/address/commons/util/StringUtil.java
@@ -55,4 +55,16 @@ public class StringUtil {
     public static boolean isUnsignedInteger(String s) {
         return s != null && s.matches("^0*[1-9]\\d*$");
     }
+
+    /**
+     * Returns true if s is null, empty or contains only (Unicode) whitespace characters.
+     */
+    public static boolean isBlank(String s) {
+        if (s == null) {
+            return true;
+        }
+
+        boolean hasOnlyWhitespace = s.codePoints().parallel().allMatch(cp -> Character.isWhitespace(cp));
+        return hasOnlyWhitespace;
+    }
 }
--- a/src/main/java/seedu/address/commons/util/StringUtil.java
+++ b/src/main/java/seedu/address/commons/util/StringUtil.java
@@ -7,6 +7,7 @@ import java.io.StringWriter;
  * Helper functions for handling strings.
  */
 public class StringUtil {
+    public static final String MULTILINE_REGEX = "(?s).*\\R.*"; // (?s) so that . matches line terminators as well
 
     /**
      * Returns true if the {@code sentence} contains the {@code word}.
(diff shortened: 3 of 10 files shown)


CanIHasReview:
# v1

@kychua submitted v1 for review.

* [v1 1/2] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/edf78d3b750d58a506ba5ddb9bf6a2009fc50ad0)
* [v1 2/2] [Update test person's name to test new name regex](https://github.com/se-edu/addressbook-level4/pull/265/commits/1f9d203d8e4a7989addeb1f76271756c9fef3314)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8164678f10e07049144c98acd52fc3572ad5c9a8...1f9d203d8e4a7989addeb1f76271756c9fef3314))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/1/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/model/person/Name.java
```diff
@@ -8,8 +8,17 @@
  */
 public class Name {
 
-    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names should be spaces or alphanumeric characters";
-    public static final String NAME_VALIDATION_REGEX = "[\\p{Alnum} ]+";
+    public static final String MESSAGE_NAME_CONSTRAINTS =
+            "Person names should only contain spaces, unicode characters, numbers "
```
kychua: "unicode characters" seems too technical for a user but I'm not sure what to replace it with

src/main/java/seedu/address/model/person/Name.java
```diff
@@ -8,8 +8,17 @@
...
+            + "or the following punctuation marks: . , ' -";
+    public static final String NAME_VALIDATION_REGEX = "^["
```
MightyCupcakes: Should use Java's Pattern to ensure consistency with the rest of the code base (like Parser)

MightyCupcakes: Should use Java's Pattern to ensure consistency with the rest of the code base (like Parser)

src/main/java/seedu/address/model/person/Name.java
chao1995: @kychua The issue only requests for allowing hyphen and comma in the names, I will we only add support for these two in this PR. We can support unicode in a separate PR.

src/main/java/seedu/address/model/person/Name.java
chao1995: I agree, but we can do this in another PR.

chao1995: I agree, but we can do this in another PR.

kychua: Yup, since this affects other classes as well

CanIHasReview:
# v2

@kychua submitted v2 for review.

* [v2 1/2] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/c72a077758b9e4f89da6a4db990e952bb3d825b1)
* [v2 2/2] [Update test person's name to test new name regex](https://github.com/se-edu/addressbook-level4/pull/265/commits/973bb8b5f14d76cd0404f1ad9c5d4695e688946d)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/da4dc8e78218b35ccf991108acf1c56ca8492f57...973bb8b5f14d76cd0404f1ad9c5d4695e688946d)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/commit/0c857fd680e94e31714b90d9e98e2fc0fde764f9))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/2/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/model/person/Name.java
```diff
@@ -8,9 +8,19 @@
  */
```
damithc: Commit message:
```
The original name validation only accepts alphanumeric and whitespace
characters. This is too restrictive and rejects valid names as well,
such as non-english names, hyphenated names and names containing dots,
commas or apostrophes.
```
No need for `The original`.

Add a paragraph to explain your solution. e.g. `Update regular expression to include ...`

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -397,7 +397,8 @@ public void execute_find_matchesIfAnyKeywordPresent() throws Exception {
     class TestDataHelper {
 
         Person adam() throws Exception {
-            Name name = new Name("Adam Brown");
+            // unicode format based on http://stackoverflow.com/questions/4237581/comparing-unicode-characters-in-junit
+            Name name = new Name("Adam,  o'\u00E9-\u00F6 Jr. 2");
```
damithc: Can be squashed with other commit. No need to mention updating tests in the commit message as it is expected anyway.

CanIHasReview:
# v3

@kychua submitted v3 for review.

* [v3 1/1] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/cec8d8e53d555ab74ed768dcc5c2a3b56da1a61a)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/941d0afebe52dc6bd86b60bd4a11861f48273900...cec8d8e53d555ab74ed768dcc5c2a3b56da1a61a)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/commit/c34622a0f28bdc85c212588b12ad3274f75ca59e))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/3/head:BRANCHNAME
```

</details>

kychua:
To fix: commit message should specify that the unicode issues (case insensitive comparison, precomposed/decomposed characters) are referring to Java's `equalsIgnoreCase` method

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -399,7 +399,8 @@ public void execute_find_matchesIfAnyKeywordPresent() throws Exception {
     class TestDataHelper {
 
         Person adam() throws Exception {
-            Name name = new Name("Adam Brown");
+            // unicode format based on http://stackoverflow.com/questions/4237581/comparing-unicode-characters-in-junit
+            Name name = new Name("Adam,  o'\u00E9-\u00F6 \u0041 \u030A Jr. 2");
```
damithc: While you are following the selected answer from the S/O, may be using UTF-8 encoding (the other answer) is makes the test code more reader friendly? Can be done as a separate issue.

CanIHasReview:
# v4

@kychua submitted v4 for review.

* [v4 1/1] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/48ad9dc502937b86bdc430706cb4d1cfc1950246)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/2221d24f2e49a9ec8a9c0d45b9dcb03694d25b0f...48ad9dc502937b86bdc430706cb4d1cfc1950246)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/35bd07ae4247e2ac7b01ff459cd6fbc5eda2d0af~1...35bd07ae4247e2ac7b01ff459cd6fbc5eda2d0af))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/4/head:BRANCHNAME
```

</details>

kychua:
Updated commit message to specify that the unicode issues are referring to Java's `String.equalsIgnoreCase()`

damithc:
@kychua can resolve conflicts and submit a new iteration?

kychua:
@yamgent

Didn't the original regex already reject strings of whitespace characters?
 The name string is trimmed before validating and empty strings are rejected by the regex.

> The current regex allows space characters. However, this means that a
string of whitespace characters (e.g. "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;") are permitted.
>
> Modify the regex such that blank strings (such as " ") are no longer
allowed.
>
> An alternative would be to edit the validation function to check for
the length of the trimmed string seperately (i.e. test.trim().length()
== 0). However, that would not be elegant since we will now have two
different things doing our validation (regex and the code checking
length), which feels unnecessary compared to just using regex alone.

yamgent:
@kychua The change in the regex was made because the unit test is testing `Name#isValidName()`, not `Name#Name()`.

The callers of the methods may be trimming the string, but `Name#isValidName()` itself does not trim the string, hence lies the problem.

kychua:
@yamgent I see. Then why do we not trim in the validation method instead? I don't think checking length is necessary, i.e. this seems to work:
```java
System.out.println(" ".trim().matches("[\\p{Alnum} ]+"));  // prints false
```
Or is it to avoid trimming twice?

yamgent:
@kychua That would cause another issue, which is that you can now start a name with a whitespace character. This one was not explained properly, and the regex wasn't clear enough for this requirement, so I apologies for the confusion.

```java
System.out.println(" John".matches("[\\p{Alnum} ]+"));  // prints true
```

kychua:
@yamgent I see, thanks!

CanIHasReview:
# v5

@kychua submitted v5 for review.

* [v5 1/1] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/316d358c2da931ee20ae34324519a8a462cc6203)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/f40ee42cb2e29a99c9c6b8757b64b0b0f4284d67...316d358c2da931ee20ae34324519a8a462cc6203)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/d8fabb7acdef212403c0d0e19f8195275964c562~1...d8fabb7acdef212403c0d0e19f8195275964c562))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/5/head:BRANCHNAME
```

</details>

pyokagan:
I assume @ndt93 won't be reviewing this :-P

pyokagan:
Stepping back a bit, why do we need to validate names? Why do we see a need to control what users can or cannot put inside the "name" field? @damithc

damithc:
> Stepping back a bit, why do we need to validate names? Why do we see a need to control what users can or cannot put inside the "name" field of their own personal address book? @damithc

Yes, there is no real need to block certain chars from the name. A more important need is to ensure a name _can_ include any valid char (people gets upset when a software says their name is 'invalid'. :-p)  For example, the name `Lee-Roberts` should be accepted even if `-R` is a valid command switch.  So, I'm OK to remove name validation altogether and let users put anything in that field. 

On a related note, name validation is not entirely evil as it can help detect user errors  (e.g user swapping values of two fields by mistake) as long as we handle possibly invalid values in a user-friendly way (e.g. warn instead of reject)

kychua:
>Hmm, what is the point of bringing up String#equalsIgnoreCase() here? We use
regexes for name validation, right?

Find command uses String#equalsIgnoreCase().

> On another note, this probably means that two names which look the same may not
be detected as the same by our code. (via ReadOnlyPerson#isSameStateAs()).
This means that we won't properly detect duplicate names in other languages,
although I wonder if it's worth the effort to fix that.

Just realised ReadOnlyPerson#isSameStateAs() doesn't even do caseless comparisons... but yes, same issue as FindCommand.

The precomposed/decomposed marks issue can probably be fixed by using Java's Normalizer#normalize to convert the input to standard form.

The case-insensitive comparison seems harder. The ICU library [seems to be able to handle this](http://stackoverflow.com/a/34966206) but that adds a dependency. Also, I think Pierce pointed out that in Japanese, the same word can be written in different ways so we aren't really going to be able to support other languages even if we manage to fix the issues. Is it enough if we guarantee users can find their names if they use an identical search term?

damithc:
@kychua 
> Is it enough if we guarantee users can find their names if they use an identical search term?

What's the current situation for English names and non-English names?

kychua:
@damithc 
> What's the current situation (as per current state of this PR) for English names and non-English names?

English names work for case insensitive comparisons. Non-English names might or might not work for case-insensitive searches. Case sensitive searches should work fine if the exact same input method is used, but not if one uses precomposed characters and the other uses decomposed characters.

damithc:
> English names work for case insensitive comparisons. Non-English names might or might not work for case-insensitive searches. Case sensitive searches should work fine if the exact same input method is used, but not if one uses precomposed characters and the other uses decomposed characters.

OK. Probably good enough for this stage.

CanIHasReview:
# v6

@kychua submitted v6 for review.

* [v6 1/1] [Make name validation regex less restrictive](https://github.com/se-edu/addressbook-level4/pull/265/commits/db116b1784c614dafd114c5e95da84bd5802b1c6)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...db116b1784c614dafd114c5e95da84bd5802b1c6)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/2ffa5e19fbfa97074dc6bd7a835c46de05b7e87a~1...2ffa5e19fbfa97074dc6bd7a835c46de05b7e87a))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/6/head:BRANCHNAME
```

</details>

damithc:
@kychua is the long commit message still relevant?

kychua:
@damithc 

> is the long commit message still relevant?

I can make it more concise but the problems with unicode still exist because we are still allowing unicode characters...

src/main/java/seedu/address/model/person/Name.java
```diff
@@ -8,15 +8,11 @@
...
-    /*
-     * The first character of the address must not be a whitespace,
-     * otherwise " " (a blank string) becomes a valid input.
+    public static final String MESSAGE_NAME_CONSTRAINTS = "Person names can take any value and should not be blank";
+    /**
+     * Names must start with a non-whitespace character. This also implies that names must not be blank e.g. "  ".
      */
-    public static final String NAME_VALIDATION_REGEX = "[\\p{Alnum}][\\p{Alnum} ]*";
-
+    public static final String NAME_VALIDATION_REGEX = "[^\\s].*";
```
pyokagan: At this point, we could just use simple string operations instead of regex?

pyokagan:
> @damithc How do you want to proceed with this?

Forgot to put my personal opinion, I think `TODO` comments are fine, given that we aren't particularly mission critical :-P

damithc:
> I think TODO comments are fine, given that we aren't particularly mission critical :-P

Agreed, a `TODO` is enough for now. Also agree that it's good to document 'known issues' too. May be as an appendix to the User Guide?

pyokagan:
@kychua I just realized: while it is a good idea to be as permissive as possible wrt to characters in names, there are still some characters that we should not allow. For example, newlines, since names with newlines probably won't render correctly in our UI. So maybe keeping a list of illegal characters (maybe via a regex) would be a good idea?

pyokagan:
@kychua Oh, to add on, I think something better than just writing `TODO` is to have failing test cases that prove that the bugs hypothesized in the commit message do actually exist.

CanIHasReview:
# v7

@kychua submitted v7 for review.

* [v7 1/3] [Name: change validation to reject only blank names](https://github.com/se-edu/addressbook-level4/pull/265/commits/2af13c05f485c3801c54f118608d527d6aca5b3a)
* [v7 2/3] [Name: remove line terminators in name before validating](https://github.com/se-edu/addressbook-level4/pull/265/commits/32f962372633f9005b3df06cfa0014b3ca9f913b)
* [v7 3/3] [Address: remove unnecessary regex usage](https://github.com/se-edu/addressbook-level4/pull/265/commits/3b9aaac273f0b3c3ac9f005c0e13dddf5febb595)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...3b9aaac273f0b3c3ac9f005c0e13dddf5febb595)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/compare/d16bfe4d89f0c84d5d18ecbf571d5131d712bdd2~1...d16bfe4d89f0c84d5d18ecbf571d5131d712bdd2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/7/head:BRANCHNAME
```

</details>

CanIHasReview:
# v8

@kychua submitted v8 for review.

* [v8 1/4] [testutil: Add ExpectedFailure annotation for tests](https://github.com/se-edu/addressbook-level4/pull/265/commits/5a4f4d396c909c94465aa0a9edb2db23bea24895)
* [v8 2/4] [Name: change validation to reject only blank names](https://github.com/se-edu/addressbook-level4/pull/265/commits/9af2a1637874f785b2ce27b749f5da8f30e0c418)
* [v8 3/4] [Name: remove line terminators in name before validating](https://github.com/se-edu/addressbook-level4/pull/265/commits/5201cf3714a61d9e42a5f426dcd2a65d93d15fe7)
* [v8 4/4] [Address: remove unnecessary regex usage](https://github.com/se-edu/addressbook-level4/pull/265/commits/9bcd3869de1af97cf67c7c58fd8199365421ff54)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/e1eafd1145cb5fdcc81bbacecf9e5145903b3fe2...9bcd3869de1af97cf67c7c58fd8199365421ff54)) ([:chart_with_upwards_trend: Interdiff between v7 and v8](https://github.com/se-edu/addressbook-level4/compare/e8918b5af79b1a5030720bed67bdbca1485f6b2b~1...e8918b5af79b1a5030720bed67bdbca1485f6b2b))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/265/8/head:BRANCHNAME
```

</details>

kychua:
@pyokagan 
> while it is a good idea to be as permissive as possible wrt to characters in names, there are still some characters that we should not allow. For example, newlines, since names with newlines probably won't render correctly in our UI. So maybe keeping a list of illegal characters (maybe via a regex) would be a good idea?

I stripped removed line breaks instead of marking multi-line names invalid since JavaFx strips the newline character. Can't think of what other characters to ban...

kychua:
# v8

* Added ExpectedFailure annotation to mark tests -> by right it should be shown somewhere that these tests failed but that it's expected, but I don't know how/where to show it
* Added tests to log bugs
* Added `StringUtil#isBlank(String)` to support unicode whitespace instead of using `str.trim().isEmpty()`
* Handle multi-line names by removing line breaks
* Update `Address` validation to use `StringUtil#isBlank(String)` instead of regex

Zhiyuan-Amos:
Should we close this?

eugenepeh:
Think we can put it on hold till @kychua decides to resume it

Zhiyuan-Amos:
@kychua do you plan to resume on it? :P

kychua:
@Zhiyuan-Amos 
> do you plan to resume on it?

nope, I won't be working on this anymore.

Issue #246, opened 2017-01-13
User/Dev Guide: ToC links broken


damithc:
Links from Table of Contents are broken

damithc:
Apparently, fixing the ToC links on the .md version breaks them on the GitHub Pages version :-p

damithc:
Reverted the fix because it is more important for the links to work in the GitHub Pages version. This issue should become irrelevant after we migrate to AsciiDoc

damithc:
Lowering priority because this will not be a problem after we move to asciidoctor.

PR #268, opened 2017-01-24
Find a way to convert markdown documents to html and pdf #175

### LINKED ISSUE #175: Find a way to convert markdown documents to html and pdf
Students might have to use more styling (colors etc.) in the docs and submit documents in pdf format. It would be good if we can find a way to do those two things with minimum effort.

### Option 1: Github Flavored Markdown
* Use default themes and GH pages to render to html
* Print to a PDF writer to convert to PDF

Pros:
* Easy to edit

Cons: 
* No extra stuff like TOC, admonition blocks etc.
* Default themes don't fit very well

### Option 2: AsciiDocs
* Use Github's support to render to basic
* Optional: Run script locally to generate fancy html
* Run script locally to generate PDF

Pros:
* Lots of bells and whistles (TOC, admonition, etc.)
* Proper pdf generation

Cons:
* Need to install python etc.
* Syntax highlighting is not as good as GFMD
* More complex syntax, less editor support

### Option 3: Strapdown
* Html files, but the body can contain markdown
* Render as html and print to a pdf writer

Pros:
* Can be combined with bootstrap
* Can preview locally without running a script

Cons:
* Slow to render
* **Can't seem to get it to work when served from GitHub :-p**

kychua:
Fixes #175 #246 

Requires:
- creating gh-pages branch
- storing personal access token in travis

Ongoing:
- clean up new asciidoc files
- add documentation for how to use asciidocs
- clean up commits

(Currently, deploying to github pages is under `after_success` in travis, which means build passes even if deploying fails. I tried moving it up to `scripts` but it keeps resulting in an error:
travis_error)


damithc:
@kychua no need to convert all current docs right away until we are dead sure this is the approach we want to take. Just a sampling would do.

> storing personal access token in travis

Is this a security problem?

kychua:
@damithc 
> no need to convert all current docs right away until we are dead sure this is the approach we want to take. Just a sampling would do.

Ok, I'll just keep both formats?

> Is this a security problem?

Travis supports encrypted environment variables. According to Travis,
> By default, the value of these new environment variables is hidden from the export line in the logs... 
> The variables are stored  encrypted in our systems, and get decrypted when the build script is generated.
> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

.travis.yml
```diff
@@ -5,12 +5,15 @@ matrix:
...
     ./config/travis/run-checks.sh &&
-    travis_retry ./gradlew clean checkstyleMain checkstyleTest headless allTests coverage coveralls
+    travis_retry ./gradlew clean checkstyleMain checkstyleTest headless allTests coverage coveralls asciidoctor -i
 
 before_install:
           - "export DISPLAY=:99.0"
           - "sh -e /etc/init.d/xvfb start"
 
+after_success:
+  - test $TRAVIS_PULL_REQUEST == "false" && test $TRAVIS_BRANCH == "master" && bash deploy_github_pages.sh
```
pyokagan: I think you can combine the `test`s into a single `test $TRAVIS_PULL_REQUEST == "false" -a $TRAVIS_BRANCH == "master"`

Also, make sure to quote any variable expansions, otherwise word splitting would be performed which is not desirable in conditionals:
```
test "$TRAVIS_PULL_REQUEST" == "false" -a "$TRAVIS_BRANCH" == "master"
```

Finally, yoda conditionals are encouraged in shell scripting -- it avoids problems when the value of variables contain certain special values which might change the conditional:
```
test "false" == "$TRAVIS_PULL_REQUEST" -a "master" == "$TRAVIS_BRANCH"
```

damithc:
> Ok, I'll just keep both formats?

@kychua Both UG and DG need to be migrated before the PR can be merged. What I meant was, don't let the migration of the docs (if it is taking lot of time) delay you getting feedback for the rest of the stuff in the PR.

yamgent:
> Is this a security problem?

I took a look at the process of setting up personal access token in GitHub.

As long as students safeguard their personal access token (by not leaking them out in places such as printing it in the Travis build log, or putting their access token in a file and push it to the repo), and understand what they are doing in GitHub and Travis, it will be reasonably safe. If done properly, no one (including themselves) can see their personal access token after the initial setup in GitHub and Travis.

This means that students should carefully follow our instructions in setting this up, especially if they do not have a good understanding of personal access token.

damithc:
> This means that students should carefully follow our instructions in setting this up, especially if they do not have a good understanding of personal access token.

Thanks for the investigation @yamgent and @kychua

damithc:
I've set codacy to ignore asciidoctor.css

damithc:
@kychua can you give a preview of the results? May be stage it in your fork's gh_pages?

kychua:
@damithc The preview is [here](http://kychua.github.io/addressbook-level4)

damithc:
> The preview is here

Preview looks good @kychua 
As discussed, let's break this up to small PRs.

damithc:
@kychua time to move to the next step?

yamgent:
Superseded by #271.

Issue #172, opened 2016-11-27
Add missing unit tests


damithc:
Create one PR for each Test class.
* PR name can be `Add unit tests for {ClassName} class`
* Omit the `Fixes #....` phrase from the PR.
* Test class name should match class being unit tested.  
* Where dependency injection is too hard to use, the SUT need not be isolated from dependencies. i.e. Unit + integration tests

PR #261, opened 2017-01-20
Parser and ArgumentTokenizer cleanups #235

PierceAndy:
Fixes #235 
Collecting a list of small cleanups that can be made.
Cleanups have also been applied to 97c34f2b96ec34b3afd93876f3eda12455759c91 that addresses #141.

## `ArgumentTokenizer`

- `getPreamble()`: Rather than returning `Optional.empty()`, an empty string can be returned instead. (With defensive programming, callers will need to check for empty string anyway).

- `getAllValues()`: Rather than returning `Optional.empty()`, an empty list can be returned instead. (With defensive programming, callers will need to check for an empty list anyway). This has the side benefit of removing the `toSet()` helper in `Parser`.

## `Parser`

Use of regular expressions that are less readable than their less fancy counterparts:

- `PERSON_INDEX_ARGS_FORMAT` can be replaced with `String.isEmpty()`. Actually, I think `StringUtils.isUnsignedInteger()` already checks for empty string soo...

- `KEYWORDS_ARGS_FORMAT` can be replaced with `keywords.length <= 0`.

PR #253, opened 2017-01-16
Add unit tests for person's fields classes #172

yamgent:
Part of #172.

PR #273, opened 2017-01-26
Add unit tests for AddCommand

yamgent:
This is part of #172.

```
AddCommand does not have its own unit test.

Let's add a unit test for AddCommand.
```

### DIFF
--- a/src/test/java/seedu/address/logic/commands/AddCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/AddCommandTest.java
@@ -1,0 +1,137 @@
+package seedu.address.logic.commands;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Set;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+import seedu.address.commons.core.UnmodifiableObservableList;
+import seedu.address.commons.exceptions.IllegalValueException;
+import seedu.address.logic.commands.exceptions.CommandException;
+import seedu.address.model.Model;
+import seedu.address.model.ReadOnlyAddressBook;
+import seedu.address.model.person.Person;
+import seedu.address.model.person.ReadOnlyPerson;
+import seedu.address.model.person.UniquePersonList.DuplicatePersonException;
+import seedu.address.model.person.UniquePersonList.PersonNotFoundException;
+import seedu.address.testutil.PersonBuilder;
+
+public class AddCommandTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void constructor_nullPerson_throwsNullPointerException() throws Exception {
+        thrown.expect(NullPointerException.class);
+        new AddCommand(null);
+    }
+
+    @Test
+    public void execute_personAcceptedByModel_addSuccessful() throws Exception {
+        ModelStubAcceptingPersonAdded modelStub = new ModelStubAcceptingPersonAdded();
+        Person validPerson = new PersonBuilder().build();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
+
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, validPerson), commandResult.feedbackToUser);
+        assertEquals(Arrays.asList(validPerson), modelStub.personsAdded);
+    }
+
+    @Test
+    public void execute_duplicatePerson_throwsCommandException() throws Exception {
+        ModelStub modelStub = new ModelStubThrowingDuplicatePersonException();
+        Person validPerson = new PersonBuilder().build();
+
+        thrown.expect(CommandException.class);
+        thrown.expectMessage(AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        getAddCommandForPerson(validPerson, modelStub).execute();
+    }
+
+    /**
+     * Generates a new AddCommand with the details of the given person.
+     */
+    private AddCommand getAddCommandForPerson(Person person, Model model) throws IllegalValueException {
+        AddCommand command = new AddCommand(person);
+        command.setData(model);
+        return command;
+    }
+
+    /**
+     * A default model stub that have all of the methods failing.
+     */
+    private class ModelStub implements Model {
+        @Override
+        public void addPerson(Person person) throws DuplicatePersonException {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public void resetData(ReadOnlyAddressBook newData) {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public ReadOnlyAddressBook getAddressBook() {
+            fail("This method should not be called.");
+            return null;
+        }
+
+        @Override
+        public void deletePerson(ReadOnlyPerson target) throws PersonNotFoundException {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public void updatePerson(int filteredPersonListIndex, ReadOnlyPerson editedPerson)
+                throws DuplicatePersonException {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public UnmodifiableObservableList<ReadOnlyPerson> getFilteredPersonList() {
+            fail("This method should not be called.");
+            return null;
+        }
+
+        @Override
+        public void updateFilteredListToShowAll() {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public void updateFilteredPersonList(Set<String> keywords) {
+            fail("This method should not be called.");
+        }
+    }
+
+    /**
+     * A Model stub that always throw a DuplicatePersonException when trying to add a person.
+     */
+    private class ModelStubThrowingDuplicatePersonException extends ModelStub {
+        @Override
+        public void addPerson(Person person) throws DuplicatePersonException {
+            throw new DuplicatePersonException();
+        }
+    }
+
+    /**
+     * A Model stub that always accept the person being added.
+     */
+    private class ModelStubAcceptingPersonAdded extends ModelStub {
+        final ArrayList<Person> personsAdded = new ArrayList<>();
+
+        @Override
+        public void addPerson(Person person) throws DuplicatePersonException {
+            personsAdded.add(person);
+        }
+    }
+
+}
--- a/src/main/java/seedu/address/model/person/UniquePersonList.java
+++ b/src/main/java/seedu/address/model/person/UniquePersonList.java
@@ -115,7 +115,7 @@ public class UniquePersonList implements Iterable<Person> {
      * Signals that an operation would have violated the 'no duplicates' property of the list.
      */
     public static class DuplicatePersonException extends DuplicateDataException {
-        protected DuplicatePersonException() {
+        public DuplicatePersonException() {
             super("Operation would result in duplicate persons");
         }
     }


CanIHasReview:
# v1

* [v1 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/4d777377189fbd4e6922d1f13ef24bc0d2e047e8)
* [v1 2/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/a1508209ba5b99d389508a77f53ad2fb2edaa827)
* [v1 3/3] [Remove integrated test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/96f13dc111f9e8bfb953db0a7cf71a0f039704a2)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/da4dc8e78218b35ccf991108acf1c56ca8492f57...96f13dc111f9e8bfb953db0a7cf71a0f039704a2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/1/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,141 @@
...
+     * Converts the list of tags into a set of strings.
+     */
+    private Set<String> convertTagsToStringSet(UniqueTagList uniqueTagList) {
+        Set<String> stringSet = new HashSet<String>();
+        uniqueTagList.forEach((t) -> stringSet.add(t.tagName));
+        return stringSet;
+    }
+
+    /**
+     * A specially crafted model injected into add command
```
pyokagan: "Specially crafted" is kind of redundant. I'd expect that every single line of code you write is already "specially crafted" :-P

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,141 @@
...
+     * for testing add's functionality.
+     */
+    private class TestAddModel implements Model {
```
MightyCupcakes: How does this differ from the usual Model? I know this is for testing - add something like what does this model do or do not do in the header comment.

damithc: This is actually a Stub. Naming it such will help remind readers of the `Stub` concept.

damithc:
> Shouldn't you leave at least one GUI test?

Yes, there should be GUI tests to ensure that adding can be done via the GUI and the GUI itself behaves correctly for adding operations e.g. the added person appears in the list. However, GUI tests should not perform different variations of add operations if the GUI behavior is same across all those variations. Those variations should be done at lower levels. That means at least one successful add operation and at least one unsuccessful add operation should be done via GUI (because GUI behavior is different for success and failure).

CanIHasReview:
# v2

* [v2 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/b4a7d044b68c19126065533bb6a3f33aeaf46a7f)
* [v2 2/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/468fb759063617ef7f06709e4afdd8fbbf2475d6)
* [v2 3/3] [Remove integrated test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/81b357c9e6b04949f9f395c8529982e4905f7189)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/9321a79e2fc882fbfce56d31d2f8df32f11bd16c...81b357c9e6b04949f9f395c8529982e4905f7189)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/commit/36b5d9c6bd1c30ef1d6bed07852ad1a1ec0dea3d))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/2/head:BRANCHNAME
```

</details>

src/test/java/guitests/AddCommandTest.java
```diff
@@ -1,53 +0,0 @@
...
-        assertTrue(personListPanel.isListMatching(currentList));
-
-        //add to empty list
-        commandBox.runCommand("clear");
-        assertAddSuccess(td.alice);
-
-        //invalid command
-        commandBox.runCommand("adds Johnny");
-        assertResultMessage(Messages.MESSAGE_UNKNOWN_COMMAND);
-    }
```
damithc: As all these test cases have GUI implications, I'm inclined to keep all of them. i.e. AddCommandTest doesn't have excessive tests. An example of a GUI test with excessive testings is EditCommandTest

yamgent: Ok, I will keep them in the next iteration.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,170 @@
...
+    @Test
+    public void constructor_invalidEmail_throwsIllegalValueException() throws Exception {
+        thrown.expect(IllegalValueException.class);
+        new AddCommand("valid name", "61239876", "", "valid address", new HashSet<>());
+    }
+
+    @Test
+    public void constructor_invalidAddress_throwsIllegalValueException() throws Exception {
+        thrown.expect(IllegalValueException.class);
+        new AddCommand("valid name", "61239876", "valid.email@mail.com", "", new HashSet<>());
```
kychua: Seems a little repetitive... would have to change quite a few lines if say, we changed the parameters for AddCommand. Actually, would it be better if there's just one or two to test the integration, then have tests for the invalid values in the tests for `Person` and `Address`, `Email` etc. classes instead?

src/test/java/guitests/AddCommandTest.java
chao1995: Should we also keep invalid command test?

I think we should remove the invalid command test here and create a dedicated InvalidCommandTest to test all the invalid commands (in another PR)

chao1995: I think the invalid command test can be removed since it is a test against "unknown command", we should probably create a IncorrectCommandTest to test this.

But this better be done in another PR (because many other, if not all, *CommandTests have such **invalid** command tests, which is basically an **incorrect** command test and should be removed).

Also, I found this piece of code in EditCommandTest.

```java
@Test
public void edit_invalidValues_failure() {
    commandBox.runCommand("edit 1 *&");
    assertResultMessage(Name.MESSAGE_NAME_CONSTRAINTS);

    commandBox.runCommand("edit 1 p/abcd");
    assertResultMessage(Phone.MESSAGE_PHONE_CONSTRAINTS);

    commandBox.runCommand("edit 1 e/yahoo!!!");
    assertResultMessage(Email.MESSAGE_EMAIL_CONSTRAINTS);

    commandBox.runCommand("edit 1 a/");
    assertResultMessage(Address.MESSAGE_ADDRESS_CONSTRAINTS);

    commandBox.runCommand("edit 1 t/*&");
    assertResultMessage(Tag.MESSAGE_TAG_CONSTRAINTS);
}
```

Should these be considered as variations of invalid command test and thus be pushed down to lower level?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,170 @@
...
+
+        private TestPerson personInList;
+
+        protected AddModelStub(TestPerson personInList) {
+            this.personInList = personInList;
+        }
+
+        @Override
+        public void resetData(ReadOnlyAddressBook newData) {
+            // do nothing
```
chao1995: There are a lot of unused methods because AddModelStub implements Model. Can we instead make AddModelStub extend from ModelManager and then override just the `addPerson` method?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,170 @@
...
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Before
+    public void setUp() throws Exception {
+        TypicalTestPersons typicalTestPersons = new TypicalTestPersons();
+
+        newPerson = typicalTestPersons.alice;
+        duplicatePerson = typicalTestPersons.daniel;
+        model = new AddModelStub(duplicatePerson);
```
chao1995: This is a bit confusing at the first glance about why daniel is sadly a duplicate person.

I think it is clearer to add a method called `addPerson()` in `AddModelStub` and when testing duplicate person, we add some person (maybe daniel) twice. (Maybe you can use a List to keep track of all persons added in `AddModelStub`)

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,170 @@
...
+        uniqueTagList.forEach((t) -> stringSet.add(t.tagName));
+        return stringSet;
+    }
+
+    /**
+     * A model stub injected into the add command, for testing add's functionality.
+     *
+     * Only {@link #addPerson(Person)} is implemented for this stub.
+     */
+    private class AddModelStub implements Model {
```
damithc: As this is a private class, may be no need for the `Add` prefix to the class name?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
damithc: Coincidentally, I got this blog post in my email today this is probably relevant here https://testing.googleblog.com/2017/01/testing-on-toilet-keep-cause-and-effect.html

src/test/java/guitests/AddCommandTest.java
yamgent: As discussed earlier, we should consider testing these variations in LogicManager's test (lower level). 

The GUI test should only test that when we input an invalid value, an invalid message comes out. Variations are not really needed.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: Hmm... I agree that it can feel quite repetitive. However the purpose of these tests is to check that the correct Exceptions are thrown, rather than actually testing the logic of processing invalid values. (which are covered in tests for `Address`, `Email`, etc in PR #253, and has more variations in invalid values)

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: I agree. Thanks for the blog post, it is quite informative. 

@chao1995 Thanks for your suggestion.

yamgent:
As everyone has reviewed, I will now change it to `s.Ongoing`.

yamgent:
@pyokagan I clicked the 'Submit New Iteration' in @CanIHasReview for this particular PR, but nothing got posted. Works on my other PRs though.

CanIHasReview:
# v3

* [v3 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/393c612501eb01c46120d00aeb2216fe4f1965b5)
* [v3 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/718c2c054aa5fb8c2a97845d1d2a349d2fa0d249)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/941d0afebe52dc6bd86b60bd4a11861f48273900...718c2c054aa5fb8c2a97845d1d2a349d2fa0d249)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/0c9a47d141aaf88c1e8b430fe7c16a5e950ba233~1...0c9a47d141aaf88c1e8b430fe7c16a5e950ba233))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/3/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
eugenepeh: Would it be better to change the values to constants such as VALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS? Since its a bit repetitive and unnecessary to test variants in this case, also will be easier to change in addition.

yamgent: You have a point. I will refactor them out into private static variables for the next iteration. Thanks!

yamgent:
@chao1995 @MightyCupcakes Any comments?

chao1995:
No other comments from me

CanIHasReview:
# v4

* [v4 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/b0012f0b0c8d44ff029ec1177d13e64bd386eff9)
* [v4 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/bbb2ed07dd53b29656633c544b019c846967f543)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/2221d24f2e49a9ec8a9c0d45b9dcb03694d25b0f...bbb2ed07dd53b29656633c544b019c846967f543)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/27e1593d0959689fc56c60e268c116be6290a7d8~1...27e1593d0959689fc56c60e268c116be6290a7d8))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/4/head:BRANCHNAME
```

</details>

damithc:
@chao1995 this PR is pending your review.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,144 @@
...
+        command.execute();
+    }
+
+    /**
+     * Asserts that the command produced the correct message after execution.
+     */
+    public void assertCommandBehaviour(String expectedString, Command command) throws CommandException {
+        CommandResult commandResult = command.execute();
+
+        assertEquals(expectedString, commandResult.feedbackToUser);
```
damithc: We should also check to see if the person was added to the model?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,144 @@
...
+     * Converts the list of tags into a set of strings.
+     */
+    private Set<String> convertTagsToStringSet(UniqueTagList uniqueTagList) {
+        Set<String> stringSet = new HashSet<String>();
+        uniqueTagList.forEach((t) -> stringSet.add(t.tagName));
+        return stringSet;
+    }
+
+    /**
+     * A model stub injected into the add command, for testing add's functionality.
```
damithc: Why exactly are we using a stub? Need to be explained here.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * Only {@link #addPerson(Person)} is implemented for this stub.
```
damithc: This is not exactly correct. All method are implemented, only the mentioned method is overridden.

CanIHasReview:
# v5

* [v5 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/660c20126f7dffa25f65a53d68aa9d2fec8e07be)
* [v5 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/2de3d00d8541548353acd1d03e0d7108f4882a44)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/13e914d0558b8b2e402910002497651ae492b046...2de3d00d8541548353acd1d03e0d7108f4882a44)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/5c18663c0f14355e8624c33f4465ad3b5f88c3db~1...5c18663c0f14355e8624c33f4465ad3b5f88c3db))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/5/head:BRANCHNAME
```

</details>

yamgent:
Apology in advance for the first commit's commit message, will fix it in the next iteration.

yamgent:
@chao1995 Any additional comments for the current iteration?

chao1995:
> AddCommand is currently being tested through integrated testing.

Can omit the word "currently". And actually we can say `we have no unit tests for AddCommand` since AddCommand is being tested in gui testing as well.

> Let's write AddCommandTest (unit test version), which injects a stub
model into the command

injects `model stub`

> \- The ModelStub inherits from ModelManager, rather than implements the
   Model interface, to avoid having too many empty methods, since we
   only need to override the addPerson() method for isolated testing.

I don't think we need to include this because it is reflected in the code. Or if you want to compare it with some other alternatives, i.e. why not implement Model interface. But I don't think it is an alternative if the in-use approach is much better.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,135 @@
...
+        AddCommand command = getAddCommandForPerson(validPerson, modelStub);
+        command.execute();
+    }
+
+    /**
+     * Generates a new AddCommand with the details of the given person.
+     */
+    private AddCommand getAddCommandForPerson(Person person, ModelStub modelStub) throws IllegalValueException {
+        Set<String> tags = new HashSet<>();
+        person.getTags().forEach((t) -> tags.add(t.tagName));
```
yamgent: The `toSet()` method that is in the codebase accepts only lists of `String` as input.

I think I will refactor this portion of the code to a new method `toSet()` and put it in a test util class, that way subsequent command unit test PRs can make use of it.

CanIHasReview:
# v6

@yamgent submitted v6 for review.

* [v6 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/45c518fffa71ac83fe3ea0ccff426f6aae8b9e0d)
* [v6 2/3] [Add TestPerson\#getTagsAsStringsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/81dfd05e4135d5cf1f21fc7d6b7804de33876498)
* [v6 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/d7e6f1f9b48accc764349abc80429c833b548ca7)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/f40ee42cb2e29a99c9c6b8757b64b0b0f4284d67...d7e6f1f9b48accc764349abc80429c833b548ca7)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/a1fa435406953f7f0eb15a8f5420ae9ada30a5ce~1...a1fa435406953f7f0eb15a8f5420ae9ada30a5ce))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/6/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Generates a new AddCommand with the details of the given person.
+     */
+    private AddCommand getAddCommandForPerson(TestPerson person, ModelManager modelStub) throws IllegalValueException {
+        Set<String> tags = person.getTagsAsStringsSet();
+
+        AddCommand command = new AddCommand(person.getName().toString(), person.getPhone().toString(),
+                person.getEmail().toString(), person.getAddress().toString(),
+                tags);
```
damithc: Can do without this line break?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        command.setData(modelStub);
+        return command;
+    }
+
+    /**
+     * A model stub injected into the add command, that always reject person being added
```
damithc: ~~`injected into the add command, `~~

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+     * by treating it as a duplicate.
+     *
+     * The stub inherits from ModelManager instead of implementing the Model interface so that we do not have
+     * too many empty methods.
```
damithc: Probably no need for this sentence.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+                person.getEmail().toString(), person.getAddress().toString(),
+                tags);
+
+        command.setData(modelStub);
+        return command;
+    }
+
+    /**
+     * A model stub injected into the add command, that always reject person being added
+     * by treating it as a duplicate.
```
damithc: `A Model stub that always throw a DuplicatePersonException when trying to add a person`

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * The stub inherits from ModelManager instead of implementing the Model interface so that we do not have
+     * too many empty methods.
+     */
+    private class DuplicatePersonModelStub extends ModelManager {
```
damithc: Not sure if the name is explanatory enough. Can't think of a better name though. `ModelStubThrowingDuplicatePersonException`?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * A model stub injected into the add command, that always accept the person being added
+     * by treating it as a unique person.
+     *
+     * The stub inherits from ModelManager instead of implementing the Model interface so that we do not have
+     * too many empty methods.
+     */
+    private class UniquePersonModelStub extends ModelManager {
+        public void addPerson(Person person) throws DuplicatePersonException {
+            // always accept
```
damithc: To be precise, this stub pretends to accept but actually ignores :-p

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Test
+    public void execute_duplicatePerson_throwsDuplicatePersonException() throws Exception {
+        DuplicatePersonModelStub modelStub = new DuplicatePersonModelStub();
+        TestPerson validPerson = new PersonBuilder()
+                .withName(VALID_NAME)
+                .withPhone(VALID_PHONE)
+                .withEmail(VALID_EMAIL)
+                .withAddress(VALID_ADDRESS)
+                .build();
```
damithc: Probably you can create a valid add command once and use it in both cases?

src/test/java/seedu/address/testutil/TestPerson.java
```diff
@@ -78,6 +81,15 @@ public UniqueTagList getTags() {
         return tags;
     }
 
+    /**
+     * Converts the list of tags to a string set.
+     */
+    public Set<String> getTagsAsStringsSet() {
```
chao1995: I don't think this name is very good as it includes its return type. Maybe `getTagsSet()` would do?

In the JavaDoc

> Converts the list of tags to a string set.

can be `Get the tags' names as a string set`.

CanIHasReview:
# v7

@yamgent submitted v7 for review.

* [v7 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/04486e07b4253b92411f1acc131a94719b34fb48)
* [v7 2/3] [Add TestPerson\#getTagsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/de76cac0b75f7dad8626f1292ea2c50608365e1b)
* [v7 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/4c280c3335edcdaef5a42f887b88ef6bf016d1ec)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/1eeb6fb0a75d11d317390588f6799edb52f33047...4c280c3335edcdaef5a42f887b88ef6bf016d1ec)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/compare/56457aa0f73f5e98eff18a64228b74471cb7afc2~1...56457aa0f73f5e98eff18a64228b74471cb7afc2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/7/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    private AddCommand getAddCommandForPerson(TestPerson person, ModelManager modelStub) throws IllegalValueException {
+        Set<String> tags = person.getTagsSet();
+
+        AddCommand command = new AddCommand(person.getName().toString(), person.getPhone().toString(),
+                person.getEmail().toString(), person.getAddress().toString(), tags);
+
+        command.setData(modelStub);
+        return command;
+    }
```
damithc: I'm curious: does a similar method exist somewhere else in the test code?

src/test/java/seedu/address/testutil/TestPerson.java
```diff
@@ -78,6 +81,15 @@ public UniqueTagList getTags() {
...
     }
 
+    /**
+     * Gets the names of the tags as a string set.
+     */
+    public Set<String> getTagsSet() {
+        Set<String> tagsSet = new HashSet<>();
+        tags.forEach((tag) -> tagsSet.add(tag.tagName));
+        return tagsSet;
+    }
```
damithc: I wonder if there's a shorter way to do this conversion using Java 8 `stream()`, `map()` etc.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: Nope. There is a `TestPerson#getAddCommand()`, but that generates a `String` and is mainly for testing the `Parser` (it generates what the user would normally type on the command box).

src/test/java/seedu/address/testutil/TestPerson.java
yamgent: Yes, but would require UniqueTagList to have a `stream()` method though. Should we consider implementing that?

damithc: I guess no need.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
damithc: Got it.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,119 @@
...
+import seedu.address.testutil.TestPerson;
+
+public class AddCommandTest {
+
+    private static final String VALID_NAME = "valid name";
+    private static final String VALID_PHONE = "61239876";
+    private static final String VALID_EMAIL = "valid.email@mail.com";
+    private static final String VALID_ADDRESS = "valid address";
+
+    private static TestPerson validPerson;
```
chao1995: why static?

CanIHasReview:
# v8

@yamgent submitted v8 for review.

* [v8 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/06882261902eaa5fadd06fbaef93ef33e5b45ed1)
* [v8 2/3] [Add TestPerson\#getTagsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/ef5af055adc581b08c497d4c424eb15b127af1c3)
* [v8 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/793dc98fe9e7a67de737103cea33269709573340)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...793dc98fe9e7a67de737103cea33269709573340)) ([:chart_with_upwards_trend: Interdiff between v7 and v8](https://github.com/se-edu/addressbook-level4/compare/e0265d6d2e46e6f31464c581ae16d48c081dab96~1...e0265d6d2e46e6f31464c581ae16d48c081dab96))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/8/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,119 @@
...
+        thrown.expect(CommandException.class);
+        thrown.expectMessage(AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        getAddCommandForPerson(validPerson, modelStub).execute();
+    }
+
+    /**
+     * Generates a new AddCommand with the details of the given person.
+     */
+    private AddCommand getAddCommandForPerson(TestPerson person, ModelManager modelStub) throws IllegalValueException {
```
damithc: This method should not care if it is a stub. `modelStub` -> `model` ?

MightyCupcakes: Then am I right to say that the method does not need to care about `ModelManager` too? All it require is a `Model` to pass it to `Command#setData()`

Therefore:
`ModelManager modelStub` -> `Model model` ?

damithc: Good point @MightyCupcakes 
How about `TestPerson` --> `ReadOnlyPerson`?
In general it's better for a method signature to use the most general types. However, I'm not sure if its worth additional `import` statements, given this is a private method in a test class.

yamgent: The additional `import` statements might be worth the price if it aids the developer's understanding of the code. Since we do not need more than what `Model` and `ReadOnlyPerson` can provide, I think the change will help in that aspect. 

Will do the changes for the next iteration.

CanIHasReview:
# v9

@yamgent submitted v9 for review.

* [v9 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/06882261902eaa5fadd06fbaef93ef33e5b45ed1)
* [v9 2/3] [Add TestPerson\#getTagsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/ef5af055adc581b08c497d4c424eb15b127af1c3)
* [v9 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/a137970a07a40978ccdefb13246e3f41c6a79917)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...a137970a07a40978ccdefb13246e3f41c6a79917)) ([:chart_with_upwards_trend: Interdiff between v8 and v9](https://github.com/se-edu/addressbook-level4/compare/2885e75617abebbdaac48255a8ec598f29e0f895~1...2885e75617abebbdaac48255a8ec598f29e0f895))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/9/head:BRANCHNAME
```

</details>

yamgent:
# v9
Changed `ModelManager modelStub` to `Model model`.
`TestPerson` wasn't changed because the test relied on `TestPerson#getTagsSet()`.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,120 @@
...
+    private TestPerson validPerson;
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Before
+    public void setUp() throws Exception {
+        validPerson = new PersonBuilder().withName(VALID_NAME).withPhone(VALID_PHONE).withEmail(VALID_EMAIL)
+                .withAddress(VALID_ADDRESS).build();
+    }
```
damithc: Didn't notice this before. Probably can initialize the variable in line 29 and get rid of this method? JUnit instantiate the class for each test method so the variable is re-initialized for each test method anyway.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,120 @@
...
+        thrown.expectMessage(AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        getAddCommandForPerson(validPerson, modelStub).execute();
+    }
+
+    /**
+     * Generates a new AddCommand with the details of the given person.
+     */
+    private AddCommand getAddCommandForPerson(TestPerson person, Model model) throws IllegalValueException {
+        Set<String> tags = person.getTagsSet();
```
damithc: Inline this in line 97?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: I would have to create an empty `AddCommandTest()` constructor method though, because `with*()` throws exceptions, and therefore the constructor must declare that.

CanIHasReview:
# v10

@yamgent submitted v10 for review.

* [v10 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/06882261902eaa5fadd06fbaef93ef33e5b45ed1)
* [v10 2/3] [Add TestPerson\#getTagsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/ef5af055adc581b08c497d4c424eb15b127af1c3)
* [v10 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/82729ca8fa882d92cad5790753421e76e1106aab)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...82729ca8fa882d92cad5790753421e76e1106aab)) ([:chart_with_upwards_trend: Interdiff between v9 and v10](https://github.com/se-edu/addressbook-level4/compare/b1afb56f550d2f502639b2d5e2cecce7092bd0f9~1...b1afb56f550d2f502639b2d5e2cecce7092bd0f9))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/10/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
damithc: Attacking this from another angle:
A valid person is not required for many of the tests. So it's a bit misleading to have this in the `@Before` method.

An alternative is to have a `createPerson()` method that is called only when needed. Would that be better?

src/test/java/seedu/address/testutil/TestPerson.java
```diff
@@ -78,6 +81,15 @@ public UniqueTagList getTags() {
         return tags;
     }
 
+    /**
+     * Gets the names of the tags as a string set.
```
damithc: `a set of strings`?

CanIHasReview:
# v11

@yamgent submitted v11 for review.

* [v11 1/3] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/06882261902eaa5fadd06fbaef93ef33e5b45ed1)
* [v11 2/3] [Add TestPerson\#getTagsSet\(\)](https://github.com/se-edu/addressbook-level4/pull/273/commits/6bd745e783088e35c33ae7bcf6d749b89ad28d64)
* [v11 3/3] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/1020ee75f19c615387e51617475c990033986b9a)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/d741336a16662b675c117ed2aa3ade11a934f24c...1020ee75f19c615387e51617475c990033986b9a)) ([:chart_with_upwards_trend: Interdiff between v10 and v11](https://github.com/se-edu/addressbook-level4/compare/ce2b7fbf39096b5954e86d2fb3bc5988563acab7~1...ce2b7fbf39096b5954e86d2fb3bc5988563acab7))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/11/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * A Model stub that always accept the person being added.
+     */
+    private class ModelStubAcceptingPersonAdded extends ModelManager {
+        public void addPerson(Person person) throws DuplicatePersonException {
+            // pretend to accept (but actually does nothing)
+        }
+    }
+}
```
pyokagan: Need a newline here to be symmetric.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Test
+    public void execute_newPersonNotInList_addSuccessful() throws Exception {
+        ModelManager modelStub = new ModelStubAcceptingPersonAdded();
+        TestPerson validPerson = createValidPerson();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
+
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, validPerson), commandResult.feedbackToUser);
```
pyokagan: Do we need to check if the model's `addPerson()` method was called? Probably yes -- it is part of the contract of AddCommand in order to ensure the correct operation of the program.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: Er, what do you mean by that? Do you mean leave a line after `ModelStubAcceptingPersonAdded` class?

CanIHasReview:
# v12

@yamgent submitted v12 for review.

* [v12 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/f4160dbc6c6928cc3d9a95aef596df4731c79df5)
* [v12 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/9fd913cc9580744f4d295a6905a53040cfcbfa87)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...9fd913cc9580744f4d295a6905a53040cfcbfa87)) ([:chart_with_upwards_trend: Interdiff between v11 and v12](https://github.com/se-edu/addressbook-level4/compare/d53832156f7d69ba9b8749953aabeac12d72ef06~1...d53832156f7d69ba9b8749953aabeac12d72ef06))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/12/head:BRANCHNAME
```

</details>

yamgent:
# v12
* Commit `Add TestPerson#getTagsSet()` purged because it was made redundant by #261.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: @yamgent 
> Er, what do you mean by that? Do you mean leave a line after ModelStubAcceptingPersonAdded class?

Yes.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Test
+    public void constructor_invalidAddress_throwsIllegalValueException() throws Exception {
+        thrown.expect(IllegalValueException.class);
+        new AddCommand(VALID_NAME, VALID_PHONE, VALID_EMAIL, "", new UniqueTagList());
+    }
+
+    @Test
+    public void constructor_invalidTags_throwsIllegalValueException() throws Exception {
```
pyokagan: This test case is checking for `NullPointerException`?

With that said, it is indeed true that we should be checking for `NullPointerException` for all the arguments, so I'd expect to see this test case repeated for name, phone, email, address as well :-)

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Test
+    public void execute_newPersonNotInList_addSuccessful() throws Exception {
+        ModelStubAcceptingPersonAdded modelStub = new ModelStubAcceptingPersonAdded();
+        TestPerson validPerson = createValidPerson();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
+
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, validPerson), commandResult.feedbackToUser);
+        assertTrue(modelStub.isAddPersonMethodCalled());
```
pyokagan: Not sufficient to just check if the `addPerson()` method was called -- it may have passed a person with wrong data to `addPerson()`.

On another note, though, do we really need to use a stub for this test? Why not just use `ModelManager` directly and compare the contents of the address book?

pyokagan: > On another note, though, do we really need to use a stub for this test?

Ah right, it's because this PR wants to add true blue unit tests where we are fully testing `AddCommand` in isolation. I'm still not sure if we want to use this approach for testing all our classes, though, as it means we end up implementing a separate stub class for every test, and we end up with a whole load of manually-implemented stub test code we have to maintain.

pyokagan: @damithc Any thoughts on the increase in code when we stub everything in tests? Should we fully commit to this approach?

Another point is that I personally would like to make writing tests "fun" in AddressBook-Level4, to make it easier to contribute. Requiring developers to write stubs manually for everything is unfortunately not very fun, so I'd like to relax this requirement as much as possible.

damithc: > @damithc Any thoughts on the increase in code when we stub everything in tests? Should we fully commit to this approach?

Perhaps we do it the 'pure' way in one place and use the 'mixed' approach in the rest of the classes? The downside is multiple approaches makes the code inconsistent. The upside is that we get to illustrate what the 'pure' unit testing is and make the reader see the cost involved.

CanIHasReview:
# v13

@yamgent submitted v13 for review.

* [v13 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/f4160dbc6c6928cc3d9a95aef596df4731c79df5)
* [v13 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/e5a1233d6d60bd08fb0914fbe4c098e2083f26eb)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...e5a1233d6d60bd08fb0914fbe4c098e2083f26eb)) ([:chart_with_upwards_trend: Interdiff between v12 and v13](https://github.com/se-edu/addressbook-level4/compare/51d47985aa78f79480a92fb5675926135f4fc39f~1...51d47985aa78f79480a92fb5675926135f4fc39f))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/13/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: OK. Just to clarify everything, @yamgent,

* For this PR we can go ahead and use stubs. In future PRs just use whatever (integration/unit tests) is more readable.

* Stubs need to be implemented stringently -- the comment in https://github.com/se-edu/addressbook-level4/pull/273#discussion_r103876941 still applies:

> Not sufficient to just check if the addPerson() method was called -- it may have passed a person with wrong data to addPerson().

On top of that, you may also need to check that `AddCommand` does not call any *other* mutating method (e.g. `updatePerson()`, `deletePerson()`) other than `addPerson()`.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Test
+    public void constructor_invalidAddress_throwsIllegalValueException() throws Exception {
+        thrown.expect(IllegalValueException.class);
+        new AddCommand(VALID_NAME, VALID_PHONE, VALID_EMAIL, "", new UniqueTagList());
+    }
+
+    @Test
+    public void constructor_noName_throwsAssertionError() throws Exception {
```
pyokagan: `nullName`, perhaps?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: Remember to address https://github.com/se-edu/addressbook-level4/pull/273#discussion_r104316589

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+        TestPerson validPerson = createValidPerson();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
+
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, validPerson), commandResult.feedbackToUser);
+        assertTrue(modelStub.isAddPersonMethodCalled());
+    }
+
+    @Test
+    public void execute_duplicatePerson_throwsDuplicatePersonException() throws Exception {
```
pyokagan: Aren't we expecting that `AddCommand` throws a `CommandException` instead, rather than a `DuplicatePersonException`?

CanIHasReview:
# v14

@yamgent submitted v14 for review.

* [v14 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/f4160dbc6c6928cc3d9a95aef596df4731c79df5)
* [v14 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/c3ac97a0d0ca7388b1f671b96a4efb1c58dbfaf4)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...c3ac97a0d0ca7388b1f671b96a4efb1c58dbfaf4)) ([:chart_with_upwards_trend: Interdiff between v13 and v14](https://github.com/se-edu/addressbook-level4/compare/1ed0e9786aa26d7a7eb9ca018871e703ba77f762~1...1ed0e9786aa26d7a7eb9ca018871e703ba77f762))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/14/head:BRANCHNAME
```

</details>

yamgent:
## v14

* Stub now checks that other methods are not called.
* Stub now checks that the correct person is added.
* Typos in test case names are corrected and improvements are made.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+        @Override
+        public void updateFilteredListToShowAll() {
+            fail("This method should not be called.");
+        }
+
+        @Override
+        public void updateFilteredPersonList(Set<String> keywords) {
+            fail("This method should not be called.");
+        }
+    }
```
damithc: Quick question: do both stubs need to verify that other methods are not called?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,242 @@
...
+    public void constructor_nullAddress_throwsAssertionError() throws Exception {
+        thrown.expect(AssertionError.class);
+        new AddCommand(VALID_NAME, VALID_PHONE, VALID_EMAIL, null, new UniqueTagList());
+    }
+
+    @Test
+    public void constructor_nullTags_throwsNullPointerException() throws Exception {
+        thrown.expect(NullPointerException.class);
+        new AddCommand(VALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, null);
+    }
```
damithc: If you are trying to verify that the SUT constructs the `Person` object correctly, one way to do that is to examine the `Person` object created by the SUT. Above tests will pass even if the SUT is passing all `null`s to the Person constructor, right?

yamgent: Ok, will add a new test case to test that the SUT is creating the `Person` object correctly.

The null test cases are separated, because otherwise there's no way of telling which parameter is causing the exception.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
yamgent: Since both stubs are used for the purpose of checking `addPerson()`'s logic, it should be assumed that `addPerson()` might call other methods that it should not be calling regardless of how the result turns out, so they should be verified.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
damithc: Parameter validation is not part of SUT.  'invalid parameter throws exception' is just one path through the SUT, and hence should be only one test case? Passing of parameters to the Person constructor is within SUT but that part can be verified by examining the the resulting Person object.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
damithc: If a successful add operation doesn't call any other mutating methods, there's no chance a failed add operation will? The current way of verifying it in both cases is safer though, even if a bit excessive. Perhaps you can use inheritance to avoid those duplicated method implementations between the two stubs?

pyokagan: @damithc
> If a successful add operation doesn't call any other mutating methods, there's no chance a failed add operation will?

This is my position as well.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: Maybe wait for #315 first? That way you won't have to write all of these tests :-)

CanIHasReview:
# v15

@yamgent submitted v15 for review.

* [v15 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/217e5cefa34fc6c59318b82b652381ab67992548)
* [v15 2/2] [Add unit test for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/152a8cc31b03fe76823ba74c6b1f0f8555d25241)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/c3f8027f059c932c6bc6aded05897dd6eb8c087e...152a8cc31b03fe76823ba74c6b1f0f8555d25241)) ([:chart_with_upwards_trend: Interdiff between v14 and v15](https://github.com/se-edu/addressbook-level4/compare/98c977e20873bc8e858f1f46682d23872e77ebd5~1...98c977e20873bc8e858f1f46682d23872e77ebd5))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/15/head:BRANCHNAME
```

</details>

yamgent:
## v15

### Removing unnecessary test cases

* Since parameters validation is not part of SUT, test cases testing each parameter was removed (only one left behind).
* The `execute()` command checks that the correct person is added to the address book, so we are already indirectly verifying that the SUT actually created a correct person in the constructor. Therefore, the test case that I proposed to create wasn't really required (otheriwse we will have to create a new `AddCommand#getPerson()` method, which feels unnecessay).

### Unnecessary method verification for failing model stub

Both `ModelStub`s that this unit test have are using the `Model` interface. Therefore, we are forced to override the methods even if it is unnecessary. There are two possible solutions to this: 

1. Create a base `ModelStub` that has all of its methods failing, then `ModelStubThrowingDuplicatePersonException` and `ModelStubAcceptingPersonAdded` will inherit this and override `addPerson(Person)` accordingly. _(This is the current design that I pushed into the PR.)_
    * Advantage: Reduce the need to override all methods and avoid cluttering the code.
    * Disadvantage: Maybe the inheritance relationship may not make sense?

1. Extend `ModelManager` class, instead of implementing `Model` interface, for our `ModelStub`s
    * Advantage: Becomes one of the integration tests that we discussed today.
    * Disadvantage: We lose an example of a true-blue unit test.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,167 @@
...
+    public void constructor_invalidParameters_throwsIllegalValueException() throws Exception {
+        thrown.expect(IllegalValueException.class);
+        new AddCommand("", VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, new UniqueTagList());
+    }
+
+    @Test
+    public void constructor_nullParameters_throwsAssertionError() throws Exception {
+        thrown.expect(AssertionError.class);
+        new AddCommand(VALID_NAME, null, VALID_EMAIL, VALID_ADDRESS, new UniqueTagList());
+    }
```
damithc: These two test cases go through the same path in the SUT? I don't think AddCommand is aware of null or invalid parameters.

pyokagan: @damithc 
> I don't think AddCommand is aware of null or invalid parameters.

From the perspective of the caller though, it would seem as if `AddCommand` is doing the validation? So we should at least have some sanity checks for that.

pyokagan: Now, just looking at the current state of the PR, I think the tests in v14 should come back. However, again, you may wish to wait for #315 so you don't have to write so many tests :-P

damithc: > From the perspective of the caller though, it would seem as if AddCommand is doing the validation? So we should at least have some sanity checks for that.

As unit tests are written by the developer of the SUT, they can be written in white-box fashion. That means the tester knows that the validation happens outside of SUT. Otherwise we end up with lot of duplicate tests.  Just like we ensure the logic is only in one place (i.e. DRY), we should try to ensure testing of that logic is also in one place only? What do you think?

pyokagan: @damithc 
> That means the tester knows that the validation happens outside of SUT. That means the tester knows that the validation happens outside of SUT. Otherwise we end up with lot of duplicate tests. 

I'd argue that they aren't truly full *duplicates*, as e.g. one test is testing that `Person` throws the exception, while another test is testing that *`AddCommand`* is throwing the exception. By virtue of the SUTs being different they are not full duplicates.

On top of that, we need the tests because the internal implementation of the `AddCommand` constructor (*not* the `Person` constructor) may change. Motivating example: Rather than constructing the `Person` directly, we store the individual fields first, and then construct them when `execute()` is called. However, this would change behavior, as now the IllegalValueException/AssertionError will not be thrown at construction time, but rather at `execute()` instead. Without these exception tests, we will not be aware of such a subtle behavioral change unless the reviewer is sharp enough (and we really shouldn't rely on reviewers for everything :-P).

damithc: > I'd argue that they aren't truly full duplicates, as e.g. one test is testing that Person throws the exception, while another test is testing that AddCommand is throwing the exception. By virtue of the SUTs being different they are not full duplicates.

Agree. What we are testing is if the AddCommand correctly pass through the exception in the case the parameters are invalid. One test case is enough for that?


> On top of that, we need the tests because the internal implementation of the AddCommand constructor (not the Person constructor) may change. Motivating example: Rather than constructing the Person directly, we store the individual fields first, and then construct them when execute() is called. However, this would change behavior, as now the IllegalValueException/AssertionError will not be thrown at construction time, but rather at execute() instead. Without these exception tests, we will not be aware of such a subtle behavioral change unless the reviewer is sharp enough (and we really shouldn't rely on reviewers for everything :-P).

This is also a valid point.  The issue is, where do we draw the line? It's clear if we say 'unit tests are white-box and there should not be any cases covering potential bugs in code outside the SUT'.  Yes, a reviewer might miss it. Perhaps a guideline such as 'introducing new logic to (or removing logic from) a method should result in a corresponding change in its unit tests'? I'm a fan of having a lot of automated tests but also keen to hit the sweet spot of 'minimal yet sufficient' level of such tests.

May be the problem here is that we have not isolated the SUT from its dependency on `Person`? Stubbing out `Person` sounds like going to extremes but given that this test class is going to serve as an illustrations of going to such extremes ... :-p

yamgent:
Ok will hold until #315 is merged then, would be easier to resolve this when that happens :).

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: @damithc Sorry, I don't think I'll have time to respond for now. Just this though:

> May be the problem here is that we have not isolated the SUT from its dependency on Person?

As I mentioned above, I think the real problem is that `AddCommand` simply repeats the arguments of the `Person` constructor, which is icky API design hence leading to these icky tests :-P #315 will solve that problem. As of now, I'll just accept any resolution to this issue, with the assumption that #315 will be merged soonish in the future.

damithc: Yes, let's expedite #315 so that we can postpone this discussion for a future time.

damithc:
@yamgent can continue with this now?

yamgent:
@damitch Am working on it. Sorry for the delay.

CanIHasReview:
# v16

@yamgent submitted v16 for review.

* [v16 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/c98c78d122a7804e5dcc8c2683c1dc0dd90fa6b8)
* [v16 2/2] [Add unit tests for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/30c37e62b7b253e19a47eff6e02985f5e7c87a6d)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/a5d9fcc4cce6b798663a565091e38782d140b7e7...30c37e62b7b253e19a47eff6e02985f5e7c87a6d)) ([:chart_with_upwards_trend: Interdiff between v15 and v16](https://github.com/se-edu/addressbook-level4/compare/6b8b6b6f5f065dec57bca14b4dae2c7e6f908061~1...6b8b6b6f5f065dec57bca14b4dae2c7e6f908061))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/16/head:BRANCHNAME
```

</details>

yamgent:
# v16
* Adapted the test cases to the new AddCommand constructor. Not sure whether `constructor_nullParameters_throwsNullPointerException()` test case is useful anymore though.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+    private static final String VALID_NAME = "valid name";
+    private static final String VALID_PHONE = "61239876";
+    private static final String VALID_EMAIL = "valid.email@mail.com";
+    private static final String VALID_ADDRESS = "valid address";
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void constructor_nullParameters_throwsNullPointerException() throws Exception {
```
damithc: `_nullPerson_`?

CanIHasReview:
# v17

@yamgent submitted v17 for review.

* [v17 1/2] [Make DuplicatePersonException public](https://github.com/se-edu/addressbook-level4/pull/273/commits/c98c78d122a7804e5dcc8c2683c1dc0dd90fa6b8)
* [v17 2/2] [Add unit tests for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/20cd9ed30c03999027d44096d23e7c761cc6ee3a)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/a5d9fcc4cce6b798663a565091e38782d140b7e7...20cd9ed30c03999027d44096d23e7c761cc6ee3a)) ([:chart_with_upwards_trend: Interdiff between v16 and v17](https://github.com/se-edu/addressbook-level4/compare/f8d5107139a4674db5549c743c8d8319842f5e23~1...f8d5107139a4674db5549c743c8d8319842f5e23))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/17/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+        Person validPerson = createValidPerson();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
+
+        assertEquals(String.format(AddCommand.MESSAGE_SUCCESS, validPerson), commandResult.feedbackToUser);
+        assertTrue(modelStub.isAddPersonMethodCalledWith(validPerson));
+    }
+
+    @Test
+    public void execute_duplicatePerson_throwsCommandException() throws Exception {
```
Zhiyuan-Amos: @yamgent I am wondering if your test is more of a mock rather than a stub.

From mockito website: 

> A mock object is a dummy implementation for an interface or a class in which you define the output of certain method calls. Mock objects are configured to perform a certain behavior during a test. They typical record the interaction with the system and test can validate that.

So currently you are giving dummy implementations to the different methods in model, which fits the description of a mock.

Also, 

> For example, you can validate if certain methods on the mock object were called. 

This is what you're doing on line 48.

yamgent: I based my understanding of Mocks vs Stubs through this article: https://martinfowler.com/articles/mocksArentStubs.html, under the section "The Difference Between Mocks and Stubs". In particular, the `MailService` test example is a particularly interesting example for this PR.

Basically, stubs uses state verification while mock uses behavior verification. Perhaps the naming might not be appropriate, but `isAddPersonMethodCalledWith()` is doing state verification, not behavior verification.

Zhiyuan-Amos: cool i see! Thanks for your thorough explanation :>

pyokagan: fwiw we could probably make the `ModelStubAcceptingPersonAdded` stub more "stub-like" with something like:
```java
@Test
public void execute_newPersonNotInList_addSuccessful() throws Exception {
    ModelStubAcceptingPersonAdded modelStub = new ModelStubAcceptingPersonAdded();

    Person validPerson = createValidPerson();

    CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();

    assertEquals(Arrays.asList(validPerson), modelStub.personsAdded);
}

...

private class ModelStubAcceptingPersonAdded extends ModelStub {
    final ArrayList<Person> personsAdded = new ArrayList<>();

    @Override
    public void addPerson(Person person) throws DuplicatePersonException {
        personsAdded.add(person);
    }
}
```

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void constructor_nullPerson_throwsNullPointerException() throws Exception {
+        thrown.expect(NullPointerException.class);
+        new AddCommand(null);
+    }
+
+    @Test
+    public void execute_newPersonNotInList_addSuccessful() throws Exception {
```
pyokagan: Given that in this test we use a model stub which would accept all persons, shouldn't this test be named `execute_personAcceptedByModel_addSuccessful()`?

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        getAddCommandForPerson(validPerson, modelStub).execute();
+    }
+
+    /**
+     * Creates a person with valid contact details.
+     */
+    private Person createValidPerson() throws IllegalValueException {
+        return new PersonBuilder().withName(VALID_NAME).withPhone(VALID_PHONE).withEmail(VALID_EMAIL)
+                .withAddress(VALID_ADDRESS).build();
```
pyokagan: It seems that `PersonBuilder` now initializes itself with default valid values after d010443ebdd238e46d1abda1a59718027c7f473b, so we don't need this anymore.

CanIHasReview:
# v18

@yamgent submitted v18 for review.

* [v18 1/1] [Add unit tests for AddCommand](https://github.com/se-edu/addressbook-level4/pull/273/commits/e8fd8b188fc239a619c189637a24a1b8fc8f68a0)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8e4d306e4cc579b4e967b267a3c41aafcedbc700...e8fd8b188fc239a619c189637a24a1b8fc8f68a0)) ([:chart_with_upwards_trend: Interdiff between v17 and v18](https://github.com/se-edu/addressbook-level4/compare/9b44234242dff5c9d9fbffb18da9b92aeb578227~1...9b44234242dff5c9d9fbffb18da9b92aeb578227))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/273/18/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -0,0 +1,137 @@
...
+        thrown.expect(NullPointerException.class);
+        new AddCommand(null);
+    }
+
+    @Test
+    public void execute_personAcceptedByModel_addSuccessful() throws Exception {
+        ModelStubAcceptingPersonAdded modelStub = new ModelStubAcceptingPersonAdded();
+        Person validPerson = new PersonBuilder().build();
+
+        CommandResult commandResult = getAddCommandForPerson(validPerson, modelStub).execute();
```
pyokagan: It's interesting how we end up making a `getAddCommandForPerson` helper function just because commands require us to first construct them and then separately pass the model to them via `setData()`.

Imagine if this line could just be
```java
CommandResult commandResult = new AddCommand(validPerson).execute(modelStub);
```
;-)

yamgent:
The commit is rebased to update to latest master branch (actually not necessary, but I accidentally did it anyway so...)

Issue #291, opened 2017-02-16
Make ReadOnlyPerson truly read-only


pyokagan:
`ReadOnlyPerson#getTags()` returns a `UniqueTagList` which is mutable.

pyokagan:
Related: #199

PR #313, opened 2017-03-02
AddressBook#addPerson() does not copy the Person #311

### LINKED ISSUE #311: AddressBook#addPerson() does not copy the Person
This means it's possible to cause constraint violations, e.g.:
```java
addressBook.addPerson(somePerson);
somePerson.setTags(new UniqueTagList("some", "tags")); // constraint violation!
```

Also, `AddressBook#addPerson()` should be taking a `ReadOnlyPerson`.

Need to do a thorough cleaning of all codepaths, e.g. `Model#addPerson(), UniquePersonList#add()...`

yamgent:
Fixes #311.

```
When accepting a new Person, both UniquePersonList and AddressBook
insert the original object, rather than a copy of the object.

Outside classes are able to indirectly modify the list by using the
same object after inserting them into the list. This is undesirable as
UniquePersonList and AddressBook cannot guarantee that the constraints,
such as the uniqueness of the list, continue to hold, which results in
incorrect behavior for both classes.

Let's ensure that UniquePersonList and AddressBook are actually storing
copies of the Person object, rather than the original object, so that
outside classes are not able to modify them after it is inserted into
the list or book.
```

### DIFF
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -83,9 +83,10 @@ public class AddressBook implements ReadOnlyAddressBook {
      *
      * @throws UniquePersonList.DuplicatePersonException if an equivalent person already exists.
      */
-    public void addPerson(Person p) throws UniquePersonList.DuplicatePersonException {
-        syncMasterTagListWith(p);
-        persons.add(p);
+    public void addPerson(ReadOnlyPerson p) throws UniquePersonList.DuplicatePersonException {
+        Person newPerson = new Person(p);
+        syncMasterTagListWith(newPerson);
+        persons.add(newPerson);
     }
 
     /**
--- a/src/main/java/seedu/address/model/Model.java
+++ b/src/main/java/seedu/address/model/Model.java
@@ -3,7 +3,6 @@ package seedu.address.model;
 import java.util.Set;
 
 import seedu.address.commons.core.UnmodifiableObservableList;
-import seedu.address.model.person.Person;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.person.UniquePersonList;
 import seedu.address.model.person.UniquePersonList.DuplicatePersonException;
@@ -22,7 +21,7 @@ public interface Model {
     void deletePerson(ReadOnlyPerson target) throws UniquePersonList.PersonNotFoundException;
 
     /** Adds the given person */
-    void addPerson(Person person) throws UniquePersonList.DuplicatePersonException;
+    void addPerson(ReadOnlyPerson person) throws UniquePersonList.DuplicatePersonException;
 
     /**
      * Updates the person located at {@code filteredPersonListIndex} with {@code editedPerson}.
--- a/src/main/java/seedu/address/model/person/UniquePersonList.java
+++ b/src/main/java/seedu/address/model/person/UniquePersonList.java
@@ -34,12 +34,12 @@ public class UniquePersonList implements Iterable<Person> {
      *
      * @throws DuplicatePersonException if the person to add is a duplicate of an existing person in the list.
      */
-    public void add(Person toAdd) throws DuplicatePersonException {
+    public void add(ReadOnlyPerson toAdd) throws DuplicatePersonException {
         assert toAdd != null;
         if (contains(toAdd)) {
             throw new DuplicatePersonException();
         }
-        internalList.add(toAdd);
+        internalList.add(new Person(toAdd));
     }
 
     /**
--- a/src/main/java/seedu/address/model/ModelManager.java
+++ b/src/main/java/seedu/address/model/ModelManager.java
@@ -10,7 +10,6 @@ import seedu.address.commons.core.UnmodifiableObservableList;
 import seedu.address.commons.events.model.AddressBookChangedEvent;
 import seedu.address.commons.util.CollectionUtil;
 import seedu.address.commons.util.StringUtil;
-import seedu.address.model.person.Person;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.person.UniquePersonList;
 import seedu.address.model.person.UniquePersonList.PersonNotFoundException;
@@ -65,7 +64,7 @@ public class ModelManager extends ComponentManager implements Model {
     }
 
     @Override
-    public synchronized void addPerson(Person person) throws UniquePersonList.DuplicatePersonException {
+    public synchronized void addPerson(ReadOnlyPerson person) throws UniquePersonList.DuplicatePersonException {
         addressBook.addPerson(person);
         updateFilteredListToShowAll();
         indicateAddressBookChanged();
--- a/src/test/java/seedu/address/logic/commands/AddCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/AddCommandTest.java
@@ -69,7 +69,7 @@ public class AddCommandTest {
      */
     private class ModelStub implements Model {
         @Override
-        public void addPerson(Person person) throws DuplicatePersonException {
+        public void addPerson(ReadOnlyPerson person) throws DuplicatePersonException {
             fail("This method should not be called.");
         }
 
@@ -117,7 +117,7 @@ public class AddCommandTest {
      */
     private class ModelStubThrowingDuplicatePersonException extends ModelStub {
         @Override
-        public void addPerson(Person person) throws DuplicatePersonException {
+        public void addPerson(ReadOnlyPerson person) throws DuplicatePersonException {
             throw new DuplicatePersonException();
         }
     }
@@ -129,8 +129,8 @@ public class AddCommandTest {
         final ArrayList<Person> personsAdded = new ArrayList<>();
 
         @Override
-        public void addPerson(Person person) throws DuplicatePersonException {
-            personsAdded.add(person);
+        public void addPerson(ReadOnlyPerson person) throws DuplicatePersonException {
+            personsAdded.add(new Person(person));
         }
     }
 


CanIHasReview:
# v1

* [v1 1/6] [Change UniquePersonList\#add\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/cf007852b144107a7142d1b56dc3d7e339ada67c)
* [v1 2/6] [Add UniquePersonList\#stream\(\)](https://github.com/se-edu/addressbook-level4/pull/313/commits/fbe4f3008cc1757d48790108a9582f97c79ba282)
* [v1 3/6] [Update AddressBook\#syncMasterTagListWith\(\.\.\.\) to generate new Person](https://github.com/se-edu/addressbook-level4/pull/313/commits/471a763d4608d6c3be89cc2d82eb5e60ec38edf0)
* [v1 4/6] [Change AddressBook\#addPerson\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/c5a8e8de1caf2ab92459bbe11cc1c235669c6d0e)
* [v1 5/6] [Change Model\#addPerson\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/a73d64242e4fecd6e84c31a99b9dd42f5d794ce4)
* [v1 6/6] [Change tags of Adam to be in lexicographical order](https://github.com/se-edu/addressbook-level4/pull/313/commits/76bf28cdc8ccdeb437cdbf02371f4e5ee4232b87)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...76bf28cdc8ccdeb437cdbf02371f4e5ee4232b87))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/313/1/head:BRANCHNAME
```

</details>

yamgent:
`AddressBook#syncMasterTagListWith(Person)` modifies the Person object so that it uses the `Tag`s owned by `AddressBook` rather than its own copy of `Tag`s.

This PR changes `AddressBook#addPerson(Person)` to `AddressBook#addPerson(ReadOnlyPerson)`, so changes to `Tag`s are now made on the copy of the Person object rather than the original Person object.

However, this results in the test case `LogicManagerTest#execute_add_successful()` failing, because a side-effect of `AddressBook#syncMasterTagListWith(Person)` is that it sorts the tags of the Person object. Previously the `AddressBook` would modify the test case's input data tags, so the sorting effect was not detected and the test case passes.

There are multiple ways to approach this problem:
* Prevent `AddressBook` from sorting the tags in the first place. This would require some changes to the `AddressBook#syncMasterTagListWith(Person)` logic.
* Change the comparison test such that the order of the tags no longer matter (they can be in a different order, but as long as the content of the tags are the same, it should pass).
* Modify `ReadOnlyPerson#getAsText()` to sort the tags before printing.
* Implement any combination of the approaches above.

I am not sure which approach to take in this situation. (NB: [v1 6/6] is written purely to pass the test case temporarily, and will be removed once a decision is made)

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -112,6 +113,10 @@ public int hashCode() {
         return internalList.hashCode();
     }
 
+    public Stream<Person> stream() {
```
MightyCupcakes: Stream of what? Requires javadocs here since this is a public method.

damithc:
> I'm inclined to go with the solution of modifying the comparison test such that the order of the tags does not matter. But that is just me.

What if the UniqueTagList guarantees that tags are always sorted? Will that help prevent this sort of problems?

yamgent:
@damithc UniqueTagList can be sorted, although naive usage of its `add()` will cause `O(n^2)` complexity (maybe should make the list immutable?)

Also a correction: I re-read the code for `AddressBook#syncMasterTagListWith(Person)`, and realised that it was using a `HashSet`, so I was wrong in saying that it was sorted in a lexicographical order. In fact the order is entirely arbitrary.

damithc:
> @damithc UniqueTagList can be sorted, although naive usage of its add() will cause O(n^2) complexity (maybe should make the list immutable?)

Performance is not a big issue for the tag list because we don't expect it to have lot of elements. In fact the person card should show tags in sorted order, so it makes sense to keep them in sorted order all the time. However, it will make the UniqueTagList behavior differ from UniquePersonList, am I right? Rename it to SortedUniqueTagList? :-p Also, there's the question of whether we are making this PR too big if we do the sorting in this PR.

MightyCupcakes:
Yeah, I agree. Performance should not be an issue because before the number of tags becomes too huge, the UI will look really bad trying to squeeze so many tags in a `PersonCard`. 

In fact, I think there should be a upper-bound on the number of tags to make sure the UI will always look consistent

kychua:
Keeping tag list always sorted makes more sense... it's weird for the user to see the tags in random order - it's not the order they entered but it's also not alphabetical. I suppose this can be considered another PR, so tests for this PR can be changed to ignore order for now?

> In fact, I think there should be a upper-bound on the number of tags to make sure the UI will always look consistent

Would it be enough to ensure the UI is consistent for normal usage, and works decently for unlikely scenarios? As in if the user wants to have that many tags, better to have it look less consistent rather than be told it is impossible? 

Seems like long names aren't displayed properly either.. the index gets cut off when the name's too long.

damithc:
> Keeping tag list always sorted makes more sense... it's weird for the user to see the tags in random order - it's not the order they entered but it's also not alphabetical. I suppose this can be considered another PR, so tests for this PR can be changed to ignore order for now?

Or put this PR on hold until sorting is implemented. The total work is less that way?

> In fact, I think there should be a upper-bound on the number of tags to make sure the UI will always look consistent
> > Would it be enough to ensure the UI is consistent for normal usage, and works decently for unlikely scenarios? As in if the user wants to have that many tags, better to have it look less consistent rather than be told it is impossible?

Yes, better to allow with an ugly UI than block. 
 
> Seems like long names aren't displayed properly either.. the index gets cut off when the name's too long.

Long names are not unusual (especially from where I come from :-p). This should be fixed. Can you post an issue for this? @kychua

yamgent:
Ok, I think I will put this on hold, and write a `SortedUniqueTagList` in another PR.

pyokagan:
Why do we even need a list to store tags? Shouldn't a plain set be sufficient?

"Displaying the tag list as sorted" is simply a concern of the UI. The model shouldn't need to care about that.

yamgent:
> "Displaying the tag list as sorted" is simply a concern of the UI. The model shouldn't need to care about that.

That is a valid point actually. The model can represent the tags whatever way it wants, but the UI should sort it before displaying it. This would solve the issue in this PR without having to worry about other issues raised during the discussion.

damithc:
> "Displaying the tag list as sorted" is simply a concern of the UI. The model shouldn't need to care about that.

That's correct. My thinking was that by having it guaranteed as sorted at source avoids having to sort it in many different places. But we can cross that bridge when we come to it.

> This would solve the issue in this PR without having to worry about other issues raised during the discussion.

In that case, go for it.

pyokagan:
@yamgent You may actually wish to solve #291 first. The model component shouldn't be exposing `UniqueTagList` into the outside world, but rather using a well known interface like `Set` or `List`.

yamgent:
@pyokagan Ok, will take a look at it, thanks for the heads up!

CanIHasReview:
# v2

* [v2 1/5] [Change UniquePersonList\#add\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/38bba27cfacaecd88f1d61388e87375a69db0644)
* [v2 2/5] [Add UniquePersonList\#stream\(\)](https://github.com/se-edu/addressbook-level4/pull/313/commits/8657fd517617e08e2c9e6a6a5cdd83c4d67a9925)
* [v2 3/5] [Update AddressBook\#syncMasterTagListWith\(\.\.\.\) to generate new Person](https://github.com/se-edu/addressbook-level4/pull/313/commits/e40f3ee983aaa54e6af7804cc127308ae800c806)
* [v2 4/5] [Change AddressBook\#addPerson\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/b91ce6c3df67c260eab4cf9303a2c6519ad3d5dd)
* [v2 5/5] [Change Model\#addPerson\(\.\.\.\) signature from Person to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/ca23774aa00d0939f8bbed00408dcf9b3f298aa3)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/dc9fe96a38a2b5f9bb5bb4c05c3ff1db47b1effd...ca23774aa00d0939f8bbed00408dcf9b3f298aa3)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/e02905959685904537376c6e6f6f1aa9f8963cce~1...e02905959685904537376c6e6f6f1aa9f8963cce))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/313/2/head:BRANCHNAME
```

</details>

yamgent:
Oops, some requested changes not made in this version, will submit a new one.

CanIHasReview:
# v3

* [v3 1/3] [Change UniquePersonList\#add\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/de49973ce40850b86663e7dfc6d4bfed7e10669e)
* [v3 2/3] [Change AddressBook\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/54bf8ffda2ed3cae3028c9ae6c8a0ebd2cec9037)
* [v3 3/3] [Change Model\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/7171f69c0a7610d5f2cfe1acb43381bf3f019add)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/dc9fe96a38a2b5f9bb5bb4c05c3ff1db47b1effd...7171f69c0a7610d5f2cfe1acb43381bf3f019add)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/58955c5a28dcee683f0ed7be78344639054f37db~1...58955c5a28dcee683f0ed7be78344639054f37db))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/313/3/head:BRANCHNAME
```

</details>

yamgent:
# v2

With #291 solved, [v1 6/6], which was a temporary fix, is now purged and no longer necessary.

# v3

v1 was a bit too ambitious by enforcing the parameter of `AddressBook#syncMasterTagListWith(Person)` to be immutable. However, that method does not help at all in enforcing the constraint of  `AddressBook#addPerson()` having to copy the Person. Furthermore, it is not an API method. Therefore [v1 3/6] is purged. Changes in [v1 2/6] was done because it was used subsequently [v1 3/6], so it is also purged.

CanIHasReview:
# v4

* [v4 1/3] [Change UniquePersonList\#add\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/712f3a1fe67f3803c669704b19068638b907dee7)
* [v4 2/3] [Change AddressBook\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/ac2c15653558f2833e91cc190cbbe7bdd24165b3)
* [v4 3/3] [Change Model\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/4514affb91c43a8ad18d89a6c02313499cbb1a94)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/8e4d306e4cc579b4e967b267a3c41aafcedbc700...4514affb91c43a8ad18d89a6c02313499cbb1a94)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/0c22821e79be33785aa0f9e62dbab86beba98a8a~1...0c22821e79be33785aa0f9e62dbab86beba98a8a))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/313/4/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -83,9 +83,10 @@ public void resetData(ReadOnlyAddressBook newData) {
      *
```
damithc: Commit message 2: where does this constrain come from?
>This violates the constraint that the object should not be modifiable
after being added to the address book.

src/main/java/seedu/address/model/Model.java
```diff
@@ -3,7 +3,6 @@
 import java.util.Set;
```
damithc: Commit 3 message:

>The model should not be expected to change the Person object when
accepting a new Person, and should use a ReadOnlyPerson version
instead.

The `Model` is free to change the `Person` but nobody else should be changing it afterwards, right?

src/main/java/seedu/address/model/AddressBook.java
yamgent: Will rephrase that sentence since we are adding a new constraint rather than following an existing constraint.

src/main/java/seedu/address/model/Model.java
yamgent: That is correct, will rephrase the justification to make it better

CanIHasReview:
# v5

* [v5 1/3] [Change UniquePersonList\#add\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/ad970d90a8c4a9e81680c053e033f5f2255f459d)
* [v5 2/3] [Change AddressBook\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/1d9a07fd3469504f545ccc4492b2d810994137d2)
* [v5 3/3] [Change Model\#addPerson\(\.\.\.\) signature to ReadOnlyPerson](https://github.com/se-edu/addressbook-level4/pull/313/commits/9e158dd1e06328b723edeeded077f15019ac39d0)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/ef06553753b5073c012ccac9ab1082c46b9d0f51...9e158dd1e06328b723edeeded077f15019ac39d0)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/63d0f8054edfa616e82bf3aa331040731b92c6ad~1...63d0f8054edfa616e82bf3aa331040731b92c6ad))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/313/5/head:BRANCHNAME
```

</details>

yamgent:
@kychua @MightyCupcakes Do take a look again if you guys have the time. Will be merging this by Monday, but you can still leave comments if you spot any issues, even after it is merged.

yamgent:
@Zhiyuan-Amos Can also take a look and approve this? :P

PR #320, opened 2017-03-06
ArgumentTokenizer: ensure tokenizing before getting values #312

### LINKED ISSUE #312: ArgumentTokenizer: ensure tokenizing before getting values
ArgumentTokenizer's tokenized argument getter methods can be accessed before ArgumentTokenizer is given arguments to tokenize through ArgumentTokenizer#tokenize(String). 

Furthermore, the getter methods in `ArgumentTokenizer`, `getValue(Prefix)`, `getAllValues(Prefix)`, and `getPreamble()`, don't give a warning when they are called before `ArgumentTokenizer#tokenize(...)` is ever called.

This behaviour is problematic as the getters can be called before ArgumentTokenizer#tokenize(...) without any warnings, and would return empty optionals, lists, and strings that are indistinguishable from normal empty results.

[Link to discussion that highlighted this issue](https://github.com/se-edu/addressbook-level4/pull/261#discussion_r103476184)

PierceAndy:
Closes #312.

ArgumentTokenizer's tokenized argument getter methods can be accessed before ArgumentTokenizer is given arguments to tokenize through ArgumentTokenizer#tokenize(String). 

Furthermore, the getter methods in `ArgumentTokenizer`, `getValue(Prefix)`, `getAllValues(Prefix)`, and `getPreamble()`, don't give a warning when they are called before `ArgumentTokenizer#tokenize(...)` is ever called.

This behaviour is problematic as the getters can be called before ArgumentTokenizer#tokenize(...) without any warnings, and would return empty optionals, lists, and strings that are indistinguishable from normal empty results.

[Link to discussion that highlighted this issue](https://github.com/se-edu/addressbook-level4/pull/261#discussion_r103476184)

> ArgumentTokenizer's tokenized arguments getters can be called before
> arguments have been tokenized with ArgumentTokenizer#tokenize(String).
> 
> This is problematic as the getters would return empty Optionals, lists,
> and strings that are indistinguishable from normal empty results.
> 
> Let's encapsulate ArgumentTokenizer's getters in ArgumentMultimap such
> that we can enforce the sequence of method calls and simplify
> ArgumentTokenizer's logic.


CanIHasReview:
# v1

@PierceAndy submitted v1 for review.

* [v1 1/3] [ArgumentTokenizer: move tokenize\(\.\.\.\) into constructor](https://github.com/se-edu/addressbook-level4/pull/320/commits/632573aaa2cf004be1818d0579165b2d89384a3b)
* [v1 2/3] [ArgumentTokenizer: remove unnecessary resetTokenizerState\(\)](https://github.com/se-edu/addressbook-level4/pull/320/commits/2d2b597b393a3a0afec1c7620bd81f020c5eb333)
* [v1 3/3] [ArgumentTokenizer: simplify getPreamble\(\) logic](https://github.com/se-edu/addressbook-level4/pull/320/commits/9a53ec04b4296cd55ea303a1ad7065e5ebfb90c2)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...9a53ec04b4296cd55ea303a1ad7065e5ebfb90c2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/1/head:BRANCHNAME
```

</details>

PierceAndy:
@MightyCupcakes Yes, the step builder pattern and nested classes work, but I realized that they were unnecessarily complex, for the reason you mentioned. I will improve the first commit's message with what you wrote in the next iteration.

> Moving the logic into the constructor is preferred over other potential
> solutions such as the step builder pattern or nested classes, as this
> solution does not involve the creation of additional objects and is more
> efficient. Furthermore, there is no need for a more complex solution that 
> preserves ArgumentTokenizer#tokenize(...) as a separate method, as a 
> new ArgumentTokenizer object is created each time we parse a user input, 
> and there are no cases that require consecutive parsing of arguments 
> with identical prefixes other than in the test cases.

damithc:
@PierceAndy what about this alternative?

```java
Arguments arguments = new ArgumentTokenizer().tokenize(...);
String name = agruments.getValue(...)
```

There is a general consensus that doing work inside a constructor is not good. But as you said, we don't want to complicate things too much too.

PierceAndy:
@damithc Right, that would be a simple solution while reducing constructor complexity.

CanIHasReview:
# v2

@PierceAndy submitted v2 for review.

* [v2 1/1] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/8e39586b39357fe871b2ec6e43cb8b6bba3d8f00)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/65fdb4818f191c27f46846588172f644483718af...8e39586b39357fe871b2ec6e43cb8b6bba3d8f00)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/ebafbaaad9479023fa8d47c1210e48be2bd89469~1...ebafbaaad9479023fa8d47c1210e48be2bd89469))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/2/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -102,10 +66,12 @@ private void resetTokenizerState() {
...
     /**
-     * Extracts the preamble/arguments and stores them in local variables.
-     * @param prefixPositions must contain all prefixes in the {@code argsString}
+     * Extracts the preamble/arguments, and returns a map of preamble keys and argument values.
+     * @param argsString      Arguments string of the form: preamble <prefix>value <prefix>value ...
+     * @param prefixPositions 0-based positions of prefixes, contains all prefixes in the {@code argsString}.
+     * @return                A map of preamble keys and argument values.
      */
-    private void extractArguments(String argsString, List<PrefixPosition> prefixPositions) {
+    private  Map<Prefix, List<String>> extractArguments(String argsString, List<PrefixPosition> prefixPositions) {
```
chao1995: Instead of return the map, return `Arguments` directly.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -119,11 +85,13 @@ private void extractArguments(String argsString, List<PrefixPosition> prefixPosi
         prefixPositions.add(endPositionMarker);
 
         // Extract the prefixed arguments and preamble (if any)
+        Map<Prefix, List<String>> tokenizedArguments = new HashMap<>();
         for (int i = 0; i < prefixPositions.size() - 1; i++) {
             String argValue = extractArgumentValue(argsString, prefixPositions.get(i), prefixPositions.get(i + 1));
-            saveArgument(prefixPositions.get(i).getPrefix(), argValue);
+            saveArgument(tokenizedArguments, prefixPositions.get(i).getPrefix(), argValue);
```
chao1995: So as mentioned, if use getOrDefault, can replace this method with 3 lines of code.

For clarity concern, since now the Map is a local variable, adding an instance method for just saving something to a local map, doesn't feel quite right.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -35,41 +30,10 @@ public ArgumentTokenizer(Prefix... prefixes) {
...
-     */
-    public String getPreamble() {
-        Optional<String> storedPreamble = getValue(new Prefix(""));
-        return storedPreamble.orElse("");
-    }
-
-    private void resetTokenizerState() {
-        this.tokenizedArguments.clear();
+        Map<Prefix, List<String>> tokenizedArguments = extractArguments(argsString, positions);
+        return new Arguments(tokenizedArguments);
```
PierceAndy: Yeah I was debating on which return type to give `extractArguments(...)`.
@MightyCupcakes Thoughts?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
PierceAndy: Note to self: extra space before `Map`

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
MightyCupcakes: Yeah, I agree with @chao1995

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
PierceAndy: I would argue that `saveArgument(...)` is following SLAP, and extracting it out into the body of the loop makes the code harder to read and manage.

chao1995: Yes. I didn't use the SLAP term, but I mean it
> The reason not to do this maybe is for clarity concern

My reason is that this method needs a reference to a local variable (the Map). I don't feel right doing that. Plus, in my opinion adding a comment to clear enough for this piece of code (just adding a value to a map).

PierceAndy: Right, that makes sense. Because you want to avoid mutating something that you're passing by reference.

CanIHasReview:
# v3

@PierceAndy submitted v3 for review.

* [v3 1/1] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/0c63ccf622c7783f3cbc14a18593a6ea6baca91e)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/c3f8027f059c932c6bc6aded05897dd6eb8c087e...0c63ccf622c7783f3cbc14a18593a6ea6baca91e)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/eacb12a5f1f19c74bf91b163c5d878973286b731~1...eacb12a5f1f19c74bf91b163c5d878973286b731))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/3/head:BRANCHNAME
```

</details>

PierceAndy:
# v3
- Changed `extractArguments(...)` return type from `Map` to `Arguments`, and updated related comments.
- Extracted `saveArgument(...)` into `extractArguments(...)`'s for loop to avoid mutating `Map` which was being passed by reference into `saveArgument(...)`. 

@chao1995 Thanks for bringing these points up.

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Returns the preamble (text before the first valid prefix). Trims any leading/trailing spaces.
+     */
+    public String getPreamble() {
+        Optional<String> storedPreamble = getValue(new Prefix(""));
+        return storedPreamble.orElse("");
+    }
+}
```
damithc: Quick comment: 
The dependency is `ArgumentTokenizer` --> `Arguments`.  Therefore, try to write `Arguments` in a way that it is independent from `ArguementTokenizer`. Preferrably, it should not even know  the word 'tokenize'. It's just a data structure. It doesn't know how the data were created.

CanIHasReview:
# v4

@PierceAndy submitted v4 for review.

* [v4 1/1] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/8ae7da4e95ad44afb6d9a8c4fc58874f56631a34)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/c3f8027f059c932c6bc6aded05897dd6eb8c087e...8ae7da4e95ad44afb6d9a8c4fc58874f56631a34)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/33d21125e4fa4a859974370396fcbf469245d041~1...33d21125e4fa4a859974370396fcbf469245d041))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/4/head:BRANCHNAME
```

</details>

PierceAndy:
# v4
- Remove reference to tokenizing from `Arguments` class.

CanIHasReview:
# v5

@PierceAndy submitted v5 for review.

* [v5 1/2] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/634f67f26717069003cce321df3514f09f2ce669)
* [v5 2/2] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/305fbee7aa1f75bf9287f97563e0897d97c22d6b)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/c3f8027f059c932c6bc6aded05897dd6eb8c087e...305fbee7aa1f75bf9287f97563e0897d97c22d6b)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/d4c616eb7c32dc326cb4daf7de4bda40554cfccd~1...d4c616eb7c32dc326cb4daf7de4bda40554cfccd))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/5/head:BRANCHNAME
```

</details>

PierceAndy:
# v5
- Improve comments in `ArgumentTokenizer` and `Arguments` classes.
- Rename `args` variable to `argsString` to avoid gems such as `Arguments arguments = argsTokenizer.tokenize(args)`

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -33,43 +28,13 @@ public ArgumentTokenizer(Prefix... prefixes) {
     }
 
     /**
-     * @param argsString arguments string of the form: preamble <prefix>value <prefix>value ...
+     * @param argsString Arguments string of the form: {@code preamble <prefix>value <prefix>value ...}
```
MightyCupcakes: Needs a header sentence to describe what this method does since this is a public API

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Stores mapping of prefixes to their respective arguments.
+ */
+public class Arguments {
+
+    /** Prefixes mapped to their respective arguments**/
+    private final Map<Prefix, List<String>> arguments;
+
+    /**
+     * Creates Arguments object stores the mapping of prefixes to their respective arguments.
```
MightyCupcakes: `Creates an Arguments object that stores the mapping of prefixes to their respective arguments.`  ?

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param arguments Map with {@code Prefix} keys and {@code List<String>} argument values.
+     */
+    public Arguments(Map<Prefix, List<String>> arguments) {
+        this.arguments = arguments;
+    }
+
+    /**
+     * Returns last value of given prefix.
```
MightyCupcakes: `Returns the last value of the given prefix.` ?

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public Optional<String> getValue(Prefix prefix) {
+        List<String> values = getAllValues(prefix);
+        return values.isEmpty() ? Optional.empty() : Optional.of(values.get(values.size() - 1));
+    }
+
+    /**
+     * Returns all values of given prefix, if any.
```
MightyCupcakes: `Returns all values of the given prefix` ?

the `if any` is not needed since the following sentence will explain what happens if there aren't any value

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
...
+     * If the prefix does not exist or has no values, returns an empty list.
```
MightyCupcakes: `If the prefix does not exist or has no values, this will return an empty list.` ?

CanIHasReview:
# v6

@PierceAndy submitted v6 for review.

* [v6 1/2] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/af0fcc1e45125f19f133f6e2c0a7148c35616378)
* [v6 2/2] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/390b9ec5840b46dc8ed933b082d8fed5f0bb19d9~1...390b9ec5840b46dc8ed933b082d8fed5f0bb19d9))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/6/head:BRANCHNAME
```

</details>

PierceAndy:
# v6
- Update comments

src/main/java/seedu/address/logic/parser/Arguments.java
```diff
@@ -0,0 +1,54 @@
+package seedu.address.logic.parser;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import seedu.address.logic.parser.ArgumentTokenizer.Prefix;
```
damithc: Can this be avoided? e.g. by detaching Prefix class from ArgumentTokenizer

CanIHasReview:
# v7

@PierceAndy submitted v7 for review.

* [v7 1/3] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/af0fcc1e45125f19f133f6e2c0a7148c35616378)
* [v7 2/3] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)
* [v7 3/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/c524e83dbf892a623dd9d707a013fddec4554a00)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...c524e83dbf892a623dd9d707a013fddec4554a00)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/compare/28dbdef5f7634083733a4788be9b9df68871b6fc~1...28dbdef5f7634083733a4788be9b9df68871b6fc))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/7/head:BRANCHNAME
```

</details>

PierceAndy:
# v7
- Extract nested `Prefix` class out of `ArgumentTokenizer` into its own class.

CanIHasReview:
# v8

@PierceAndy submitted v8 for review.

* [v8 1/3] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/af0fcc1e45125f19f133f6e2c0a7148c35616378)
* [v8 2/3] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)
* [v8 3/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/c524e83dbf892a623dd9d707a013fddec4554a00)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...c524e83dbf892a623dd9d707a013fddec4554a00)) ([:chart_with_upwards_trend: Interdiff between v7 and v8](https://github.com/se-edu/addressbook-level4/compare/908bd6b6271f32057a67d52f465cccc9104679ed~1...908bd6b6271f32057a67d52f465cccc9104679ed))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/8/head:BRANCHNAME
```

</details>

PierceAndy:
# v7 & v8
- Extract nested `Prefix` class out of `ArgumentTokenizer` class.

CanIHasReview:
# v9

@PierceAndy submitted v9 for review.

* [v9 1/3] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/af0fcc1e45125f19f133f6e2c0a7148c35616378)
* [v9 2/3] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)
* [v9 3/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/fd134801a5564f8300032640bdc03ba953b92dcb)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...fd134801a5564f8300032640bdc03ba953b92dcb)) ([:chart_with_upwards_trend: Interdiff between v8 and v9](https://github.com/se-edu/addressbook-level4/compare/4f0581c0a45bd0c3854cdf2d3a9c8607126fbd74~1...4f0581c0a45bd0c3854cdf2d3a9c8607126fbd74))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/9/head:BRANCHNAME
```

</details>

PierceAndy:
# v7 & v9
- Extract nested `Prefix` class out of `ArgumentTokenizer` class.

src/main/java/seedu/address/logic/parser/Prefix.java
```diff
@@ -0,0 +1,35 @@
+package seedu.address.logic.parser;
+
+/**
+ * A prefix that marks the beginning of an argument in an arguments string.
+ * E.g. '/t' in 'add James /t friend'.
+ */
+public class Prefix {
+    final String prefix;
```
MightyCupcakes: Missing visibility modifiers.

src/main/java/seedu/address/logic/parser/Prefix.java
```diff
@@ -0,0 +1,35 @@
...
+
+    Prefix(String prefix) {
```
MightyCupcakes: Same here, and the rest.

src/main/java/seedu/address/logic/parser/Prefix.java
damithc: If you want to keep the members package private (i.e. without visibility modifier), may be make it clear using a header comment?

PierceAndy: This was an oversight on my part. I don't think there's a special reason to start using package private access modifiers here.

CanIHasReview:
# v10

@PierceAndy submitted v10 for review.

* [v10 1/3] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/af0fcc1e45125f19f133f6e2c0a7148c35616378)
* [v10 2/3] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/f5b5bcb80d558216b84c1f404de7cf0c1bcb9b00)
* [v10 3/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/9387a17e1c52f124c9c9aac4c68760256f69c2cd)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...9387a17e1c52f124c9c9aac4c68760256f69c2cd)) ([:chart_with_upwards_trend: Interdiff between v9 and v10](https://github.com/se-edu/addressbook-level4/compare/92ffa0e26f9c0ad151009e1d51ab19b6e8135ecb~1...92ffa0e26f9c0ad151009e1d51ab19b6e8135ecb))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/10/head:BRANCHNAME
```

</details>

PierceAndy:
# v10
- Add access modifiers to `Prefix` class.

damithc:
I wonder if extracting the Prefix should be done earlier so that the mutual dependency doesn't happen at all.
Also, note that after the code is somewhat settled, you need to add a commit to update design diagrams.

damithc:
Also, one of the commits is quite big, although I'm not sure if it can be split meaningfully.

PierceAndy:
@damithc 
> I wonder if extracting the Prefix should be done earlier so that the mutual dependency doesn't happen at all. Also, note that after the code is somewhat settled, you need to add a commit to update design diagrams.

Noted.

> Also, one of the commits is quite big, although I'm not sure if it can be split meaningfully.

Commit `[1/3] ArgumentTokenizer: extract getters into Arguments class` could be split up into two commits: the first one introducing the new `Arguments` class, and the second propagating the changes out to the rest of the code base?

damithc:
> Commit [1/3] ArgumentTokenizer: extract getters into Arguments class could be split up into two commits: the first one introducing the new Arguments class, and the second propagating the changes out to the rest of the code base?

Not sure if it is meaningful to add a class without using it. :-p Such a split can make the commit hard to understand too.  We can leave it as it is for now. I haven't done a full review yet, so I'm not entirely sure if the current commit is hard to review.

CanIHasReview:
# v11

@PierceAndy submitted v11 for review.

* [v11 1/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/506670d0a19334b6d25c33b330d780a0ff7ff502)
* [v11 2/3] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/06fe55c795da956fe619fd077dea3581c65ca860)
* [v11 3/3] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/82226eee2d9c5cde996db4f64dcbe4c320635990)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...82226eee2d9c5cde996db4f64dcbe4c320635990)) ([:chart_with_upwards_trend: Interdiff between v10 and v11](https://github.com/se-edu/addressbook-level4/compare/2873f756644b634424ef85300b91a5c2114fba4d~1...2873f756644b634424ef85300b91a5c2114fba4d))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/11/head:BRANCHNAME
```

</details>

PierceAndy:
# v11
- Shift [v10 3/3] to [v11 1/3], `Extract Prefix class out of ArgumentTokenizer class`

CanIHasReview:
# v12

@PierceAndy submitted v12 for review.

* [v12 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/506670d0a19334b6d25c33b330d780a0ff7ff502)
* [v12 2/4] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/06fe55c795da956fe619fd077dea3581c65ca860)
* [v12 3/4] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/82226eee2d9c5cde996db4f64dcbe4c320635990)
* [v12 4/4] [Update Logic class diagram](https://github.com/se-edu/addressbook-level4/pull/320/commits/c1d62cfb48447b71b844caadcf817c5ca938d188)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...c1d62cfb48447b71b844caadcf817c5ca938d188)) ([:chart_with_upwards_trend: Interdiff between v11 and v12](https://github.com/se-edu/addressbook-level4/compare/8903fd51ad7cb7963945e3412f24e2456e780ae2~1...8903fd51ad7cb7963945e3412f24e2456e780ae2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/12/head:BRANCHNAME
```

</details>

PierceAndy:
# v12
- Update Logic class diagram, to show updated dependencies between `ArgumentTokenizer`, `Arguments`, `Prefix`, and command parser classes.

MightyCupcakes:
The powerpoint file needs to be updated along with the corresponding image.

CanIHasReview:
# v13

@PierceAndy submitted v13 for review.

* [v13 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/506670d0a19334b6d25c33b330d780a0ff7ff502)
* [v13 2/4] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/06fe55c795da956fe619fd077dea3581c65ca860)
* [v13 3/4] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/82226eee2d9c5cde996db4f64dcbe4c320635990)
* [v13 4/4] [Update Logic class diagram](https://github.com/se-edu/addressbook-level4/pull/320/commits/9c2444ccb2e6ee74da6d60a19f178ba7658681cb)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...9c2444ccb2e6ee74da6d60a19f178ba7658681cb)) ([:chart_with_upwards_trend: Interdiff between v12 and v13](https://github.com/se-edu/addressbook-level4/compare/6e461d9055e2cc785d7edb4e952b61b615a805a2~1...6e461d9055e2cc785d7edb4e952b61b615a805a2))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/13/head:BRANCHNAME
```

</details>

PierceAndy:
# v13
- Update Logic class diagram in `.pptx` file
- Improve layout of classes in Logic class diagram

src/main/java/seedu/address/logic/parser/Prefix.java
```diff
@@ -0,0 +1,35 @@
+package seedu.address.logic.parser;
+
+/**
+ * A prefix that marks the beginning of an argument in an arguments string.
+ * E.g. '/t' in 'add James /t friend'.
```
MightyCupcakes: `* E.g. '/t' in 'add James /t friend'.` -> `* E.g. 't/' in 'add James t/ friend'.` ?

PierceAndy:
@chao1995 [v13 4/4] includes `Arguments` class, introduced in [v13 2/4]
[v13 4/4] could be moved to [v13 3/4], but I like the idea of updating the design diagrams after the code has settled.

CanIHasReview:
# v14

@PierceAndy submitted v14 for review.

* [v14 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/7fc6b554f26779e025c47325411c02417556a2ec)
* [v14 2/4] [ArgumentTokenizer: extract getters into Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/7a28678ad1d409abcabe69c7355748339802e9a0)
* [v14 3/4] [Update Logic class diagram with Arguments class](https://github.com/se-edu/addressbook-level4/pull/320/commits/7738761e6ebc56e05f77d86ba611cf9f87a66aee)
* [v14 4/4] [Parser: rename args variable to argsString](https://github.com/se-edu/addressbook-level4/pull/320/commits/097f6947bbf94b30e7a8985a387b02d86b1d15e5)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/297506b87fabe6de6baa7901388039ec50199802...097f6947bbf94b30e7a8985a387b02d86b1d15e5)) ([:chart_with_upwards_trend: Interdiff between v13 and v14](https://github.com/se-edu/addressbook-level4/compare/c8d06eada48f12e2682008301d2c02cd57307158~1...c8d06eada48f12e2682008301d2c02cd57307158))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/14/head:BRANCHNAME
```

</details>

PierceAndy:
# v14
- Correct typo in class header: `'/t'` -> `'t/'`
- [v13 4/4] moved to [v13 3/4]

damithc:
> One last concern from me is using Arguments as class name. Usually we use singular word for class names.

Yes, not common but not unheard of either. e.g. JDK has a class named `Arrays`.

chao1995:
> Yes, not common but not unheard of either. e.g. JDK has a class named Arrays.

`Arrays` looks like a utility class only (also `Collections`), but `Arguments` is not a utility class.

chao1995:
> Yes, not common but not unheard of either. e.g. JDK has a class named Arrays.

`Arrays` looks like a utility class only (also `Collections`), but `Arguments` is not a utility class. so it's possible that we want to create a list of `Arguments`.

damithc:
> Arrays looks like a utility class only (also Collections), but Arguments is not a utility class.

That's true.

PierceAndy:
@damithc @chao1995 How about `ArgumentMap`? Since it's effectively a mapping of prefixes to their respective arguments.

damithc:
> How about ArgumentMap? Since it's effectively a mapping of prefixes to their respective arguments.

Sure. `ArgumentMap` or `ArgumentsMap`?

chao1995:
@PierceAndy
The prefix `Collection` or `Map` would make people think it's an extension of the Java's `Collection` or `Map`.

Prof suggested `ArgumentsBundle` in last Thursday's class. I think this name is by far the best I encounter.  :)

PierceAndy:
How about `ArgumentsContainer`? It's in the spirit of both `ArgumentsCollection` and `ArgumentsBundle`, and it shouldn't mislead like the `Map` suffix does. `Bundle` as a suffix doesn't seem common in the wild.

damithc:
> How about ArgumentsContainer? It's in the spirit of both ArgumentsCollection and ArgumentsBundle, and it shouldn't mislead like the Map suffix does. Bundle as a suffix doesn't seem common in the wild.

[One other use of `ArgumentMap`](https://github.com/seleniumQuery/seleniumQuery/blob/master/src/main/java/io/github/seleniumquery/by/common/preparser/ArgumentMap.java)

Given that most suitable suffixes are already used by Java, I'm inclined to use `ArgumentMap`. It's short and to the point.

MightyCupcakes:
If we don't want to use map because of conflicts with the standard java library, why not use its counterpart `Dictionary`?

`ArgumentDictionary`

CanIHasReview:
# v15

@PierceAndy submitted v15 for review.

* [v15 1/3] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/b70234be2fad7b9e9302b8fae89f403b93d80bf9)
* [v15 2/3] [ArgumentTokenizer: extract getters into ArgumentMap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/99745d068eb767e2ed1ed2858e4d273e765d40c5)
* [v15 3/3] [Update Logic class diagram with ArgumentMap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/909d388485908bbe2bba14a899f63b339ba58059)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/f5efb728b72c193913006d96d4d5591da810af50...909d388485908bbe2bba14a899f63b339ba58059)) ([:chart_with_upwards_trend: Interdiff between v14 and v15](https://github.com/se-edu/addressbook-level4/compare/8ddc6f50eaf33400ff7ef40913423f7a7549a6d7~1...8ddc6f50eaf33400ff7ef40913423f7a7549a6d7))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/15/head:BRANCHNAME
```

</details>

PierceAndy:
# v15
- Rename `Arguments` class to `ArgumentMap`
- Update class diagrams
- Remove `[v14 4/4] Parser: rename args variable to argsString` as the potential for confusion between `Arguments` and `args` is no longer present

> There are only two hard things in Computer Science: cache invalidation and naming things. 
> -- Phil Karlton

chao1995:
> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

Another one: "There are only two hard things in Computer Science: cache invalidation, naming things, and off-by-one errors."

damithc:
> If we don't want to use map because of conflicts with the standard java library, why not use its counterpart Dictionary?

Something relevant: http://stackoverflow.com/questions/2884068/what-is-the-difference-between-a-map-and-a-dictionary

PierceAndy:
@damithc 
1. As it is, `ArgumentTokenizer#tokenize(...)` can't be static as it is dependent on the prefixes provided to `ArgumentTokenizer`'s constructor. 
We could turn `ArgumentTokenizer` into something of a one-trick utility class and make `ArgumentTokenizer#tokenize(...)` static by changing its signature to `ArgumentTokenizer#tokenize(String, Prefix...)`, i.e. tokenize a given string with the given prefixes.
This would do away with the need to create a tokenizer each time we wish to tokenize a string, but we would also lose the ability to consecutively tokenize strings with the same set of prefixes. However, given that the latter feature is not actually used in production, this seems like a reasonable trade-off.
2. It could, not that it would be used anywhere. But given how it is reasonable to expect that in a map and we're not using DIP here, we should put it in to be consistent.

damithc:
> We could turn ArgumentTokenizer into something of a one-trick utility class and make ArgumentTokenizer#tokenize(...) static by changing its signature to ArgumentTokenizer#tokenize(String, Prefix...), i.e. tokenize a given string with the given prefixes.

That's what I was thinking too. It doesn't seem like we have a need to instantiate a `ArgumentTokenizer`. A state-independent method is easier to test. 

>It could, not that it would be used anywhere. 
But given how it is reasonable to expect that in a map and we're not using DIP here, we should put it in to be consistent with how maps are designed.

Actually, it can be used inside `ArgumentTokenizer#tokenize(...)`. Instead of creating the entire `Map` and passing it to the `ArgumentMap` constructor, `ArgumentTokenizer#tokenize(...)` can `put` each argument into the `ArgumentMap` object as they are being extracted.

CanIHasReview:
# v16

@PierceAndy submitted v16 for review.

* [v16 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/df208f7718fb86027345408af2ebee35355090e1)
* [v16 2/4] [ArgumentTokenizer: extract getters into ArgumentMap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/f36045d25573acd18cbed07ef7d03d3fcabde8b0)
* [v16 3/4] [Update Logic class diagram with ArgumentMap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/088b93b75c601beb385d5cd30f831dae775d1c48)
* [v16 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/029f3ed18cfb5ca3bfdb6edb0136d9ee24613e0a)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/4c4e1ed78fd8465512c5b434a3c592da52740717...029f3ed18cfb5ca3bfdb6edb0136d9ee24613e0a)) ([:chart_with_upwards_trend: Interdiff between v15 and v16](https://github.com/se-edu/addressbook-level4/compare/ba78cdb90d008c7c87a8514aaff9defc6930a0d3~1...ba78cdb90d008c7c87a8514aaff9defc6930a0d3))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/16/head:BRANCHNAME
```

</details>

PierceAndy:
# v16
1. Make `ArgumentTokenizer#tokenize(...)` static
2. Add `ArgumentMap#put(Prefix, List<String>)`
3. Rename `argumentMap` variables to `argMap` to be consistent with `args`, `argPrefix`, `argValue`, etc
4. Improved various comments

## Further notes
1. This makes the constructor and list of prefixes attribute redundant. The constructor's arguments are now accepted by `ArgumentTokenizer#tokenize(String, Prefix...)` instead, and other functions that accessed the attribute are now passed the prefixes through their arguments instead.
2. This also has the benefit of getting rid `ArgumentTokenizer`'s dependency on Java's Map and HashMap, and fully relying on `ArgumentMap` instead.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -118,19 +86,27 @@ private void extractArguments(String argsString, List<PrefixPosition> prefixPosi
...
+        // Map prefixes to their argument values (if any)
+        ArgumentMap argMap = new ArgumentMap();
         for (int i = 0; i < prefixPositions.size() - 1; i++) {
+            // Extract prefixes and their arguments
+            Prefix argPrefix = prefixPositions.get(i).getPrefix();
             String argValue = extractArgumentValue(argsString, prefixPositions.get(i), prefixPositions.get(i + 1));
-            saveArgument(prefixPositions.get(i).getPrefix(), argValue);
+
+            // Append extracted argument to any previously encountered arguments with the same prefix
+            List<String> argValues = new ArrayList<>(argMap.getAllValues(argPrefix));
```
damithc: Argument accumulation for the same prefix can be encapsulated inside `ArgumentMap` class. i.e. use `put(Prfex, String)` instead of `put(Prfex, List<String>)`

PierceAndy: I toyed with the idea for a bit, but I realized it would be inconsistent with how Java's `Map#put(...)` works, which would trip up developers. `Map#put(...)` is defined to replace the original value. Even though we're not implementing `Map` in `ArgumentMap`, I think it's better to be consistent with `put(...)`'s expected behaviour.

`Map#merge(key, value, bifunction)` was a candidate method to name the accumulating function after, but this would require the calling function to provide the implementation for the bifunction, which defeats the entire purpose.

The best alternative would be to use another method name, something like `ArgumentMap#add(Prefix, String)` or `ArgumentMap#append(Prefix, String)`.

chao1995: You can keep put(Prefix, List<String>), but add another method `insert(Prefix, String)`. In C++, there is a [multimap](http://www.cplusplus.com/reference/map/multimap/insert/) class, which allows multiple values for the same key, insert is the name they use.

chao1995: There is a [MultipleMap](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/Multimap.html) class in Guava, which allows multiple values for the same key.

The name of choice is `putAll` and `put`.

So I think we can follow it. Using `putAll(Prefix, List<String>)`, but add an `put(Prefix, String)`. So instead of creating a new ArrayList every prefix, encapsulate this in `ArgumentMap`.

PierceAndy: If we were to go with MultiMaps and (once again) rename `ArgumentMap` -> `ArgumentMultiMap`, I think it would be safe to use `put` as a method name that accumulates values over the same `Prefix` key.

chao1995: Maybe `ArgumentMap` is enough... Just in case `put` is too misleading, we can borrow `insert` from C++'s multimap.

PierceAndy: I think it's fine to stick to Java's terminologies for now. I feel that `Multimap` is a good solution, as it fits the description of our implementation of `ArgumentMap` well, and it lets others know that it's not just a simple map.

CanIHasReview:
# v17

@PierceAndy submitted v17 for review.

* [v17 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/df208f7718fb86027345408af2ebee35355090e1)
* [v17 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/721fa9869141f1307a6a5203b4fa3cbeb53dc82e)
* [v17 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/1b043208a5758ff4772b61a688f6f859283c4c6f)
* [v17 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/01ad42bff4e35d3889728058d903deeeddf5e569)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/4c4e1ed78fd8465512c5b434a3c592da52740717...01ad42bff4e35d3889728058d903deeeddf5e569)) ([:chart_with_upwards_trend: Interdiff between v16 and v17](https://github.com/se-edu/addressbook-level4/compare/7f11efd321c7f772c35e5c8abbc063d9f68ca480~1...7f11efd321c7f772c35e5c8abbc063d9f68ca480))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/17/head:BRANCHNAME
```

</details>

PierceAndy:
# v17
- Rename `ArgumentMap` class -> `ArgumentMultimap`
- Rename `argMap` variables -> `argMultimap`
- Encapsulate argument accumulation for the same prefix inside `ArgumentMultimap` class with `ArgumentMultimap#put(Prefix, String)`.
- Minor comment fixes

MightyCupcakes:
If you are going to use `MultiMap`, might have to explain more about its functionality in the header javadocs of the class since it is not clear at first why it is called multimap.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Associates the specified argument value with the specified prefix key in this map.
+     * If the map previously contained a mapping for the key, the new value is appended to the list of existing values.
+     *
+     * @param prefix   Prefix key with which the specified argument value is to be associated
+     * @param argValue Argument value to be associated with the specified prefix key
+     */
+    public void put(Prefix prefix, String argValue) {
+        List<String> argValues = new ArrayList<>(getAllValues(prefix));
```
chao1995: This part adds O(N) time complexity. It is unnecessary to initialize another ArrayList.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Returns all values of the given prefix.
+     * If the prefix does not exist or has no values, this will return an empty list.
+     */
+    public List<String> getAllValues(Prefix prefix) {
+        if (!argMultimap.containsKey(prefix)) {
+            return Collections.emptyList();
+        }
+        return new ArrayList<>(argMultimap.get(prefix));
```
chao1995: These four lines can be simplified to one line with the use of `getOrDefault`. And again, don't initialize another ArrayList.

MightyCupcakes: But that will allow modifications to the internal list itself?

chao1995:
```java
import java.util.ArrayList;
import java.util.HashMap;

public class TestMap {
    public static void main(String[] args) {
        HashMap<Integer, ArrayList<Integer>> m = new HashMap<>();
        ArrayList<Integer> arr = new ArrayList<>();
        arr.add(2);
        arr.add(3);
        m.put(1, arr);
        m.get(1).add(3);
        System.out.println(arr); // prints [2, 3, 3]
    }
}
```
I just test it. This is the bahaviour of Map.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
PierceAndy: I thought about removing the wrapping ArrayLists, but it was throwing errors when I did so. The problem was with `Collections.emptyList()` from `ArgumentMultimap#getAllValues(...)` as it was immutable, but wrapping it in an ArrayList turns it into a normal empty list.

CanIHasReview:
# v18

@PierceAndy submitted v18 for review.

* [v18 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/df208f7718fb86027345408af2ebee35355090e1)
* [v18 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/65347037ad3449266cf06e18697a1ab09947e92b)
* [v18 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/55987465b6b977783ec034328c49d0852e34728b)
* [v18 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/9b4bb8e9f370c164fd47550b8a9749081b9b1346)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/4c4e1ed78fd8465512c5b434a3c592da52740717...9b4bb8e9f370c164fd47550b8a9749081b9b1346)) ([:chart_with_upwards_trend: Interdiff between v17 and v18](https://github.com/se-edu/addressbook-level4/compare/e8b6136614250f42a50bf5801c1c0ad8b6688ed1~1...e8b6136614250f42a50bf5801c1c0ad8b6688ed1))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/18/head:BRANCHNAME
```

</details>

PierceAndy:
# v18
- Update `ArgumentMultimap` header comments
- Remove unnecessary ArrayList instantiations
- Change possible return value of `ArgumentMultimap#getAllValues(...)` from immutable `Collections.emptyList()` to `new ArrayList()`

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
PierceAndy: I made a mistake with v18.

@MightyCupcakes highlighted the reason behind the wrapping and non-usage of `getOrDefault(...)`. Even though removing the wrapping would follow HashMap's implementation, we don't strictly have to follow it as ours is a Multimap.

The wrapping + non-usage of `getOrDefault(...)` is because using `getOrDefault(prefix, new ArrayList<>())` leads to inconsistent behaviour. If The prefix key exists, the list that is returned when modified, also modifies the internal list of the Multimap. However, if the key doesn't exist, the new empty list that is returned is not actually mapped, and modifying it doesn't modify the internal list of the Multimap.

We should be consistent with how `getAllValues(...)` return type behaves, and I think it would be better if it doesn't allow modifications to the internal list.

chao1995: > However, if the key doesn't exist, the new empty list that is returned is not actually mapped

Maybe can change to

```java
if (!argMultimap.containsKey(prefix)) {
        argMultimap.put(prefix, new ArrayList<>());
}
return argMultimap.get(prefix);
```

CanIHasReview:
# v19

@PierceAndy submitted v19 for review.

* [v19 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/df208f7718fb86027345408af2ebee35355090e1)
* [v19 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/58c419d2fcb110f4191e8efd0835f69f3dc63217)
* [v19 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/036197bd16a648efc12cc34ac8ab705537d52d0a)
* [v19 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/0914be7286a4f8cfec12cb63c2d1e1823b235d73)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/4c4e1ed78fd8465512c5b434a3c592da52740717...0914be7286a4f8cfec12cb63c2d1e1823b235d73)) ([:chart_with_upwards_trend: Interdiff between v18 and v19](https://github.com/se-edu/addressbook-level4/compare/99c8d636b05fbdd0951f37e50f9c2d7697a04ff6~1...99c8d636b05fbdd0951f37e50f9c2d7697a04ff6))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/19/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
MightyCupcakes: I don't think a getter should create a key in the map if the said key does not exist. It may lead to exceed the expected size of the map and cause a complete rehashing which is needlessly expensive for a map get operation which is supposed to be O(1)

PierceAndy:
# v19
- Revert incorrect use of `HashMap#getOrDefault(...)` in `ArgumentMultimap#getAllValues(Prefix)`

## Notes
The non-usage of `getOrDefault(...)` is because using `getOrDefault(prefix, new ArrayList<>())` leads to inconsistent behaviour. If the prefix key exists, the list that is returned when modified, also modifies the internal list of the `ArgumentMultimap`. However, if the key doesn't exist, the new empty list that is returned is not actually mapped, and modifying it doesn't modify the internal list of the `ArgumentMultimap`.

We should be consistent with how `getAllValues(...)` return type behaves, and it would be better if it doesn't allow modifications to the internal list. Therefore, the returned list is also wrapped in a new ArrayList.

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -22,58 +18,32 @@
...
-     * Creates an ArgumentTokenizer that can tokenize arguments string as described by prefixes
+     * Creates an ArgumentTokenizer object that can tokenize arguments strings as described by the given prefixes.
      */
     public ArgumentTokenizer(Prefix... prefixes) {
         this.prefixes = Arrays.asList(prefixes);
     }
 
     /**
-     * @param argsString arguments string of the form: preamble <prefix>value <prefix>value ...
+     * Tokenizes an arguments string and returns an ArgumentMultimap object that maps prefixes to their respective
```
damithc: `{@code ArgumentMultimap}`?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -22,58 +18,32 @@
...
-     */
-    public String getPreamble() {
-        Optional<String> storedPreamble = getValue(new Prefix(""));
-        return storedPreamble.orElse("");
-    }
-
-    private void resetTokenizerState() {
-        this.tokenizedArguments.clear();
+        ArgumentMultimap argMultimap = extractArguments(argsString, positions);
+        return argMultimap;
```
damithc: `return extractArguments(argsString, positions);` ?

src/main/java/seedu/address/logic/parser/ArgumentTokenizer.java
```diff
@@ -86,7 +56,11 @@ private void resetTokenizerState() {
     }
 
     /**
-     * Finds all positions in an arguments string at which a given {@code prefix} appears
+     * Finds all zero-based prefix positions of the given {@code prefix} in the given arguments string.
+     *
+     * @param argsString Arguments string of the form: {@code preamble <prefix>value <prefix>value ...}
+     * @param prefix     Prefix to find in the arguments string
+     * @return           List of zero-based prefix positions of the given prefix in the given arguments string
```
damithc: Consider using `{@see}` to refer to other methods instead of repeating the same header comment, especially for private methods.

src/main/java/seedu/address/logic/parser/AddCommandParser.java
```diff
@@ -23,9 +23,8 @@
      * and returns an AddCommand object for execution.
      */
     public Command parse(String args) {
-        ArgumentTokenizer argsTokenizer =
-                new ArgumentTokenizer(PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS, PREFIX_TAG);
-        ArgumentMultimap argMultimap = argsTokenizer.tokenize(args);
+        ArgumentMultimap argMultimap = ArgumentTokenizer.tokenize(args,
```
damithc: Line break at higher level (i.e. `=` or `.`)?

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -29,9 +29,8 @@
      */
     public Command parse(String args) {
         assert args != null;
-        ArgumentTokenizer argsTokenizer =
-                new ArgumentTokenizer(PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS, PREFIX_TAG);
-        ArgumentMultimap argMultimap = argsTokenizer.tokenize(args);
+        ArgumentMultimap argMultimap = ArgumentTokenizer.tokenize(args,
+                PREFIX_PHONE, PREFIX_EMAIL, PREFIX_ADDRESS, PREFIX_TAG);
```
damithc: Same here.

CanIHasReview:
# v20

@PierceAndy submitted v20 for review.

* [v20 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/df208f7718fb86027345408af2ebee35355090e1)
* [v20 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/0ebf916d7d1a9d6aa7ec8b38264fa96e11822c07)
* [v20 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/ea0c126d531ef3843d32d2314d50586a20a0a17d)
* [v20 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/780ef5888c318ecda209dc969230e691689cab02)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/4c4e1ed78fd8465512c5b434a3c592da52740717...780ef5888c318ecda209dc969230e691689cab02)) ([:chart_with_upwards_trend: Interdiff between v19 and v20](https://github.com/se-edu/addressbook-level4/compare/05848185cf8c9df8852856764902dbd30e9fa169~1...05848185cf8c9df8852856764902dbd30e9fa169))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/20/head:BRANCHNAME
```

</details>

PierceAndy:
# v20
- Improve comments and formatting

damithc:
@PierceAndy Also propose a merge commit message.

PierceAndy:
@damithc Proposed merge commit message added to first post.

PierceAndy:
@damithc Proposed merge commit message:

> ArgumentTokenizer's tokenized arguments getters can be called before
> arguments have been tokenized with ArgumentTokenizer#tokenize(String).
> 
> This is problematic as the getters would return empty Optionals, lists,
> and strings that are indistinguishable from normal empty results.
> 
> Let's encapsulate ArgumentTokenizer's getters in ArgumentMultimap such
> that we can enforce the sequence of method calls and simplify
> ArgumentTokenizer's logic.

CanIHasReview:
# v21

@PierceAndy submitted v21 for review.

* [v21 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/9398ec60da751da81eacb976fb18daa284fa846d)
* [v21 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/2be974cc339fa4745f139869915246ac0086d8b8)
* [v21 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/f32fc3394a0c968b03c696621335a5268fc1ae1a)
* [v21 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/eaae7e37035bd1df0c8757dda97090df0b1bd1ac)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/e1eafd1145cb5fdcc81bbacecf9e5145903b3fe2...eaae7e37035bd1df0c8757dda97090df0b1bd1ac)) ([:chart_with_upwards_trend: Interdiff between v20 and v21](https://github.com/se-edu/addressbook-level4/compare/0f0db5c9cf6f89a0f7913af7b9ecec1e5e0d3f5b~1...0f0db5c9cf6f89a0f7913af7b9ecec1e5e0d3f5b))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/21/head:BRANCHNAME
```

</details>

PierceAndy:
# v21
- Rebase master

CanIHasReview:
# v22

@PierceAndy submitted v22 for review.

* [v22 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cce7b1a7813517114eb96e2e10ebf6f37b768971)
* [v22 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cc9189f059e8e7b332ffd068d7bed59980bc3095)
* [v22 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cc8c831ce4dab26328f74237ea68e6a7c2ad7f9c)
* [v22 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/9c87cb50a8c584f77d24aed70fb551019b7f8ba7)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/27cb4621dca1bf474b2838352a90873f3fbcb4ef...9c87cb50a8c584f77d24aed70fb551019b7f8ba7)) ([:chart_with_upwards_trend: Interdiff between v21 and v22](https://github.com/se-edu/addressbook-level4/compare/93d43a8b532702365a2fb7a2cec89a4a0458473c~1...93d43a8b532702365a2fb7a2cec89a4a0458473c))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/22/head:BRANCHNAME
```

</details>

PierceAndy:
# v22
- Rebase with master

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /** Prefixes mapped to their respective arguments**/
+    private final Map<Prefix, List<String>> argMultimap;
+
+    /**
+     * Creates an ArgumentMultimap object that stores the mapping of prefixes to their respective arguments.
+     */
+    public ArgumentMultimap() {
+        argMultimap = new HashMap<>();
+    }
```
damithc: Can/should we avoid this constructor by initializing the map in line 19?

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return new ArrayList<>(argMultimap.get(prefix));
+    }
+
+    /**
+     * Returns the preamble (text before the first valid prefix). Trims any leading/trailing spaces.
+     */
+    public String getPreamble() {
+        Optional<String> storedPreamble = getValue(new Prefix(""));
+        return storedPreamble.orElse("");
```
damithc: Can merge these two lines?

CanIHasReview:
# v23

@PierceAndy submitted v23 for review.

* [v23 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cce7b1a7813517114eb96e2e10ebf6f37b768971)
* [v23 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cc9189f059e8e7b332ffd068d7bed59980bc3095)
* [v23 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/d82f95aab8a8997e1f0cbde643f6d7d97d720736)
* [v23 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/77aa2e6743a5b172b1c17ba8cff0c9b200d030c2)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/27cb4621dca1bf474b2838352a90873f3fbcb4ef...77aa2e6743a5b172b1c17ba8cff0c9b200d030c2)) ([:chart_with_upwards_trend: Interdiff between v22 and v23](https://github.com/se-edu/addressbook-level4/compare/070afa6fe5726886c5f9899840d6b2c9f3610917~1...070afa6fe5726886c5f9899840d6b2c9f3610917))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/23/head:BRANCHNAME
```

</details>

PierceAndy:
# v23
- Minor improvements to `ArgumentMultimap` class

PierceAndy:
@MightyCupcakes A couple of minor changes since your last review.

CanIHasReview:
# v24

@PierceAndy submitted v24 for review.

* [v24 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cce7b1a7813517114eb96e2e10ebf6f37b768971)
* [v24 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cc9189f059e8e7b332ffd068d7bed59980bc3095)
* [v24 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/7bb0af6594af38a0bca446d47c6b7654f5a1a98e)
* [v24 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/31411b92e416d633bc1e5dc6873951411a42e6df)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/27cb4621dca1bf474b2838352a90873f3fbcb4ef...31411b92e416d633bc1e5dc6873951411a42e6df)) ([:chart_with_upwards_trend: Interdiff between v23 and v24](https://github.com/se-edu/addressbook-level4/compare/decf81b3d51bf3938999c981888c72c2ebb00b7e~1...decf81b3d51bf3938999c981888c72c2ebb00b7e))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/24/head:BRANCHNAME
```

</details>

PierceAndy:
# v24
- Fix minor typo.

src/main/java/seedu/address/logic/parser/ArgumentMultimap.java
```diff
@@ -16,14 +16,7 @@
...
     /** Prefixes mapped to their respective arguments**/
-    private final Map<Prefix, List<String>> argMultimap;
-
-    /**
-     * Creates an ArgumentMultimap object that stores the mapping of prefixes to their respective arguments.
-     */
-    public ArgumentMultimap() {
-        argMultimap = new HashMap<>();
-    }
+    private final Map<Prefix, List<String>> argMultimap = new HashMap<>();
```
MightyCupcakes: Should be in commit 2 instead of this commit

CanIHasReview:
# v25

@PierceAndy submitted v25 for review.

* [v25 1/4] [Extract Prefix class out of ArgumentTokenizer class](https://github.com/se-edu/addressbook-level4/pull/320/commits/cce7b1a7813517114eb96e2e10ebf6f37b768971)
* [v25 2/4] [ArgumentTokenizer: extract getters into ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/96c1f6221d8f799c26e6f9166eceaf4f8a3c0a89)
* [v25 3/4] [Update Logic class diagram with ArgumentMultimap class](https://github.com/se-edu/addressbook-level4/pull/320/commits/0d2e7b284fdd56c4526b7bf097682a7fb789b22f)
* [v25 4/4] [ArgumentTokenizer: make tokenize\(\.\.\.\) static](https://github.com/se-edu/addressbook-level4/pull/320/commits/b3baf264f5b368654e18e774d604629e1ff7007e)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/27cb4621dca1bf474b2838352a90873f3fbcb4ef...b3baf264f5b368654e18e774d604629e1ff7007e)) ([:chart_with_upwards_trend: Interdiff between v24 and v25](https://github.com/se-edu/addressbook-level4/compare/3924e4c3d4748136c45a9ece0ade64e4a2202979~1...3924e4c3d4748136c45a9ece0ade64e4a2202979))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/320/25/head:BRANCHNAME
```

</details>

PierceAndy:
# v25
- Move erroneous changes from [v24 3/4] to [v24 2/4]

PR #329, opened 2017-03-08
Make ReadOnlyPerson observable #292

### LINKED ISSUE #292: Make ReadOnlyPerson observable
Currently we use ugly hacks all over the place in order to ensure that the UI is updated when we modify a Person object.

e.g. https://github.com/se-edu/addressbook-level4/blob/f40ee42cb2e29a99c9c6b8757b64b0b0f4284d67/src/main/java/seedu/address/model/person/UniquePersonList.java#L61-L63 
```java
        // TODO: The code below is just a workaround to notify observers of the updated person.
        // The right way is to implement observable properties in the Person class.
        // Then, PersonCard should then bind its text labels to those observable properties.
```

MightyCupcakes:
Closes #292

### DIFF
--- a/src/main/java/seedu/address/model/person/ObservablePerson.java
+++ b/src/main/java/seedu/address/model/person/ObservablePerson.java
@@ -1,0 +1,21 @@
+package seedu.address.model.person;
+
+import javafx.beans.property.ReadOnlyProperty;
+import seedu.address.commons.core.UnmodifiableObservableList;
+import seedu.address.model.tag.Tag;
+
+/**
+ * An observable interface for a Person in the addressbook.
+ */
+public interface ObservablePerson {
+
+    ReadOnlyProperty<String> nameProperty();
+    ReadOnlyProperty<String> addressProperty();
+    ReadOnlyProperty<String> phoneProperty();
+    ReadOnlyProperty<String> emailProperty();
+
+    /**
+     * Returns an unmodifiable observable list of the underlying tag list of this {@code Person}
+     */
+    UnmodifiableObservableList<Tag> tagProperty();
+}
--- a/src/main/java/seedu/address/model/person/Person.java
+++ b/src/main/java/seedu/address/model/person/Person.java
@@ -2,7 +2,12 @@ package seedu.address.model.person;
 
 import java.util.Objects;
 
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.ReadOnlyProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import seedu.address.commons.core.UnmodifiableObservableList;
 import seedu.address.commons.util.CollectionUtil;
+import seedu.address.model.tag.Tag;
 import seedu.address.model.tag.UniqueTagList;
 
 /**
@@ -11,11 +16,15 @@ import seedu.address.model.tag.UniqueTagList;
  */
 public class Person implements ReadOnlyPerson {
 
+    private final ObjectProperty<String> nameProperty;
+    private final ObjectProperty<String> addressProperty;
+    private final ObjectProperty<String> emailProperty;
+    private final ObjectProperty<String> phoneProperty;
+
     private Name name;
     private Phone phone;
     private Email email;
     private Address address;
-
     private UniqueTagList tags;
 
     /**
@@ -27,6 +36,12 @@ public class Person implements ReadOnlyPerson {
         this.phone = phone;
         this.email = email;
         this.address = address;
+
+        nameProperty = new SimpleObjectProperty<>(name.fullName);
+        phoneProperty = new SimpleObjectProperty<>(phone.value);
+        emailProperty = new SimpleObjectProperty<>(email.value);
+        addressProperty = new SimpleObjectProperty<>(address.value);
+
         this.tags = new UniqueTagList(tags); // protect internal tags from changes in the arg list
     }
 
@@ -39,7 +54,9 @@ public class Person implements ReadOnlyPerson {
 
     public void setName(Name name) {
         assert name != null;
+
         this.name = name;
+        nameProperty.set(this.name.fullName);
     }
 
     @Override
@@ -47,9 +64,16 @@ public class Person implements ReadOnlyPerson {
         return name;
     }
 
+    @Override
+    public ReadOnlyProperty<String> nameProperty() {
+        return nameProperty;
+    }
+
     public void setPhone(Phone phone) {
         assert phone != null;
+
         this.phone = phone;
+        phoneProperty.set(this.phone.value);
     }
 
     @Override
@@ -57,9 +81,16 @@ public class Person implements ReadOnlyPerson {
         return phone;
     }
 
+    @Override
+    public ReadOnlyProperty<String> phoneProperty() {
+        return phoneProperty;
+    }
+
     public void setEmail(Email email) {
         assert email != null;
+
         this.email = email;
+        emailProperty.set(this.email.value);
     }
 
     @Override
@@ -67,9 +98,16 @@ public class Person implements ReadOnlyPerson {
         return email;
     }
 
+    @Override
+    public ReadOnlyProperty<String> emailProperty() {
+        return emailProperty;
+    }
+
     public void setAddress(Address address) {
         assert address != null;
+
         this.address = address;
+        addressProperty.set(this.address.value);
     }
 
     @Override
@@ -78,8 +116,18 @@ public class Person implements ReadOnlyPerson {
     }
 
     @Override
+    public ReadOnlyProperty<String> addressProperty() {
+        return addressProperty;
+    }
+
+    @Override
     public UniqueTagList getTags() {
         return new UniqueTagList(tags);
+    }
+
+    @Override
+    public UnmodifiableObservableList<Tag> tagProperty() {
+        return tags.asObservableList();
     }
 
     /**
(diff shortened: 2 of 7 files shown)


src/main/java/seedu/address/model/person/Person.java
```diff
@@ -37,56 +49,64 @@ public Person(ReadOnlyPerson source) {
         this(source.getName(), source.getPhone(), source.getEmail(), source.getAddress(), source.getTags());
     }
 
+    private void populatePropertyMap() {
```
Zhiyuan-Amos: Since you have a `propertyMap` which already has the mapping to the values of the `Person`, do you still need the the object variables `address`, `email` etc?

Sorry if this comment is a newbie question; not too familiar with javafx.

Zhiyuan-Amos: For example, `getPhone()` can be replaced with `propertyMap.get(property).get()`.

MightyCupcakes: good point I didn't really thought of that

CanIHasReview:
# v1

@MightyCupcakes submitted v1 for review.

* [v1 1/3] [Make ReadOnlyPerson observable](https://github.com/se-edu/addressbook-level4/pull/329/commits/0cd396dab0aeab685662a0d74564f4ead02fe113)
* [v1 2/3] [PersonCard: Bind text labels to their respective Person properties](https://github.com/se-edu/addressbook-level4/pull/329/commits/cf5279298a48f082c22153efb322014bd32bf96c)
* [v1 3/3] [UniquePersonList: Remove workaround to notify UI of changes in Person](https://github.com/se-edu/addressbook-level4/pull/329/commits/9b8f0218de99c6a7e03f5e92877006ea72fcb4d0)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/c3f8027f059c932c6bc6aded05897dd6eb8c087e...9b8f0218de99c6a7e03f5e92877006ea72fcb4d0))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/329/1/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -34,9 +55,57 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
...
+        personPropertyConsumerMap.put(PersonProperty.ADDRESS, (person) -> address.setText(person.getAddress().value));
+        personPropertyConsumerMap.put(PersonProperty.EMAIL, (person) -> email.setText(person.getEmail().value));
+        personPropertyConsumerMap.put(PersonProperty.NAME, (person) -> name.setText(person.getName().fullName));
+        personPropertyConsumerMap.put(PersonProperty.PHONE, (person) -> phone.setText(person.getPhone().value));
+        personPropertyConsumerMap.put(PersonProperty.TAG, this::initTags);
+    }
+
+    private void bindListeners(ReadOnlyPerson person) {
+        for (PersonProperty property : personProperties) {
+            removeListenersForPerson(person);
```
MightyCupcakes: This is a misplaced. Will be moved out the loop

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -34,9 +55,57 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
...
+            ChangeListener<Object> changeListener =
+                (o, oldVal, newVal) -> personPropertyConsumerMap.get(property).accept(person);
+
+            person.registerListener(property, changeListener);
+
+            LISTENER_MAP.put(person, new WeakReference<>(changeListener));
+        }
+    }
+
+    private void removeListenersForPerson (ReadOnlyPerson person) {
```
acjh: Remove space between `removeListenersForPerson` and `(`

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -34,9 +55,57 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
...
+        if (!LISTENER_MAP.containsKey(person)) {
+            // If there is no listeners currently registered for this person, exit.
```
acjh: - "If there `are` no listeners"
- "currently" is unnecessary

src/main/java/seedu/address/model/person/ObservablePerson.java
```diff
@@ -0,0 +1,25 @@
+package seedu.address.model.person;
+
+import javafx.beans.value.ChangeListener;
+
+/**
+ * An observable interface for a Person in the addressbook.
```
acjh: "`A`ddressbook"

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -34,9 +55,57 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
         address.setText(person.getAddress().value);
         email.setText(person.getEmail().value);
         initTags(person);
+
+        personPropertyConsumerMap = Maps.newEnumMap(PersonProperty.class);
+
+        populatePersonPropertyConsumerMap();
```
acjh: Empty line after related method call, rather than before?

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -103,6 +110,16 @@ public void resetData(ReadOnlyPerson replacement) {
     }
 
     @Override
+    public void registerListener(PersonProperty property, ChangeListener<Object> listener) {
```
acjh: Preserve the name of the method it wraps? (i.e. rename this to `addListener`)

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -103,6 +110,16 @@ public void resetData(ReadOnlyPerson replacement) {
...
+        propertyMap.get(property).addListener(listener);
+    }
+
+    @Override
+    public void removeListener(PersonProperty property, ChangeListener<Object> listener) {
+        propertyMap.get(property).removeListener(listener);
+    }
```
PierceAndy: It might be good to include comments on what happens when you add the same listener more than once with `registerListener(...)`, or when you try to remove a listener that does not exist with `removeListener(...)`?

Right now to find out, you have to refer to the official API documents for `ChangeListener`.

MightyCupcakes:
@yamgent I thought about it but you can always cast it and violate the read-only property. Since it looks like we are trying to really enforce the read-only properties (especially for UniqueTagList because we have to wrap the underlying taglist for the listeners to be notified of changes), it seem appropriate to also plug that possible leak.

MightyCupcakes:
@yamgent 

We can do that but the tag list in the UI is not a text property. Every tag is a new label

src/main/java/seedu/address/ui/PersonCard.java
PierceAndy: To be consistent with `bindListeners(ReadOnlyPerson)`, how about `unbindListeners(ReadOnlyPerson)`?
Alternatively, `addListeners(ReadOnlyPerson)` and `removeListeners(ReadOnlyPerson)`.

src/main/java/seedu/address/ui/PersonCard.java
PierceAndy: Code is already self-documenting.

yamgent:
@MightyCupcakes You make a valid point about the tag list. Was kind of hoping we would be able to cut down some code by binding the property, but the tag list still would need a listener here. Do you think we should stay with the current design, or would a hybrid (tag list listener, everything else property) be good?

MightyCupcakes:
@yamgent 

To actually bind the textproperty to observed object, we still need another method for the `ObservablePerson` interface. Then you got a bunch of attributes using this method and Tag alone using two methods. I am not sure but this seems a little messy and it makes implementing `ReadOnlyPerson` even more troublesome that it is. New developers might wonder why we are not standardising everything and this again has to be in the commit message (and it is already so long).

Since the usage of listeners cannot be avoided, then I suggest we stick to using it for all. If not we are implementing working code and then avoid using it for a seemingly easier way.

@PierceAndy  thoughts?

pyokagan:
The idiomatic JavaFX way for working with an observable value `X` is to:

1. have a `Xproperty()` method which exposes the observable property

2. have a `getX()` getter method

3. have a `setX()` setter method if the property is writable.

c.f. https://dominiks.gitbooks.io/javafx/content/java8/properties.html

It may be good to just follow this convention. A major reason is because JavaFX has lots of supporting infrastructure for dealing with observable values, e.g. the [`Bindings`](https://docs.oracle.com/javase/8/javafx/api/javafx/beans/binding/Bindings.html) utility class, so it would be good to tap into that.

pyokagan:
@MightyCupcakes 
> We can do that for every Person attribute except for the tag list in the UI as it is not a text property. Every tag is a new label and added as a child node to the main flowpane

We need a way to transform an `ObservableList<Tag>` into an `ObservableList<Label>` or something. Perhaps look at implementing a helper class like a [`MappedList`](https://gist.github.com/TomasMikula/8883719). It is sad that JavaFX still doesn't come with such a utility included ([the bug has been open since 2014](https://bugs.openjdk.java.net/browse/JDK-8091967)), especially since mapping is such a fundamental operation.

pyokagan:
@MightyCupcakes I'm having second thoughts about the validity of the issue #292 , wrt with how we can ensure correct functionality when interacting with transformation lists (e.g. `FilteredList`). You may wish to put this PR on hold (until I find time to investigate) or investigate the validity of the issue further yourself.

PR #332, opened 2017-03-09
fix developer guide issue

R-o-y:
fix developer guide issue: hyperlink cannot go to the corresponding section


acjh:
An identical fix was tried and reverted. Please see @damithc's comments in #246.

Issue #358, opened 2017-03-24
Fix errors in GUI tests when upgrading to the new version (4.0.6-alpha) of TestFx


chao1995:
An extension of #353.

Previously #354 fixes the GUI tests error by fixing the TestFx to an older version (4.0.5-alpha). However, we need to find out what is the reason causing GUI tests to fail when upgrading to the new version (4.0.6-alpha) of TestFx.

VeryLazyBoy:
In 4.0.6-alpha, the method `getParameters` in `MainApp#getApplicationParameter(...)` is supposed to return a object of `Application.Parameter` with empty fields when there is no parameters passed to the application. However, it seems that `getParameters` gives a null value instead in this new version of TestFX, which causes a `NullPointerException`.

chao1995:
@VeryLazyBoy Thanks for looking into this! I saw you closed the PR. why?

VeryLazyBoy:
Because I think, instead of changing our own code to comply with TestFX, choosing a stable version of it might be a right way?  I really don't understand why TestFX has changed the behavior of `getParameters`. According to Javadoc of `javafx.application.Application#getParameters`, `getParameters` shouldn't return null when it is called inside `init`. Perhaps, our guitest failures are due to TestFX's own fault?

damithc:
@VeryLazyBoy you can file a bug report with TestFx and see what they say.

VeryLazyBoy:
I got a reply from TestFX. In 4.0.6-alpha, to increase compatibility of Java 9, they remove a private API which is used to register parameters of JavaFX application. As a result, the parameters for application are not correctly passed. It seems that they are finding a way to rectify the problem. **Here is the link to** [**their reply**](https://github.com/TestFX/TestFX/issues/380)

pyokagan:
Hrm, a new TestFX version hasn't been released yet?

Anyway, I'm personally interested in upgrading to 4.0.6-alpha because it makes headless mode in Linux work. With 4.0.5-alpha, it would just error out with `IllegalAccessException` :-(

Quick patch:
```diff
diff --git a/build.gradle b/build.gradle
--- a/build.gradle
+++ b/build.gradle
@@ -40,8 +40,8 @@ allprojects {
         jacksonVersion = '2.7.0'
         jacksonDataTypeVersion = '2.7.4'
         junitVersion = '4.12'
-        testFxVersion = '4.0.5-alpha'
-        monocleVersion = '1.8.0_20'
[...]

PR #359, opened 2017-03-24
Fix errors in GUI tests when upgrading to the new version (4.0.6-alpha) of TestFx #358

VeryLazyBoy:


VeryLazyBoy:
I think this PR is unnecessary.

Issue #254, opened 2017-01-16
Move some of the EditCommand tests to lower levels


damithc:
Currently the EditCommand is tested mostly at GUI level which is costly. Most of these tests should be moved to lower level unit/integration tests.

Zhiyuan-Amos:
@damithc we will close this issue, and we will have a separate issue for improving the system tests? :>

PR #392, opened 2017-05-03
Improve Edit Command Tests #254

Zhiyuan-Amos:
Part of #254.


damithc:
In the interest of merging/reviewing faster, may be you want to create separate PRs for adding tests at different levels? After those PRs are merged, we can delete redundant tests at other levels.

PR #393, opened 2017-05-03
Add more tests for EditCommand #254 

Zhiyuan-Amos:
Part of #254

### DIFF
--- a/src/test/java/seedu/address/logic/commands/EditCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/EditCommandTest.java
@@ -7,10 +7,14 @@ import static org.junit.Assert.fail;
 import static seedu.address.testutil.EditCommandTestUtil.DESC_AMY;
 import static seedu.address.testutil.EditCommandTestUtil.DESC_BOB;
 import static seedu.address.testutil.EditCommandTestUtil.VALID_NAME_BOB;
+import static seedu.address.testutil.EditCommandTestUtil.VALID_PHONE_BOB;
+import static seedu.address.testutil.EditCommandTestUtil.VALID_TAG_HUSBAND;
 import static seedu.address.testutil.TypicalPersons.INDEX_FIRST_PERSON;
 import static seedu.address.testutil.TypicalPersons.INDEX_SECOND_PERSON;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 
 import org.junit.Test;
@@ -38,7 +42,7 @@ public class EditCommandTest {
     private Model model = new ModelManager(new TypicalPersons().getTypicalAddressBook(), new UserPrefs());
 
     @Test
-    public void execute_validCommand_succeeds() throws Exception {
+    public void execute_allFieldsSpecifiedUnfilteredList_success() throws Exception {
         Person editedPerson = new PersonBuilder().build();
         EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(editedPerson).build();
         EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON, descriptor);
@@ -52,7 +56,57 @@ public class EditCommandTest {
     }
 
     @Test
-    public void execute_duplicatePerson_throwsCommandException() throws Exception {
+    public void execute_someFieldsSpecifiedUnfilteredList_success() throws Exception {
+        Index indexLastPerson = Index.fromOneBased(model.getFilteredPersonList().size());
+        ReadOnlyPerson lastPerson = model.getFilteredPersonList().get(indexLastPerson.getZeroBased());
+
+        PersonBuilder personInList = new PersonBuilder(lastPerson);
+        Person editedPerson = personInList.withName(VALID_NAME_BOB).withPhone(VALID_PHONE_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+
+        EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB)
+                .withPhone(VALID_PHONE_BOB).withTags(VALID_TAG_HUSBAND).build();
+        EditCommand editCommand = prepareCommand(indexLastPerson, descriptor);
+
+        String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
+
+        Model expectedModel = new ModelManager(new AddressBook(model.getAddressBook()), new UserPrefs());
+        expectedModel.updatePerson(lastPerson, editedPerson);
+
+        assertCommandSuccess(editCommand, expectedMessage, expectedModel);
+    }
+
+    @Test
+    public void execute_noFieldSpecifiedUnfilteredList_success() throws Exception {
+        EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON, new EditPersonDescriptor());
+        ReadOnlyPerson editedPerson = model.getFilteredPersonList().get(INDEX_FIRST_PERSON.getZeroBased());
+
+        String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
+
+        Model expectedModel = new ModelManager(new AddressBook(model.getAddressBook()), new UserPrefs());
+
+        assertCommandSuccess(editCommand, expectedMessage, expectedModel);
+    }
+
+    @Test
+    public void execute_filteredList_success() throws Exception {
+        showFirstPersonOnly();
+
+        ReadOnlyPerson personInFilteredList = model.getFilteredPersonList().get(INDEX_FIRST_PERSON.getZeroBased());
+        Person editedPerson = new PersonBuilder(personInFilteredList).withName(VALID_NAME_BOB).build();
+        EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON,
+                new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB).build());
+
+        String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
+
+        Model expectedModel = new ModelManager(new AddressBook(model.getAddressBook()), new UserPrefs());
+        expectedModel.updatePerson(model.getFilteredPersonList().get(0), editedPerson);
+
+        assertCommandSuccess(editCommand, expectedMessage, expectedModel);
+    }
+
+    @Test
+    public void execute_duplicatePersonUnfilteredList_failure() throws Exception {
         Person firstPerson = new Person(model.getFilteredPersonList().get(INDEX_FIRST_PERSON.getZeroBased()));
         EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(firstPerson).build();
         EditCommand editCommand = prepareCommand(INDEX_SECOND_PERSON, descriptor);
@@ -60,10 +114,37 @@ public class EditCommandTest {
     }
 
     @Test
-    public void execute_invalidPersonIndex_throwsCommandException() throws Exception {
-        Index outOfBoundIndex = Index.fromZeroBased(model.getFilteredPersonList().size());
+    public void execute_duplicatePersonFilteredList_failure() throws Exception {
+        showFirstPersonOnly();
+
+        // edit person in filtered list into a duplicate in address book
+        ReadOnlyPerson personInList = model.getAddressBook().getPersonList().get(INDEX_SECOND_PERSON.getZeroBased());
+        EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON,
+                new EditPersonDescriptorBuilder(personInList).build());
+        assertCommandFailure(editCommand, EditCommand.MESSAGE_DUPLICATE_PERSON);
+    }
+
+    @Test
+    public void execute_invalidPersonIndexUnfilteredList_failure() throws Exception {
+        Index outOfBoundIndex = Index.fromOneBased(model.getFilteredPersonList().size() + 1);
         EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB).build();
         EditCommand editCommand = prepareCommand(outOfBoundIndex, descriptor);
+        assertCommandFailure(editCommand, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+    }
+
+    /**
+     * Edit filtered list where index is larger than size of filtered list,
+     * but smaller than size of address book
+     */
+    @Test
+    public void execute_invalidPersonIndexFilteredList_failure() throws Exception {
+        showFirstPersonOnly();
+        Index outOfBoundIndex = INDEX_SECOND_PERSON;
+        // ensures that outOfBoundIndex is still in bounds of address book list
+        assertTrue(outOfBoundIndex.getZeroBased() < model.getAddressBook().getPersonList().size());
+
+        EditCommand editCommand = prepareCommand(outOfBoundIndex,
+                new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB).build());
         assertCommandFailure(editCommand, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
     }
 
@@ -99,6 +180,17 @@ public class EditCommandTest {
         EditCommand editCommand = new EditCommand(index, descriptor);
         editCommand.setData(model, new CommandHistory());
         return editCommand;
+    }
+
+    /**
+     * Updates the filtered list to show only the first person in the {@code model}'s address book.
+     */
+    private void showFirstPersonOnly() {
+        ReadOnlyPerson person = model.getAddressBook().getPersonList().get(0);
+        final String[] splitName = person.getName().fullName.split("\\s+");
+        model.updateFilteredPersonList(new HashSet<>(Arrays.asList(splitName)));
+
+        assertTrue(model.getFilteredPersonList().size() == 1);
     }
 
     /**
--- a/src/test/java/seedu/address/logic/commands/EditCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/EditCommandTest.java
@@ -12,13 +12,11 @@ import static seedu.address.testutil.TypicalPersons.INDEX_SECOND_PERSON;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Optional;
 
 import org.junit.Test;
 
 import seedu.address.commons.core.Messages;
 import seedu.address.commons.core.index.Index;
-import seedu.address.commons.exceptions.IllegalValueException;
 import seedu.address.logic.CommandHistory;
 import seedu.address.logic.commands.EditCommand.EditPersonDescriptor;
 import seedu.address.logic.commands.exceptions.CommandException;
@@ -42,7 +40,7 @@ public class EditCommandTest {
     @Test
     public void execute_validCommand_succeeds() throws Exception {
         Person editedPerson = new PersonBuilder().build();
-        EditPersonDescriptor descriptor = createEditPersonDescriptor(editedPerson);
+        EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(editedPerson).build();
         EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON, descriptor);
 
         String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
@@ -56,7 +54,7 @@ public class EditCommandTest {
     @Test
     public void execute_duplicatePerson_throwsCommandException() throws Exception {
         Person firstPerson = new Person(model.getFilteredPersonList().get(INDEX_FIRST_PERSON.getZeroBased()));
-        EditPersonDescriptor descriptor = createEditPersonDescriptor(firstPerson);
+        EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(firstPerson).build();
         EditCommand editCommand = prepareCommand(INDEX_SECOND_PERSON, descriptor);
         assertCommandFailure(editCommand, EditCommand.MESSAGE_DUPLICATE_PERSON);
     }
@@ -101,20 +99,6 @@ public class EditCommandTest {
         EditCommand editCommand = new EditCommand(index, descriptor);
         editCommand.setData(model, new CommandHistory());
         return editCommand;
-    }
-
-    /**
-     * Returns an {@code EditPersonDescriptor} with fields containing {@code person}'s details
-     */
-    private EditPersonDescriptor createEditPersonDescriptor(ReadOnlyPerson person) throws IllegalValueException {
-        EditPersonDescriptor descriptor = new EditPersonDescriptor();
-        descriptor.setName(Optional.of(person.getName()));
-        descriptor.setPhone(Optional.of(person.getPhone()));
-        descriptor.setEmail(Optional.of(person.getEmail()));
-        descriptor.setAddress(Optional.of(person.getAddress()));
-        descriptor.setTags(Optional.of(person.getTags()));
-
-        return descriptor;
     }
 
     /**
--- a/src/test/java/seedu/address/logic/commands/EditCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/EditCommandTest.java
@@ -16,7 +16,6 @@ import java.util.Optional;
 
 import org.junit.Test;
 
-import javafx.collections.transformation.FilteredList;
 import seedu.address.commons.core.Messages;
 import seedu.address.commons.core.index.Index;
 import seedu.address.commons.exceptions.IllegalValueException;
@@ -26,7 +25,6 @@ import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.AddressBook;
 import seedu.address.model.Model;
 import seedu.address.model.ModelManager;
-import seedu.address.model.ReadOnlyAddressBook;
 import seedu.address.model.UserPrefs;
 import seedu.address.model.person.Person;
 import seedu.address.model.person.ReadOnlyPerson;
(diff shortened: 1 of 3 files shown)


CanIHasReview:
# v1

@Zhiyuan-Amos submitted v1 for review.

* [v1 1/2] [Change PersonBuilder constructor](https://github.com/se-edu/addressbook-level4/pull/393/commits/c472a0e920ffc38aaa4ea7c7ee76f241c89d9370)
* [v1 2/2] [Move out Integration Tests in EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/ebf05d28371cc809b1ad80d702249434bff6877d)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/dc9fe96a38a2b5f9bb5bb4c05c3ff1db47b1effd...ebf05d28371cc809b1ad80d702249434bff6877d))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/393/1/head:BRANCHNAME
```

</details>

Zhiyuan-Amos:
@yamgent is there a need to make each commit so bite-sized? ><

yamgent:
@Zhiyuan-Amos Yes, this is so that we retain the ability to revert one part of the commit without affecting the rest.

yamgent:
Just for clarification, the reason why I mention splitting is because two different things are happening here:

* edit_allFieldsSpecified_success() is actually being deleted, rather than being transferred, because it is already in `EditCommandIntegrationTest`, so it is a duplicate.
* The rest are being transfered.

yamgent:
Actually, it seems that you are also removing `edit_allFieldsSpecified_success()` in PR #395?

Zhiyuan-Amos:
@yamgent #392 see prof's message :) reply you the rest tmr :P

yamgent:
>@yamgent #392 see prof's message :) reply you the rest tmr :P

Ok my bad. :P Then perhaps you should do the removing of `edit_allFieldsSpecified_success()` in PR #395 instead of this one.

Zhiyuan-Amos:
@yamgent I forgot that I am removing `edit_allFieldsSpecified_success()` in PR #395. First time managing so many PRs ><

CanIHasReview:
# v2

* [v2 1/3] [Change PersonBuilder’s Constructor](https://github.com/se-edu/addressbook-level4/pull/393/commits/84141c049fd42728d95065e26a9509a56d438081)
* [v2 2/3] [Extract Integration Tests for EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/5361a3ad97858cf6b07c5d2416ffa5cba2162699)
* [v2 3/3] [Rename test in EditCommandIntegrationTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/a0264e315d16ed670215eef2306b0e69e05aa46f)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/dc9fe96a38a2b5f9bb5bb4c05c3ff1db47b1effd...a0264e315d16ed670215eef2306b0e69e05aa46f)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/f22b9be5392a5382f1ef5e737a9ad779c56ee00d~1...f22b9be5392a5382f1ef5e737a9ad779c56ee00d))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/393/2/head:BRANCHNAME
```

</details>

damithc:
Same comment as other PR: let's not touch the GUI test. 

Previously we wrote `EditCommandIntegrationTest` to just do integration testing (hence only few test cases in it). However, because we are not going to do 'pure' unit and integration tests (except for `add` command) I think these are the test classes we need for the `edit` command:

1. `EditCommandGuiTest` - To be renamed from current GUI test. Should only test paths relevant to the GUI. i.e. not really a system test or an acceptance test. Only an integration test. system/acceptance tests are not automated yet because GUI tests are costly.
2. `EditCommandParserTest` - Should only test paths relevant to parsing.
3. `EditCommandTest` - Should contain both unit and integration tests for edit command. Uses the real model (no stubbing). 

1. To be addressed by a separate PR, after 2 and 3 are done.
2. Addressed by an ongoing PR.
3. Addressed by this PR. That means need to add all missing unit tests.

What do you guys think? adding @chao1995 because he was involved in this class before.

Zhiyuan-Amos:
Yup, will remove changes in GUI.

CanIHasReview:
# v3

@Zhiyuan-Amos submitted v3 for review.

* [v3 1/5] [Rename EditCommandIntegrationTest to EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/30aaeaa1e14b1b6b87aaa8b196ca3f411bc9f51d)
* [v3 2/5] [Change PersonBuilder’s Constructor](https://github.com/se-edu/addressbook-level4/pull/393/commits/b78cb0b0f34cfd7a1bc2a30089a473d3d7214a3b)
* [v3 3/5] [Create Unit Tests for EditCommand](https://github.com/se-edu/addressbook-level4/pull/393/commits/018183ddbc0901ea1867cd36a265a8a41dfc165d)
* [v3 4/5] [Order tests in EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/626885d95fc2496e3a0a959b4c6ef812237e0c1c)
* [v3 5/5] [Rename test in EditCommandTest](https://github.com/se-edu/addressbook-level4/pull/393/commits/6f12a9eeecf684a19a8449cf0b54101ce1406d61)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/dc9fe96a38a2b5f9bb5bb4c05c3ff1db47b1effd...6f12a9eeecf684a19a8449cf0b54101ce1406d61)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/b870644c9e4574019178ef51d36c45dce93218da~1...b870644c9e4574019178ef51d36c45dce93218da))
<details>

```shell
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/393/3/head:BRANCHNAME
```

</details>

Zhiyuan-Amos:
do i need to rename the branch as well? AFAIK it's not possible to do so without deleting the current branch i.e will have to open a new PR.

damithc:
> do i need to rename the branch as well? AFAIK it's not possible to do so without deleting the current branch i.e will have to open a new PR.

I guess no need. The branch is deleted in the end anyway.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -34,6 +35,30 @@
     private Parser parser = new Parser();
 
     @Test
+    public void execute_notAllFieldsSpecified_success() throws Exception {
```
yamgent: Should we segregate the methods so that it is easier to tell which are unit tests and which are integration tests?

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -77,8 +102,21 @@ private Command prepareCommand(String userInput) {
     /**
      * Executes the given {@code command}, confirms that <br>
      * - the result message matches {@code expectedMessage} <br>
+     * This is used for unit tests.
+     * @see #assertCommandSuccess(Command, String, ReadOnlyAddressBook, List)
```
yamgent: Not sure if adding the `@see` is necessary. Otherwise it should be noted that that command is the integration counterpart.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -23,18 +24,56 @@
...
 
     private static final int ZERO_BASED_INDEX_FIRST_PERSON = 0;
     private static final int ZERO_BASED_INDEX_SECOND_PERSON =  ZERO_BASED_INDEX_FIRST_PERSON + 1;
 
     private Model model = new ModelManager(new TypicalPersons().getTypicalAddressBook(), new UserPrefs());
     private Parser parser = new Parser();
 
+    /*
+     * Unit Tests
+     */
```
damithc: In the context of this class, what's the difference between a unit test and an integration test? My initial impression is each test is a unit+integration test.

Zhiyuan-Amos: That was my initial idea too actually, but I thought you wanted me to separate them :P

Erm, unit test tests primarily the code path for EditCommand. Whereas integration test tests primarily the code path of EditCommand + Model. The distinction isn't that much though.

yamgent: My current understanding of the code is that the unit tests here do not bother about what happens to the Model at all, while in the integration tests, the Model correctness is also verified.

Zhiyuan-Amos:
@yamgent 

> We can also add unit tests for failure cases as well.

Do you have any examples? ><

yamgent:
> @yamgent
> 
> We can also add unit tests for failure cases as well.
> Do you have any examples? ><

Sorry, I forgot that we are not testing the Parser here, ignore that comment.

chao1995:
@damithc Hi Prof, a few doubts on your previous comment.

> However, because we are not going to do 'pure' unit and integration tests (except for add command) I think these are the test classes we need for the edit command:
> ...
> 3. EditCommandTest - Should contain both unit and integration tests for edit command. Uses the real model (no stubbing).

If we use real model, is it still unit tests? And if we are going to write unit tests for EditCommand, what is the reason that we want to mix unit and integration tests in the same file?

damithc:
> If we use real model, is it still unit tests? And if we are going to write unit tests for EditCommand, what is the reason that we want to write the unit test in the same file as integration test? Since I think having names like EditCommandIntegrationTest and EditCommandUnitTest will be clearer.

@chao1995 
No it will not be strictly unit tests or integration tests. They are a mixture.
As we saw from unit tests for add command ( #273 ), pure unit testing requires stubbing which requires a lot of extra work. So, we decided to do it the 'pure' way only for one command (i.e. add command) and use the hybrid approach for other commands. I missed this point when you wrote the EditCommandIntegrationTest (which was a pure integration test). Now we are morphing it into contain both unit+integration test (i.e. this PR).

yamgent:
I find it quite hard to define what it means to have an "unpure" unit test actually. It seems that when we use the real Model, we might as well verify the Model correctness as well. So for example, `execute_notAllFieldsSpecified_success()` can also check that the person have been updated in the Model correctly. Don't really see a reason not to do so otherwise.

chao1995:
> I find it quite hard to define what it means to have an "unpure" unit test actually. It seems that when we use the real Model, we might as well verify the Model correctness as well. So for example, execute_notAllFieldsSpecified_success() can also check that the person have been updated in the Model correctly.

I think "unpure" refers to that we are doing unit tests while **using the real Model** instead of a stub. However, the unit tests won't involve checks on the real Model because by right that's a stub Model.

> Don't really see a reason not to do so otherwise.

I think the reason not to do so is that we will check the correctness of the Model in the integration test.

yamgent:
> However, the unit tests won't involve checks on the real Model because by right that's a stub Model.

Although that way we won't know whether the command actually called the correct methods in Model without the stub. And we don't do that in integration either, we just observe that the data in Model are correct and assumed that the correct methods are called. So I guess we are doing an "incomplete" unit test?

chao1995:
> Although that way we won't know whether the command actually called the correct methods in Model without the stub. And we don't do that in integration either, we just observe that the data in Model are correct and assumed that the correct methods are called. So I guess we are doing an "incomplete" unit test?

Good point. According to http://stackoverflow.com/a/17810004/3522482, checking whether the correct methods are called (behaviour verification) sounds more like what a Mock should do. On the other hand, a stub is usually used to do state verification, e.g. checking the Model state.

I think I was wrong saying that "the unit tests won't involve checks on the real Model because by right that's a stub Model" since unit tests do need to verify some states. But if we do check the state of the Model in unit tests, it becomes the integration test.

My conclusion is that we don't do unit tests on `execute()` command at all. So the unit test is "incomplete", but I guess that's what we want. Since for the mixture, it's impossible to make the unit tests complete because that means using both a stub/mock Model and a real Model at the same time. If we really want to complete it, we have to separate unit tests from integration tests.

chao1995:
> Although that way we won't know whether the command actually called the correct methods in Model without the stub. And we don't do that in integration either, we just observe that the data in Model are correct and assumed that the correct methods are called. So I guess we are doing an "incomplete" unit test?

Good point. According to http://stackoverflow.com/a/17810004/3522482, checking whether the correct methods are called (i.e. behaviour verification) sounds more like what a Mock should do. On the other hand, a stub is usually used to do state verification, e.g. checking the Model state.

I think I was wrong saying that "the unit tests won't involve checks on the real Model because by right that's a stub Model" since unit tests do need to verify some states. But if we do check the state of the Model in unit tests, it becomes an integration test since the Model is real.

And for the mixture, it's impossible to make the unit tests complete because that means using both a stub/mock Model and a real Model at the same time. If we really want to complete it, we have to separate unit tests from integration tests.

My conclusion is that we don't do unit tests on `execute()` command at all. So the unit test is "incomplete", but I guess that's what we can do now.

In the future, if we use Mockito, writing a mock Model doesn't require any new code, then it's quite easy to separate unit and integration tests.

Zhiyuan-Amos:
@damithc shall I try playing around with mockito?

damithc:
> I think "unpure" refers to that we are doing unit tests while using the real Model instead of a stub.

That's correct.

> @damithc shall I try playing around with mockito?

No need for now.  While mocking is the better way for unit testing, it is not covered in level 4.

-------
Let's say Foo depends on Bar and we are testing Foo.

Pure unit testing:
* Stub out Bar but ensure that Foo calls Bar correctly by intercepting/recording method calls received by the stub. We do that in #273 but it's a lot of extra work. The better way to do it is using mocks but we don't want to use mocks here because mocking is not covered in level 4 topics. 

Pure integration testing:
* Ensure Foo still works when it is connected up with the real Bar. One test case for each different way Foo interacts with Bar. This should be a sub-set of unit test cases because multiple unit test cases can have the same interaction with Bar but differ in what happens inside the Foo. 

Hybrid (unit + integration) testing:
* We use real Bar instead of stubs or mocks. Test cases will be similar to those used in pure unit tests. However, because we cannot intercept/record calls from Foo to Bar (like we did in pure unit tests), we need to indirectly verify the interaction was correct by observing what Bar did. That is, we assume Foo did the right thing if it caused Bar to do the right thing. This means testing of Foo now depends on the correctness of Bar. This is why hybrid testing is not as good as pure testing.
So yes, I think all tests in EditCommandTest should check the model too. It's not enough to check the result message alone.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -53,6 +58,77 @@ public void execute_validCommand_succeeds() throws Exception {
...
+        expectedAddressBook.updatePerson(ZERO_BASED_INDEX_FIRST_PERSON, editedPerson);
+
+        List<ReadOnlyPerson> expectedFilteredList = new ArrayList<>(expectedAddressBook.getPersonList());
+        String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
+
+        assertCommandSuccess(editCommand, expectedMessage, expectedAddressBook, expectedFilteredList);
+
+        // edit filtered list where index is larger than size of filtered list,
+        // but smaller than size of address book -> failure
+        updatefilteredListShowOnlyFirstPersonInAddressBook();
```
yamgent: I feel that these test cases are better well served if they are each like more than 5 lines long. Currently it is very difficult to keep track of what is going on, especially when the `editedPerson` variable keeps getting reused.

yamgent: If the initialization of the address book is done multiple times, consider creating a `setUp()` method that creates special `addressbooks` that are properly set up to what we want already. That way we don't have to repeatedly reset the addressbook before each test case (that would require splitting up the test cases into their own individual methods).

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
```diff
@@ -22,6 +23,16 @@ public EditPersonDescriptorBuilder(EditPersonDescriptor descriptor) {
...
+    public EditPersonDescriptorBuilder(ReadOnlyPerson person) throws IllegalValueException {
+        this();
+        descriptor.setName(Optional.of(person.getName()));
+        descriptor.setPhone(Optional.of(person.getPhone()));
+        descriptor.setEmail(Optional.of(person.getEmail()));
+        descriptor.setAddress(Optional.of(person.getAddress()));
+        descriptor.setTags(Optional.of(person.getTags()));
+    }
+
+
```
limmlingg: Extra line here

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -56,6 +60,38 @@ public void execute_validCommand_succeeds() throws Exception {
     }
 
     @Test
+    public void execute_someFieldsSpecified_success() throws Exception {
+        ReadOnlyAddressBook addressBook = model.getAddressBook();
+        PersonBuilder personInList = new PersonBuilder(addressBook.getPersonList().get(ZERO_BASED_INDEX_FIRST_PERSON));
+        Person editedPerson = personInList.withName(VALID_NAME_BOB).withPhone(VALID_PHONE_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+
+        EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(editedPerson).build();
```
yamgent: This is still modifying all the fields though.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -138,6 +183,12 @@ private EditCommand prepareCommand(int oneBasedIndex, EditPersonDescriptor descr
         return editCommand;
     }
 
+    private void updateFilteredListShowOnlyFirstPersonInAddressBook() {
```
yamgent: `showFirstPersonOnly()`

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
```diff
@@ -22,6 +23,18 @@ public EditPersonDescriptorBuilder(EditPersonDescriptor descriptor) {
         this.descriptor = new EditPersonDescriptor(descriptor);
     }
 
+    /**
+     * Returns an {@code EditPersonDescriptor} with fields containing {@code person}'s details
+     */
+    public EditPersonDescriptorBuilder(ReadOnlyPerson person) throws IllegalValueException {
```
yamgent: instead of a constructor, perhaps this should be a method `withPerson(ReadOnlyPerson person)`?

Then your commit message should mention that it belongs to `EditPersonDescriptorBuilder` because `EditPersonDescriptorBuilder` is a helper class for all our `EditPersonDescriptor` needs.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -42,31 +42,102 @@
     private Model model = new ModelManager(new TypicalPersons().getTypicalAddressBook(), new UserPrefs());
 
     @Test
-    public void execute_validCommand_succeeds() throws Exception {
+    public void execute_allFieldsSpecified_succeeds() throws Exception {
```
limmlingg: Should be `success`?

damithc:
I guess last commit will have to change significantly after #471 is merged?

Zhiyuan-Amos:
@damithc @limmlingg @MightyCupcakes ready for review! (CanIHasReview isn't working)

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -52,6 +56,78 @@ public void execute_validCommand_succeeds() throws Exception {
...
+
+        String expectedMessage = String.format(EditCommand.MESSAGE_EDIT_PERSON_SUCCESS, editedPerson);
+
+        Model expectedModel = new ModelManager(new AddressBook(model.getAddressBook()), new UserPrefs());
+
+        assertCommandSuccess(editCommand, expectedMessage, expectedModel);
+    }
+
+    @Test
+    public void execute_listIsFilteredOutOfBounds_failure() throws Exception {
```
damithc: `listIsFilteredOutOfBounds` is hard to decipher. `editPersonNotInFilteredList`? Same issue with the next two test cases.

Zhiyuan-Amos:
@damithc made the changes already!

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -52,6 +56,78 @@ public void execute_validCommand_succeeds() throws Exception {
...
+        Index outOfBoundIndex = Index.fromZeroBased(model.getFilteredPersonList().size());
+        assertTrue(outOfBoundIndex.getZeroBased() <= model.getAddressBook().getPersonList().size());
+
+        EditCommand editCommand = prepareCommand(outOfBoundIndex,
+                new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB).build());
+        assertCommandFailure(editCommand, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+    }
+
+    @Test
+    public void execute_editPersonInFilteredList_failure() throws Exception {
```
damithc: `_editCausesDuplicate_` is more informative?

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -52,6 +56,78 @@ public void execute_validCommand_succeeds() throws Exception {
...
+
+        // edit person in filtered list into a duplicate in address book
+        ReadOnlyPerson personInList = model.getAddressBook().getPersonList().get(INDEX_SECOND_PERSON.getZeroBased());
+        EditCommand editCommand = prepareCommand(INDEX_FIRST_PERSON,
+                new EditPersonDescriptorBuilder(personInList).build());
+        assertCommandFailure(editCommand, EditCommand.MESSAGE_DUPLICATE_PERSON);
+    }
+
+    @Test
+    public void execute_editPersonInFilteredList_success() throws Exception {
```
damithc: What does this case test that is not tested by the first three cases?

damithc: Some interesting boundary cases: edit the first/last person in the list (but can be covered by the first 3 cases)

src/test/java/seedu/address/logic/commands/EditCommandTest.java
Zhiyuan-Amos: that's because there's a test below `execute_duplicatePerson_throwsCommandException`, as such I want to differentiate between the test for a filtered list vs an unfiltered list. 

Perhaps naming it to be `..._duplicatePersonInFilteredList_...` would be better? then for `..._duplicatePerson_...` i can change it to `..._duplicatePersonInUnfilteredList_...`

damithc: Oh I didn't realize unfiltered list is an equivalence class. Yes, it is better to mention both `filteredList` and `unfilteredList` in the 2nd part of the name. Or else, you can omit 'list type' from the test name but in that case test both filtered list and unfiltered list variations inside the test case.

src/test/java/seedu/address/logic/commands/EditCommandTest.java
Zhiyuan-Amos: Wang Leng suggested that this can be merged into the first 3 cases, and together with your suggestion, perhaps we can do this:
1st test case: Test for 1st person, unfiltered list
2nd test case: Test for last person, unfiltered list.
3rd test case: Test for 1st person, filtered list

src/test/java/seedu/address/logic/commands/EditCommandTest.java
Zhiyuan-Amos: Yup! Guess my naming could have been improved ><

Zhiyuan-Amos:
@damithc updated!

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -52,22 +56,96 @@ public void execute_validCommand_succeeds() throws Exception {
...
+        Index outOfBoundIndex = Index.fromOneBased(model.getFilteredPersonList().size() + 1);
         EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder().withName(VALID_NAME_BOB).build();
         EditCommand editCommand = prepareCommand(outOfBoundIndex, descriptor);
         assertCommandFailure(editCommand, Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
     }
 
     @Test
+    public void execute_invalidPersonIndexFilteredList_failure() throws Exception {
+        // edit filtered list where index is larger than size of filtered list,
+        // but smaller than size of address book
```
damithc: This comment applies to the next three lines only? If you are describing the entire test case, may be give as a header comment?

Zhiyuan-Amos: Noted! Do i put it below the @Test or before it?

damithc: Not sure myself. :-p

Zhiyuan-Amos: https://stackoverflow.com/questions/3109950/codestyle-put-javadoc-before-or-after-annotation 

I will do it this way then:
```
// header comments
@Test
public void...
```

Zhiyuan-Amos: for your review prof :>

damithc: Sure. Use `/** ... */` though

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -42,37 +42,111 @@
...
     }
 
+    /**
+     * Edit filtered list where index is larger than size of filtered list,
+     * but smaller than size of address book
+     */
+    @Test
+    public void execute_invalidPersonIndexFilteredList_failure() throws Exception {
+        showFirstPersonOnly();
+        Index outOfBoundIndex = Index.fromZeroBased(model.getFilteredPersonList().size());
```
yamgent: `fromOneBased`?

src/test/java/seedu/address/logic/commands/EditCommandTest.java
```diff
@@ -42,37 +42,111 @@
...
+        assertTrue(outOfBoundIndex.getZeroBased() <= model.getAddressBook().getPersonList().size());
```
yamgent: `<`?

Zhiyuan-Amos:
@yamgent done, for your review :>

PR #398, opened 2017-05-04
Hack out Undo / Redo commands

Zhiyuan-Amos:
This is done in a hackish manner, and the rough architecture is laid out here.

The undoing of Delete, Clear and Edit commands are the same (I just stored the previous AddressBook state) because properly implementing it requires much more editing of Model / ModelManager / AddressBook which will get quite messy, and is unnecessary now since we are just looking through the architecture.

The main highlights are:

1. Each `ReversibleCommand` stores the values that it requires to undo itself. For example, when properly implemented, the `DeleteCommand` will store `deletedPerson` and `indexOfDeletedPerson`.

2. Parser now has a `History` object (referencing the `HistoryManager` in `LogicManager`), and Parser passes this `History` object to `UndoCommandParser` and `RedoCommandParser`. This may seem kinda weird, but my opinion is that this is the best way to do it. Other possible ways include:

- Remove entirely `UndoCommandParser` and `RedoCommandParser` since they don't really do much. We can put the logic of these parsers into the commands' constructor.

- Let `LogicManager` do the logic of `UndoCommandParser` and `RedoCommandParser`; after the line `Command command = parser.parse..` in `LogicManager`, then `LogicManager` checks if the command is `undo / redo`. If true, `LogicManager` will then execute undo / redo respectively. This means more logic inside `LogicManager`, and it seems like `LogicManager` will handle a lot of low-level things which seems wrong. The plus point however is that Parser has no need to store a reference to the `History` object.

### DIFF
--- a/src/main/java/seedu/address/logic/HistoryManager.java
+++ b/src/main/java/seedu/address/logic/HistoryManager.java
@@ -1,0 +1,37 @@
+package seedu.address.logic;
+
+import java.util.ArrayList;
+import java.util.Optional;
+
+import seedu.address.logic.commands.ReversibleCommand;
+
+public class HistoryManager implements History {
+    private ArrayList<ReversibleCommand> commandHistory;
+    private int pointer; // points to the command to be undone in commandHistory
+
+    public HistoryManager() {
+        commandHistory = new ArrayList<ReversibleCommand>();
+        pointer = -1;
+    }
+
+    public void push(ReversibleCommand e) {
+        // if after undoing a few commands, you did not redo all of them, and now you are executing
+        // new command
+        if (pointer != commandHistory.size() - 1) {
+            commandHistory.subList(pointer + 1, commandHistory.size()).clear();
+        }
+
+        commandHistory.add(e);
+        pointer = commandHistory.size() - 1;
+    }
+
+    public Optional<ReversibleCommand> getPreviousCommand() {
+        return pointer >= 0 ? Optional.of(commandHistory.get(pointer--)) :
+            Optional.empty();
+    }
+
+    public Optional<ReversibleCommand> getNextCommand() {
+        return pointer < commandHistory.size() - 1 ? Optional.of(commandHistory.get(++pointer)) :
+            Optional.empty();
+    }
+}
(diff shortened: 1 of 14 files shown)


src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,37 @@
+package seedu.address.logic;
+
+import java.util.ArrayList;
+import java.util.Optional;
+
+import seedu.address.logic.commands.ReversibleCommand;
+
+public class HistoryManager implements History {
+    private ArrayList<ReversibleCommand> commandHistory;
```
yamgent: Possible to implement the logic with two stacks instead? One 'undo' stack and one 'redo' stack.

src/main/java/seedu/address/logic/parser/Parser.java
```diff
@@ -27,6 +30,11 @@
      * Used for initial separation of command word and args.
      */
     private static final Pattern BASIC_COMMAND_FORMAT = Pattern.compile("(?<commandWord>\\S+)(?<arguments>.*)");
+    private History history;
+
+    public Parser(History history) {
```
yamgent: Why does the parser need to know about the history?

src/main/java/seedu/address/logic/parser/RedoCommandParser.java
```diff
@@ -0,0 +1,27 @@
...
+    public RedoCommandParser(History history) {
+        this.history = history;
+    }
+
+    public Command parse() {
+        Optional<ReversibleCommand> command = history.getNextCommand();
+        if (command.isPresent()) {
+            return new RedoCommand(command.get());
+        } else {
+            return new IncorrectCommand(RedoCommand.MESSAGE_FAILURE);
```
yamgent: These logic should be inside `RedoCommand` rather than be in the parser.

src/main/java/seedu/address/logic/parser/UndoCommandParser.java
```diff
@@ -0,0 +1,27 @@
...
+    public UndoCommandParser(History history) {
+        this.history = history;
+    }
+
+    public Command parse() {
+        Optional<ReversibleCommand> command = history.getPreviousCommand();
+        if (command.isPresent()) {
+            return new UndoCommand(command.get());
+        } else {
+            return new IncorrectCommand(UndoCommand.MESSAGE_FAILURE);
```
yamgent: These logic should be inside `UndoCommand` rather than be in the parser.

Zhiyuan-Amos:
@yamgent we will discuss this with prof on Monday :)

PR #422, opened 2017-05-18
Implement undo and redo functionalities #390

### LINKED ISSUE #390: User should be allowed to Undo and Redo certain commands
in the event where the user may have accidentally entered an undesirable command, in which user is able to reverse the command easily.

I'm not sure whether this is the right place to discuss the implementation of Undo and Redo.
Reference: [Command Pattern](http://gameprogrammingpatterns.com/command.html) (refer to the section titled Undo and Redo)

Creating an abstract class `ReversibleCommand` which extends `Command`. `ReversibleCommand` contains an abstract method `undo` and a default method `redo` (which basically calls `execute`). Only commands that changes the model will inherit from `ReversibleCommand` i.e Add, Edit, Delete, Clear. 

From the reference above, we can implement a `historyArray` in `LogicManager`, which upon every command execution, pushes the command into the `historyArray`. Each command will now have to store slightly more information: 

1. Upon executing `AddCommand`, `AddCommand` also has to store the index of person added on the list. When executing undo, execute `DeleteCommand` at the index.

2. Upon executing `DeleteCommand`, store the details of the person being deleted. When executing undo, execute the equivalent `AddCommand`. However, over here we have to support adding at a particular index i.e we cannot solely execute `AddCommand` (new MoveCommand?). It's not necessary to have to execute `AddCommand` as well, since I think we can just manipulate the model to add the person at a particular index.

3. Upon executing `ClearCommand`, store the entire model. When executing undo, restore the entire model.

4. Upon executing `EditCommand`, store the original information of the person being edited. When executing undo, execute `EditCommand` using the original information.

Then, we have an index to point to the `historyArray` so we know which command to undo / redo next. An alternative implementation is to use `undoStack` and `redoStack`.

Zhiyuan-Amos:
Fixes #390 WIP, tests are not done yet

### DIFF
--- a/src/main/java/seedu/address/logic/HistoryManager.java
+++ b/src/main/java/seedu/address/logic/HistoryManager.java
@@ -1,0 +1,71 @@
+package seedu.address.logic;
+
+import java.util.Stack;
+
+import seedu.address.logic.commands.Command;
+import seedu.address.logic.commands.RedoCommand;
+import seedu.address.logic.commands.ReversibleCommand;
+import seedu.address.logic.commands.UndoCommand;
+import seedu.address.logic.commands.exceptions.CommandException;
+
+/**
+ * Stores the history of commands executed.
+ */
+public class HistoryManager implements History {
+    private Stack<Command> undoStack;
+    private Stack<Command> redoStack;
+
+    public HistoryManager() {
+        undoStack = new Stack<>();
+        redoStack = new Stack<>();
+    }
+
+    /**
+     * All commands in {@code redoStack} will be removed upon method call.
+     */
+    @Override
+    public void addNewExecutedCommand(Command command) {
+        if (command instanceof UndoCommand || command instanceof RedoCommand) {
+            return;
+        }
+
+        if (!redoStack.empty()) {
+            redoStack.clear();
+        }
+
+        undoStack.push(command);
+    }
+
+    @Override
+    public void undoPreviousReversibleCommand() throws CommandException {
+        Command toUndo;
+        do {
+            if (undoStack.empty()) {
+                throw new CommandException(UndoCommand.MESSAGE_FAILURE);
+            }
+            toUndo = undoStack.pop();
+            redoStack.push(toUndo);
+        } while (!(toUndo instanceof ReversibleCommand));
+
+        ((ReversibleCommand) toUndo).rollback();
+    }
+
+    @Override
+    public void executeNextReversibleCommand() throws CommandException {
+        Command toRedo;
+        do {
+            if (redoStack.empty()) {
+                throw new CommandException(RedoCommand.MESSAGE_FAILURE);
+            }
+            toRedo = redoStack.pop();
+            undoStack.push(toRedo);
+        } while (!(toRedo instanceof ReversibleCommand));
+
+        try {
+            toRedo.execute();
+        } catch (CommandException ce) {
+            throw new AssertionError("The command has been successfully executed previously; " +
+                    "it should not fail now");
+        }
+    }
+}
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -21,8 +21,10 @@ public class LogicManager extends ComponentManager implements Logic {
 
     private final Model model;
     private final Parser parser;
+    private final History history;
 
     public LogicManager(Model model, Storage storage) {
+        history = new HistoryManager();
         this.model = model;
         this.parser = new Parser();
     }
@@ -31,8 +33,11 @@ public class LogicManager extends ComponentManager implements Logic {
     public CommandResult execute(String commandText) throws CommandException {
         logger.info("----------------[USER COMMAND][" + commandText + "]");
         Command command = parser.parseCommand(commandText);
-        command.setData(model);
-        return command.execute();
+        command.setData(model, history);
+        CommandResult result = command.execute();
+        history.addNewExecutedCommand(command);
+
+        return result;
     }
 
     @Override
--- a/src/main/java/seedu/address/logic/commands/DeleteCommand.java
+++ b/src/main/java/seedu/address/logic/commands/DeleteCommand.java
@@ -4,13 +4,14 @@ import seedu.address.commons.core.Messages;
 import seedu.address.commons.core.UnmodifiableObservableList;
 import seedu.address.commons.util.IndexUtil;
 import seedu.address.logic.commands.exceptions.CommandException;
+import seedu.address.model.AddressBook;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.person.UniquePersonList.PersonNotFoundException;
 
 /**
  * Deletes a person identified using it's last displayed index from the address book.
  */
-public class DeleteCommand extends Command {
+public class DeleteCommand extends ReversibleCommand {
 
     public static final String COMMAND_WORD = "delete";
 
@@ -40,6 +41,7 @@ public class DeleteCommand extends Command {
         ReadOnlyPerson personToDelete = lastShownList.get(IndexUtil.oneToZeroIndex(targetIndex));
 
         try {
+            setData(new AddressBook(model.getAddressBook()));
             model.deletePerson(personToDelete);
         } catch (PersonNotFoundException pnfe) {
             assert false : "The target person cannot be missing";
--- a/src/main/java/seedu/address/logic/commands/RedoCommand.java
+++ b/src/main/java/seedu/address/logic/commands/RedoCommand.java
@@ -1,0 +1,25 @@
+package seedu.address.logic.commands;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import seedu.address.logic.commands.exceptions.CommandException;
+
+/**
+ * Redo the previously undone command.
+ */
+public class RedoCommand extends Command {
+
+    public static final String COMMAND_WORD = "redo";
+    public static final String MESSAGE_SUCCESS = "Redo success!";
+    public static final String MESSAGE_FAILURE = "There's no more commands left to be redone!";
+
+    @Override
+    public CommandResult execute() throws CommandException {
+        checkNotNull(model);
+        checkNotNull(history);
+
+        history.executeNextReversibleCommand();
+
+        return new CommandResult(MESSAGE_SUCCESS);
+    }
+}
--- a/src/test/java/seedu/address/logic/HistoryManagerTest.java
+++ b/src/test/java/seedu/address/logic/HistoryManagerTest.java
@@ -1,0 +1,199 @@
+package seedu.address.logic;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import javafx.collections.transformation.FilteredList;
+import seedu.address.logic.commands.AddCommand;
+import seedu.address.logic.commands.ClearCommand;
+import seedu.address.logic.commands.Command;
+import seedu.address.logic.commands.ListCommand;
+import seedu.address.logic.commands.RedoCommand;
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.logic.commands.UndoCommand;
+import seedu.address.logic.commands.exceptions.CommandException;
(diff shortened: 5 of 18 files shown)


src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+    private Stack<Command> toUndo;
+    private Stack<Command> toRedo;
+
+    public HistoryManager() {
+        toUndo = new Stack<Command>();
+        toRedo = new Stack<Command>();
+    }
+
+    @Override
+    public void push(Command command) {
```
yamgent: `addNewExecutedCommand()`?

src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Stack;
+
+import seedu.address.logic.commands.Command;
+import seedu.address.logic.commands.ReversibleCommand;
+
+/**
+ * Stores the history of commands executed.
+ */
+public class HistoryManager implements History {
+    private Stack<Command> toUndo;
```
yamgent: Rename this to `undoStack` and `redoStack`?

src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public HistoryManager() {
+        toUndo = new Stack<Command>();
+        toRedo = new Stack<Command>();
+    }
+
+    @Override
+    public void push(Command command) {
+        // if after undoing a few commands, you did not redo all of them, and now you are executing
+        // new command
```
yamgent: This should be in the header comments. Something like `All commands that can be redo will be removed upon execution of a new command`.

src/main/java/seedu/address/logic/LogicManager.java
```diff
@@ -31,8 +35,13 @@ public LogicManager(Model model, Storage storage) {
     public CommandResult execute(String commandText) throws CommandException {
         logger.info("----------------[USER COMMAND][" + commandText + "]");
         Command command = parser.parseCommand(commandText);
-        command.setData(model);
-        return command.execute();
+        command.setData(model, history);
+        CommandResult result = command.execute();
+        if (!(command instanceof UndoCommand || command instanceof RedoCommand)) {
```
yamgent: Only check whether it is an instance of `ReversibleCommand`? All commands that returns false for such a check will not be added to the HistoryManager. That way we do not have to check the specific type here.

src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        toUndo.push(command);
+    }
+
+    @Override
+    public Optional<ReversibleCommand> getPreviousReversibleCommand() {
+        Command toUndo;
+        do {
+            if (this.toUndo.empty()) {
```
yamgent: ~~this.~~ toUndo. Avoid using the `this` keyword unless absolutely necessary (usually only used in getters/setters/constructors for variable assignment).

src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+                return Optional.empty();
+            }
+            toUndo = this.toUndo.pop();
+            this.toRedo.push(toUndo);
```
yamgent: The side effect seems surprising especially when the method name seems to imply that it is an accessor method. It also seems to be doing the job "halfway" (updates the stack but not actually reversing the command).

It should either be just purely a getter method, or it should do all the way (actually reverse the command here).

src/main/java/seedu/address/logic/commands/DeleteCommand.java
```diff
@@ -40,6 +41,7 @@ public CommandResult execute() throws CommandException {
         ReadOnlyPerson personToDelete = lastShownList.get(IndexUtil.oneToZeroIndex(targetIndex));
 
         try {
+            previousAddressBook = new AddressBook(model.getAddressBook());
```
yamgent: This line of code ideally should be moved to a method in the `ReversibleCommand`, so that we do not have to copy the same line over multiple classes.

src/main/java/seedu/address/logic/LogicManager.java
yamgent: Also what happens if the command fails? Does it still get pushed into the stack?

src/main/java/seedu/address/logic/commands/RedoCommand.java
```diff
@@ -0,0 +1,36 @@
...
+ * Redo the previously undone command.
+ */
+public class RedoCommand extends Command {
+
+    public static final String COMMAND_WORD = "redo";
+    public static final String MESSAGE_SUCCESS = "Redo success!";
+    public static final String MESSAGE_FAILURE = "There's no more commands left to be redone!";
+
+    @Override
+    public CommandResult execute() throws CommandException {
```
Zhiyuan-Amos: i don't think Exception will be thrown here

src/test/java/seedu/address/logic/parser/ParserTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+public class ParserTest {
+    Parser parser = new Parser();
+
+    @Test
+    public void parseCommand_undoCommandWord_returnsUndoCommand() {
+        assertTrue(parser.parseCommand(UndoCommand.COMMAND_WORD) instanceof UndoCommand);
+        assertTrue(parser.parseCommand("  " + UndoCommand.COMMAND_WORD + " ") instanceof UndoCommand);
+        assertFalse(parser.parseCommand("Undo") instanceof UndoCommand);
+        assertFalse(parser.parseCommand("undo 3") instanceof UndoCommand);
```
Zhiyuan-Amos: this is a true actually

src/test/java/seedu/address/logic/parser/ParserTest.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @Test
+    public void parseCommand_redoCommandWord_returnsRedoCommand() {
+        assertTrue(parser.parseCommand(RedoCommand.COMMAND_WORD) instanceof RedoCommand);
+        assertTrue(parser.parseCommand("  " + RedoCommand.COMMAND_WORD + " ") instanceof RedoCommand);
+        assertFalse(parser.parseCommand("Redo") instanceof RedoCommand);
+        assertFalse(parser.parseCommand("redo 3") instanceof RedoCommand);
```
Zhiyuan-Amos: this is a true actually

src/main/java/seedu/address/logic/commands/RedoCommand.java
```diff
@@ -0,0 +1,36 @@
...
+        checkNotNull(model);
+        checkNotNull(history);
+
+        Optional<ReversibleCommand> nextCommand = history.getNextReversibleCommand();
+        if (!nextCommand.isPresent()) {
+            throw new CommandException(MESSAGE_FAILURE);
+        }
+
+        try {
+            nextCommand.get().execute();
```
yamgent: I was wondering that since we are already doing 'undo' using snapshots of the model to do a rollback, can 'redo' also do the same as well? This would also eliminate the possibility of having commands suddenly failing when trying to redo.

src/main/java/seedu/address/logic/LogicManager.java
Zhiyuan-Amos: It doesn't; command.execute() only returns a `CommandResult` if the execution is successful, otherwise it will throw a `CommandException`.

src/main/java/seedu/address/logic/commands/RedoCommand.java
Zhiyuan-Amos: yup we will do that for consistency for now. In the future PR will revert it back to this so that we don't have to keep an extra variable to store the snapshot.

Also, if commands suddenly fail when trying to redo, throwing an `AssertionError` makes sense because this case isn't supposed to happen and the fault is with the programmer, not the user.

src/main/java/seedu/address/logic/commands/DeleteCommand.java
Zhiyuan-Amos: As discussed, will look into putting this inside `HistoryManager` i.e when the command is pushed into `HistoryManager`, it will store the value for `previousAddressBook`

Zhiyuan-Amos: nope i think can't cos when the command is pushed into `HistoryManager`, the command would have already executed the the address book already updated. :<

src/main/java/seedu/address/logic/commands/RedoCommand.java
Zhiyuan-Amos: Edit: We will not make any changes as spoken :>

src/test/java/seedu/address/logic/HistoryManagerTest.java
```diff
@@ -0,0 +1,199 @@
...
+        assertUndoFailure();
+    }
+
+    @Test
+    public void executeNextReversibleCommand_emptyRedoStack_throwsCommandException() throws Exception {
+        assertRedoFailure();
+    }
+
+    @Test
+    public void undoPreviousReversibleCommandAndExecuteNextReversibleCommand_nonReversibleAndReversibleCommands() throws
```
yamgent: The method name is long, but the issue here is I am not really sure what this test case is supposed to be doing. :P

Perhaps omit the second part, so that it just mention the methods that are being tested?

src/test/java/seedu/address/logic/HistoryManagerTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * Confirms that execution of {@code HistoryManager#undoPreviousReversibleCommand} fails and the error message
+     * equals to {@value UndoCommand#MESSAGE_FAILURE}
+     */
+    private void assertUndoFailure() {
+        try {
+            history.undoPreviousReversibleCommand();
+            fail();
```
yamgent: `fail(String)` allows us to add a message. Mention why the test program should not end up executing at this position.

src/test/java/seedu/address/logic/commands/UndoAndRedoCommandTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private static UndoCommand undoCommand;
+    private static RedoCommand redoCommand;
+
+    @Before
+    public void setUpBeforeEachTest() throws Exception {
+        model = new ModelManager(new TypicalPersons().getTypicalAddressBook(), new UserPrefs());
+        history = new HistoryManager();
+        deleteCommand.setData(model, history);
+        undoCommand.setData(model, history);
+        redoCommand.setData(model, history);
```
yamgent: If the variables are not shared across all methods, perhaps they can just be local?

src/main/java/seedu/address/logic/HistoryManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+        undoStack.push(command);
+    }
+
+    @Override
+    public void undoPreviousReversibleCommand() throws CommandException {
+        Command toUndo;
+        do {
+            if (undoStack.empty()) {
+                throw new CommandException(UndoCommand.MESSAGE_FAILURE);
```
yamgent: HistoryManager should throws its own Exception (e.g. `OutOfRevisibleCommandException`?). `UndoCommand` and `RedoCommand` should be the one throwing the `CommandException`.

src/test/java/seedu/address/logic/HistoryManagerTest.java
```diff
@@ -0,0 +1,199 @@
...
+import seedu.address.testutil.TypicalPersons;
+
+/**
+ * The methods in HistoryManager cannot be tested independently. For HistoryManager#addNewExecutedCommand(Command)
+ * to be tested, both HistoryManager#undoPreviousReversibleCommand and HistoryManager#executeNextReversibleCommand
+ * need to be executed as well. For HistoryManager#executeNextReversibleCommand to be tested,
+ * HistoryManager#undoPreviousReversibleCommand needs to be executed which requires
+ * HistoryManager#addNewExecutedCommand(Command) to be executed.
+ *
+ * As such, when testing for each method, we have to assume that other methods are working as intended. Also, we have
```
yamgent: Maybe we could create a way to examine the history stack?

PR #447, opened 2017-05-25
Create a new Index class #438

### LINKED ISSUE #438: Create a new Index class
The codebase uses a mixture of one-based index and zero-based index, due to the difference in the GUI (one-based) and programming semantics (zero-based).

The conversion between the two indices are done in different location for different commands, making its usage inconsistent.

Let's create a new Index class, which allow us to avoid having to deal with the index type in the first place.

Also the conversion should only be done at the boundary of the component.

index

yamgent:
Fixes #438

```
The usages of zero-based indices and one-based indices are mixed
throughout the codebase. Both uses int, and are only differentiated by
different suffixes for the variable names (e.g int zeroBasedIndex,
int oneBasedIndex).

The index base can be easily confused and is prone to error. The
conversion also takes place at different areas of the codebase and is
not consistent.

Let's create an Index class that understands how to convert to a
both zero-based and one-based index, and update the codebase to use
Index instead of int when handling user inputs.

Index from the point when a new user input is being parsed, so that if
the current component wants to communicate with another component, it
can send an Index to avoid having to know what base the other component
is using for its index. However, after receiving the Index, that
component can convert it back to an int if the index will not be passed
to a different component again.
```

### DIFF
--- a/src/main/java/seedu/address/commons/core/index/Index.java
+++ b/src/main/java/seedu/address/commons/core/index/Index.java
@@ -1,0 +1,54 @@
+package seedu.address.commons.core.index;
+
+/**
+ * Represents a zero-based or one-based index.
+ *
+ * {@code Index} should be used right from the start (when parsing in a new user input), so that if the current
+ * component wants to communicate with another component, it can send an {@code Index} to avoid having to know what
+ * base the other component is using for its index. However, after receiving the {@code Index}, that component can
+ * convert it back to an int if the index will not be passed to a different component again.
+ */
+public class Index {
+    private int zeroBasedIndex;
+
+    /**
+     * Index can only be created by calling {@link Index#fromZeroBased(int)} or
+     * {@link Index#fromOneBased(int)}.
+     */
+    private Index(int zeroBasedIndex) {
+        if (zeroBasedIndex < 0) {
+            throw new IndexOutOfBoundsException();
+        }
+
+        this.zeroBasedIndex = zeroBasedIndex;
+    }
+
+    public int getZeroBased() {
+        return zeroBasedIndex;
+    }
+
+    public int getOneBased() {
+        return zeroBasedIndex + 1;
+    }
+
+    /**
+     * Creates a new {@code Index} using a zero-based index.
+     */
+    public static Index fromZeroBased(int zeroBasedIndex) {
+        return new Index(zeroBasedIndex);
+    }
+
+    /**
+     * Creates a new {@code Index} using a one-based index.
+     */
+    public static Index fromOneBased(int oneBasedIndex) {
+        return new Index(oneBasedIndex - 1);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        return other == this // short circuit if same object
+                || (other instanceof Index // instanceof handles nulls
+                && this.zeroBasedIndex == ((Index) other).zeroBasedIndex); // state check
+    }
+}
--- a/src/main/java/seedu/address/commons/events/ui/JumpToListRequestEvent.java
+++ b/src/main/java/seedu/address/commons/events/ui/JumpToListRequestEvent.java
@@ -1,5 +1,6 @@
 package seedu.address.commons.events.ui;
 
+import seedu.address.commons.core.index.Index;
 import seedu.address.commons.events.BaseEvent;
 
 /**
@@ -9,8 +10,8 @@ public class JumpToListRequestEvent extends BaseEvent {
 
     public final int targetIndex;
 
-    public JumpToListRequestEvent(int targetIndex) {
-        this.targetIndex = targetIndex;
+    public JumpToListRequestEvent(Index targetIndex) {
+        this.targetIndex = targetIndex.getZeroBased();
     }
 
     @Override
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -10,6 +10,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import seedu.address.commons.core.index.Index;
 import seedu.address.commons.exceptions.IllegalValueException;
 import seedu.address.commons.util.StringUtil;
 import seedu.address.model.person.Address;
@@ -26,16 +27,16 @@ public class ParserUtil {
     public static final String MESSAGE_INVALID_INDEX = "Index is not a non-zero unsigned integer.";
 
     /**
-     * Parses {@code index} into an integer and returns it. Leading and trailing whitespaces will be trimmed.
+     * Parses {@code oneBasedIndex} into an {@code Index} and returns it. Leading and trailing whitespaces will be
+     * trimmed.
      * @throws IllegalValueException if the specified index is invalid (not non-zero unsigned integer).
      */
-    public static int parseIndex(String index) throws IllegalValueException {
-        String trimmedIndex = index.trim();
+    public static Index parseIndex(String oneBasedIndex) throws IllegalValueException {
+        String trimmedIndex = oneBasedIndex.trim();
         if (!StringUtil.isNonZeroUnsignedInteger(trimmedIndex)) {
             throw new IllegalValueException(MESSAGE_INVALID_INDEX);
         }
-        return Integer.parseInt(trimmedIndex);
-
+        return Index.fromOneBased(Integer.parseInt(trimmedIndex));
     }
 
     /**
--- a/src/main/java/seedu/address/model/Model.java
+++ b/src/main/java/seedu/address/model/Model.java
@@ -3,6 +3,7 @@ package seedu.address.model;
 import java.util.Set;
 
 import seedu.address.commons.core.UnmodifiableObservableList;
+import seedu.address.commons.core.index.Index;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.person.UniquePersonList;
 import seedu.address.model.person.UniquePersonList.DuplicatePersonException;
@@ -24,14 +25,13 @@ public interface Model {
     void addPerson(ReadOnlyPerson person) throws UniquePersonList.DuplicatePersonException;
 
     /**
-     * Updates the person located at {@code filteredPersonListIndex} with {@code editedPerson}.
+     * Updates the person located at {@code index} with {@code editedPerson}.
      *
      * @throws DuplicatePersonException if updating the person's details causes the person to be equivalent to
      *      another existing person in the list.
-     * @throws IndexOutOfBoundsException if {@code filteredPersonListIndex} < 0 or >= the size of the filtered list.
+     * @throws IndexOutOfBoundsException if {@code index} is out of range of the filtered list.
      */
-    void updatePerson(int filteredPersonListIndex, ReadOnlyPerson editedPerson)
-            throws UniquePersonList.DuplicatePersonException;
+    void updatePerson(Index index, ReadOnlyPerson editedPerson) throws UniquePersonList.DuplicatePersonException;
 
     /** Returns the filtered person list as an {@code UnmodifiableObservableList<ReadOnlyPerson>} */
     UnmodifiableObservableList<ReadOnlyPerson> getFilteredPersonList();
--- a/src/main/java/seedu/address/model/ModelManager.java
+++ b/src/main/java/seedu/address/model/ModelManager.java
@@ -10,6 +10,7 @@ import javafx.collections.transformation.FilteredList;
 import seedu.address.commons.core.ComponentManager;
 import seedu.address.commons.core.LogsCenter;
 import seedu.address.commons.core.UnmodifiableObservableList;
+import seedu.address.commons.core.index.Index;
 import seedu.address.commons.events.model.AddressBookChangedEvent;
 import seedu.address.commons.util.StringUtil;
 import seedu.address.model.person.ReadOnlyPerson;
@@ -73,11 +74,11 @@ public class ModelManager extends ComponentManager implements Model {
     }
 
     @Override
-    public void updatePerson(int filteredPersonListIndex, ReadOnlyPerson editedPerson)
+    public void updatePerson(Index index, ReadOnlyPerson editedPerson)
             throws UniquePersonList.DuplicatePersonException {
         requireNonNull(editedPerson);
 
-        int addressBookIndex = filteredPersons.getSourceIndex(filteredPersonListIndex);
+        int addressBookIndex = filteredPersons.getSourceIndex(index.getZeroBased());
         addressBook.updatePerson(addressBookIndex, editedPerson);
         indicateAddressBookChanged();
     }
--- a/src/test/java/seedu/address/commons/core/index/IndexTest.java
+++ b/src/test/java/seedu/address/commons/core/index/IndexTest.java
@@ -1,0 +1,93 @@
+package seedu.address.commons.core.index;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import org.junit.Test;
+
+public class IndexTest {
+
+    @Test
+    public void createOneBasedIndex() {
+        // invalid index
+        assertCreateOneBasedFailure(0);
+
+        // check equality using the same base
+        assertEquals(1, Index.fromOneBased(1).getOneBased());
+        assertEquals(5, Index.fromOneBased(5).getOneBased());
+
+        // convert from one-based index to zero-based index
+        assertEquals(0, Index.fromOneBased(1).getZeroBased());
+        assertEquals(4, Index.fromOneBased(5).getZeroBased());
+    }
+
+    @Test
+    public void createZeroBasedIndex() {
+        // invalid index
+        assertCreateZeroBasedFailure(-1);
+
+        // check equality using the same base
+        assertEquals(0, Index.fromZeroBased(0).getZeroBased());
+        assertEquals(5, Index.fromZeroBased(5).getZeroBased());
+
+        // convert from zero-based index to one-based index
+        assertEquals(1, Index.fromZeroBased(0).getOneBased());
+        assertEquals(6, Index.fromZeroBased(5).getOneBased());
+    }
+
+    /**
+     * Executes {@code Index#fromZeroBased(int)} with {@code invalidZeroBasedIndex}, confirms that an
+     * {@code IndexOutOfBoundsException} is thrown.
+     */
+    private void assertCreateZeroBasedFailure(int invalidZeroBasedIndex) {
+        assertCreateFailure(invalidZeroBasedIndex, true);
+    }
+
+    /**
+     * Executes {@code Index#fromOneBased(int)} with {@code invalidOneBasedIndex}, confirms that an
+     * {@code IndexOutOfBoundsException} is thrown.
+     */
+    private void assertCreateOneBasedFailure(int invalidOneBasedIndex) {
+        assertCreateFailure(invalidOneBasedIndex, false);
+    }
+
+    /**
+     * Executes either {@code Index#fromZeroBased(int)} (if it is zero based), or {@code Index#fromOneBased(int)}
+     * (if it is one based), and confirms that an {@code IndexOutOfBoundsException} is thrown.
+     */
+    private void assertCreateFailure(int invalidIndex, boolean isZeroBased) {
+        try {
+            if (isZeroBased) {
+                Index.fromZeroBased(invalidIndex);
+            } else {
+                Index.fromOneBased(invalidIndex);
+            }
+            fail("The expected IndexOutOfBoundsException was not thrown.");
+        } catch (IndexOutOfBoundsException ie) {
+            // expected behaviour
+        }
+    }
+
+    @Test
+    public void equals() {
+        final Index fifthPersonIndex = Index.fromOneBased(5);
+
+        // same values -> returns true
+        assertTrue(fifthPersonIndex.equals(Index.fromOneBased(5)));
+        assertTrue(fifthPersonIndex.equals(Index.fromZeroBased(4)));
+
+        // same object -> returns true
+        assertTrue(fifthPersonIndex.equals(fifthPersonIndex));
+
+        // null -> returns false
+        assertFalse(fifthPersonIndex.equals(null));
+
+        // different types -> returns false
+        assertFalse(fifthPersonIndex.equals(5.0f));
+
+        // different index -> returns false
+        assertFalse(fifthPersonIndex.equals(Index.fromOneBased(1)));
+    }
+}
--- a/src/test/java/seedu/address/logic/LogicManagerTest.java
+++ b/src/test/java/seedu/address/logic/LogicManagerTest.java
@@ -10,6 +10,8 @@ import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 import static seedu.address.model.util.SampleDataUtil.getTagSet;
+import static seedu.address.testutil.TypicalPersons.INDEX_SECOND_PERSON;
+import static seedu.address.testutil.TypicalPersons.INDEX_THIRD_PERSON;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -27,6 +29,7 @@ import org.junit.rules.TemporaryFolder;
 import com.google.common.eventbus.Subscribe;
 
 import seedu.address.commons.core.EventsCenter;
+import seedu.address.commons.core.index.Index;
 import seedu.address.commons.events.model.AddressBookChangedEvent;
 import seedu.address.commons.events.ui.JumpToListRequestEvent;
 import seedu.address.commons.events.ui.ShowHelpRequestEvent;
@@ -73,7 +76,7 @@ public class LogicManagerTest {
     //These are for checking the correctness of the events raised
     private ReadOnlyAddressBook latestSavedAddressBook;
     private boolean helpShown;
-    private int targetedJumpIndex;
+    private Index targetedJumpIndex;
 
     @Subscribe
     private void handleLocalModelChangedEvent(AddressBookChangedEvent abce) {
@@ -87,7 +90,7 @@ public class LogicManagerTest {
 
     @Subscribe
     private void handleJumpToListRequestEvent(JumpToListRequestEvent je) {
-        targetedJumpIndex = je.targetIndex;
+        targetedJumpIndex = Index.fromZeroBased(je.targetIndex);
     }
 
     @Before
@@ -103,7 +106,7 @@ public class LogicManagerTest {
 
         latestSavedAddressBook = new AddressBook(model.getAddressBook()); // last saved assumed to be up to date
         helpShown = false;
-        targetedJumpIndex = -1; // non yet
+        targetedJumpIndex = null;
     }
 
     @After
@@ -317,7 +320,7 @@ public class LogicManagerTest {
             model.addPerson(p);
         }
 
-        assertCommandException(commandWord + " 3", expectedMessage);
+        assertCommandException(commandWord + " " + INDEX_THIRD_PERSON.getOneBased(), expectedMessage);
     }
 
     @Test
@@ -341,7 +344,7 @@ public class LogicManagerTest {
 
         assertCommandSuccess(SelectCommand.COMMAND_WORD + " 2",
                 String.format(SelectCommand.MESSAGE_SELECT_PERSON_SUCCESS, 2), expectedModel);
-        assertEquals(1, targetedJumpIndex);
+        assertEquals(INDEX_SECOND_PERSON, targetedJumpIndex);
         assertEquals(model.getFilteredPersonList().get(1), threePersons.get(1));
     }
 
--- a/src/main/java/seedu/address/logic/commands/DeleteCommand.java
+++ b/src/main/java/seedu/address/logic/commands/DeleteCommand.java
@@ -2,7 +2,7 @@ package seedu.address.logic.commands;
 
 import seedu.address.commons.core.Messages;
 import seedu.address.commons.core.UnmodifiableObservableList;
-import seedu.address.commons.util.IndexUtil;
+import seedu.address.commons.core.index.Index;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.person.ReadOnlyPerson;
 import seedu.address.model.person.UniquePersonList.PersonNotFoundException;
@@ -21,9 +21,9 @@ public class DeleteCommand extends Command {
 
     public static final String MESSAGE_DELETE_PERSON_SUCCESS = "Deleted Person: %1$s";
 
-    public final int targetIndex;
+    public final Index targetIndex;
 
-    public DeleteCommand(int targetIndex) {
+    public DeleteCommand(Index targetIndex) {
         this.targetIndex = targetIndex;
     }
 
@@ -33,11 +33,11 @@ public class DeleteCommand extends Command {
 
         UnmodifiableObservableList<ReadOnlyPerson> lastShownList = model.getFilteredPersonList();
 
-        if (lastShownList.size() < targetIndex) {
+        if (targetIndex.getZeroBased() >= lastShownList.size()) {
             throw new CommandException(Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
         }
 
-        ReadOnlyPerson personToDelete = lastShownList.get(IndexUtil.oneToZeroIndex(targetIndex));
+        ReadOnlyPerson personToDelete = lastShownList.get(targetIndex.getZeroBased());
 
         try {
             model.deletePerson(personToDelete);
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -1,7 +1,6 @@
 package seedu.address.logic.commands;
 
 import static java.util.Objects.requireNonNull;
-import static seedu.address.commons.util.AppUtil.checkArgument;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_ADDRESS;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_EMAIL;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
@@ -12,8 +11,8 @@ import java.util.Optional;
 import java.util.Set;
 
 import seedu.address.commons.core.Messages;
+import seedu.address.commons.core.index.Index;
 import seedu.address.commons.util.CollectionUtil;
-import seedu.address.commons.util.IndexUtil;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.model.person.Address;
 import seedu.address.model.person.Email;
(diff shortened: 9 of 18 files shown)


src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -71,11 +72,11 @@ public synchronized void addPerson(ReadOnlyPerson person) throws UniquePersonLis
     }
 
     @Override
-    public void updatePerson(int filteredPersonListIndex, ReadOnlyPerson editedPerson)
+    public void updatePerson(Index index, ReadOnlyPerson editedPerson)
```
damithc: Shouldn't this be `ZeroBasedIndex`?

yamgent: Yes it should be `ZeroBasedIndex`. I just realize that I didn't use them at all. I will also modify the other classes like `UniquePersonList`.

CanIHasReview:
# v1

* [v1 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/21dde07f9869173debb7dd3a9523b3c655994cd7)
* [v1 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/417cab4975ce559a4071c3c0e504e5a6a44e464e)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...417cab4975ce559a4071c3c0e504e5a6a44e464e))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/1/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/model/ModelManager.java
yamgent: Per #438 the interfaces no longer exist.

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
...
+ * Allows the index to be zero-based or one-based.
+ */
+public class Index {
+    private int zeroBasedIndex;
+
+    private Index(int zeroBasedIndex) {
+        this.zeroBasedIndex = zeroBasedIndex;
+    }
+
+    /**
```
Zhiyuan-Amos: header comment for this method and the following method seems unnecessary because the method name seems clear imo

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
...
+     * Returns the index that starts from 0.
+     */
+    public int getZeroBasedIndex() {
+        return this.zeroBasedIndex;
```
Zhiyuan-Amos: unnecessary `this`, same for below.

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
+package seedu.address.commons.core.index;
+
+/**
+ * Allows the index to be zero-based or one-based.
+ */
+public class Index {
+    private int zeroBasedIndex;
+
+    private Index(int zeroBasedIndex) {
```
Zhiyuan-Amos: perhaps can give a short comment here on why is this a private constructor. `this class can only be instantiated by calling Index#createFromZeroBased and Index#createFromOneBased`.

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
+package seedu.address.commons.core.index;
+
+/**
+ * Allows the index to be zero-based or one-based.
```
Zhiyuan-Amos: Phrasing seems weird. Perhaps can change it to something like "Stores an index and allows it to be represented in zero-based or one-based"

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
...
+    private int zeroBasedIndex;
+
+    private Index(int zeroBasedIndex) {
+        this.zeroBasedIndex = zeroBasedIndex;
+    }
+
+    /**
+     * Returns the index that starts from 0.
+     */
+    public int getZeroBasedIndex() {
```
Zhiyuan-Amos: as discussed, `index.getZeroBasedIndex()` seems pretty verbose. We can perhaps do it better by having a `public static final int zeroBased`, so that we can get the value by simply calling `index.zeroBased`.

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Returns the index that starts from 1.
+     */
+    public int getOneBasedIndex() {
+        return this.zeroBasedIndex + 1;
+    }
+
+    /**
+     * Creates a new {@code Index} if the parameter passed in is zero-based.
```
Zhiyuan-Amos: `...if the parameter passed in is zero-based.` sounds weird. Perhaps something like `... {@code Index} with {@code this.zeroBased} = {@code zeroBased}`. Same for below.

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public static Index createFromZeroBased(int zeroBased) {
```
Zhiyuan-Amos: maybe `createZeroBasedIndex` sounds better

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class IndexTest {
+    @Test
+    public void createOneBasedIndex() {
+        // get back the same index type
+        assertEquals(1, Index.createFromOneBased(1).getOneBasedIndex());
+        assertEquals(5, Index.createFromOneBased(5).getOneBasedIndex());
```
Zhiyuan-Amos: maybe should have 1 negative value test and 1 positive value test. 

Edit: For `Index`, the constructor should perhaps ensure that the values given are not invalid values i.e when calling `createFromZeroBased`, the input cannot be smaller than 0.

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class IndexTest {
+    @Test
+    public void createOneBasedIndex() {
+        // get back the same index type
```
Zhiyuan-Amos: the comment a bit weird >< Maybe can say something like `// Check equality using Index#getOneBasedIndex`

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,52 @@
...
+        assertEquals(5, Index.createFromZeroBased(5).getZeroBasedIndex());
+
+        // convert to one-based index
+        assertEquals(1, Index.createFromZeroBased(0).getOneBasedIndex());
+        assertEquals(6, Index.createFromZeroBased(5).getOneBasedIndex());
+    }
+
+    @Test
+    public void equals() {
+        final Index indexFifthPerson = Index.createFromOneBased(5);
```
Zhiyuan-Amos: fifthPersonIndex sounds a little better. idk

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+        // same values -> returns true
+        Index indexWithSameValues = Index.createFromOneBased(5);
```
Zhiyuan-Amos: fifthPersonIndexCopy sounds better.

src/main/java/seedu/address/commons/core/index/Index.java
pyokagan: Or `Represents a zero-based or one-based index.`

src/main/java/seedu/address/commons/core/index/Index.java
pyokagan: > We can perhaps do it better by having a public static final int zeroBased,

Would need to store the index twice though. Some might consider that un-DRY.

But anyway, accessing fields directly is not consistent with the rest of our code base (ignoring those other outliers) which lowers API discoverability.

I do think `getZeroBasedIndex()` should be shortened to `getZeroBased()` though. Callers will likely be naming their variables `somethingIndex`, so you end up with `somethingIndex.getZeroBased()`, which is clear enough.

src/main/java/seedu/address/commons/events/ui/JumpToListRequestEvent.java
```diff
@@ -9,8 +10,8 @@
 
     public final int targetIndex;
 
-    public JumpToListRequestEvent(int targetIndex) {
-        this.targetIndex = targetIndex;
+    public JumpToListRequestEvent(Index targetIndex) {
```
Zhiyuan-Amos: not sure whether we should take in `int` or `Index`, because it seems like there won't be a conversion going on. 

For e.g, one of the reasons why Parser, Commands use `Index` is because there's a need to convert from one-based to zero-based. But not so over here.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -86,7 +89,7 @@ private void handleShowHelpRequestEvent(ShowHelpRequestEvent she) {
 
     @Subscribe
     private void handleJumpToListRequestEvent(JumpToListRequestEvent je) {
-        targetedJumpIndex = je.targetIndex;
+        targetedJumpIndex = Index.createFromZeroBased(je.targetIndex);
```
Zhiyuan-Amos: again, not sure whether should we use Index or int here

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -102,7 +105,7 @@ public void setUp() {
 
         latestSavedAddressBook = new AddressBook(model.getAddressBook()); // last saved assumed to be up to date
         helpShown = false;
-        targetedJumpIndex = -1; // non yet
+        targetedJumpIndex = Index.createFromZeroBased(-1); // none yet
```
Zhiyuan-Amos: In your previous commit, I stated that we should restrict the values of `Index`. So do take note of that in relation to this.

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -71,11 +72,11 @@ public synchronized void addPerson(ReadOnlyPerson person) throws UniquePersonLis
...
             throws UniquePersonList.DuplicatePersonException {
         assert editedPerson != null;
 
-        int addressBookIndex = filteredPersons.getSourceIndex(filteredPersonListIndex);
+        int addressBookIndex = filteredPersons.getSourceIndex(index.getZeroBasedIndex());
```
Zhiyuan-Amos: as discussed previously, maybe from Model component onwards can take in an `int` instead of `Integer` :>

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -364,7 +368,7 @@ public void execute_select_jumpsToCorrectPerson() throws Exception {
                 String.format(SelectCommand.MESSAGE_SELECT_PERSON_SUCCESS, 2),
                 expectedAb,
                 expectedAb.getPersonList());
-        assertEquals(1, targetedJumpIndex);
+        assertEquals(INDEX_SECOND_PERSON, targetedJumpIndex);
```
Zhiyuan-Amos: i'm wondering what does this test accomplish XD

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -22,10 +23,10 @@ public static String getAddCommand(Person person) {
     }
 
     /**
-     * Returns an edit command string for editing the person at {@code zeroBasedIndex} to match {@code editedPerson}.
+     * Returns an edit command string for editing the person at {@code index} to match {@code editedPerson}.
      */
-    public static String getEditCommand(int zeroBasedIndex, Person editedPerson) {
-        return EditCommand.COMMAND_WORD + " " + (zeroBasedIndex + 1) + " " + getPersonDetails(editedPerson);
+    public static String getEditCommand(Index index, Person editedPerson) {
+        return EditCommand.COMMAND_WORD + " " + index.getOneBasedIndex() + " " + getPersonDetails(editedPerson);
```
Zhiyuan-Amos: perhaps header comment should be more descriptive because people calling this method don't know whether to give it an index of zero-based or one-based. 

Then again, I'm removing this in my `Tidy EditCommand` PR. :>

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -22,16 +23,15 @@
...
      */
-    public static int parseIndex(String index) throws IllegalValueException {
+    public static Index parseIndex(String index) throws IllegalValueException {
         String trimmedIndex = index.trim();
         if (!StringUtil.isUnsignedInteger(trimmedIndex)) {
             throw new IllegalValueException("Index is not a non-zero unsigned integer.");
         }
-        return Integer.parseInt(trimmedIndex);
-
+        return Index.createFromOneBased(Integer.parseInt(trimmedIndex));
```
yamgent: I solved it by renaming `String index` to `String oneBasedIndex`.

src/main/java/seedu/address/commons/events/ui/JumpToListRequestEvent.java
yamgent: It should be considered as crossing the boundary since the model sends this event to the UI.

src/test/java/seedu/address/logic/LogicManagerTest.java
yamgent: `targetedJumpIndex` is assigned a value when `JumpToListRequestEvent` is sent by the model. This line of code is testing that the correct event with the correct number is assigned.

src/test/java/seedu/address/logic/LogicManagerTest.java
yamgent: Again, it should be considered as crossing the boundary since the model sends this event to the UI.

src/test/java/seedu/address/testutil/PersonUtil.java
yamgent: The beauty of `Index` is that we should not even have to care that it is oneBased or zeroBased.

If we just want the first person, we can do it in any kind of lingo that we are comfortable with: either `Index.createFromZeroBased(0)` or `Index.createFromOneBased(1)`. Both makes sense that they are referring to the first person, right?

CanIHasReview:
# v2

* [v2 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/4464e11101cbd3579c663674b3547c4789b117a5)
* [v2 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/4320f75ab54a51f4b7de4df9f2b03106bb203ac4)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...4320f75ab54a51f4b7de4df9f2b03106bb203ac4)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/81119a51e126a8bd0ee9fb982611fcfe98f903f6~1...81119a51e126a8bd0ee9fb982611fcfe98f903f6))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/2/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,49 @@
...
+ * Represents a zero-based or one-based index.
+ */
+public class Index {
+    private int zeroBasedIndex;
+
+    /**
+     * Index can only be created by calling {@link Index#createFromZeroBased(int)} or
+     * {@link Index#createFromOneBased(int)}.
+     */
+    private Index(int zeroBasedIndex) throws IndexOutOfBoundsException {
```
Zhiyuan-Amos: no need to put `throws...`

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertEquals(1, Index.createFromOneBased(1).getOneBased());
+        assertEquals(5, Index.createFromOneBased(5).getOneBased());
+
+        // convert to zero-based index
+        assertEquals(0, Index.createFromOneBased(1).getZeroBased());
+        assertEquals(4, Index.createFromOneBased(5).getZeroBased());
+    }
+
+    @Test
+    public void createOneBasedIndex_invalidIndex_throwsOutOfBoundsException() {
```
Zhiyuan-Amos: as spoken, let's move this test into `createOneBasedIndex`, and create helper private method `assertCreateOneBasedFailure`

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.rules.ExpectedException;
+
+public class IndexTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void createOneBasedIndex() {
+        // check equality for the same base
```
Zhiyuan-Amos: check equality **using** the same base? Idk which sounds better :<

CanIHasReview:
# v3

* [v3 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/b4370af80b50ae9ddda4605287a24ef46fe9d596)
* [v3 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/00759ba59f1cede7ee3d4b5e1301422103be83c3)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...00759ba59f1cede7ee3d4b5e1301422103be83c3)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/009dacfa3bc14ea480f84efaa323b273bb0b7908~1...009dacfa3bc14ea480f84efaa323b273bb0b7908))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/3/head:BRANCHNAME
```

</details>

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,93 @@
...
+import static org.junit.Assert.fail;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class IndexTest {
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
```
Zhiyuan-Amos: not used :>

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,93 @@
...
+            // expected behaviour
+        }
+    }
+
+    @Test
+    public void equals() {
+        final Index fifthPersonIndex = Index.createFromOneBased(5);
+
+        // same values -> returns true
+        Index fifthPersonIndexCopy = Index.createFromOneBased(5);
```
Zhiyuan-Amos: should add another test which creates from zero based.

CanIHasReview:
# v4

* [v4 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/fe1cd9c93b94a55d2b04591971f12f04494a003c)
* [v4 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/21c345dfa11e51a05d00468fc359f3ff13019960)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...21c345dfa11e51a05d00468fc359f3ff13019960)) ([:chart_with_upwards_trend: Interdiff between v3 and v4](https://github.com/se-edu/addressbook-level4/compare/c626f35a8f007f80df3bc1563b9af2b8786142da~1...c626f35a8f007f80df3bc1563b9af2b8786142da))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/4/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public int getOneBased() {
+        return zeroBasedIndex + 1;
+    }
+
+    /**
+     * Creates a new {@code Index} using a zero-based index.
+     */
+    public static Index createFromZeroBased(int zeroBasedIndex) {
```
damithc: We can probably get rid of `create`. `fromZeroBased` is still self-explanatory.

src/test/java/seedu/address/logic/LogicManagerTest.java
```diff
@@ -338,7 +341,8 @@ private void assertIndexNotFoundBehaviorForCommand(String commandWord) throws Ex
             model.addPerson(p);
         }
 
-        assertCommandException(commandWord + " 3", expectedMessage);
+        assertCommandException(commandWord + " " + INDEX_THIRD_PERSON.getOneBased(),
+                expectedMessage);
```
damithc: Doesn't look like this line needs a line break.

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,88 @@
...
+     * {@code IndexOutOfBoundsException} is thrown.
+     */
+    private void assertCreateZeroBasedFailure(int invalidZeroBasedIndex) {
+        try {
+            Index.createFromZeroBased(invalidZeroBasedIndex);
+            fail("Expected IndexOutOfBoundsException to be thrown.");
+        } catch (IndexOutOfBoundsException ie) {
+            // expected behaviour
+        }
+    }
```
damithc: These two methods look very similar. Can unify?

damithc:
Commit 2 message:
>Let's update the entire codebase to switch from using int to Index, to
avoid having to do conversion at different areas of the codebase.

This statement is not correct?

Merge commit message:
>Using int as index is still allowed when dealing with other objects
within the same component. However, when crossing between different
components (e.g. between Model and UI), the bases of the index may
change. Therefore, the Index class should be used for such scenarios.

Can be more clear where to use `Index` and where to use `int` and where to convert?

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,88 @@
...
+        // check equality using the same base
+        assertEquals(0, Index.createFromZeroBased(0).getZeroBased());
+        assertEquals(5, Index.createFromZeroBased(5).getZeroBased());
+
+        // convert from zero-based index to one-based index
+        assertEquals(1, Index.createFromZeroBased(0).getOneBased());
+        assertEquals(6, Index.createFromZeroBased(5).getOneBased());
+    }
+
+
```
Zhiyuan-Amos: extra line

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private void assertCreateOneBasedFailure(int invalidOneBasedIndex) {
+        try {
+            Index.createFromOneBased(invalidOneBasedIndex);
+            fail("Expected IndexOutOfBoundsException to be thrown.");
+        } catch (IndexOutOfBoundsException ie) {
+            // expected behaviour
+        }
+    }
+
+
```
Zhiyuan-Amos: extra line

src/test/java/seedu/address/commons/core/index/IndexTest.java
Zhiyuan-Amos: @yamgent maybe can try `assertCreateFailure(int invalidIndex, boolean isZeroBased)`

CanIHasReview:
# v5

* [v5 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/7c3a5b7100da58e60f67daa2aa54077d6fb245b5)
* [v5 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/61b70f86a9482ae12f32f7c8953f4ad79a96041f)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...61b70f86a9482ae12f32f7c8953f4ad79a96041f)) ([:chart_with_upwards_trend: Interdiff between v4 and v5](https://github.com/se-edu/addressbook-level4/compare/ab9303969ef169c0e5165163cd05d8aa17b164e8~1...ab9303969ef169c0e5165163cd05d8aa17b164e8))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/5/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -22,16 +23,16 @@
 public class ParserUtil {
 
     /**
-     * Parses {@code index} into an integer and returns it. Leading and trailing whitespaces will be trimmed.
+     * Parses {@code oneBasedIndex} into an Index and returns it. Leading and trailing whitespaces will be trimmed.
      * @throws IllegalValueException if the specified index is invalid (not non-zero unsigned integer).
+     * Parses
```
Zhiyuan-Amos: what happened?

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -22,16 +23,16 @@
...
      */
-    public static int parseIndex(String index) throws IllegalValueException {
-        String trimmedIndex = index.trim();
+    public static Index parseIndex(String oneBasedIndex) throws IllegalValueException {
+        String trimmedIndex = oneBasedIndex.trim();
         if (!StringUtil.isUnsignedInteger(trimmedIndex)) {
             throw new IllegalValueException("Index is not a non-zero unsigned integer.");
         }
-        return Integer.parseInt(trimmedIndex);
-
```
Zhiyuan-Amos: why you delete new line here :<

src/main/java/seedu/address/model/Model.java
```diff
@@ -24,13 +25,13 @@
...
+     * Updates the person located at {@code index} with {@code editedPerson}.
      *
      * @throws DuplicatePersonException if updating the person's details causes the person to be equivalent to
      *      another existing person in the list.
-     * @throws IndexOutOfBoundsException if {@code filteredPersonListIndex} < 0 or >= the size of the filtered list.
+     * @throws IndexOutOfBoundsException if {@code index} is out of range of the filtered list.
      */
-    void updatePerson(int filteredPersonListIndex, ReadOnlyPerson editedPerson)
+    void updatePerson(Index index, ReadOnlyPerson editedPerson)
             throws UniquePersonList.DuplicatePersonException;
```
Zhiyuan-Amos: see if can put in previous line :>

damithc:
>Index should be used right from the start (when parsing in a new user
input), so that if the current component wants to communicate with
another component, it can send an Index to avoid having to know what
base the other component is using for its index. However, after
receiving the Index, that component can convert it back to an int if
the index will not be passed to a different component again.

In addition to the merge commit message, should we also document this in dev guide or the header comment of the `Index` class? What do you guys think?

Zhiyuan-Amos:
I think commenting in the header comment is good!

yamgent:
Agreed, will do so for the next iteration.

CanIHasReview:
# v6

@yamgent submitted v6 for review.

* [v6 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/deed3fa0e04c972a50f927fc1ce9525ad8305b48)
* [v6 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/13244910753257d8074590f78ae8553cb0f73400)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...13244910753257d8074590f78ae8553cb0f73400)) ([:chart_with_upwards_trend: Interdiff between v5 and v6](https://github.com/se-edu/addressbook-level4/compare/38d62d979736c0eca266e426d2bacb29f13a8a22~1...38d62d979736c0eca266e426d2bacb29f13a8a22))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/6/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -22,16 +23,15 @@
 public class ParserUtil {
 
     /**
-     * Parses {@code index} into an integer and returns it. Leading and trailing whitespaces will be trimmed.
+     * Parses {@code oneBasedIndex} into an Index and returns it. Leading and trailing whitespaces will be trimmed.
```
Zhiyuan-Amos: {@code Index}

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,54 @@
+package seedu.address.commons.core.index;
+
+/**
+ * Represents a zero-based or one-based index.
+ *
+ * {@code Index} should be used right from the start (when parsing in a new user input), so that if the current
+ * component wants to communicate with another component, it can send an {@code Index} to avoid having to know what
+ * base the other component is using for its index. However, after receiving the {@code Index}, that component can
+ * convert it back to an int if the index will not be passed to a different component again.
```
Zhiyuan-Amos: perhaps this line can phrase more similarly to that on the 2nd commit message.

CanIHasReview:
# v7

@yamgent submitted v7 for review.

* [v7 1/2] [Create Index class](https://github.com/se-edu/addressbook-level4/pull/447/commits/deed3fa0e04c972a50f927fc1ce9525ad8305b48)
* [v7 2/2] [Update codebase to use Index instead of int](https://github.com/se-edu/addressbook-level4/pull/447/commits/6b39861e1c022381f6232cc01bab61af924b66f8)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/cfc8ecc72a73a989ecc58427e6ddd1c564361408...6b39861e1c022381f6232cc01bab61af924b66f8)) ([:chart_with_upwards_trend: Interdiff between v6 and v7](https://github.com/se-edu/addressbook-level4/compare/0b178ea7232e77f8685f82be03c62d8f1430df82~1...0b178ea7232e77f8685f82be03c62d8f1430df82))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/447/7/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/commons/core/index/Index.java
```diff
@@ -0,0 +1,54 @@
+package seedu.address.commons.core.index;
+
+/**
+ * Represents a zero-based or one-based index.
```
chao1995: @yamgent But then given an index, we still have no idea whether it's an one-based index or a zero-base index, right?

Zhiyuan-Amos: Yup I agree with you. However, this approach helps devs to understand that we are using the same 'object' vs `IndexUtil.convertToZeroBased`. The latter doesn't have the connotation that the `int` are actually the same thing.

damithc: From what I remember, the benefit of this approach is that each class only needs to know what base is used by itself and it can pass the index object to other classes without bothering about what base is used by the receiving class. @yamgent is that correct?

yamgent: Yes, the beauty of this class is that we do not have to care whether components interacting with each other are using compatible bases for their indices. They can just pass around an `Index` class, instead of having to decide whether it is appropriate or necessary to use `IndexUtil#oneToZeroBased()` during the interaction.

It is true that we don't know whether it is an one-based index or a zero-based index. But that question need not be answered until someone really needs the raw index, and by that time, we are able to cater for both ways.

chao1995: Yes, but the thing is the dev needs to remember what the class needs, or by commenting somewhere.

I think the most self-explaining way is to ask for an object that is of type `OneBasedIndex` in the GUI, when it needs one, and asking for `ZeroBasedIndex` objects in the backend logic. `OneBasedIndex` can have an instance method `toZeroBased` to return a `ZeroBasedIndex` object and vice versa.

damithc:
> I think the most self-explaining way is to ask for an object that is of type OneBasedIndex in the GUI when it needs one, and asking for ZeroBasedIndex objects in the backend logic. OneBasedIndex can have an instance method toZeroBased to return a ZeroBasedIndex object and vice versa.
> 
> Why is this approach not the desired approach? Is it inconvenient in some way?

I think we considered this approach briefly. We discarded it because specifying the type of index expected in the API makes the abstraction 'leaky'. API user should not be told if the component uses one-based or zero-based index because that is an internal implementation detail. Furthermore, the component should be able to change its internal implementation without affecting the API. If the type of index is specified in the API, that limits the freedom of the component to change its internal implementation later.

But you are right in that this approach necessitates the developer of the component to remember which type of index it uses, (and there is a chance that him mixing up the two types and introducing a bug). We can mitigate that by adding a wrapper around the component to translates `Index` objects to `ZeroBasedIndex` objects before passing it to the component but we didn't want to go that far.

chao1995:
@damithc I see. If an approach doesn't have subclasses, yet still allows us to differentiate whether an `Index` is one-based or zero-based, will it be better?

The approach is by introducing a state to the `Index` class, say an instance variable `isZeroBased`, then when we do the conversion `void toOneBased()`, we first check whether the index `isZeroBased`.

damithc:
> The approach is by introducing a state to the Index class, say an instance variable isZeroBased, then when we do the conversion void toOneBased(), we first check whether the index is zero based by looking at the state isZeroBased.

The current view is that an `Index` object represents the position of the person in a list (we should have called it `Position`) and therefore immutable.  It should not get modified as it is being passed around because what it represents has not changed. It encapsulates both one-based and zero-based interpretations of the position at the same time. It's up to the client object to ask for the interpretation the client object uses.

In my previous response when I said 
>But you are right in that this approach necessitates the developer of the component to remember which type of index it uses, (and there is a chance that him mixing up the two types and introducing a bug). We can mitigate that by adding a wrapper around the component to translates Index objects to ZeroBasedIndex objects before passing it to the component but we didn't want to go that far.

I did not mean an actual conversion. The design I had in mind was 
```java
class Index implements ZeroBasedindex, OneBasedIndex {...}

interface ZeroBasedIndex{
    getZeroBasedIndex()
}

interface OneBasedIndex{
    getOneBasedIndex()
}

//This class accepts Index so the caller doesn't have to know which base is used by this class
class ModelWrapper{

    model Model;

    public deletePerson(Index index){ 
        model.deletePerson(index);
    }
}

//This class only accepts ZeroBasedIndex so the code has access to getZeroBasedIndex method only.
class Model{  
    deletePerson(ZeroBasedIndex index){ 
    ... 
   }
}
```

chao1995:
@damith Thank you prof for the thorough explanations!

My previous worry is that there is no way for us to determine whether an `Index`, with its internal integer value being 1, should be interpreted as a zero-based or one-based index.

Will this case be possible to happen, where someone passes in an intended-to-be-zero-based `Index` to `ModelWrapper`?

If it could happen, `Model` may end up deleting the wrong person, e.g. minus 1 (which is already an zero-based index when passed in) to get zero based index 0, then ending up deleting the first person instead of the second.

damithc:
As I understand ( @yamgent and @Zhiyuan-Amos can correct if I'm wrong), there are two places where an error can be made.
1. When creating the `Index` object: When user types `delete 1`, we should create the `Index` object using `Index.fromOneBased(1)`. 
2. When reading from the `Index` object: For example, if the `Model` receives the same `Index` object, it should use `getZeroBasedIndex()` which will return `0`. 

As you can see, there's no room for introducing errors along the way because it's the same immutable object that is being passed around. The conversions happen at the origin and at the destination (where an error is less likely), not somewhere midway which was the case before.

PR #462, opened 2017-06-02
WIP: Access previous user input

Zhiyuan-Amos:
Hello prof @damithc , is it possible for you to review the architecture of the most recent **commit** (I rebased off my other history branch)? Would like to clear the architecture with you before doing a more thorough job (i.e header comments, tests etc). Have already cleared with @yamgent and @limmlingg .

Basically the `CommandHistory` works like a `ListIterator`, just that we did not have it implement `ListIterator` because that way `CommandHistory` has to override quite a number of methods. But it essentially works like an iterator. The only issue with this design is that the `CommandHistory` will have an `int index`, which would seem rather weird if it doesn't implement `ListIterator` i.e it isn't clear that we can iterate through `CommandHistory`.

The rationale for not allowing `CommandBox` to access `CommandHistory` directly (i.e store a reference to `CommandHistory`, or store a reference to `CommandHistory.iterator` is because `Logic` works as a `Facade`, so `CommandBox` should access `CommandHistory` through `Logic`.


damithc:
It's a bit odd for the CommandHistory object to know which command is being displayed in the GUI. What if there are two places in the GUI where you can use the up/down key to get previous inputs? 

I assume you plan to evolve CommandHistory to keep a `List<UserCommand>` objects in future instead of just `List<String>`, in which case you can move one step in that direction by having this API for `CommandHistory` (and also `Logic`).

```
getMostRecent(): UserCommand  -- returns the most recent command that was input by the user
getPrevious(UserCommand refUserCommand) -- returns the command that was entered prior to the supplied refUserCommand
getNext(UserCommand refUserCommand)
```

What do you think?

PR #501, opened 2017-06-16
Remove Logic#getFilteredPersonList()

### LINKED ISSUE #500: Remove Logic#getFilteredPersonList
`MainWindow` accesses the `filteredPersonList` of the `Model` through `Logic`.

This is a bad design of the API as Logic should only be involved in executing operations, not returning a component of the Model.

Zhiyuan-Amos:
Fixes #500

### DIFF
--- a/src/main/java/seedu/address/ui/UiManager.java
+++ b/src/main/java/seedu/address/ui/UiManager.java
@@ -19,6 +19,7 @@ import seedu.address.commons.events.ui.PersonPanelSelectionChangedEvent;
 import seedu.address.commons.events.ui.ShowHelpRequestEvent;
 import seedu.address.commons.util.StringUtil;
 import seedu.address.logic.Logic;
+import seedu.address.model.Model;
 import seedu.address.model.UserPrefs;
 
 /**
@@ -32,13 +33,15 @@ public class UiManager extends ComponentManager implements Ui {
     private static final String ICON_APPLICATION = "/images/address_book_32.png";
 
     private Logic logic;
+    private Model model;
     private Config config;
     private UserPrefs prefs;
     private MainWindow mainWindow;
 
-    public UiManager(Logic logic, Config config, UserPrefs prefs) {
+    public UiManager(Logic logic, Model model, Config config, UserPrefs prefs) {
         super();
         this.logic = logic;
+        this.model = model;
         this.config = config;
         this.prefs = prefs;
     }
@@ -52,7 +55,7 @@ public class UiManager extends ComponentManager implements Ui {
         primaryStage.getIcons().add(getImage(ICON_APPLICATION));
 
         try {
-            mainWindow = new MainWindow(primaryStage, config, prefs, logic);
+            mainWindow = new MainWindow(primaryStage, config, prefs, logic, model);
             mainWindow.show(); //This should be called before creating other UI parts
             mainWindow.fillInnerParts();
 
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -70,7 +70,7 @@ public class MainApp extends Application {
 
         logic = new LogicManager(model);
 
-        ui = new UiManager(logic, config, userPrefs);
+        ui = new UiManager(logic, model, config, userPrefs);
 
         initEventsCenter();
     }
--- a/src/main/java/seedu/address/logic/Logic.java
+++ b/src/main/java/seedu/address/logic/Logic.java
@@ -1,10 +1,8 @@
 package seedu.address.logic;
 
-import javafx.collections.ObservableList;
 import seedu.address.logic.commands.CommandResult;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.logic.parser.exceptions.ParseException;
-import seedu.address.model.person.ReadOnlyPerson;
 
 /**
  * API of the Logic component
@@ -18,8 +16,5 @@ public interface Logic {
      * @throws ParseException If an error occurs during parsing.
      */
     CommandResult execute(String commandText) throws CommandException, ParseException;
-
-    /** Returns the filtered list of persons */
-    ObservableList<ReadOnlyPerson> getFilteredPersonList();
 
 }
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -2,7 +2,6 @@ package seedu.address.logic;
 
 import java.util.logging.Logger;
 
-import javafx.collections.ObservableList;
 import seedu.address.commons.core.ComponentManager;
 import seedu.address.commons.core.LogsCenter;
 import seedu.address.logic.commands.Command;
@@ -11,7 +10,6 @@ import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.logic.parser.Parser;
 import seedu.address.logic.parser.exceptions.ParseException;
 import seedu.address.model.Model;
-import seedu.address.model.person.ReadOnlyPerson;
 
 /**
  * The main LogicManager of the app.
@@ -39,10 +37,5 @@ public class LogicManager extends ComponentManager implements Logic {
         } finally {
             history.add(commandText);
         }
-    }
-
-    @Override
-    public ObservableList<ReadOnlyPerson> getFilteredPersonList() {
-        return model.getFilteredPersonList();
     }
 }
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -119,7 +119,7 @@ public class MainWindow extends UiPart<Region> {
 
     void fillInnerParts() {
         browserPanel = new BrowserPanel(browserPlaceholder);
-        personListPanel = new PersonListPanel(getPersonListPlaceholder(), logic.getFilteredPersonList());
+        personListPanel = new PersonListPanel(getPersonListPlaceholder(), model.getFilteredPersonList());
         new ResultDisplay(getResultDisplayPlaceholder());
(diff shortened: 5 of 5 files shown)


CanIHasReview:
# v1

@Zhiyuan-Amos submitted v1 for review.

* [v1 1/2] [UiManager & MainWindow: Add Model as instance variable](https://github.com/se-edu/addressbook-level4/pull/501/commits/279869bcc3683bc9053545abd7deda77e3e1ad62)
* [v1 2/2] [Remove Logic\#getFilteredPersonList](https://github.com/se-edu/addressbook-level4/pull/501/commits/c106528ef784388b756075300bcebe906fa5eecf)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/5cc85fcf11b40d1660d0b6be20a3ef90bc71c1c1...c106528ef784388b756075300bcebe906fa5eecf))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/501/1/head:BRANCHNAME
```

</details>

Zhiyuan-Amos:
@yamgent just fyi, if you wanna bold the code segment, you can do:
```
<pre>
I am not bolded <b>I am bolded</b>
</pre>
```

It will turn out like this:

<pre>
I am not bolded <b>I am bolded</b>
</pre>

CanIHasReview:
# v2

* [v2 1/2] [UiManager & MainWindow: Add Model as instance variable](https://github.com/se-edu/addressbook-level4/pull/501/commits/c7ecbc73a346c80ea3c1306a9cdc45c5df67d7c0)
* [v2 2/2] [Remove Logic\#getFilteredPersonList\(\)](https://github.com/se-edu/addressbook-level4/pull/501/commits/8854621c4252ccde1dcf3ea7539cc9337721dd6f)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/5cc85fcf11b40d1660d0b6be20a3ef90bc71c1c1...8854621c4252ccde1dcf3ea7539cc9337721dd6f)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/78357da6dcdf0ccfcee8fcd8c7974caf9e862f4c~1...78357da6dcdf0ccfcee8fcd8c7974caf9e862f4c))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/501/2/head:BRANCHNAME
```

</details>

yamgent:
@limmlingg For your review, especially with the England. :)

CanIHasReview:
# v3

@Zhiyuan-Amos submitted v3 for review.

* [v3 1/2] [UiManager & MainWindow: Add Model as instance variable](https://github.com/se-edu/addressbook-level4/pull/501/commits/a965fa7fd42a512de72dd9fd999556925fc08629)
* [v3 2/2] [Remove Logic\#getFilteredPersonList\(\)](https://github.com/se-edu/addressbook-level4/pull/501/commits/cbcf68a2ed2b7286273ff07c5e8e166d5a4e81de)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/5cc85fcf11b40d1660d0b6be20a3ef90bc71c1c1...cbcf68a2ed2b7286273ff07c5e8e166d5a4e81de)) ([:chart_with_upwards_trend: Interdiff between v2 and v3](https://github.com/se-edu/addressbook-level4/compare/a4ac339b70493f7888c263c23aa00428190d8fcd~1...a4ac339b70493f7888c263c23aa00428190d8fcd))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/501/3/head:BRANCHNAME
```

</details>

Zhiyuan-Amos:
@yamgent @limmlingg for your review; need to merge quickly this cos there's other dependencies :>

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -32,13 +33,15 @@
     private static final String ICON_APPLICATION = "/images/address_book_32.png";
 
     private Logic logic;
+    private Model model;
     private Config config;
     private UserPrefs prefs;
     private MainWindow mainWindow;
 
-    public UiManager(Logic logic, Config config, UserPrefs prefs) {
+    public UiManager(Logic logic, Model model, Config config, UserPrefs prefs) {
```
pyokagan: Such a constructor does not guarantee that the Model used by `logic` is the same as the `model` passed into the constructor.

It would be better to implement a `logic.getModel()` instead.

Also, doesn't this mean that the UI layer can completely bypass the logic layer and directly modify the model instead? Maybe make a `ReadOnlyModel` access control interface?

chao1995: > It would be better to implement a logic.getModel() instead.

`logic.getModel()` seems like a violation of Law of Demeter (which simply put means "Ask for thing, don't look for things.").

I agree by structuring this method this way, the two `Model` can be different objects and thus get out of sync. But what if we make sure the application has only one model (singleton), which I guess it should be?

chao1995: > It would be better to implement a logic.getModel() instead.

`logic.getModel()` seems like a violation of Law of Demeter, which advocates "[ask for things, don't look for things.](https://www.youtube.com/watch?v=RlfLCWKxHJ0)",  if we try to get the model from logic in this constructor.

I agree by structuring this method this way, the two `Model` can be different objects and thus get out of sync. But we won't face this problem if we make sure the application has only one model (singleton), which I guess it should be?

Zhiyuan-Amos: I'm actually thinking whether it's necessary for `MainApp` to create `Logic` and then pass it to `UiManager`. This is because `Logic` is specific to the UI component only; the UI component can create it's own `Logic` upon receiving the `Model`.

I'm thinking that MainWindow, being the parent View Controller, should create the `Logic` and pass to children View Controllers (i.e `CommandBox`). That way, we won't have the issue of MainApp passing to UiManager a Logic that is configured to the wrong Model.

Zhiyuan-Amos: @chao1995 making it a Singleton means everyone else can touch it :< Like even the Storage component etc can touch it.

@pyokagan I'm thinking that having a ReadOnlyModel is a little weird because a View Controller ought to be able to update its own Model accordingly (MVC). We are abstracting out the majority of the logic into a LogicManager class (otherwise, MainWindow would be rather bloated), thus it should be understood that even though MainWindow can update the Model by itself, it should actually do it through the LogicManager.

pyokagan: @Zhiyuan-Amos 
> a View Controller ought to be able to update its own Model accordingly (MVC).



> even though MainWindow can update the Model by itself, it should actually do it through the LogicManager.

These seem contradictory?

Zhiyuan-Amos: We can remove the LogicManager; the purpose of LogicManager is just to make MainWindow (the VC) less bloated.

chao1995: > making it a Singleton means everyone else can touch it :< Like even the Storage component etc can touch it.

By `Singleton`, did you mean the Singleton pattern? which actually is different from what I mean by `singleton`. You can have `singleton` without using the Singleton patter, with proper dependency injection.

Zhiyuan-Amos: I see! Thanks for your suggestion Huang Chao :> However, I'm not sure how do we do that (both Wang Leng and myself never heard of this before ><), do you know of any guides where we can find out more about this?

damithc:
> You can have singleton without using the Singleton pattern, via the help of proper dependency injection.

Mechanism to do the above: instantiate only one object and pass it to every other object that needs it.

src/main/java/seedu/address/ui/UiManager.java
chao1995: Yes. I learn about these testing philosophy mostly from Google Testing Blog. About singleton, you can look at this one: https://testing.googleblog.com/2008/08/by-miko-hevery-so-you-join-new-project.html. And this one is also a good read: https://testing.googleblog.com/2008/08/by-miko-hevery-so-you-decided-to.html.

By the way, search Google Tech Talks given by Misko Hevery (author of AngularJS) on YouTube. He had done a few great talks (though the contents are more or less the same as in  the blog, though I think watching videos is less arduous).

chao1995: @Zhiyuan-Amos Yes. I learn about these testing philosophy mostly from Google Testing Blog. About Singleton, you can take a look at this blog post: https://testing.googleblog.com/2008/08/by-miko-hevery-so-you-join-new-project.html. And this one is also a good read: https://testing.googleblog.com/2008/08/by-miko-hevery-so-you-decided-to.html.

By the way, search Google Tech Talks given by Misko Hevery (author of AngularJS) on YouTube. He had done quite a few great talks (the contents are more or less the same as in  the blog, though I think watching videos is less arduous). :)

Zhiyuan-Amos: Ah I see, thanks much for sharing the articles with me :> 

However, the 2nd article also assumes the Singleton pattern is being used i.e `ModelManager` has to be a implement the Singleton pattern as well?

chao1995: > However, the 2nd article also assumes the Singleton pattern is being used i.e ModelManager has to be a implement the Singleton pattern as well?

Not really. It wants to avoid using Singleton Pattern. So instead of `Server.getInstnace()` (which is a Singleton pattern example), you now don't **look for** a singleton instance anymore, instead you **ask for** in your method arguments, so somebody else will care about this for you and make sure this is singleton. Usually this **somebody else** is some Factory/Dependency injection framework.

damithc:
Let's put this on hold as well, as it pertains to a change in the GUI architecture.

Zhiyuan-Amos:
Closing this PR as it pertains to a change in architecture (which requires quite a big amount of work)

PR #548, opened 2017-07-06
Addressbook-Level 3: Remove GUI

yamgent:
```
addressbook-level3 does not have detailed GUI component.

Let's remove them.
```


yamgent:
@damithc is this the right direction to go for addressbook-level3?

yamgent:
As discussed, I will close this, come up with a workflow document. I will demonstrate a workflow in a separate repository so that hopefully we will be able to visualize it better (rather that polluting this repository and accidentally make undesirable changes).

Issue #577, opened 2017-07-21
MainWindow: Remove Placeholders


Zhiyuan-Amos:
`MainWindow` consists of multiple placeholders. These placeholders act
as container view for the UI parts / View Controllers.

These placeholders are unnecessary. JavaFX provides simpler ways to
access sub-View Controllers that do not require us to maintain these
placeholders [1].

[1] See bullet point titled: Improved access to sub-controllers in FXML
http://docs.oracle.com/javafx/2/fxml_get_started/whats_new2.htm
Additional information:
https://stackoverflow.com/questions/9717852/how-to-pass-object-created-
in-fxml-controller1-to-controller2-of-inner-fxml-cont/10718683#10718683

pyokagan:
@Zhiyuan-Amos Is this still really an issue?

Zhiyuan-Amos:
Doesn't seem like it I guess; the current design is working fine. I'm not as familiar with FXML like you so I'll give it a pass :P

PR #237, opened 2016-12-24
Simplify construction of UiParts #184

pyokagan:
Fixes #184 
Currently there is quite a lot of [boilerplate](https://github.com/se-edu/addressbook-level4/blob/61b1f6fa5d259677efcd8df61b73b7569b7d5715/src/main/java/seedu/address/ui/CommandBox.java#L32-L64) involved if a developer wants to implement a new `UiPart`:

```java
    public static CommandBox load(Stage primaryStage, AnchorPane commandBoxPlaceholder, Logic logic) {
        CommandBox commandBox = UiPartLoader.loadUiPart(primaryStage, commandBoxPlaceholder, new CommandBox());
        commandBox.configure(logic);
        commandBox.addToPlaceholder();
        return commandBox;
    }

    public void configure(Logic logic) {
        this.logic = logic;
        registerAsAnEventHandler(this);
    }

    private void addToPlaceholder() {
        SplitPane.setResizableWithParent(placeHolderPane, false);
        placeHolderPane.getChildren().add(commandTextField);
        FxViewUtil.applyAnchorBoundaryParameters(commandPane, 0.0, 0.0, 0.0, 0.0);
        FxViewUtil.applyAnchorBoundaryParameters(commandTextField, 0.0, 0.0, 0.0, 0.0);
    }

    @Override
    public void setNode(Node node) {
        commandPane = (AnchorPane) node;
    }

    @Override
    public String getFxmlPath() {
        return FXML;
    }

    @Override
    public void setPlaceholder(AnchorPane pane) {
        this.placeHolderPane = pane;
}
```

In short, the developer needs to override:

* `setNode()`
* `setFxmlPath()`
* potentially `setPlaceholder()`

and think of what is the sane thing to do for each of these methods. Additionally, to make the API pleasant to use, the developer also needs to implement a static `load()` method (and in that method, we need to decide which of the three overloaded `UiPartLoader::loadUiPart()` methods we should call)

The reason why is because the current implementation (though `UiPartLoader`) [puts the responsibility on creating new controller instances on the `FXMLLoader`](https://github.com/se-edu/addressbook-level4/blob/master/src/main/java/seedu/address/ui/UiPartLoader.java#L27), but the `FXMLLoader` does not allow us to pass arguments to the constructed controller. This leads to a lot of boilerplate code just to work around this limitation.

With the changes in this PR, we construct the controller *first* and then pass it to the `FXMLLoader`, we can move the FXML loading directly into the `UiPart` constructor and reduce the above code to just:
[...]

PR #578, opened 2017-07-21
MainWindow: Remove browserPlaceholder #577

Zhiyuan-Amos:
Part of #577. @damithc This PR is just a sample (because only one of the placeholders are removed). If this PR is merged, we have to remove all other placeholders as well (which I can do in this PR as well actually)

For more information, take a look at the header `Nested Controllers`: https://docs.oracle.com/javafx/2/api/javafx/fxml/doc-files/introduction_to_fxml.html#class_instance_elements.

If I am not wrong, we should only use `FXMLLoader` to manually create a View Controller when we are presenting another scene e.g presenting the `HelpWindow`.

### DIFF
--- a/src/main/java/seedu/address/ui/UiPart.java
+++ b/src/main/java/seedu/address/ui/UiPart.java
@@ -32,6 +32,9 @@ public abstract class UiPart<T> {
         requireNonNull(fxmlFileUrl);
         fxmlLoader = new FXMLLoader(fxmlFileUrl);
         fxmlLoader.setController(this);
+        if (this instanceof BrowserPanel) {
+            return;
+        }
         try {
             fxmlLoader.load();
         } catch (IOException e) {
--- a/src/main/resources/view/MainWindow.fxml
+++ b/src/main/resources/view/MainWindow.fxml
@@ -45,11 +45,7 @@
       <StackPane fx:id="personListPanelPlaceholder" VBox.vgrow="ALWAYS"/>
     </VBox>
 
-    <StackPane fx:id="browserPlaceholder" prefWidth="340" >
-      <padding>
-        <Insets top="10" right="10" bottom="10" left="10" />
-      </padding>
-    </StackPane>
+    <fx:include fx:id="browserPanel" source="BrowserPanel.fxml"/>
   </SplitPane>
 
   <StackPane fx:id="statusbarPlaceholder" VBox.vgrow="NEVER" />
--- a/src/main/java/seedu/address/ui/BrowserPanel.java
+++ b/src/main/java/seedu/address/ui/BrowserPanel.java
@@ -8,8 +8,10 @@ import com.google.common.eventbus.Subscribe;
 import javafx.event.Event;
 import javafx.fxml.FXML;
 import javafx.scene.layout.Region;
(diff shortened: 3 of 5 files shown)


src/main/java/seedu/address/ui/UiPart.java
```diff
@@ -32,6 +32,9 @@ public UiPart(URL fxmlFileUrl) {
         requireNonNull(fxmlFileUrl);
         fxmlLoader = new FXMLLoader(fxmlFileUrl);
         fxmlLoader.setController(this);
+        if (this instanceof BrowserPanel) {
+            return;
+        }
```
damithc: Super class should not know about sub classes. Is this branch a temporary workaround which will be removed before merging the full PR?

Zhiyuan-Amos: Yes, this is a temporary workaround. In the future I foresee `UiPart` being an interface with only 1 method `getRoot()`. Thus, this constructor will be removed.

Zhiyuan-Amos:
> BTW, it needs to be done in one shot because we cannot have a codebase where super class refers sub class.

Noted prof. I will probably do this after I am done with CS2103R and CP3108A. Also, I have included above the Java Guide that I'm using to guide my PR.

pyokagan:
@Zhiyuan-Amos Note that you can't pass arguments to controllers when using `fx:controller`, so it'll likely not work with the other ui parts.

pyokagan:
We used to use `fx:controller`, but it led to a lot of boilerplate. See #237 and its commit 549e2ce6c770a17bdacfa82e2160b02853629f92

src/main/resources/view/MainWindow.fxml
```diff
@@ -45,11 +45,7 @@
       <StackPane fx:id="personListPanelPlaceholder" VBox.vgrow="ALWAYS"/>
     </VBox>
 
-    <StackPane fx:id="browserPlaceholder" prefWidth="340" >
-      <padding>
-        <Insets top="10" right="10" bottom="10" left="10" />
```
Zhiyuan-Amos: i think i lost some information here by simply removing this.

Zhiyuan-Amos:
Much thanks for your thoughts @pyokagan! Your previous PR was very insightful.

> Note that you can't pass arguments to controllers when using fx:controller

Yes, this one I am aware of that. One of the possible workarounds is that `MainWindow` has to call setter methods for the different components. So for example, `CommandBox` has a `setData(Logic)` method, and `MainWindow` has to call that. It's somewhat similar to the way `Command` does it currently: Parse the `Command` first, then call `setData` later. Just that not all GUI components have to implement this.

One of the motivation behind this PR is that we need to have `placeholders` for the different components, which may not appear to be intuitive to other developers (please correct me if I am wrong). So in our current implementation, we have a `StackPane` which contains another `StackPane`, which then contains the `WebView`. By using `<fx:include>` and `<fx:controller>` instead, we remove the need of having such `placeholders`.

I think your comments made me think deeper about this issue. One problem is that now it's not possible to test for the UI components separately as initializing the View Controller doesn't initialize the FXML components now (since we are no longer calling `super(FXML)`). So that's one thing I have to think further about.

PR #590, opened 2017-07-28
Logic: Add alias for commands

yamgent:


damithc:
Just one alias is enough? Do we also want to update docs/tests etc.?

PR #591, opened 2017-07-28
Model: Add deleteTag(Tag)

yamgent:

### DIFF
--- a/src/main/java/seedu/address/model/ModelManager.java
+++ b/src/main/java/seedu/address/model/ModelManager.java
@@ -2,6 +2,7 @@ package seedu.address.model;
 
 import static seedu.address.commons.util.CollectionUtil.requireAllNonNull;
 
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -79,6 +82,20 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedPerson)
         indicateAddressBookChanged();
     }
 
+    @Override
+    public void deleteTag(Tag tag) throws PersonNotFoundException, DuplicatePersonException {
+        for (int i = 0; i < addressBook.getPersonList().size(); i++) {
+            ReadOnlyPerson oldPerson = addressBook.getPersonList().get(i);
+
+            Person newPerson = new Person(oldPerson);
+            Set<Tag> newTags = newPerson.getTags();
```
nbriannl: getTags() returns an immutable tag set, which throws UnsupportedOperationException if modification is attempted.
making the subsequent newTags.remove(tag) method wrong.

nbriannl: `Set<Tag> newTags = new HashSet<Tag>(newPerson.getTags());`

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -79,6 +82,20 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedPerson)
...
+            newTags.remove(tag);
```
nbriannl: For reference, .remove(tag) on immutable newTags will throw Exception.

PR #594, opened 2017-07-28
Storage: Add backupAddressBook(ReadOnlyAddressBook)

yamgent:
Part of #784. Rebased, fixed the compilation error.

Proposed commit message:
```
Storage saves to a single file path.

Users may want to do a backup of the address book in a different
location.

Let's add Storage#backupAddressBook(ReadOnlyAddressBook).
```

_Note:_ This PR is used as an example for future developers. Do not merge!

### DIFF
--- a/src/main/java/seedu/address/storage/StorageManager.java
+++ b/src/main/java/seedu/address/storage/StorageManager.java
@@ -78,6 +78,10 @@ public class StorageManager extends ComponentManager implements Storage {
         addressBookStorage.saveAddressBook(addressBook, filePath);
     }
 
+    @Override
+    public void backupAddressBook(ReadOnlyAddressBook addressBook) throws IOException {
+        addressBookStorage.backupAddressBook(addressBook);
+    }
 
     @Override
     @Subscribe
--- a/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
+++ b/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
@@ -6,6 +6,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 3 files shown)


src/main/java/seedu/address/storage/StorageManager.java
```diff
@@ -77,6 +77,10 @@ public void saveAddressBook(ReadOnlyAddressBook addressBook, String filePath) th
         addressBookStorage.saveAddressBook(addressBook, filePath);
     }
 
+    @Override
+    public void backupAddressBook(ReadOnlyAddressBook addressBook) throws IOException {
+        saveAddressBook(addressBook, addressBookStorage.getAddressBookFilePath() + "-backup.xml");
```
Zhiyuan-Amos: This method should be calling `addressBookStorage#backupAddressBook(ReadOnlyAddressBook)` instead :P

Zhiyuan-Amos:
@yamgent this PR is failing 1 test :<

yamgent:
@Zhiyuan-Amos I rebased the thing, and Travis didn't report any problems. Which test is failing for you?

Zhiyuan-Amos:
@yamgent When I ran the test last night, I think `ErrorDialogGuiTest` (smth like that, I recall seeing the words `ErrorDialog`) was failing. Since it isn't failing now then that's great ^^

yamgent:
Changelog:

* Because we changed the way we handled paths (from `String` to `Path`), we must now use the `Paths` class to generate the backup file path:

https://github.com/se-edu/addressbook-level4/blob/dea261987f512a22eb0e604e28eed6287e682cc1/src/main/java/seedu/address/storage/XmlAddressBookStorage.java#L66

src/main/java/seedu/address/storage/XmlAddressBookStorage.java
```diff
@@ -60,6 +61,11 @@ public Path getAddressBookFilePath() {
         }
     }
 
+    @Override
+    public void backupAddressBook(ReadOnlyAddressBook addressBook) throws IOException {
+        saveAddressBook(addressBook, Paths.get(filePath.toString() + ".backup"));
```
Zhiyuan-Amos: I'm wondering whether the backup Path should be stored in a private final field, to be assigned in the constructor.

This is so that we don't have to call Paths.get whenever this method is called.

src/main/java/seedu/address/storage/XmlAddressBookStorage.java
```diff
@@ -22,10 +23,12 @@
 
     private static final Logger logger = LogsCenter.getLogger(XmlAddressBookStorage.class);
 
-    private Path filePath;
+    private final Path filePath;
+    private final Path backupFilePath;
 
     public XmlAddressBookStorage(Path filePath) {
         this.filePath = filePath;
+        this.backupFilePath = Paths.get(filePath.toString() + ".backup");
```
yamgent: Shouldn't use `this.` here, my bad.

PR #596, opened 2017-07-29
Ui: Show total persons in status bar

yamgent:


charlesgoh:
We need to update StatusBarFooterTest.java as well because the setUp function calls StatusBarFooter with only one argument. We have to add in the second arg (i.e. size of personlist)

PR #379, opened 2017-04-05
Checkstyle: enforce header comments on classes and methods #319

eugenepeh:
Addition to #319

PR #631, opened 2017-08-16
Main folder: add missing API header comments #379

eugenepeh:
Part of #379

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -80,6 +80,11 @@ public class MainApp extends Application {
         return applicationParameters.get(parameterName);
     }
 
+    /**
+     * Returns a {@code ModelManager} with the data from {@code storage}'s address book and {@code userPrefs}. <br>
+     * The data from the sample address book will be used instead if {@code storage}'s address book is not found,
+     * or an empty address book will be used instead if errors occur when reading {@code storage}'s address book.
+     */
     private Model initModelManager(Storage storage, UserPrefs userPrefs) {
         Optional<ReadOnlyAddressBook> addressBookOptional;
         ReadOnlyAddressBook initialData;
@@ -104,6 +109,13 @@ public class MainApp extends Application {
         LogsCenter.init(config);
     }
 
+    /**
+     * Returns a {@code Config} using the file at {@code configFilePath}. <br>
+     * The default file path {@code Config#DEFAULT_CONFIG_FILE} will be used instead
+     * if the {@code configFilePath} is null. <br>
+     * The {@code ConfigUtil}'s config file is also updated in case it was missing to begin with or
+     * there are new/unused fields.
+     */
     protected Config initConfig(String configFilePath) {
         Config initializedConfig;
         String configFilePathUsed;
@@ -135,6 +147,13 @@ public class MainApp extends Application {
         return initializedConfig;
     }
 
+    /**
+     * Returns a {@code UserPrefs} using the file at {@code storage}'s user prefs file path. <br>
+     * Returns Returns a new {@code UserPrefs} with default configuration if errors occur when
+     * reading from the file. <br>
+     * The {@code storage}'s user prefs is also updated in case it was missing to begin with or
+     * there are new/unused fields.
+     */
     protected UserPrefs initPrefs(UserPrefsStorage storage) {
         String prefsFilePath = storage.getUserPrefsFilePath();
         logger.info("Using prefs file : " + prefsFilePath);
--- a/src/main/java/seedu/address/commons/core/LogsCenter.java
+++ b/src/main/java/seedu/address/commons/core/LogsCenter.java
@@ -61,6 +61,10 @@ public class LogsCenter {
         return getLogger(clazz.getSimpleName());
     }
 
+    /**
+     * Adds the {@code consoleHandler} to the {@code logger}. <br>
+     * Creates the {@code consoleHandler} if it is null.
+     */
     private static void addConsoleHandler(Logger logger) {
         if (consoleHandler == null) {
             consoleHandler = createConsoleHandler();
@@ -68,6 +72,9 @@ public class LogsCenter {
         logger.addHandler(consoleHandler);
     }
 
+    /**
+     * Remove all the handlers from {@code logger}.
+     */
     private static void removeHandlers(Logger logger) {
         Handler[] handlers = logger.getHandlers();
         for (Handler handler : handlers) {
@@ -75,6 +82,10 @@ public class LogsCenter {
         }
     }
 
+    /**
+     * Adds the {@code fileHandler} to the {@code logger}. <br>
+     * Creates {@code fileHandler} if it is null.
+     */
     private static void addFileHandler(Logger logger) {
         try {
             if (fileHandler == null) {
@@ -86,6 +97,10 @@ public class LogsCenter {
         }
     }
 
+    /**
+     * Creates a {@code fileHandler} for the log file.
+     * @throws IOException if there are problems opening the file.
+     */
     private static FileHandler createFileHandler() throws IOException {
         FileHandler fileHandler = new FileHandler(LOG_FILE, MAX_FILE_SIZE_IN_BYTES, MAX_FILE_COUNT, true);
         fileHandler.setFormatter(new SimpleFormatter());
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -60,6 +60,9 @@ public class AddressBook implements ReadOnlyAddressBook {
         this.tags.setTags(tags);
     }
 
+    /**
+     * Resets the existing data of this address book with {@code newData}.
+     */
     public void resetData(ReadOnlyAddressBook newData) {
         requireNonNull(newData);
         try {
@@ -139,6 +142,10 @@ public class AddressBook implements ReadOnlyAddressBook {
         persons.forEach(this::syncMasterTagListWith);
     }
 
+    /**
+     * Removes {@code key} from this address book.
+     * @throws PersonNotFoundException if the {@code key} is not in this address book.
+     */
     public boolean removePerson(ReadOnlyPerson key) throws PersonNotFoundException {
         if (persons.remove(key)) {
             return true;
--- a/src/main/java/seedu/address/model/tag/UniqueTagList.java
+++ b/src/main/java/seedu/address/model/tag/UniqueTagList.java
@@ -99,6 +99,9 @@ public class UniqueTagList implements Iterable<Tag> {
         return internalList.iterator();
     }
 
+    /**
+     * Returns the backing list as an {@code UnmodifiableObservableList}.
+     */
     public UnmodifiableObservableList<Tag> asObservableList() {
         assert CollectionUtil.elementsAreUnique(internalList);
         return new UnmodifiableObservableList<>(internalList);
@@ -112,6 +115,10 @@ public class UniqueTagList implements Iterable<Tag> {
                         && this.internalList.equals(((UniqueTagList) other).internalList));
     }
 
+    /**
+     * Returns true if the element in this list is equal to the elements in {@code other}.
+     * The elements do not have to be in the same order.
+     */
     public boolean equalsOrderInsensitive(UniqueTagList other) {
         assert CollectionUtil.elementsAreUnique(internalList);
         assert CollectionUtil.elementsAreUnique(other.internalList);
--- a/src/main/java/seedu/address/model/util/SampleDataUtil.java
+++ b/src/main/java/seedu/address/model/util/SampleDataUtil.java
@@ -14,6 +14,9 @@ import seedu.address.model.person.Phone;
 import seedu.address.model.person.exceptions.DuplicatePersonException;
 import seedu.address.model.tag.Tag;
 
+/**
+ * Contains utility methods for populating {@code AddressBook} with sample data.
+ */
 public class SampleDataUtil {
     public static Person[] getSamplePersons() {
         try {
--- a/src/main/java/seedu/address/ui/CommandBox.java
+++ b/src/main/java/seedu/address/ui/CommandBox.java
@@ -15,6 +15,9 @@ import seedu.address.logic.commands.CommandResult;
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.logic.parser.exceptions.ParseException;
 
+/**
+ * The UI component that is responsible for receiving user command inputs.
+ */
 public class CommandBox extends UiPart<Region> {
 
     public static final String ERROR_STYLE_CLASS = "error";
@@ -35,6 +38,9 @@ public class CommandBox extends UiPart<Region> {
         historySnapshot = logic.getHistorySnapshot();
     }
 
+    /**
+     * Handles {@code keyEvent}.
+     */
     @FXML
     private void handleKeyPress(KeyEvent keyEvent) {
         switch (keyEvent.getCode()) {
@@ -89,6 +95,9 @@ public class CommandBox extends UiPart<Region> {
         commandTextField.positionCaret(commandTextField.getText().length());
     }
 
+    /**
+     * Handles the Enter button pressed event.
+     */
     @FXML
     private void handleCommandInputChanged() {
         try {
@@ -109,6 +118,9 @@ public class CommandBox extends UiPart<Region> {
         }
     }
 
+    /**
+     * Initializes the history snapshot.
+     */
     private void initHistory() {
         historySnapshot = logic.getHistorySnapshot();
         // add an empty string to represent the most-recent end of historySnapshot, to be shown to
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -39,6 +39,16 @@ public class HelpWindow extends UiPart<Region> {
         browser.getEngine().load(userGuideUrl);
     }
 
+    /**
+     * Shows the help window.
+     * @throws IllegalStateException if this method is called on a thread
+     *     other than the JavaFX Application Thread.
+     * @throws IllegalStateException if this method is called during
+     *     animation or layout processing.
+     * @throws IllegalStateException if this method is called on the
+     *     primary stage.
+     * @throws IllegalStateException if {@code dialogStage} is already showing.
+     */
     public void show() {
         logger.fine("Showing help page about the application.");
         dialogStage.showAndWait();
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -122,6 +122,9 @@ public class MainWindow extends UiPart<Region> {
         });
     }
 
+    /**
+     * Fills up all the placeholders of this window.
+     */
     void fillInnerParts() {
         browserPanel = new BrowserPanel();
         browserPlaceholder.getChildren().add(browserPanel.getRoot());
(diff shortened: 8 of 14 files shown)


CanIHasReview:
# v1

@eugenepeh submitted v1 for review.

* [v1 1/1] [Main folder: add missing API header comments](https://github.com/se-edu/addressbook-level4/pull/631/commits/70e396ed61ed54ba75bc12ae7967405603f747f7)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/ddcea63541d5f73f6b3d7c6c7676d4db2bb1af9a...70e396ed61ed54ba75bc12ae7967405603f747f7))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/631/1/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -60,6 +60,9 @@ public void setTags(Set<Tag> tags) {
         this.tags.setTags(tags);
     }
 
+    /**
+     * Replace existing {@code UniquePersonList} and {@code UniqueTagList} with {@code newData}'s.
```
Zhiyuan-Amos: This header comment is explaining the inner workings of the method :P It should simply give a high-level view of what this method is doing e.g. `Resets the existing data of this address book with {@code newData}`.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -139,6 +142,10 @@ private void syncMasterTagListWith(UniquePersonList persons) {
         persons.forEach(this::syncMasterTagListWith);
     }
 
+    /**
+     * Removes a person, that is equivalent to {@code key}, from {@code UniquePersonList}.
+     * @throws PersonNotFoundException if an equivalent person is not found and deleted.
```
Zhiyuan-Amos: Removes {@code key} from this address book. 
@throws PersonNotFoundException if {@code key} is not in this address book.

Since `UniquePersonList` is a private variable, I don't think people can see that AddressBook is implemented with a {@code UniquePersonList} anws :P

src/main/java/seedu/address/model/tag/UniqueTagList.java
```diff
@@ -99,6 +99,9 @@ public void add(Tag toAdd) throws DuplicateTagException {
         return internalList.iterator();
     }
 
+    /**
+     * Returns the {@code internalList} as {@code UnmodifiableObservableList}.
```
Zhiyuan-Amos: Same as what I mentioned above: `internalList` is a private variable :P So, do something like this instead:

`Returns the backing list as an {@code UnmodifiableObservableList}.`

src/main/java/seedu/address/model/tag/UniqueTagList.java
```diff
@@ -112,6 +115,10 @@ public boolean equals(Object other) {
                         && this.internalList.equals(((UniqueTagList) other).internalList));
     }
 
+    /**
+     * Returns true if this {@code internalList} is equal to {@code other}'s internalList.
+     * The list order is not sensitive.
```
Zhiyuan-Amos: Returns true if the elements in this ~~{@code internalList}~~ list is equal to the elements in {@code other}~~'s internalList~~.
The elements do not have to be in the same order.

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -14,6 +14,9 @@
 import seedu.address.model.person.exceptions.DuplicatePersonException;
 import seedu.address.model.tag.Tag;
 
+/**
+ * Contain utility methods for populating AddressBook with sample data.
```
Zhiyuan-Amos: Contain**s** ... {@code AddressBook}

src/main/java/seedu/address/ui/CommandBox.java
```diff
@@ -15,6 +15,9 @@
 import seedu.address.logic.commands.exceptions.CommandException;
 import seedu.address.logic.parser.exceptions.ParseException;
 
+/**
+ * The Command Box of the App.
```
Zhiyuan-Amos: I know you forked this from BrowserPanel, but we can try to be more descriptive over here :P

src/main/java/seedu/address/ui/CommandBox.java
```diff
@@ -35,6 +38,9 @@ public CommandBox(Logic logic) {
         historySnapshot = logic.getHistorySnapshot();
     }
 
+    /**
+     * Handles the key press event, {@code keyEvent}, of the Command Box.
```
Zhiyuan-Amos: `Handles {@code keyEvent}` will do.

src/main/java/seedu/address/ui/CommandBox.java
```diff
@@ -109,6 +118,9 @@ private void handleCommandInputChanged() {
         }
     }
 
+    /**
+     * Initializes the history snapshot.
```
Zhiyuan-Amos: I'm wondering if `Initializes {@code historySnapshot}` will do. @yamgent since this is a private method, is it ok to use `{@code}` on a private variable as well?

src/main/java/seedu/address/ui/HelpWindow.java
```diff
@@ -39,6 +39,16 @@ public HelpWindow() {
...
+    /**
+     * Shows the help window.
+     * @throws IllegalStateException if this method is called on a thread
+     *     other than the JavaFX Application Thread.
+     * @throws IllegalStateException if this method is called during
+     *     animation or layout processing.
+     * @throws IllegalStateException if this method is called on the
+     *     primary stage.
+     * @throws IllegalStateException if {@code dialogStage} is already showing.
+     */
```
Zhiyuan-Amos: @damithc do we have to be so detailed in our documentation concerning unchecked exceptions?

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -122,6 +122,9 @@ private void setAccelerator(MenuItem menuItem, KeyCombination keyCombination) {
         });
     }
 
+    /**
+     * Fill up all the placeholders of the Main Window.
```
Zhiyuan-Amos: Fill**s** ..... of **this window**.

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -8,6 +8,9 @@
 import javafx.scene.layout.Region;
 import seedu.address.model.person.ReadOnlyPerson;
 
+/**
+ * An Ui component that display information of a {@code Person}.
```
Zhiyuan-Amos: An U**I** component that display**s** the information of a {@code Person}.

src/main/java/seedu/address/ui/PersonListPanel.java
```diff
@@ -51,6 +51,9 @@ private void setEventHandlerForSelectionChangeEvent() {
                 });
     }
 
+    /**
+     * Highlights the {@code PersonCard} at the {@code index}.
+     */
     private void scrollTo(int index) {
```
Zhiyuan-Amos: You should mention both scrolling and selecting. :) I'm not sure whether we should use the term `highlight`.

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -83,6 +83,9 @@ void showAlertDialogAndWait(Alert.AlertType type, String title, String headerTex
         showAlertDialogAndWait(mainWindow.getPrimaryStage(), type, title, headerText, contentText);
     }
 
+    /**
+     * Shows an alert dialog with the given parameters and wait for user response.
```
Zhiyuan-Amos: Shows an alert dialog with the given parameters ~~and wait for user response~~. This method only returns after the user has closed the alert dialog.

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -95,6 +98,9 @@ private static void showAlertDialogAndWait(Stage owner, AlertType type, String t
         alert.showAndWait();
     }
 
+    /**
+     * Shows an alert dialog with the message of {@code e} and exits the application after user confirmation.
```
Zhiyuan-Amos: after ~~user confirmation~~ the user has closed the alert dialog.

src/main/java/seedu/address/ui/PersonListPanel.java
```diff
@@ -64,6 +67,9 @@ private void handleJumpToListRequestEvent(JumpToListRequestEvent event) {
         scrollTo(event.targetIndex);
     }
 
+    /**
+     * A ListView item view that display the graphics of a {@code PersonCard}.
```
Zhiyuan-Amos: Custom {@code ListCell} that displays the graphics of a {@code PersonCard}.

src/main/java/seedu/address/MainApp.java
```diff
@@ -104,6 +108,9 @@ private void initLogging(Config config) {
         LogsCenter.init(config);
     }
 
+    /**
+     * Initializes a {@code Config} with the file at {@code configFilePath}.
```
eugenepeh: > Also, have to specify what happens when configFilePath is null. Maybe something like If {@code configFilePath} is null, the default file path {@code Config#DEFAULT_CONFIG_FILE} will be used.

That's what I thought initially but wouldn't that be explaining too much of the code and also the underlying implementation? There's also logging involved so I am not too sure if we should explain what if the `{@code configFilePath} is null`.

eugenepeh: It's also a bit weird if the programmer be like "hmm I would like to use the DEFAULT_CONFIG_FILE" 
-> I shall use the method like this `initConfig(null);`

Zhiyuan-Amos: > There's also logging involved

Logging is fine because we don't mention elsewhere in the codebase in the header comments whenever we are logging. 

> explaining too much of the code

It's necessary so that developers know the behaviour of this method if they pass in a null (one of the purpose of method headers are to explain the behaviour of the method). So for e.g., `Model#updateFilteredPersonList(Predicate<ReadOnlyPerson>)` mention that a NPE will be thrown if the predicate is null.

> It's also a bit weird if the programmer be like "hmm I would like to use the DEFAULT_CONFIG_FILE" -> I shall use the method like this initConfig(null);

Haha you are perfectly right :P It should be a `initConfig()` instead actually. However, since the method is implemented this way, we have no choice but to write the appropriate header comments according to how this method is implemented currently. Unless you wanna raise PRs to fix this method ><

eugenepeh: >It's necessary so that developers know the behaviour of this method if they pass in a null (one of the purpose of method headers are to explain the behaviour of the method). So for e.g., Model#updateFilteredPersonList(Predicate<ReadOnlyPerson>) mention that a NPE will be thrown if the predicate is null.

Alright, will it be better if I follow `@throw` exception style? where we put the expected outcome before the condition.

E.g.

`The default file path {@code Config#DEFAULT_CONFIG_FILE} will be used instead 
if the {@code configFilePath} is null.`

Zhiyuan-Amos: Yup this looks good! ^^

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -86,6 +95,10 @@ private static void addFileHandler(Logger logger) {
         }
     }
 
+    /**
+     * Creates a new {@code fileHandler}.
+     * @throws IOException if there are problems opening the file.
```
yamgent: Creates a file handler for the log file.

src/main/java/seedu/address/ui/CommandBox.java
yamgent: I am not sure whether we are eventually converting everything to `{@link}`, but if we do, `{@link}` will not be allowing linking with variables inside the method.

eugenepeh: there are some methods' header comments do that. such as `private void navigateToNextInput()`

yamgent: Those have to be changed then. It is weird to reference variables inside methods, since javadocs are meant to be for the users of the API and most of the time these users won't have access to the implementation of the methods, so the variable referencing is not useful.

Zhiyuan-Amos: Alright cool. That way I suppose this header comment is fine then ^^

CanIHasReview:
# v2

@eugenepeh submitted v2 for review.

* [v2 1/1] [Main folder: add missing API header comments](https://github.com/se-edu/addressbook-level4/pull/631/commits/17e7cdb75c9bf5c87078a0a020136ec1d811bfb7)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/ddcea63541d5f73f6b3d7c6c7676d4db2bb1af9a...17e7cdb75c9bf5c87078a0a020136ec1d811bfb7)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/4ab3fdefc6aa5b13979bc9a7dab46cdbfeb496d7~1...4ab3fdefc6aa5b13979bc9a7dab46cdbfeb496d7))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/631/2/head:BRANCHNAME
```

</details>

src/main/java/seedu/address/MainApp.java
```diff
@@ -135,6 +145,10 @@ protected Config initConfig(String configFilePath) {
         return initializedConfig;
     }
 
+    /**
+     * Returns a {@code UserPrefs} using the file at {@code prefsFilePath} stored in {@code storage}. <br>
```
Zhiyuan-Amos: `prefsFilePath` is unknown to the reader. So something like `Returns **the** {@code UserPrefs} at {@code storage}'s user prefs file path.

src/main/java/seedu/address/MainApp.java
```diff
@@ -135,6 +145,10 @@ protected Config initConfig(String configFilePath) {
...
+     * Returns the default {@code UserPrefs} if {@code DataConversionException} or {@code IOException} occurred.
```
Zhiyuan-Amos: Same issue as above ^^ Also, perhaps we should be more specific on what's the `default {@code UserPrefs}`. Maybe something like `Returns the {@code UserPrefs} constructed by the empty {@code UserPrefs}'s constructor...` Otherwise developers will be thinking "What's the default {@code UserPrefs}?"

src/main/java/seedu/address/commons/core/LogsCenter.java
```diff
@@ -61,20 +61,31 @@ public static Logger getLogger(String name) {
...
         if (consoleHandler == null) {
             consoleHandler = createConsoleHandler();
         }
         logger.addHandler(consoleHandler);
     }
 
+    /**
+     * Remove all the handlers from {@code logger}.
+     */
     private static void removeHandlers(Logger logger) {
```
Zhiyuan-Amos: If you update this method to use lambda: `Arrays.stream(handlers).forEach(logger::removeHandler);`

We can remove the header comments :P. That needs to be in another PR though.

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -14,6 +14,9 @@
 import seedu.address.model.person.exceptions.DuplicatePersonException;
 import seedu.address.model.tag.Tag;
 
+/**
+ * Contains utility methods for populating {@code AddressBook} with sample data.
+ */
 public class SampleDataUtil {
```
Zhiyuan-Amos: just curious, public static methods don't require header comments?

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -83,6 +83,10 @@ void showAlertDialogAndWait(Alert.AlertType type, String title, String headerTex
         showAlertDialogAndWait(mainWindow.getPrimaryStage(), type, title, headerText, contentText);
     }
 
+    /**
+     * Shows an alert dialog with the given parameters.
+     * This method only returns after the user has closed the alert dialog.
```
Zhiyuan-Amos: I think the header comments can be more specific. Maybe something like: `Shows an alert dialog on {@code owner} with the given parameters.`

src/main/java/seedu/address/ui/UiManager.java
```diff
@@ -95,6 +99,10 @@ private static void showAlertDialogAndWait(Stage owner, AlertType type, String t
         alert.showAndWait();
     }
 
+    /**
+     * Shows an alert dialog with the message of {@code e} and exits the application after
```
Zhiyuan-Amos: Can be more specific too. `Shows an **error** alert dialog that displays {@code title} and the error message of {@code e}, and...`

src/main/java/seedu/address/model/util/SampleDataUtil.java
eugenepeh: its not that public static don't require, its the getters and setters :P

src/main/java/seedu/address/commons/core/LogsCenter.java
eugenepeh: True but if we do that, we need to fix all the others (if there's any) as well :P
Have any of the addressbook's code started using lambda yet?

Zhiyuan-Amos: > True but if we do that, we need to fix all the others (if there's any) as well :P

I would imagine this to be a really boring PR ><

> Have any of the addressbook's code started using lambda yet?

I converted a lot of the code to lambda expressions actually :P

src/main/java/seedu/address/model/util/SampleDataUtil.java
Zhiyuan-Amos: Ah I see. Some public getter methods are complex though, for e.g. those in `SampleDataUtil` class, and perhaps having a header comment that would be helpful. Do you think we should raise an issue to improve the Checkstyle, since I think you're done with your 2103R right :P

eugenepeh: I believe, it is still possible to do something like that by adding another method rule that does not ignore getters and setters and set a different length limit e.g. 5. But it's a very gray area where the programmer may deems that there is no more useful information which he can add to the header comment. So in the end checkstyle only can assist but not foolproof. :P

>I think you're done with your 2103R right i.e. this is your last PR or smth like that :P

Pretty much :P but i will still continue to contribute from time to time

src/main/java/seedu/address/commons/core/LogsCenter.java
eugenepeh: Opened an issue for this #634 but let's get this PR approved first before I work on that one :P

src/main/java/seedu/address/MainApp.java
eugenepeh: Hmm empty constructor sounds weird though because it does comes with the default `guiSettings`, `addressBookFilePath` and `addressBookName`. 
How about `Returns a new {@code UserPrefs} with default configuration`?

Zhiyuan-Amos: Yup sounds good! :+1:

Zhiyuan-Amos:
Just curious @eugenepeh, why didn't you merge this PR with #379? You can have the first commit whereby you update all the header comments (i.e. this PR) and in the next commit you can update the checkstyle? That way you can be sure that you didn't miss anything out isn't it :P

eugenepeh:
@Zhiyuan-Amos it was too big initially and along with all ongoing PRs, most likely it will end up with a lot of collisions :P. Other contributors suggested that I should break it into smaller PRs, so it will be easier on the reviewers.

Zhiyuan-Amos:
@eugenepeh ah I see. Since we won't be merging any PRs soon, feel free to combine this PR with #379 if it makes your work easier ^^

eugenepeh:
continue in #379

Issue #551, opened 2017-07-11
Expand the scope of system tests


yamgent:
The "system" tests in `test/java/guitests/` currently act somewhat as a unit test for GUI and commands.

With PR #536, GUI unit tests are done in a seperate architecture and test cases.
With PRs associated to #172, commands themselves already have their appropriate unit tests.

As such, the system tests should be more thorough. For example, after performing a valid delete, `DeleteCommandTest` should also ensure that the StatusBar's time is updated, and that StorageManager has also saved the file.

Zhiyuan-Amos:
@damithc do we have to write system tests for:
1. `ExitCommand` (I think we definitely can't write for this :P)
1. `HelpCommand`
1. `HistoryCommand`
1. `ListCommand`
1. `RedoCommand`
1. `UndoCommand`

For `UndoCommand` and `RedoCommand`, we have integrated them with other system tests. For the remaining 4 commands, arguably we need a system test to ensure that upon execution of these commands, the system is updated / not updated accordingly.

damithc:
1. `ExitCommand` : we should, but probably can't, so it's ok.
1. `HelpCommand` : yes
1. `HistoryCommand` : yes
1. `ListCommand` : yes
1. `RedoCommand`/`UndoCommand` : if they are thoroughly tested by other commands, can be omitted. But it is unlikely other tests will push the limits of these commands, so I guess yes, we can have a system test for these too, but can be a lower priority.

Zhiyuan-Amos:
Noted. I will put the emphasis on porting the existing tests from `guitests` to `systemtests`, before implementing for the commands listed above. ^^

yamgent:
> ExitCommand (I think we definitely can't write for this :P)

The test will type in the `exit` keyword, and then test that the `MainWindow`/`MainApp` is closed?

Zhiyuan-Amos:
> The test will type in the exit keyword, and then test that the MainWindow/MainApp is closed.

After investigating this, @yamidark and I realized that it's not possible to test that the MainWindow / MainApp is closed because the JVM shuts down after the `exit` command is executed; the test will end immediately. See [here](https://github.com/se-edu/addressbook-level4/pull/832#issuecomment-366839340) for the detailed discussion.

PR #620, opened 2017-08-08
Improve system tests #551

Zhiyuan-Amos:
Part of #551

```
There are problems with system tests. For e.g. the CommandBox's style
isn't verified during the tests and the design of AddressBookSystemTest
and AppStateAsserts are hard to understand.

This means that our tests are not thorough enough and future developers
may find it hard to understand the system tests.

Let's resolve these issues.
```

PR #636, opened 2017-08-21
Add SelectCommandSystemTest #551

Zhiyuan-Amos:
Part of #551


Zhiyuan-Amos:
Was looking through this PR, and I'm wondering whether the `MainWindowHandle` in `AddressBookSystemTest` should actually be placed inside `TestApp`, since the UI components all belong to `TestApp`.

Zhiyuan-Amos:
Add test: `FindCommand` followed by `SelectCommand`

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,104 @@
...
+        /* Case: invalid index (1) -> rejected */
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+    }
+
+    /**
+     * Executes {@code command} and verifies that the command box displays an empty string, the result display
+     * box displays the success message of executing select command with the {@code expectedSelectedCardIndex}
+     * of the selected person, and the model related components equal to the current model.
+     * These verifications are done by
+     * {@code AddressBookSystemTest#assertCommandExecution(String, String, String, Model)}. Also verifies that
```
eugenepeh: could use a `<br>` here after `{@code AddressBookSystemTest#assertCommandExecution(String, String, String, Model)}.`
to improve readability.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,104 @@
...
+        Index personCount = Index.fromOneBased(getTypicalPersons().size());
+        command = SelectCommand.COMMAND_WORD + " " + personCount.getOneBased();
+        assertCommandSuccess(command, personCount);
+
+        /* Case: select the middle card in the list -> selected */
+        Index middleIndex = Index.fromOneBased(personCount.getOneBased() / 2);
+        command = SelectCommand.COMMAND_WORD + " " + middleIndex.getOneBased();
+        assertCommandSuccess(command, middleIndex);
+
+        /* Case: invalid index (size + 1) -> rejected */
```
eugenepeh: Repeated test case

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -33,13 +37,45 @@ public void select_nonEmptyList() throws Exception {
         command = SelectCommand.COMMAND_WORD + " " + personCount.getOneBased();
         assertCommandSuccess(command, personCount);
 
+        /* Case: undo previous selection -> rejected */
```
eugenepeh: Does both undo and redo works on select command? Or is it just to make sure it doesn't work?

Zhiyuan-Amos: Just to make sure that it doesn't work ^^

src/test/java/systemtests/SelectCommandSystemTest.java
Zhiyuan-Amos: Ok noted! When is the general case that I should use <br>?

Zhiyuan-Amos:
@eugenepeh thanks for your review! Yes, I have removed `assertNoCardSelected();` and `assertCardSelected(Index)`. This is because we are verifying card selections using relative comparisons: `assertSelectedCardUnchanged()`, `assertSelectedCardChanged(Index)` and `assertCardDeselected()` (`assertCardDeselected()` only implemented in #638, see v3 [4/5]). 

Relative comparisons allow improves code readability between command executions. For e.g., when a command is executed and we have `assertSelectedCardUnchanged()`, we can easily tell that this command doesn't affect the selection of the card. Compared to `assertNoCardSelected()`, we have to see whether the previous command also has `assertNoCardSelected()` before we can deduce that this command doesn't affect the selection of the card.
(At least this is why I think we are using relative comparisons instead of absolute comparisons) ^^

src/test/java/systemtests/SelectCommandSystemTest.java
eugenepeh: We should use a `<br>` or an `<ul>` when we're highlighting multiple points (especially if its an exceptionally long paragraph).

Zhiyuan-Amos:
I noticed that sometimes I didn't inline the variables (https://github.com/se-edu/addressbook-level4/pull/636/files#diff-728e4a68f3c355109800be4dfa7cc8b4R45), and sometimes I inline the variables (https://github.com/se-edu/addressbook-level4/pull/636/files#diff-728e4a68f3c355109800be4dfa7cc8b4R62). Which one should I do? ><

@yamgent

CanIHasReview:
# v2

* [v2 1/4] [AddressBookSystemTest: Update header comments](https://github.com/se-edu/addressbook-level4/pull/636/commits/40ea507905f6e136114690e6df381a2bd49a47cc)
* [v2 2/4] [Convert SelectCommandTest to SelectCommandSystemTest](https://github.com/se-edu/addressbook-level4/pull/636/commits/4b67ea75729eed59ef9b863ea3db48bddeadf31d)
* [v2 3/4] [SelectCommandSystemTest: Add additional tests](https://github.com/se-edu/addressbook-level4/pull/636/commits/fa962cc793f00abba5211006ab573a9e320281ed)
* [v2 4/4] [AddressBookSystemTest: Update assertSelectedCardChange\(Index\)](https://github.com/se-edu/addressbook-level4/pull/636/commits/c8aa559991475332204d5c3c08ea16a51c2ed374)

([:books: Archive](https://github.com/se-edu/addressbook-level4/compare/e32a1e791badb4acfe80365177b0600826dcfee1...c8aa559991475332204d5c3c08ea16a51c2ed374)) ([:chart_with_upwards_trend: Interdiff between v1 and v2](https://github.com/se-edu/addressbook-level4/compare/c04606b77d4ee6b42f7b3b1861e4c611f48babe8~1...c04606b77d4ee6b42f7b3b1861e4c611f48babe8))
<details>

```sh
git fetch https://github.com/se-edu/addressbook-level4.git refs/pr/636/2/head:BRANCHNAME
```

</details>

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,138 @@
...
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.logic.commands.UndoCommand;
+import seedu.address.model.AddressBook;
+import seedu.address.model.Model;
+import seedu.address.model.ModelManager;
+import seedu.address.model.UserPrefs;
+
+public class SelectCommandSystemTest extends AddressBookSystemTest {
+    @Test
+    public void select_nonEmptyList() throws Exception {
```
Zhiyuan-Amos: remove _nonEmptyList()

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,138 @@
...
+                String.format(MESSAGE_INVALID_COMMAND_FORMAT, SelectCommand.MESSAGE_USAGE));
+
+        /* Case: invalid arguments (extra argument) -> rejected */
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " 1 abc",
+                String.format(MESSAGE_INVALID_COMMAND_FORMAT, SelectCommand.MESSAGE_USAGE));
+
+        /* Case: mixed case command word -> rejected */
+        assertCommandFailure("SeLeCt 1", MESSAGE_UNKNOWN_COMMAND);
+
+        /* Case: select from empty list -> rejected */
```
Zhiyuan-Amos: `empty address book` to be more precise

eugenepeh: hmm... how about replacing all the `list` with `person list`? there's a lot of `list` in the comments above ^

Zhiyuan-Amos: `person list` is a good idea.

I'll make a distinction between what the user sees (`person list`), and the actual model (`address book`) so that it's clearer. E.g., `select from the person list` and `clears the address book`. ^^

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -60,12 +98,17 @@ private void assertCommandSuccess(String command, Index expectedSelectedCardInde
...
         String expectedResultMessage = String.format(
                 MESSAGE_SELECT_PERSON_SUCCESS, expectedSelectedCardIndex.getOneBased());
+        int currentSelectedCardIndex = getPersonListPanel().getSelectedCardIndex();
 
         executeCommand(command);
         assertApplicationDisplaysExpected("", expectedResultMessage, expectedModel);
 
-        waitUntilBrowserLoaded(getBrowserPanel());
-        assertSelectedCardChanged(expectedSelectedCardIndex);
+        if (currentSelectedCardIndex != expectedSelectedCardIndex.getZeroBased()) {
```
yamgent: `currentSelectedCardIndex` can be quite confusing (readers might think that it is the current index **after** the command is executed), so needs a better name.

Also, after changing the name, perhaps the if-else logic will be easier to read if we deal with `==` first?

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,97 @@
...
+import seedu.address.model.ModelManager;
+import seedu.address.model.UserPrefs;
+
+public class SelectCommandSystemTest extends AddressBookSystemTest {
+    @Test
+    public void select_nonEmptyList() throws Exception {
+        int invalidIndex = getTestApp().getModel().getFilteredPersonList().size() + 1;
+
+        /* Case: invalid index (size + 1) -> rejected */
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex);
```
yamgent: Any reason why this is the first one of all tests? :P

Zhiyuan-Amos: no reason  ><

Zhiyuan-Amos:
@yamgent @eugenepeh gentle reminder :)

Zhiyuan-Amos:
@yamgent @eugenepeh gentle reminder :)

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -55,6 +54,20 @@ public void select() throws Exception {
         /* Case: select the current selected card -> selected */
         assertCommandSuccess(command, middleIndex);
 
+        /* Case: filtered person list, select index in bounds of address book but out of bounds of person list
+        * -> rejected */
+        executeCommand(FindCommand.COMMAND_WORD + " Meier");
```
eugenepeh: I'm thinking, should we put `Meier` in `TypicalPersons` class as `typicalSearchPerson` or `typicalSearchTerm` or something so that future changes on test data can be made all in 1 place.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assert getTestApp().getModel().getAddressBook().getPersonList().size() == 0;
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+    }
+
+    /**
+     * Executes {@code command} and verifies that the command box displays an empty string, the result display
+     * box displays the success message of executing select command with the {@code expectedSelectedCardIndex}
+     * of the selected person, and the model related components equal to the current model.
+     * These verifications are done by
+     * {@code AddressBookSystemTest#assertCommandExecution(String, String, String, Model)}.<br>
```
yamgent: `AddressBookSystemTest#assertCommandExecution`

This method doesn't exist, and is not used in the logic (should be `assertApplicationDisplaysExpected` instead?).

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertStatusBarUnchanged();
+
+        clockRule.setInjectedClockToCurrentTime();
+    }
+
+    /**
+     * Executes {@code command} and verifies that the command box displays {@code command}, the result display
+     * box displays {@code Messages#MESSAGE_INVALID_PERSON_DISPLAYED_INDEX} and the model related components equal to
+     * the current model. These verifications are done by
+     * {@code AddressBookSystemTest#assertCommandExecution(String, String, String, Model)}.<br>
```
yamgent: Same as above.

Zhiyuan-Amos:
> Looks good to me, only thing I would like to solve is the assertCommandSuccess... would be nice if all SystemTest classes follow a single model.

This was discussed here ^^ https://github.com/se-edu/addressbook-level4/pull/620#discussion_r131940717

src/test/java/systemtests/SelectCommandSystemTest.java
Zhiyuan-Amos: @yamgent this sounds good, though I'm not sure what's a good name for it because we somehow have to specify that this search term by default should return 2 results XD

yamgent: Typical person alone makes it quite hard to understand what we are supposed to expect from the person. If we really need to extract the string constant then can just make the content as the name:
```java
public static String MEIER = "Meier";
```

Zhiyuan-Amos:
Made some minor changes ^^ @yamgent for your approval and to tag prof if you're ok!

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,145 @@
...
+import seedu.address.logic.commands.RedoCommand;
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.logic.commands.UndoCommand;
+import seedu.address.model.Model;
+
+public class SelectCommandSystemTest extends AddressBookSystemTest {
+    @Test
+    public void select() {
+        /* Case: select the first card in the person list, command with leading spaces and trailing spaces
+        * -> selected */
```
damithc: This `*` is in the wrong place? Should be like this instead?
```java
/* Case: select the first card in the person list, command with leading spaces and trailing spaces
 * -> selected 
 */

Comment (damithc) [reply src/test/java/systemtests/SelectCommandSystemTest.java]: On a related note, is it better to use this instead? I remember reading somewhere that there is no good reason to use `/* ... */` style comments.

```java
\\Case: select the first card in the person list, command with leading spaces and trailing spaces
\\    -> selected 
```

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        /* Case: filtered person list, select index in bounds of address book but out of bounds of person list
+        * -> rejected */
+        executeCommand(FindCommand.COMMAND_WORD + " " + MEIER);
+        assert getTestApp().getModel().getFilteredPersonList().size()
+                < getTestApp().getModel().getAddressBook().getPersonList().size();
+        invalidIndex = getTestApp().getModel().getAddressBook().getPersonList().size();
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex, MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+
+        /* Case: filtered person list, select index in bounds of address book and person list -> selected */
```
damithc: `in bounds` -> `within bounds`?

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,145 @@
...
+     * error style.
+     * @see AddressBookSystemTest#assertApplicationDisplaysExpected(String, String, Model)
+     */
+    private void assertCommandFailure(String command, String expectedResultMessage) {
+        Model expectedModel = prepareModelFilteredList(getTestApp().getModel().getFilteredPersonList());
+
+        executeCommand(command);
+        assertApplicationDisplaysExpected(command, expectedResultMessage, expectedModel);
+        assertSelectedCardUnchanged();
+        assertCommandBoxStyleError();
```
damithc: clearer to say `assertCommandBoxShowsErrorStyle()` ?

src/test/java/seedu/address/testutil/TypicalPersons.java
```diff
@@ -14,6 +14,7 @@
 public class TypicalPersons {
 
     public static final ReadOnlyPerson ALICE, BENSON, CARL, DANIEL, ELLE, FIONA, GEORGE, HOON, IDA;
+    public static final String MEIER = "Meier"; // search term used for filtering
```
damithc: This comment is about how the value is being used. Instead, explain what the object represents. e.g. 
`String KEYWORD_MATCHING_MEIER = "Meier"; // A keyword that matches MEIER`

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -48,6 +53,8 @@
     private static final List<String> COMMAND_BOX_ERROR_STYLE =
             Arrays.asList("text-input", "text-field", CommandBox.ERROR_STYLE_CLASS);
 
+    private static final Predicate<ReadOnlyPerson> PREDICATE_SHOW_NO_PERSONS = unused -> false;
```
damithc: A predicate does not know anything about 'showing'. `PREDICATE_MATCHING_NO_PERSONS`?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -113,6 +120,31 @@ protected void executeCommand(String command) {
...
+     * @see AddressBookSystemTest#prepareModelFilteredList(List)
+     */
+    protected Model prepareModelFilteredList(ReadOnlyPerson... displayedPersons) {
+        return prepareModelFilteredList(Arrays.asList(displayedPersons));
+    }
+
+    /**
+     * Returns a {@code Model} backed by {@code TestApp}'s address book, displaying only {@code displayedPersons}.
+     */
+    protected Model prepareModelFilteredList(List<ReadOnlyPerson> displayedPersons) {
```
damithc: See if this can be private.

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -113,6 +120,31 @@ protected void executeCommand(String command) {
...
+        Optional<Predicate<ReadOnlyPerson>> predicate =
+                displayedPersons.stream().map(this::personEquals).reduce(Predicate::or);
+        model.updateFilteredPersonList(predicate.orElse(PREDICATE_SHOW_NO_PERSONS));
+        return model;
+    }
+
+    /**
+     * Returns a predicate that evaluates to true if this {@code ReadOnlyPerson} equals to {@code other}.
+     */
+    private Predicate<ReadOnlyPerson> personEquals(ReadOnlyPerson other) {
```
damithc: `getPredicateMatching(...)`?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -113,6 +120,31 @@ protected void executeCommand(String command) {
     }
 
     /**
+     * @see AddressBookSystemTest#prepareModelFilteredList(List)
+     */
+    protected Model prepareModelFilteredList(ReadOnlyPerson... displayedPersons) {
+        return prepareModelFilteredList(Arrays.asList(displayedPersons));
+    }
```
damithc: Looks like this method is both a command (i.e. update the filtered list in the model) and a query (i.e. get the model). See if they can be separated. https://en.wikipedia.org/wiki/Command%E2%80%93query_separation

Zhiyuan-Amos: Resolved it in v5 [4/5] by updating `TestApp#getModel()`.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,145 @@
...
+import seedu.address.logic.commands.RedoCommand;
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.logic.commands.UndoCommand;
+import seedu.address.model.Model;
+
+public class SelectCommandSystemTest extends AddressBookSystemTest {
+    @Test
+    public void select() {
+        // Case: select the first card in the person list, command with leading spaces and trailing spaces
+        //     -> selected
```
damithc: Given we are using /* */ style in other cases, let's use the same here. Just format it correctly. 
```java
/* aaa
 * bbb
 */
```

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        /* Case: invalid index (size + 1) -> rejected */
+        int invalidIndex = getTestApp().getModel().getFilteredPersonList().size() + 1;
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex, MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
+
+        /* Case: select the current selected card -> selected */
+        assertCommandSuccess(command, middleIndex);
+
+        // Case: filtered person list, select index within bounds of address book but out of bounds of person list
+        //     -> rejected
```
damithc: And here.

src/test/java/systemtests/SelectCommandSystemTest.java
Zhiyuan-Amos: Ah ok! Sorry for this noob-ish mistake, cos we usually do:
```
/**
 *
 */
``` 
(There's 2 * on the first line). So now that there's only 1 star, I wasn't sure what's our formatting convention.

damithc: /** is a javadoc comment whereas /* is a regular multi-line comment.

PR #216, opened 2016-12-18
Update GUI tests to verify 'last updated' feature #114

chao1995:
Fixes #114 

Current StatusBarFooterTest includes
- after a successful command, the sync status text has been updated
- after an invalid command, the sync status text is not updated

Some issues:
- only tested AddCommand.

Testing for save location in the status bar can be added later in [another PR](https://github.com/se-edu/addressbook-level4/pull/212) using SaveAsCommand, but some methods (e.g `getSaveLocation`) have already been included in this PR.

Ready for review.

Proposed merge commit message:

> Let's add GUI tests to verify the 'last updated' feature of the status
bar footer.

PR #645, opened 2017-08-24
Tidy up System Tests #644

### LINKED ISSUE #642: AddressBookSystemTest: Missing wait for browser to load
When TestApp is initialised, the default browser url page is loaded.

There is no wait to ensure that this page is loaded before 
assertApplicationStartingStateIsCorrect() is called. This causes a race
condition: if the page has not been successfully loaded before
assertApplicationStartingStateIsCorrect(), this assertion method will
fail.

Let's teach AddressBookSystemTest to wait for the browser to load 
before calling assertApplicationStartingStateIsCorrect().

### LINKED ISSUE #644: Tidy up System Tests
There are some areas of code in system tests that have minor bugs and
code issues. For e.g. AddressBookSystemTest's header comments is 
outdated and TestApp#getModel() is returning the wrong model.

Let's fix these minor issues.

Zhiyuan-Amos:
Fixes #644, also fixes #642 

```
There are some areas of code in system tests that have minor bugs and
code issues. For e.g. AddressBookSystemTest's header comments is
outdated and TestApp#getModel() is returning the wrong model.

Let's fix these minor issues.
```

### DIFF
--- a/src/test/java/guitests/guihandles/BrowserPanelHandle.java
+++ b/src/test/java/guitests/guihandles/BrowserPanelHandle.java
@@ -16,7 +16,7 @@ public class BrowserPanelHandle extends NodeHandle<Node> {
 
     public static final String BROWSER_ID = "#browser";
 
-    private boolean isWebViewLoaded = false;
+    private boolean isWebViewLoaded = true;
 
     private URL lastRememberedUrl;
 
@@ -26,7 +26,9 @@ public class BrowserPanelHandle extends NodeHandle<Node> {
         WebView webView = getChildNode(BROWSER_ID);
         WebEngine engine = webView.getEngine();
         new GuiRobot().interact(() -> engine.getLoadWorker().stateProperty().addListener((obs, oldState, newState) -> {
-            if (newState == Worker.State.SUCCEEDED) {
+            if (newState == Worker.State.RUNNING) {
+                isWebViewLoaded = false;
+            } else if (newState == Worker.State.SUCCEEDED) {
                 isWebViewLoaded = true;
             }
         }));
@@ -54,11 +56,10 @@ public class BrowserPanelHandle extends NodeHandle<Node> {
         return !lastRememberedUrl.equals(getLoadedUrl());
     }
 
-    public boolean getIsWebViewLoaded() {
+    /**
+     * Returns true if the browser is done loading a page, or if this browser has yet to load any page.
+     */
+    public boolean isLoaded() {
         return isWebViewLoaded;
-    }
-
-    public void setIsWebViewLoaded(boolean isWebViewLoaded) {
-        this.isWebViewLoaded = isWebViewLoaded;
     }
 }
--- a/src/test/java/guitests/guihandles/WebViewUtil.java
+++ b/src/test/java/guitests/guihandles/WebViewUtil.java
@@ -19,10 +19,9 @@ public class WebViewUtil {
     }
 
     /**
-     * Sleeps the thread till the {@code browserPanelHandle}'s {@code WebView} is successfully loaded.
+     * If the {@code browserPanelHandle}'s {@code WebView} is loading, sleeps the thread till it is successfully loaded.
      */
     public static void waitUntilBrowserLoaded(BrowserPanelHandle browserPanelHandle) {
-        new GuiRobot().waitForEvent(browserPanelHandle::getIsWebViewLoaded);
-        browserPanelHandle.setIsWebViewLoaded(false);
+        new GuiRobot().waitForEvent(browserPanelHandle::isLoaded);
     }
 }
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -102,6 +102,10 @@ public abstract class AddressBookSystemTest {
      */
     protected void executeCommand(String command) throws Exception {
         rememberStates();
+        // Injects a fixed clock before executing a command so that the time stamp shown in the status bar
+        // after each command is predictable and also different from the previous command.
+        clockRule.setInjectedClockToCurrentTime();
+
         mainWindowHandle.getCommandBox().run(command);
 
         waitUntilBrowserLoaded(getBrowserPanel());
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -35,8 +35,8 @@ import seedu.address.model.Model;
 import seedu.address.ui.CommandBox;
 
 /**
- * A system test class for AddressBook, which provides access to handles of GUI components, and
- * verifies that the starting state of the application is correct.
+ * A system test class for AddressBook, which provides access to handles of GUI components and helper methods
+ * for test verification.
  */
 public abstract class AddressBookSystemTest {
     @ClassRule
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -1,5 +1,6 @@
 package systemtests;
 
+import static guitests.guihandles.WebViewUtil.waitUntilBrowserLoaded;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static seedu.address.ui.BrowserPanel.DEFAULT_PAGE;
@@ -96,10 +97,13 @@ public abstract class AddressBookSystemTest {
 
     /**
      * Executes {@code command} in the application's {@code CommandBox}.
+     * Method returns after UI components have been updated.
      */
     protected void executeCommand(String command) throws Exception {
         rememberStates();
         mainWindowHandle.getCommandBox().run(command);
+
+        waitUntilBrowserLoaded(getBrowserPanel());
     }
 
     /**
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -62,6 +62,7 @@ public abstract class AddressBookSystemTest {
         testApp = setupHelper.setupApplication();
         mainWindowHandle = setupHelper.setupMainWindowHandle();
 
+        waitUntilBrowserLoaded(getBrowserPanel());
         assertApplicationStartingStateIsCorrect();
     }
 
--- a/src/test/java/seedu/address/TestApp.java
+++ b/src/test/java/seedu/address/TestApp.java
@@ -13,7 +13,6 @@ import seedu.address.commons.util.FileUtil;
 import seedu.address.commons.util.XmlUtil;
 import seedu.address.model.AddressBook;
 import seedu.address.model.Model;
-import seedu.address.model.ModelManager;
 import seedu.address.model.ReadOnlyAddressBook;
 import seedu.address.model.UserPrefs;
 import seedu.address.storage.UserPrefsStorage;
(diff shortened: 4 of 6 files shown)


src/test/java/guitests/guihandles/WebViewUtil.java
```diff
@@ -19,10 +19,9 @@ public static URL getLoadedUrl(WebView webView) throws MalformedURLException {
     }
 
     /**
-     * Sleeps the thread till the {@code browserPanelHandle}'s {@code WebView} is successfully loaded.
+     * If the {@code browserPanelHandle}'s {@code WebView} is loading, sleeps the thread till it is successfully loaded.
      */
     public static void waitUntilBrowserLoaded(BrowserPanelHandle browserPanelHandle) {
```
Zhiyuan-Amos: I'm not sure if the method name tallies with the header comments.

Zhiyuan-Amos:
Marking this with critical priority because it involves slight architectural changes that will affect the other PRs concerning system tests.

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -96,9 +98,20 @@ public ResultDisplayHandle getResultDisplay() {
...
         rememberStates();
+        executeCommandNow(command);
+
+        waitUntilBrowserLoaded(getBrowserPanel());
+    }
+
+    /**
+     * Updates the application's clock to the current time and executes {@code command}.
+     */
+    private void executeCommandNow(String command) {
```
Zhiyuan-Amos: can't think of a better method name ><

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -16,7 +16,8 @@
 
     public static final String BROWSER_ID = "#browser";
 
-    private boolean isWebViewLoaded = false;
+    // This variable is a substitute for using Semaphore as Semaphore isn't taught in syllabus
```
yamgent: I think can remove. This is not a proper substitute for a semaphore so the comment is kinda wrong anyway (the variable is not even protected by multi-thread access :P).

src/test/java/systemtests/AddressBookSystemTest.java
yamgent: I think this name is OK.

Zhiyuan-Amos:
@yamgent changes requested made :P

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -96,9 +98,20 @@ public ResultDisplayHandle getResultDisplay() {
...
+        clockRule.setInjectedClockToCurrentTime();
         mainWindowHandle.getCommandBox().run(command);
     }
```
damithc: Why does this need to be a separate method? Looks like it can be inlined.

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -96,9 +98,20 @@ public ResultDisplayHandle getResultDisplay() {
...
      */
     protected void executeCommand(String command) throws Exception {
         rememberStates();
+        executeCommandNow(command);
+
+        waitUntilBrowserLoaded(getBrowserPanel());
+    }
+
+    /**
+     * Updates the application's clock to the current time and executes {@code command}.
```
damithc: Isn't it misleading that `Updates the application's clock to the current time` is not mentioned in the `executeCommand` method's header comment?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -96,9 +98,20 @@ public ResultDisplayHandle getResultDisplay() {
...
+     */
+    private void executeCommandNow(String command) {
+        clockRule.setInjectedClockToCurrentTime();
```
damithc: When I'm reading the code of this class, I'm a bit puzzled by this line. Might need some explanation on why we are doing this here. Still, it is better to have this line here rather than in the child classes.

On a related note, I've lost track of how the clock injection works now. Can give a brief abstract explanation of how it works, for my own knowledge?

Zhiyuan-Amos: 1. When we run a test method, `ClockRule` injects a fixed clock. A fixed clock works in such a way that it always prints out a fixed time.
1. In our current implementation, after each command is executed (we execute multiple commands in each test method), the status bar is always being updated to the fixed time.
1. As such, we have only effectively tested that upon first command execution, the timing is updated correctly. However, we have not tested that subsequent commands will update the clock accordingly as well.
1. Thus, we have to inject a new fixed clock every time before we execute a command, which causes the status bar to be updated to a new timing after each command is executed.

Other possible solution:
We can choose to not inject a clock (the reason why `StatusBarFooter` has a `setClock(Clock)` method is solely for testing purposes. Using this solution allows us to remove unnecessary code in the main code base. Previous discussion was done here: https://github.com/se-edu/addressbook-level4/pull/216#discussion_r96589149), but we give a +1 second leeway for verifying the timing. E.g. if the command was executed on 18:51:01, we should expect the test to verify by 18:51:02.

src/test/java/systemtests/AddressBookSystemTest.java
Zhiyuan-Amos: Will inline the method as per your suggestion below. This will then no longer be an issue.

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -96,10 +98,14 @@ public ResultDisplayHandle getResultDisplay() {
 
     /**
      * Executes {@code command} in the application's {@code CommandBox}.
+     * Method returns after UI components have been updated.
      */
     protected void executeCommand(String command) throws Exception {
         rememberStates();
+        clockRule.setInjectedClockToCurrentTime();
```
damithc: We can put an explanatory comment here. e.g.
```java
//Inject a fixed clock before executing a command so that the time stamp shown in the status bar 
//after each command is predictable and also different from the previous command.
```

Regarding whether we should inject a clock at all,
* I prefer not to use the buffer/leeway approach (e.g. 1 second) because in theory it is not foolproof. Although injecting a clock is a bit of overkill, it can serve as a good example that technique. 
* If clock injection is used in StatusBar unit tests, it's OK (even preferable, because system tests should use the system as it is) not to use clock injection in system tests and instead use a slightly weaker verification such as checking if the new time stamp is later than the previous time stamp. But this can be a separate PR.

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -26,6 +26,10 @@ public BrowserPanelHandle(Node browserPanelNode) {
         WebView webView = getChildNode(BROWSER_ID);
         WebEngine engine = webView.getEngine();
         new GuiRobot().interact(() -> engine.getLoadWorker().stateProperty().addListener((obs, oldState, newState) -> {
+            if (newState == Worker.State.RUNNING) {
+                isWebViewLoaded = false;
+            }
+
             if (newState == Worker.State.SUCCEEDED) {
                 isWebViewLoaded = true;
             }
```
damithc: Shouldn't this be `else if`?

src/test/java/systemtests/AddressBookSystemTest.java
Zhiyuan-Amos: > I prefer not to use the buffer/leeway approach (e.g. 1 second) because in theory it is not foolproof.

We use a similar approach when callers call `GuiRobot#waitForEvent(BooleanSupplier event, int timeOut)` (though not entirely the same) 

For example, we wait for 5 seconds for the `BrowserPanelHandle` to load when `WebViewUtil#waitUntilBrowserLoaded(BrowserPanelHandle)` is called. If after 5 seconds and the url isn't loaded, a `RunTimeException` is thrown. So it isn't really foolproof as well, because perhaps the person's internet connection may be cut off for a few seconds, and the test will fail as well. :(

damithc: Yes, that one is bad too. :(
Also, this kind of indiscriminate waits adds up and lengthen the test duration. So, if they are used at all, they should not be used in code that is executed frequently during tests.

Zhiyuan-Amos: > Also, this kind of indiscriminate waits adds up and lengthen the test duration.

Actually this isn't an indiscriminate wait because the wait only happens if _we are expecting the Browser to load a url_. As such, even though we call `waitUntilBrowserLoaded(BrowserPanel)` every time we execute a command, there will not be a pause if the Browser isn't loading a url ^^ (though speaking of that, the naming of the method sounds weird because there isn't the connotation of "only wait if condition met".)

And if we are expecting the Browser to load a url, there are periodic checks i.e. we do not usually spend 5 seconds waiting.

Though I get it that this kind of testing is still not foolproof i.e. not determinant. If we are expecting the browser to load a new url, then we should keep waiting until it is loaded, instead of throwing a `RunTimeException` upon timeout. @yamgent thoughts?

damithc: By _indiscriminate_ I meant we wait for 5 seconds even if the browser takes only 2 seconds to load. The safety margin is actually time being wasted. A few times is ok, but 1 second wasted in a method called 60 times is a minute wasted.

Zhiyuan-Amos:
```
public void waitForEvent(BooleanSupplier event, int timeOut) {
    int timePassed = 0;
    final int retryInterval = 50;

    while (!event.getAsBoolean()) {
        sleep(retryInterval);
        timePassed += retryInterval;

        if (timePassed >= timeOut) {
            throw new EventTimeoutException();
        }
    }
}
```

We don't actually wait for 5 seconds even though the browser takes only 2 seconds to load. 
The method above from `GuiRobot` checks every 50 milliseconds, and breaks from the while loop if the condition is true. So the worst case scenario is when the browser finished loading in 2 seconds, and the next verification happens 50 ms after so the method only returns after 2.05s total :P (which is quite a short wait).

The 5 seconds is the timeout value i.e. if the browser hasn't loaded after 5s, an Exception will be thrown.

Sorry if I confused you prof :P

damithc: That kind of 'retry until timeout' is certainly fine.

yamgent: > If we are expecting the browser to load a new url, then we should keep waiting until it is loaded, instead of throwing a RunTimeException upon timeout. @yamgent thoughts?

The timeout is implemented as a safeguard in the event that the event never occurs (e.g. the browser never finishes loading for whatever possible reasons). The timeout is supposed to be lenient in that regard by providing ample time for the event to occur, while still preventing the test from stalling.

Zhiyuan-Amos: > The timeout is supposed to be lenient in that regard by providing ample time for the event to occur, while still preventing the test from stalling.

Ah yes you are right. Having this allows Travis & Appveyor to fail immediately instead of stalling till the timeout occurs for each server. 

> I prefer not to use the buffer/leeway approach (e.g. 1 second) because in theory it is not foolproof.

Sorry prof, so what do you mean that it isn't foolproof? >< Though I'm definitely ok with the idea of having a "slightly weaker verification such as checking if the new time stamp is later than the previous time stamp."

damithc: I meant it's always possible (in theory) for the task to take longer than the time we allow for it. That means the code is working correctly but the test fails.

PR #652, opened 2017-09-01
Remove UnmodifiableObservableList #649

### LINKED ISSUE #649: Remove UnmodifiableObservableList
JavaFX has a class ~~ReadOnlyListPropertyBase<E>~~ `ReadOnlyListWrapper<E>` which works similar to `UnmodifiableObservableList`.

Zhiyuan-Amos:
Fixes #649 

```
FXCollections#unmodifiableObservableList(ObservableList<E>) returns an
unmodifiable wrapper list on top of provided observable list. This 
works similarly to UnmodifiableObservableList in our code base.

There is no need for us to use our own version of
UnmodifiableObservableList.

Let's update classes to call this method in FXCollections to create an
unmodifiable ObservableList instead of using 
UnmodifiableObservableList. Then, let's delete 
UnmodifiableObservableList.
```


Zhiyuan-Amos:
@damithc @yamgent made a PR to roughly show how things will look like. Take special note on commit 4.

Zhiyuan-Amos:
Probably have to add comments on `Model`'s components to inform callers that the `ObservableList` returned is unmodifiable.

Zhiyuan-Amos:
V.3: Made the changes mentioned in https://github.com/se-edu/addressbook-level4/pull/652#issuecomment-326717337

@PierceAndy & @MightyCupcakes, are both of you still available to do reviews? ^^

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -83,11 +84,12 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedPerson)
     //=========== Filtered Person List Accessors =============================================================
 
     /**
-     * Return a list of {@code ReadOnlyPerson} backed by the internal list of {@code addressBook}
+     * Return an unmodifiable view of the list of {@code ReadOnlyPerson} backed by the internal list of
```
PierceAndy: Return -> Returns

src/main/java/seedu/address/logic/Logic.java
```diff
@@ -19,8 +19,8 @@
      */
     CommandResult execute(String commandText) throws CommandException, ParseException;
 
-    /** Returns the filtered list of persons */
-    ObservableList<ReadOnlyPerson> getFilteredPersonList();
+    /** Returns an unmodifiable view of the filtered list of persons */
+    ObservableList<? extends ReadOnlyPerson> getFilteredPersonList();
```
MightyCupcakes: Any reason for changing this to use a generic wildcard?

Zhiyuan-Amos: There will be a compilation error otherwise. That's because `UniquePersonList` internally has a `ObservableList<Person>`, and in the method `asObservableList()`, we are returning a `ObservableList<ReadOnlyPerson>`. Since the types are different, there's a compilation error. As such, `asObservableList()` has to be updated to return `ObservableList<? extends ReadOnlyPerson>`. 

This change is then propagated up all the way, thus in this case `Logic` is affected as well :>

Zhiyuan-Amos:
@PierceAndy @MightyCupcakes thank you for your reviews! I've made the changes requested ^^

src/main/java/seedu/address/logic/Logic.java
pyokagan: @Zhiyuan-Amos 
> Since the types are different, there's a compilation error. As such, asObservableList() has to be updated to return ObservableList<? extends ReadOnlyPerson>.

Isn't this a good reason *for* UnmodifiableObservableList? So that we do not need to return wildcard types unnecessarily?

See https://stackoverflow.com/questions/22815023/generic-wildcard-types-should-not-be-used-in-return-parameters for why it might not be a good idea to return wildcard types.

Zhiyuan-Amos: > Isn't this a good reason for UnmodifiableObservableList? So that we do not need to return wildcard types unnecessarily?

Yup I agree that having `UnmodifiableObservableList` solves this problem, however having so many additional lines of code (400 lines) for this 1 purpose seems a bit of an over-kill to me.

There's another link that say's it's fine to return bounded wildcard types: https://stackoverflow.com/questions/8966502/java-bounded-wildcard-in-return-type.

Though of course senpai Paul has the final say ^^. Another way to implement this is that I an map the `ObservableList<Person>` in `UniquePersonList` to a `ObservableList<ReadOnlyPerson>` as a field? Do you think this way is fine?

Zhiyuan-Amos: @pyokagan since Effective Java is quite an authoritative source I suppose, I'll go and change my implementation. Much thanks for your comment :)

Zhiyuan-Amos:
@PierceAndy & @MightyCupcakes, do let me know if you will still be taking a look at this PR. Otherwise I'll ask prof to review straight. Much thanks! :)

PierceAndy:
v5 looks fine, but my review on [v3 1/4]'s commit message has not been addressed yet.

Zhiyuan-Amos:
@damithc for your review :)

Zhiyuan-Amos:
Updated commit message of [3/4] as well. @PierceAndy for your final review before merging, much thanks! ^^

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -93,8 +96,11 @@ public void setPersons(List<? extends ReadOnlyPerson> persons) throws DuplicateP
         setPersons(replacement);
     }
 
-    public UnmodifiableObservableList<Person> asObservableList() {
-        return new UnmodifiableObservableList<>(internalList);
+    /**
+     * Returns the backing list as an unmodifiable {@code ObservableList}.
+     */
+    public ObservableList<ReadOnlyPerson> asObservableList() {
```
Zhiyuan-Amos: @damithc we should rename this method to be `asUnmodifiableObservableList`?

damithc: Not sure myself. Are `ObservableList`s usually also unmodifiable? If they are, we don't have to mention it specifically. If they are not, it may be helpful to rename as suggested.

Zhiyuan-Amos: Some `ObservableList` are modifiable. I only know that `FilteredList` is unmodifiable, and `FXCollections.observableArrayList()` returns a modifiable `ObservableList`.

Since there's no guarantee that the `ObservableList` is unmodifiable, I think we should rename it.

PR #664, opened 2017-09-08
Cleanup DeleteCommandSystemTest

Zhiyuan-Amos:
```
DeleteCommandSystemTest can be cleaner. For example, the header 
comments can be updated to be written in point forms, and the test 
cases can be segmented into different categories.

This causes it to be harder for future developer to understand the 
code.

Let's clean up DeleteCommandSystemTest.
```

### DIFF
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -33,6 +33,9 @@ import seedu.address.MainApp;
 import seedu.address.TestApp;
 import seedu.address.commons.core.EventsCenter;
 import seedu.address.commons.core.index.Index;
+import seedu.address.logic.commands.FindCommand;
+import seedu.address.logic.commands.ListCommand;
+import seedu.address.logic.commands.SelectCommand;
 import seedu.address.model.Model;
 import seedu.address.ui.CommandBox;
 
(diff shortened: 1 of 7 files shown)


src/test/java/systemtests/DeleteCommandSystemTest.java
```diff
@@ -29,6 +29,8 @@
 
     @Test
     public void delete() {
+        /* ----------------- Performing delete operation while an unfiltered list is being shown -------------------- */
```
eugenepeh: woah that's alot of effort to keep them all same length, is there better way to do this? 😱

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -113,6 +115,29 @@ protected void executeCommand(String command) {
...
+    protected void showPersonsWithName(String keyword) {
+        executeCommand(FindCommand.COMMAND_WORD + " " + keyword);
+        assert getModel().getFilteredPersonList().size() < getModel().getAddressBook().getPersonList().size();
+    }
+
+    /**
+     * Selects the person at {@code index} of the displayed list.
+     */
+    protected void selectPerson(Index index) {
+        getPersonListPanel().select(index.getZeroBased());
```
eugenepeh: is the assertion done within this?

src/test/java/systemtests/DeleteCommandSystemTest.java
Zhiyuan-Amos: I don't really keep them all the same length actually XD Just rough approximate.

src/test/java/systemtests/AddressBookSystemTest.java
Zhiyuan-Amos: there's no need for assertion because I'm selecting using `PersonListPanelHandle#select(Index)`; the method assures that the card at the index is selected.

Whereas for the other methods, cos I'm causing `executeCommand(String command)`, there's no assurance that the command executed will perform what I want it to do, so the assertion is required.

Zhiyuan-Amos: In the future I intend to make API methods to update the actual model directly without having to go through execution of commands.

Zhiyuan-Amos:
@damithc @eugenepeh @yamgent:

I have updated [6/7] because I remembered that `executeCommand(String)` doesn't only execute a command, it also does things like `rememberStates()` etc. As such, I will do the conversion as mentioned https://github.com/se-edu/addressbook-level4/pull/664#discussion_r138782762 all together in the future.

See https://github.com/se-edu/addressbook-level4/pull/664#discussion_r138781391 and interdiff for a bit more context.

Zhiyuan-Amos:
Sorry @damithc I chose "rebased and merge" and I didn't know that it will actually merge each commit 1 by 1 :O should I revert and redo this?

damithc:
Too late now. Never mind. Don't choose that option in future. If you want to rebase, you need to rebase locally.

Issue #294, opened 2017-02-19
TestUtil#assertThrows(): support checked exceptions as well


pyokagan:
`TestUtil#assertThrows()` currently does not support functions which throw checked exceptions.

e.g.
```java
TestUtil.assertThrows(IllegalValueException.class, () -> { throw new IllegalValueException(""); });
```
does not compile.

It would be useful if it did.

pyokagan:
Marking as low priority since we do not actually need it now. Just putting this down as an issue so that we do not forget it.

pyokagan:
Fixed in 21a1955a4c0b4cd085107a092dec13f66e86af66

Issue #634, opened 2017-08-18
Replace all verbose loops with lambda expression


eugenepeh:
Sometimes, defining a whole class to compare by the ith element is a bit verbose if we're only ever going to use it once to sort a vector. For example, `LogsCenter#removeHandlers(Logger)` can be simplified to `Arrays.stream(handlers).forEach(logger::removeHandler);`

Let's replace all these verbose instructions with simple lambda expressions.

PR #667, opened 2017-09-12
Improve separation of concerns between parser/storage layer and model layer

pyokagan:
```
The model classes Name, Email, Address, Phone and Tag perform input validation
and string trimming even though that is not the responsibility of the model
layer.

Fix this by moving the input validation and string trimming logic from these
classes to the parser/storage layer where they should rightfully be.
```

### DIFF
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -52,6 +52,9 @@ public class ParserUtil {
     public static Name parseName(String name) throws IllegalValueException {
         requireNonNull(name);
         String trimmedName = name.trim();
+        if (!Name.isValidName(trimmedName)) {
+            throw new IllegalValueException(Name.MESSAGE_NAME_CONSTRAINTS);
+        }
         return new Name(trimmedName);
     }
 
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -44,12 +44,24 @@ public class ParserUtil {
     }
 
     /**
+     * Parses a {@code String name} into a {@code Name}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given {@code name} is invalid.
+     */
+    public static Name parseName(String name) throws IllegalValueException {
+        requireNonNull(name);
+        String trimmedName = name.trim();
+        return new Name(trimmedName);
+    }
+
+    /**
      * Parses a {@code Optional<String> name} into an {@code Optional<Name>} if {@code name} is present.
      * See header comment of this class regarding the use of {@code Optional} parameters.
      */
     public static Optional<Name> parseName(Optional<String> name) throws IllegalValueException {
         requireNonNull(name);
-        return name.isPresent() ? Optional.of(new Name(name.get())) : Optional.empty();
+        return name.isPresent() ? Optional.of(parseName(name.get())) : Optional.empty();
     }
 
     /**
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -140,13 +140,25 @@ public class ParserUtil {
     }
 
     /**
+     * Parses a {@code String tag} into a {@code Tag}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given {@code tag} is invalid.
+     */
+    public static Tag parseTag(String tag) throws IllegalValueException {
+        requireNonNull(tag);
+        String trimmedTag = tag.trim();
+        return new Tag(trimmedTag);
+    }
+
+    /**
      * Parses {@code Collection<String> tags} into a {@code Set<Tag>}.
      */
     public static Set<Tag> parseTags(Collection<String> tags) throws IllegalValueException {
         requireNonNull(tags);
         final Set<Tag> tagSet = new HashSet<>();
         for (String tagName : tags) {
-            tagSet.add(new Tag(tagName));
+            tagSet.add(parseTag(tagName));
         }
         return tagSet;
     }
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -148,6 +148,9 @@ public class ParserUtil {
     public static Tag parseTag(String tag) throws IllegalValueException {
         requireNonNull(tag);
         String trimmedTag = tag.trim();
+        if (!Tag.isValidTagName(trimmedTag)) {
+            throw new IllegalValueException(Tag.MESSAGE_TAG_CONSTRAINTS);
+        }
         return new Tag(trimmedTag);
     }
 
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -124,6 +124,9 @@ public class ParserUtil {
     public static Email parseEmail(String email) throws IllegalValueException {
         requireNonNull(email);
         String trimmedEmail = email.trim();
+        if (!Email.isValidEmail(trimmedEmail)) {
+            throw new IllegalValueException(Email.MESSAGE_EMAIL_CONSTRAINTS);
+        }
         return new Email(trimmedEmail);
     }
 
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -68,12 +68,24 @@ public class ParserUtil {
     }
 
     /**
+     * Parses a {@code String phone} into a {@code Phone}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given {@code phone} is invalid.
+     */
+    public static Phone parsePhone(String phone) throws IllegalValueException {
+        requireNonNull(phone);
+        String trimmedPhone = phone.trim();
+        return new Phone(trimmedPhone);
+    }
+
+    /**
      * Parses a {@code Optional<String> phone} into an {@code Optional<Phone>} if {@code phone} is present.
      * See header comment of this class regarding the use of {@code Optional} parameters.
      */
     public static Optional<Phone> parsePhone(Optional<String> phone) throws IllegalValueException {
         requireNonNull(phone);
-        return phone.isPresent() ? Optional.of(new Phone(phone.get())) : Optional.empty();
+        return phone.isPresent() ? Optional.of(parsePhone(phone.get())) : Optional.empty();
     }
 
     /**
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -92,12 +92,24 @@ public class ParserUtil {
     }
 
     /**
+     * Parses a {@code String address} into an {@code Address}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given {@code address} is invalid.
+     */
+    public static Address parseAddress(String address) throws IllegalValueException {
+        requireNonNull(address);
+        String trimmedAddress = address.trim();
+        return new Address(trimmedAddress);
+    }
+
+    /**
      * Parses a {@code Optional<String> address} into an {@code Optional<Address>} if {@code address} is present.
      * See header comment of this class regarding the use of {@code Optional} parameters.
      */
     public static Optional<Address> parseAddress(Optional<String> address) throws IllegalValueException {
         requireNonNull(address);
-        return address.isPresent() ? Optional.of(new Address(address.get())) : Optional.empty();
+        return address.isPresent() ? Optional.of(parseAddress(address.get())) : Optional.empty();
     }
 
     /**
--- a/src/main/java/seedu/address/logic/parser/ParserUtil.java
+++ b/src/main/java/seedu/address/logic/parser/ParserUtil.java
@@ -116,12 +116,24 @@ public class ParserUtil {
     }
 
     /**
+     * Parses a {@code String email} into an {@code Email}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given {@code email} is invalid.
+     */
+    public static Email parseEmail(String email) throws IllegalValueException {
+        requireNonNull(email);
+        String trimmedEmail = email.trim();
+        return new Email(trimmedEmail);
+    }
+
+    /**
      * Parses a {@code Optional<String> email} into an {@code Optional<Email>} if {@code email} is present.
      * See header comment of this class regarding the use of {@code Optional} parameters.
      */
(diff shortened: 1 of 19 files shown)


src/main/java/seedu/address/model/person/Address.java
```diff
@@ -22,15 +18,11 @@
...
-     * Validates given address.
-     *
-     * @throws IllegalValueException if given address string is invalid.
+     * Constructs a valid address.
      */
-    public Address(String address) throws IllegalValueException {
+    public Address(String address) {
         requireNonNull(address);
-        if (!isValidAddress(address)) {
-            throw new IllegalValueException(MESSAGE_ADDRESS_CONSTRAINTS);
```
Zhiyuan-Amos: Have been thinking about this issue also, and I'm wondering whether we should remove this. I'm not too familiar with Java API, but at least for `URL`, the constructor throws a checked Exception whenever it is constructed with invalid parameters. This also agrees with what I read at https://stackoverflow.com/questions/6086334/is-it-good-practice-to-make-the-constructor-throw-an-exception.

pyokagan: > at least for URL, the constructor throws a checked Exception whenever it is constructed with invalid parameters.

1. Java standard library may not always be indicative of best practices. The URL class was introduced *JDK 1.0*, where the best practices were not so widely known yet. Various design problems in the JDK had to be corrected later (e.g. `java.time` was introduced in JDK 1.8 to replace the Calendar, Date APIs.)

2. In the case of the URL class, a `MalformedUrlException` is thrown if it cannot find a protocol handler, which depends on *external* system configuration and IO. This is an okay-ish reason for using a checked exception, but is not applicable to the Address class which does not depend on external configuration or IO.

> This also agrees with what I read at https://stackoverflow.com/questions/6086334/is-it-good-practice-to-make-the-constructor-throw-an-exception.

???? The top voted post doesn't talk about checked exceptions, and the second top voted posts recommends not to use checked exceptions.

pyokagan: > his is an okay-ish reason for using a checked exception,

Correction: since the URL constructor may load the protocol handler class from disk (i.e. performs IO), it is a *good* reason to use checked exceptions. However, I wish they provided an alternative URL class that doesn't require an internal protocol handler class. (Just parsed the protocol as-is without interpretation)

Zhiyuan-Amos: > ???? The top voted post doesn't talk about checked exceptions

Since OP was using `Exception`, and the answer didn't correct that, seems reasonable to assume that throwing checked Exception is fine.

Alright then, thanks, that has been insightful.

damithc:
> Likely a WIP for a long time.

So no chance of pushing this through over this weekend then? 😛 Just checking.

pyokagan:
@damithc 
> So no chance of pushing this through over this weekend then?

Yep, unfortunately.

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -29,6 +29,15 @@
 
     public static final String MESSAGE_INVALID_INDEX = "Index is not a non-zero unsigned integer.";
     public static final String MESSAGE_INSUFFICIENT_PARTS = "Number of parts must be more than 1.";
+    public static final String MESSAGE_NAME_CONSTRAINTS =
+            "Person names should only contain alphanumeric characters and spaces, and it should not be blank";
```
pyokagan: Probably should not move the constraint messages to `ParserUtil` as well. Otherwise, others may forget to update them when e.g. `isValidName()` is updated.

pyokagan:
[18/18] isn't necessary. Just because streams allow you to write everything in one line doesn't mean we have to.

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -10,9 +10,9 @@
     /**
      * Asserts that the {@code executable} throws the {@code expected} Exception.
```
Zhiyuan-Amos: header comments should be updated to `callable`.

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -43,13 +43,25 @@ public static Index parseIndex(String oneBasedIndex) throws IllegalValueExceptio
         return Index.fromOneBased(Integer.parseInt(trimmedIndex));
     }
 
+    /**
+     * Parses a {@code String name} into a {@code Name}.
+     * Leading and trailing whitespaces will be trimmed.
+     *
+     * @throws IllegalValueException if the given name string is invalid.
```
Zhiyuan-Amos: should we markup `name` by doing "... if the given {@code name} ~~string~~ is invalid"?

src/main/java/seedu/address/logic/parser/ParserUtil.java
```diff
@@ -43,13 +43,25 @@ public static Index parseIndex(String oneBasedIndex) throws IllegalValueExceptio
...
+     */
+    public static Name parseName(String name) throws IllegalValueException {
```
Zhiyuan-Amos: There should be tests for this method for having an invalid value (i.e. throws `IllegalValueException`) & a valid value. The valid value test can be partitioned into: trailing whitespace, leading whitespace, no trailing or leading whitespace.

src/test/java/seedu/address/testutil/PersonBuilder.java
```diff
@@ -57,11 +52,7 @@ public PersonBuilder withName(String name) {
      * Parses the {@code tags} into a {@code Set<Tag>} and set it to the {@code Person} that we are building.
      */
     public PersonBuilder withTags(String ... tags) {
-        try {
```
Zhiyuan-Amos: Since you've grouped similar changes with their respective `Move validation step from * to ParserUtil#parse*(String)` commit, I guess this should be merged with commit 11 as well.

src/main/java/seedu/address/model/person/Address.java
```diff
@@ -22,15 +21,11 @@
     public final String value;
 
     /**
-     * Validates given address.
-     *
-     * @throws IllegalValueException if given address string is invalid.
+     * Constructs a valid address.
```
damithc: Should we mention what happens if the address is invalid, or alternatively, the fact that method assumes the address to be valid. The current comment might be misinterpreted as 'constructs a valid address no matter what'.

src/main/java/seedu/address/logic/parser/ParserUtil.java
pyokagan: > The valid value test can be partitioned into: trailing whitespace, leading whitespace

I think both leading and trailing whitespace in one test will be fine. If the method doesn't trim leading whitespace OR doesn't trim trailing whitespace then it will fail anyway.

src/test/java/seedu/address/testutil/PersonBuilder.java
pyokagan: Not really, the changes in commit 12 are due to Name AND Email AND Phone AND Address AND Tag not throwing IllegalValueException any more and thus can only be done in the last commit, while the changes in the previous commits are due to only Name OR Email OR Phone... etc. not throwing IllegalValueExceptionm otherwise the Java compiler would complain. So I'd like to keep them separate.

src/main/java/seedu/address/model/person/Address.java
pyokagan: How about this?

```diff
diff --git a/src/main/java/seedu/address/model/person/Name.java b/src/main/java/seedu/address/model/person/Name.java
index 8c6566f..319cfb5 100644
--- a/src/main/java/seedu/address/model/person/Name.java
+++ b/src/main/java/seedu/address/model/person/Name.java
@@ -21,7 +21,10 @@ public class Name {
     public final String fullName;
 
     /**
-     * Constructs a valid name.
+     * Constructs a {@code Name}.
+     *
+     * @param name A valid name.
+     * @throws IllegalArgumentException if the given {@code name} is invalid.
      */
     public Name(String name) {
         requireNonNull(name);
```

damithc: Perhaps omit the `@throws` clause because it is an unchecked exception (i.e. follow a policy of not declaring unchecked exceptions)? As we have already specified that `@param name` is a valid name, we can argue that it is implied that an unchecked exception will result if that contract is broken by passing an invalid name. If we start documenting unchecked exceptions, we might open a whole another can of worms. What do you think?

pyokagan: OK, that's fine.

pyokagan:
@Zhiyuan-Amos 
> XmlAdaptedTag#toModelType()
>     XmlAdaptedPerson#toModelType()
> 
> No longer throws IllegalValueException as well.

Ah, interesting. These should probably still throw IllegalValueException since there is a chance that the data stored in the XML file is corrupted, but we'll need to add `isValidX()` calls to them as well to make sure they do their job properly.

pyokagan:
@Zhiyuan-Amos @damithc Thanks for the reviews.

pyokagan:
@damithc Could you take another look at this iteration? Thanks!

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String INVALID_EMAIL = "example.com";
+    private static final String INVALID_TAG = "#friend";
+
+    private static final String VALID_NAME = BENSON.getName().toString();
+    private static final String VALID_PHONE = BENSON.getPhone().toString();
+    private static final String VALID_EMAIL = BENSON.getEmail().toString();
+    private static final String VALID_ADDRESS = BENSON.getAddress().toString();
+
+    private final List<XmlAdaptedTag> validTags = BENSON.getTags().stream()
+            .map(tag -> new XmlAdaptedTag(tag))
```
Zhiyuan-Amos: IntelliJ tells me we can use method reference for this :P

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void toModelType_validPersonDetails_returnsPerson() throws Exception {
+        XmlAdaptedPerson person = new XmlAdaptedPerson(BENSON);
+        assertEquals(BENSON, person.toModelType());
+    }
+
+    @Test
+    public void toModelType_invalidName_throwsIllegalValueException() {
+        XmlAdaptedPerson person =
+                new XmlAdaptedPerson(INVALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, validTags);
+        Assert.assertThrows(IllegalValueException.class, () -> person.toModelType());
```
Zhiyuan-Amos: We can use method reference `person::toModelType`.

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String INVALID_ADDRESS = " ";
+    private static final String INVALID_EMAIL = "example.com";
+    private static final String INVALID_TAG = "#friend";
+
+    private static final String VALID_NAME = BENSON.getName().toString();
+    private static final String VALID_PHONE = BENSON.getPhone().toString();
+    private static final String VALID_EMAIL = BENSON.getEmail().toString();
+    private static final String VALID_ADDRESS = BENSON.getAddress().toString();
+
+    private final List<XmlAdaptedTag> validTags = BENSON.getTags().stream()
```
Zhiyuan-Amos: can we make this `static` too?

src/test/java/seedu/address/logic/parser/ParserUtilTest.java
```diff
@@ -83,9 +86,16 @@ public void parseName_optionalEmpty_returnsOptionalEmpty() throws Exception {
     @Test
     public void parseName_validValue_returnsName() throws Exception {
```
Zhiyuan-Amos: Should we update the 2nd clause to be `validValueWithoutWhitespace`, since we have a test below which has `validValueWithWhitespace` as the 2nd clause?

src/test/java/seedu/address/logic/parser/ParserUtilTest.java
```diff
@@ -194,6 +194,31 @@ public void parseEmail_validValueWithWhitespace_returnsEmail() throws Exception
         assertEquals(Optional.of(expectedEmail), ParserUtil.parseEmail(Optional.of(emailWithWhitespace)));
     }
 
+    @Test
+    public void parseTag_null_throwsNullPointerException() throws Exception {
```
Zhiyuan-Amos: I think we can use `Assert.assertThrows` here ^^

Zhiyuan-Amos:
On a side note, this PR closes #294.

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
pyokagan: Oh, I'm still trying to get the hang of intellij lol

src/test/java/seedu/address/logic/parser/ParserUtilTest.java
pyokagan: On second thought, I think I'll just merge the test cases.

pyokagan: On third thought, I'll just take your suggestion.

My main concern is that naming becomes difficult if more variants are added in the future, but I'll just leave it for the future people to handle.

src/test/java/seedu/address/logic/parser/ParserUtilTest.java
pyokagan: Conscious decision for consistency with other tests in the file :-)

Perhaps we may wish to get rid of all `thrown.expect` in the code base, but that's another topic.

Zhiyuan-Amos: I think I'm not seeing something rightly, cos from your `parseAddress_null_throwsNullPointerException()` test case, it's using `Assert.assertThrows` too ><

```
public void parseAddress_null_throwsNullPointerException() {
 	Assert.assertThrows(NullPointerException.class, () -> ParserUtil.parseAddress((String) null));
	Assert.assertThrows(NullPointerException.class, () -> ParserUtil.parseAddress((Optional<String>) null));
}
```

pyokagan: `parseIndex_invalidInput_throwsIllegalValueExceptio`, `parseTags_null_throwsNullPointerException` , `parseTags_collectionWithInvalidTags_throwsIllegalValueException` etc

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -60,7 +55,7 @@ public static ReadOnlyAddressBook getSampleAddressBook() {
     /**
      * Returns a tag set containing the list of strings given.
      */
-    public static Set<Tag> getTagSet(String... strings) throws IllegalValueException {
+    public static Set<Tag> getTagSet(String... strings) {
```
Zhiyuan-Amos: happened to notice that once `IllegalValueException` is no longer thrown, we can simplify this method using streams. Shall I raise an issue for this?

pyokagan: Already covered by #634

pyokagan:
I'm submitting another iteration. Found that `EditPersonDescriptorBuilder`'s code can be simplified after the changes in this PR.

Issue #319, opened 2017-03-06
Checkstyle: enforce more rules of the oss java coding standard


eugenepeh:
Currently several rules of the java coding standard stated in the 
https://oss-generic.github.io/process/codingStandards/CodingStandard-Java.html
are not being enforced. 
Issue #283 has summarized a couple of them.

Those enforceable rules should be enforced.

eugenepeh:
Hi Prof. Damith @damithc ,
Currently we have about 41 classes and 84 public methods (excluding single statement methods and those with @Override/@Test annotation) missing header comments, and most are self explanatory. 

This rule is enforceable using JavadocType and JavadocMethod. However, it will require header comment for all classes, public methods (including trivial / self explanatory). Should we still strictly enforce this rule (with a severe error) below?

> However, you MUST write header comments for all classes, public methods, and all non-trivial private methods. The code, even if it is self-explanatory, can only tell the reader HOW the code works, not WHAT the code is supposed to do.

yamgent:
I feel that for this one a warning should suffice? Unless we are ok with putting `suppressWarning` or `//CHECKSTYLE:OFF` all around these method, which may make it a bit disorganized.

damithc:
We can start with a warning. Hopefully, we can fix the warnings incrementally until we can enforce the rule fully. Also, we can exempt getters and setters.

PR #669, opened 2017-09-14
Checkstyle: enforce single statement per line #319

eugenepeh:
Addition to #319 
Ready for review

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -224,6 +224,8 @@
       <property name="allowEmptyLoopBody" value="true"/>
     </module>
 
+    <module name="OneStatementPerLine"/>
+
     <module name="UpperEll">
       <!-- Checks that long constants are defined with an upper ell.-->
       <property name="severity" value="error"/>


config/checkstyle/suppressions.xml
```diff
@@ -5,5 +5,14 @@
 "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">
 
 <suppressions>
+  <!--
+  CODING CHECKS
+  -->
+  <suppress checks="MultipleVariableDeclarations" 
+    files="[\\/]src[\\/]test[\\/]java[\\/]seedu[\\/]address[\\/]testutil[\\/]TypicalPersons\.java"/>
```
eugenepeh: This suppress is meant for the multiple declarations of `ReadOnlyPerson` in [TypicalPersons](https://github.com/se-edu/addressbook-level4/blob/master/src/test/java/seedu/address/testutil/TypicalPersons.java#L27)

config/checkstyle/suppressions.xml
```diff
@@ -5,5 +5,14 @@
 "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">
 
 <suppressions>
+  <!--
+  CODING CHECKS
+  -->
+  <suppress checks="MultipleVariableDeclarations"
+    files="[\\/]src[\\/]test[\\/]java[\\/]seedu[\\/]address[\\/]testutil[\\/]TypicalPersons\.java"/>
```
eugenepeh: This suppress is meant for the multiple declarations of ReadOnlyPerson in [TypicalPersons](https://github.com/se-edu/addressbook-level4/blob/master/src/test/java/seedu/address/testutil/TypicalPersons.java#L27).

pyokagan: Why not fix that violation?

eugenepeh: hmm, will require to declare them all in single statements each, wouldn't that make it rather ugly?

pyokagan: Well, I was asking since your commit message says that:

> It's very difficult to read multiple statements on one line. Also,
> single statement per line means fewer merge conflicts to clean up.

which says that fixing this violation would make the code more readable...

Actually, now that I read the commit message again, I see that its referring to multiple statements per line, which is covered by the `OneStatementPerLine` rule, but doesn't explain the reason for adding `MultipleVariableDeclarations` as well.

So, why do we need `MultipleVariableDeclarations`, if this example shows that having multiple variable declarations in a single statement makes the code prettier?

damithc: Google coding style [says one variable per line](https://google.github.io/styleguide/javaguide.html#s4.8.2-variable-declarations). Perhaps we follow the same but suppress it for special situations such as the example quoted from TypicalPersons?.

eugenepeh: > So, why do we need MultipleVariableDeclarations, if this example shows that having multiple variable declarations in a single statement makes the code prettier?

good point, I was thinking of the scenario where people declare and initialize variables on the same line; e.g. `int a = 1, b = 2, c = 3;`
Also having each variable on a separate line would help to facilitate the job of version control tools. Several variables are on the same line risk having conflicts for unrelated modifications by different developers.

`TypicalPersons` on the other hand is a special scenario where I'm stuck at how to make it looks better.

pyokagan: @eugenepeh 

> TypicalPersons on the other hand is a special scenario where I'm stuck at how to make it looks better.

I think something like this should be possible?

```java
    public static final ReadOnlyPerson ALICE = new PersonBuilder()
            .withName("Alice Pauline")
            .withAddress("123, Jurong West Ave 6, #08-111")
            .withEmail("alice@example.com")
            .withPhone("85355255")
            .withTags("friends")
            .build();

    public static final ReadOnlyPerson BENSON = new PersonBuilder()
            .withName("Benson Meier")
            .withAddress("311, Clementi Ave 2, #02-25")
            .withEmail("johnd@example.com")
            .withPhone("98765432")
            .withTags("owesMoney", "friends")
            .build();
...
```

Anyway, the commit message needs to explain why we are enforcing `MultipleVariableDeclarations`. If it is because we are following the Google style guide, then that needs to be stated.

eugenepeh: @pyokagan ahh, thanks. you're right, it is possible now that the `IllegalValueException` are captured in the `PersonBuilder`.

Will update the commit message too.

eugenepeh:
#671

PR #671, opened 2017-09-16
Checkstyle: enforce one variable declaration per line #319

eugenepeh:
Addition to #319

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -224,6 +224,9 @@
       <property name="allowEmptyLoopBody" value="true"/>
     </module>
 
+    <!-- Checks that each variable declaration is in its own statement and on its own line. -->
+    <module name="MultipleVariableDeclarations"/>
+
     <module name="OneStatementPerLine"/>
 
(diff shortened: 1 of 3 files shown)


eugenepeh:
@pyokagan understood that they're 2 different thing but this rule triggers even if you do something like this
> int a; int b; 

even though this check is already covered by single statement per line.

pyokagan:
@eugenepeh 

> understood that they're 2 different thing but this rule triggers an error even if you do something like this
> 
>     int a; int b;
> 
> even though this check is already covered by single statement per line.
> so it is actually covering by per line basis rather than just per statement.

Oh, so sorry. Yes, you're absolutely right, and I see that you have noted that correctly in the comment in `checkstyle.xml`.

Still, please modify the commit message to mention both `statement` and `line` please, thanks. :-)

eugenepeh:
@pyokagan oops... I just tested and realize... 
> int a,
    b;

This works too. So I don't think I can mention statement 😝

pyokagan:
@eugenepeh Update the commit message to include the examples of code that you provided above which pass or fail the check.

PR #679, opened 2017-09-18
SampleDataTest: Fix buggy implementation of getDataFileLocation() #678

### LINKED ISSUE #678: SampleDataTest: Fix buggy implementation of getDataFileLocation()
getDataFileLocation() must return a non-existent file location as part
of this method's post-condition.

However, this isn't ensured in the current implementation. As a result,
the test case in SampleDataTest may fail.

Let's update getDataFileLocation() to ensure that it returns a 
non-existent file location.

Zhiyuan-Amos:
Fixes #678

### DIFF
--- a/src/test/java/guitests/SampleDataTest.java
+++ b/src/test/java/guitests/SampleDataTest.java
@@ -2,6 +2,10 @@ package guitests;
 
 import static seedu.address.ui.testutil.GuiTestAssert.assertListMatching;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
 import org.junit.Test;
 
 import seedu.address.model.AddressBook;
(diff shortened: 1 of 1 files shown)


src/test/java/guitests/SampleDataTest.java
```diff
@@ -19,7 +23,21 @@ protected AddressBook getInitialData() {
...
-        return TestUtil.getFilePathInSandboxFolder("SomeFileThatDoesNotExist1234567890.xml");
+        String filePath = TestUtil.getFilePathInSandboxFolder("SomeFileThatDoesNotExist1234567890.xml");
+        deleteFileIfExists(filePath);
+        return filePath;
+    }
+
+    /**
+     * Deletes the file at {@code filePath} if it exists.
+     */
+    private void deleteFileIfExists(String filePath) {
```
Zhiyuan-Amos: Thought of placing this method in `FileUtil`, but `FileUtil` is a class that's used by the main code base. As such, I left this method here instead.

src/test/java/guitests/SampleDataTest.java
```diff
@@ -19,7 +23,21 @@ protected AddressBook getInitialData() {
...
+        File file = new File(filePath);
+        try {
+            Files.deleteIfExists(file.toPath());
```
pyokagan: Can use `Paths.get()`?

src/test/java/guitests/SampleDataTest.java
```diff
@@ -19,7 +23,21 @@ protected AddressBook getInitialData() {
...
+        } catch (IOException ioe) {
+            throw new AssertionError();
```
pyokagan: `new AssertionError(ioe)` please.

PR #689, opened 2017-09-19
EditCommandSystemTest: Replace assert with junit.Assert

Zhiyuan-Amos:
```
We use assert to show that a precondition is met before executing a
test.

Asserts should not be used in test cases as they can be disabled.

Let’s replace assert statements with junit.Assert
```

### DIFF
--- a/src/test/java/systemtests/EditCommandSystemTest.java
+++ b/src/test/java/systemtests/EditCommandSystemTest.java
@@ -1,5 +1,7 @@
 package systemtests;
 
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static seedu.address.logic.commands.CommandTestUtil.ADDRESS_DESC_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.ADDRESS_DESC_BOB;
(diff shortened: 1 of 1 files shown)


damithc:
Based on [this](https://objectcomputing.com/resources/publications/sett/march-2014-junit-not-just-another-pretty-assert/) assumption failures are silent (i.e. do not result in test failures). Probably that's not the behavior we want?

Zhiyuan-Amos:
image

You can see which tests are ignored. :)

damithc:
Does travis fail when a test is ignored?
Although I said OK to this earlier, now I feel like we should not bring in a relatively unknown JUnit feature into our code. It also looks like Assumptions are meant to be written against external resources, not things provided by our own code. Perhaps just use normal JUnit assertions (not java `assert`)?

Zhiyuan-Amos:
> Does travis fail when a test is ignored?

Ah nope it doesn't fail. Have to manually read from the logs to see if any tests have been skipped or not.

> Perhaps just use normal JUnit assertions (not java assert)?

The reason for using `Assume` is so that we can distinguish the `post-condition` from the `expected test result`. (which was the reason why I used `assert` instead of `assertTrue` earlier as well). I will update it then.

PR #690, opened 2017-09-19
Clean up SelectCommandSystemTest

Zhiyuan-Amos:
```
SelectCommandSystemTest can be cleaner. For example, the header 
comments can be updated to be written in point forms, and the test 
cases can be segmented into different categories.

This causes it to be harder for future developer to understand the 
code.

Let's clean up SelectCommandSystemTest.
```

### DIFF
--- a/src/test/java/systemtests/SelectCommandSystemTest.java
+++ b/src/test/java/systemtests/SelectCommandSystemTest.java
@@ -20,6 +20,8 @@ import seedu.address.model.Model;
 public class SelectCommandSystemTest extends AddressBookSystemTest {
     @Test
     public void select() {
+        /* ------------------------ Perform select operations on the shown unfiltered list -------------------------- */
+
         /* Case: select the first card in the person list, command with leading spaces and trailing spaces
          * -> selected
          */
@@ -46,18 +48,16 @@ public class SelectCommandSystemTest extends AddressBookSystemTest {
         command = SelectCommand.COMMAND_WORD + " " + middleIndex.getOneBased();
         assertCommandSuccess(command, middleIndex);
 
-        /* Case: invalid index (size + 1) -> rejected */
-        int invalidIndex = getModel().getFilteredPersonList().size() + 1;
-        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex, MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
-
         /* Case: select the current selected card -> selected */
         assertCommandSuccess(command, middleIndex);
+
+        /* ------------------------ Perform select operations on the shown filtered list ---------------------------- */
 
         /* Case: filtered person list, select index within bounds of address book but out of bounds of person list
          * -> rejected
          */
         showPersonsWithName(KEYWORD_MATCHING_MEIER);
-        invalidIndex = getModel().getAddressBook().getPersonList().size();
+        int invalidIndex = getModel().getAddressBook().getPersonList().size();
         assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex, MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
 
         /* Case: filtered person list, select index within bounds of address book and person list -> selected */
@@ -66,6 +66,8 @@ public class SelectCommandSystemTest extends AddressBookSystemTest {
         command = SelectCommand.COMMAND_WORD + " " + validIndex.getOneBased();
         assertCommandSuccess(command, validIndex);
 
+        /* ----------------------------------- Perform invalid select operations ------------------------------------ */
+
         /* Case: invalid index (0) -> rejected */
         assertCommandFailure(SelectCommand.COMMAND_WORD + " " + 0,
                 String.format(MESSAGE_INVALID_COMMAND_FORMAT, SelectCommand.MESSAGE_USAGE));
@@ -73,6 +75,10 @@ public class SelectCommandSystemTest extends AddressBookSystemTest {
         /* Case: invalid index (-1) -> rejected */
         assertCommandFailure(SelectCommand.COMMAND_WORD + " " + -1,
                 String.format(MESSAGE_INVALID_COMMAND_FORMAT, SelectCommand.MESSAGE_USAGE));
+
+        /* Case: invalid index (size + 1) -> rejected */
+        invalidIndex = getModel().getFilteredPersonList().size() + 1;
+        assertCommandFailure(SelectCommand.COMMAND_WORD + " " + invalidIndex, MESSAGE_INVALID_PERSON_DISPLAYED_INDEX);
 
         /* Case: invalid arguments (alphabets) -> rejected */
         assertCommandFailure(SelectCommand.COMMAND_WORD + " abc",
--- a/src/test/java/systemtests/SelectCommandSystemTest.java
+++ b/src/test/java/systemtests/SelectCommandSystemTest.java
@@ -1,5 +1,6 @@
 package systemtests;
 
+import static org.junit.Assert.assertTrue;
 import static seedu.address.commons.core.Messages.MESSAGE_INVALID_COMMAND_FORMAT;
 import static seedu.address.commons.core.Messages.MESSAGE_INVALID_PERSON_DISPLAYED_INDEX;
 import static seedu.address.commons.core.Messages.MESSAGE_UNKNOWN_COMMAND;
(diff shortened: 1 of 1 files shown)


Zhiyuan-Amos:
Will update [3/3] as per https://github.com/se-edu/addressbook-level4/pull/689#issuecomment-330435595

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -99,14 +99,17 @@ public void select() {
...
     /**
-     * Executes {@code command} and verifies that the command box displays an empty string, the result display
-     * box displays the success message of executing select command with the {@code expectedSelectedCardIndex}
-     * of the selected person, and the model related components equal to the current model.
-     * These verifications are done by
+     * Executes {@code command} and in addition,<br>
+     * 1. Asserts that the command box displays an empty string.<br>
+     * 2. Asserts that the result display box displays the success message of executing select command with the
+     * {@code expectedSelectedCardIndex} of the selected person.<br>
+     * 3. Asserts that the model related components equal to the current model.<br>
```
eugenepeh: I see some changes in the header comments, e.g. `status bar remain unchanged` -> `sync status changes.`,
perhaps commit message can also add in `reflect latest changes`?

current `model`. -> current `model's`.
Also what's the 2 models you're comparing?

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -130,12 +133,14 @@ private void assertCommandSuccess(String command, Index expectedSelectedCardInde
     }
 
     /**
-     * Executes {@code command} and verifies that the command box displays {@code command}, the result display
-     * box displays {@code expectedResultMessage} and the model related components equal to the current model.
-     * These verifications are done by
+     * Executes {@code command} and in addition,<br>
+     * 1. Asserts that the command box displays {@code command}.<br>
+     * 2. Asserts that result display box displays {@code expectedResultMessage}.<br>
+     * 3. Asserts that the model related components equal to the current model.<br>
```
eugenepeh: same as above.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -20,6 +20,8 @@
 public class SelectCommandSystemTest extends AddressBookSystemTest {
     @Test
     public void select() {
+        /* ----------------- Performing select operation while an unfiltered list is being shown -------------------- */
```
eugenepeh: `Perform` select `operations` on the shown unfiltered list?
would that be better? cause `being shown` sounds a bit weird when the list isn't really changing

Zhiyuan-Amos: Looks good! Now I have to change the existing classes too ><

src/test/java/systemtests/SelectCommandSystemTest.java
Zhiyuan-Amos: This line: `Asserts that the model related components equal to the current model.`
Refers to: `the model and storage contains the same person objects as {@code expectedModel} and the person list panel displays the persons in the model correctly.` (in assertApplicationDisplaysExpected).

Now that you mention it, I agree that it's pretty confusing. At least for `SelectCommandSystemTest`, I think I can write something like `Asserts that the components that use the data in the model remain unchanged`?

Zhiyuan-Amos: > perhaps commit message can also add in reflect latest changes?

Where should I add this in?

eugenepeh: How about "Asserts that the `UI` components, that use the `test model` data, remain unchanged"?

Here I assume your components mean UI components cause I'm still a little confuse. 
Actually, we can also do without ", that use the `test model` data," and go directly with
`Asserts that the UI components remain unchanged after command execution.`

Zhiyuan-Amos:
@eugenepeh RFR! (ready for review ^^)

src/test/java/systemtests/SelectCommandSystemTest.java
Zhiyuan-Amos: Ah so sorry didn't make this change in v2, just saw it >< 

> "Asserts that the UI components, that use the test model data, remain unchanged"?

This won't do because the `storage` and the `model` itself aren't UI components. :P hang on, lemme spend a while thinking.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -20,6 +20,8 @@
 public class SelectCommandSystemTest extends AddressBookSystemTest {
     @Test
     public void select() {
+        /* ------------------------ Perform select operation on the shown unfiltered list --------------------------- */
```
eugenepeh: `operations` :D
same for the rest

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -130,12 +133,14 @@ private void assertCommandSuccess(String command, Index expectedSelectedCardInde
...
     /**
-     * Executes {@code command} and verifies that the command box displays {@code command}, the result display
-     * box displays {@code expectedResultMessage} and the model related components equal to the current model.
-     * These verifications are done by
+     * Executes {@code command} and in addition,<br>
+     * 1. Asserts that the command box displays {@code command}.<br>
+     * 2. Asserts that result display box displays {@code expectedResultMessage}.<br>
+     * 3. Asserts that the components that use model's data remain unchanged.<br>
+     * 4. Asserts that the browser url, selected card and status bar remain unchanged.<br>
+     * 5. Asserts that the command box has the error style.<br>
```
eugenepeh: above wording uses `default style class`, should standardize

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -99,14 +99,17 @@ public void select() {
...
     /**
-     * Executes {@code command} and verifies that the command box displays an empty string, the result display
-     * box displays the success message of executing select command with the {@code expectedSelectedCardIndex}
-     * of the selected person, and the model related components equal to the current model.
-     * These verifications are done by
+     * Executes {@code command} and in addition,<br>
+     * 1. Asserts that the command box displays an empty string.<br>
+     * 2. Asserts that the result display box displays the success message of executing select command with the
+     * {@code expectedSelectedCardIndex} of the selected person.<br>
+     * 3. Asserts that the components that use model's data remain unchanged.<br>
```
eugenepeh: the `components` have to be a bit more specific, although you already eliminated browser url, selected card, status bar, command box etc...

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -99,14 +99,17 @@ public void select() {
...
+     * 4. Asserts that the resulting browser url and selected card update accordingly depending on the card at
```
eugenepeh: would this be better? 
`Asserts that the browser url and selected card are updated according to the data on the card at`

wait doesn't sounds right... this might be better
`Asserts that selected card is at {@code expectedSelectedCardIndex} and the browser url is updated accordingly.`

src/test/java/systemtests/SelectCommandSystemTest.java
pyokagan: Hrm, that makes me wonder. As an end-to-end system test, should we be checking the model directly? :-P

Zhiyuan-Amos: Sorry senpai what do you mean? ><

pyokagan: Well, if we are doing black-box testing of the entire application as a whole, then the "model" is considered an internal detail which we should not have direct access to.

pyokagan: But this is completely off-topic from this PR, so no need to care about that.

damithc: > Well, if we are doing black-box testing of the entire application as a whole, then the "model" is considered an internal detail which we should not have direct access to.

Yes, that's a fair point. We can remove 'model verification' aspect from the system tests.

Zhiyuan-Amos:
@damithc prof for your review.

src/test/java/systemtests/SelectCommandSystemTest.java
```diff
@@ -93,14 +100,17 @@ public void select() {
...
-     * These verifications are done by
+     * Executes {@code command} and in addition,<br>
+     * 1. Asserts that the command box displays an empty string.<br>
+     * 2. Asserts that the result display box displays the success message of executing select command with the
+     * {@code expectedSelectedCardIndex} of the selected person.<br>
+     * 3. Asserts that {@code Model}, {@code Storage} and {@code PersonListPanel} remain unchanged.<br>
+     * 4. Asserts that selected card is at {@code expectedSelectedCardIndex} and the browser url is updated
+     * accordingly.<br>
+     * 5. Asserts that the status bar remains unchanged.<br>
+     * 6. Asserts that the command box has the default style class.<br>
```
damithc: Looks like `Asserts that` is common to all 6 points. extract it out? `Executes the {@code command} and asserts,<br>`

damithc: You can order the 6 points so that it follows the physical arrangement of the GUI. e.g. point 1,2,6 are about the same part of the GUI.

PR #692, opened 2017-09-19
Guitests: Move tests to other packages #684

### LINKED ISSUE #684: Guitests: Move tests to other packages
1. `HelpWindowTest` looks like it can be renamed to `HelpCommandTest` and moved to `systemtests`.
1. `ErrorDialogGuiTest` and `SampleDataTest` can be shifted to `systemtests` as well. 
1. `StatusBarFooterTest` is superseded and can be removed as we now have system tests.

This is because `AddressBookGuiTest` and `AddressBookSystemTest` are very similar classes.

Zhiyuan-Amos:
Fixes #684

```
AddressBookGuiTest and AddressBookSystemTest are very similar.

It’s not necessary to have both of them.

Let's remove AddressBookGuiTest by shifting out the children classes to
inherit from AddressBookSystemTest instead.
```


Zhiyuan-Amos:
There's still a `guitests` packages which contains `GuiRobot` and the `guihandles`. Not quite sure where to move them.

docs/DeveloperGuide.adoc
```diff
@@ -390,7 +390,7 @@ Tests can be found in the `./src/test/java` folder.
 
 We have two types of tests:
 
-.  *GUI Tests* - These are _System Tests_ that test the entire App by simulating user actions on the GUI. These are in the `guitests` package.
+.  *GUI Tests* - These are _unit tests_ targeting the individual components of the GUI and _system tests_ that test the entire App by simulating user actions on the GUI. These are in the `seedu.address.ui` and `systemtests` package.
```
damithc: Like this?
These are,
* _unit tests_ targeting the individual components of the GUI (these tests are in the `seedu.address.ui` package)
* _system tests_ that test the entire App by simulating user actions on the GUI (these tests are in the `systemtests` package)

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -76,6 +78,25 @@ public void tearDown() throws Exception {
         EventsCenter.clearSubscribers();
     }
 
+    /**
+     * Override this in child classes to set the initial local data.
```
damithc: You may want to add an overview statement too, in keeping with our header comment style. e.g. `Returns the data to be ...`

Zhiyuan-Amos:
@eugenepeh ready for review! ^^

src/test/java/systemtests/HelpCommandTest.java
```diff
@@ -1,19 +1,25 @@
...
 
+import guitests.GuiRobot;
 import guitests.guihandles.HelpWindowHandle;
 import seedu.address.logic.commands.HelpCommand;
 
-public class HelpWindowTest extends AddressBookGuiTest {
+/**
+ * TODO: This test is incomplete.
+ */
+public class HelpCommandTest extends AddressBookSystemTest {
```
eugenepeh: How come didn't follow other SystemTest classes naming convention and name it as `HelpCommandSystemTest`?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -76,6 +78,21 @@ public void tearDown() throws Exception {
...
+     * Return null to use the data in the file specified in {@link #getDataFileLocation()}
```
eugenepeh: Don't understand what it means >_< so if it returns null, I have to grab the data in {@link #getDataFileLocation()} myself?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -76,6 +78,21 @@ public void tearDown() throws Exception {
...
+     */
+    protected AddressBook getInitialData() {
+        return TypicalPersons.getTypicalAddressBook();
+    }
+
+    /**
+     * Override this in child classes to set the data file location.
+     */
+    protected String getDataFileLocation() {
+        return TestApp.SAVE_LOCATION_FOR_TESTING;
```
eugenepeh: hmm... if TestApp extends MainApp... how come we don't have a standard method for getting the data file location?

src/test/java/systemtests/AddressBookSystemTest.java
Zhiyuan-Amos: All these I grabbed from `AddressBookGuiTest` (which is deleted in this PR). The reason why I just copy and pasted is cos... I have no idea how it works too >< For now can we just copy and paste, cos that will inflate the size of the PR ><

src/test/java/systemtests/AddressBookSystemTest.java
Zhiyuan-Amos: ah you are saying that we should use a method instead of grabbing the constant is it?

src/test/java/systemtests/AddressBookSystemTest.java
eugenepeh: hmm seems like this header comment is from a year ago... will have to consult prof but would be nice if we can update it in this PR

src/test/java/systemtests/AddressBookSystemTest.java
eugenepeh: yup out of scope, i'm just raising the question :P

Zhiyuan-Amos:
Thanks for your thorough review Eugene ^^

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -76,6 +79,28 @@ public void tearDown() throws Exception {
...
     }
 
+    /**
+     * Override this in child classes to set the initial local data.
+     * If null is returned,
+     * 1. and the file in {@link #getDataFileLocation()} is an existing file, use the data in it.
+     * 2. and the file in {@link #getDataFileLocation()} is not an existing file, the initial data will be
+     *    {@link SampleDataUtil#getSampleAddressBook()}.
+     */
+    protected AddressBook getInitialData() {
```
eugenepeh: hmm... not sure why the original header comment for this method is so complicated...
since `TypicalPersons.getTypicalAddressBook();` will always return a new `AddressBook`, 
I don't see any scenario where it can be null.

Perhaps can simplify it to like `Returns an {@code AddressBook} with default sample data.` if the child classes and the callers do not require the information.

Zhiyuan-Amos: Oh you can override this method in children classes to return `null`. So for example, `SampleDataTest` overrides this method to return `null`. This then allows `SampleDataTest` to verify if indeed the sample data is loaded or not when you start loading the app without any previously saved file.

eugenepeh: I think we can override the parent class javadoc comments too in child classes, rather than dump everything on parent class and nobody in the world know what's going on, would that be better?

Zhiyuan-Amos: Good point!

Zhiyuan-Amos:
👍 thanks again Eugene!

@damithc would you like to review this? You gave some cosmetic changes above (https://github.com/se-edu/addressbook-level4/pull/692#pullrequestreview-63562424).

Zhiyuan-Amos:
Oh @eugenepeh regarding your comment at https://github.com/se-edu/addressbook-level4/pull/692#pullrequestreview-65759570, I'm thinking that it shouldn't be shifted up as header comments because header comments shouldn't concern themselves with internal implementation; it should only describe what the method is doing. 

Since `// return null to force test app to load data from file only` describes the internal implementation, we should leave this inside right? ^^

eugenepeh:
@Zhiyuan-Amos hmm but its behavior is different from parent. Also whoever use this method should be aware of the null value and the reason behind it, so I would think its fine, can use more opinion here though.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -1,19 +1,25 @@
...
 
 import org.junit.Test;
 
+import guitests.GuiRobot;
 import guitests.guihandles.HelpWindowHandle;
 import seedu.address.logic.commands.HelpCommand;
 
-public class HelpWindowTest extends AddressBookGuiTest {
+/**
+ * TODO: This test is incomplete.
```
damithc: Perhaps mention in what way it is incomplete? e.g. need more test cases

Zhiyuan-Amos:
Made the changes! ^^

Issue #675, opened 2017-09-18
Possible state leak in AddCommandSystemTest


damithc:
Windows, from Intellij: this test fails when run with other tests, but passes when run by itself. 
```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at systemtests.AddressBookSystemTest.assertSelectedCardUnchanged(AddressBookSystemTest.java:177)
	at systemtests.AddCommandSystemTest.assertCommandSuccess(AddCommandSystemTest.java:229)
	at systemtests.AddCommandSystemTest.assertCommandSuccess(AddCommandSystemTest.java:217)
	at systemtests.AddCommandSystemTest.assertCommandSuccess(AddCommandSystemTest.java:200)
	at systemtests.AddCommandSystemTest.add(AddCommandSystemTest.java:142)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
[...]

PR #695, opened 2017-09-20
Delete ui tests

Zhiyuan-Amos:
Debugging #675. @damithc can verify if `AddCommandTest` passes locally?


damithc:
Passed both times I tried.

Zhiyuan-Amos:
> Passed both times I tried.

This means that you executed the tests twice and both works right? Not that you executed the tests on master branch, and executed the tests on this branch, and both works right? Just clarifying ^^

If it is the former:
As Wang Leng guesses that this bug is caused by an error with TestFX, I can think of 2 ways to resolve this:

1. Simply mention in `DeveloperGuide` that the tests will fail locally because of a bug with TestFX. This doesn't resolve the issue, but at least we inform the developers.
2. Force Gradle to execute `seedu.address.ui` package last. This is a workaround, because the reason why `AddCommandTest` fails is that it is executed after `seedu.address.ui`. (the error is caused by `seedu.address.ui`).

Zhiyuan-Amos:
Oh right yes, cos Gradle starts a new JVM per test or something like that? Ok will do so.

damithc:
Note that I'm not using Gradle to run tests. In my IDEA set up (done based on the instruction in our dev guide), tests are not delegated to gradle. They run within IDEA as per normal.

Zhiyuan-Amos:
Thanks for the note prof.

I tried running all tests using both `Gradle Test Runner` (i.e. tests that are delegated to gradle) and `Platform Test Runner` (i.e. tests that aren't delegated to gradle) on `master branch`, and they both fail. However, when I execute `allTests` using Gradle, the tests pass on `master branch`.

(I tried running all tests `on this branch` using `Gradle Test Runner`, `Platform Test Runner` and executing `allTests` using Gradle, and they all pass.)

As such, the only way is to pass all tests on `master branch` is to execute `allTests` using Gradle.

Zhiyuan-Amos:
Solution written by Paul https://github.com/se-edu/addressbook-level4/issues/675#issuecomment-330775347

PR #717, opened 2017-10-04
Update developer guide with user stories and NFR

shuang-yang:


Zhiyuan-Amos:
@shuang-yang wrong repo?

PR #747, opened 2017-12-01
Correctly model the scene graph hierarchy of MainWindow and HelpWindow

pyokagan:
```
As their names suggest, the root object of MainWindow and HelpWindow should be
their Stage. However, this is presently not so. The root object of MainWindow
and HelpWindow are both Regions, and the Stage and Scene which they manage is
sneakily handled in their constructors.

This violates the architecture of the UiPart system, which dictates that
the return value of UiPart#getRoot(), as well as the root node of the FXML
file, should be the root object of the scene graph which the UiPart
manages. This derivation from the architecture may confuse readers.

Fix this by using the <fx:root> feature of FXML files to achieve the
aforementioned proper UiPart structure. The user interface from the user's
point of view does not change.
```


pyokagan:
This PR contains a lot of whitespace-only changes in the FXML files so reviewers might want to use `?w=1` to review them.

pyokagan:
Merge with master would break a test:
```
systemtests.HelpCommandSystemTest > openHelpWindow FAILED

    java.lang.AssertionError at HelpCommandSystemTest.java:66
```
Looks like I'll need to submit another version.

pyokagan:
[2/9] is the new commit that fixes the test failure by tweaking `SystemTestSetupHelper` to create a new primary stage each time. cc @Zhiyuan-Amos

Zhiyuan-Amos:
sorry that I didn't review this; was quite packed. This looks really amazing :O 💯

PR #748, opened 2017-12-03
ListElementPointer(List<String>): Update code to tally with comments

Zhiyuan-Amos:
```
The constructor’s header comments state that this object is backed by
a defensive copy of the list that is passed in as parameter. However,
the code does not reflect this behaviour.

The constructor should use create a defensive copy of the list to
prevent unintentional modifications to the backing list whenever the
list passed in as parameter is modified.

Let’s update ListElementPointer(List<String>) to create a defensive copy
of the list.
```

### DIFF
--- a/src/main/java/seedu/address/logic/ListElementPointer.java
+++ b/src/main/java/seedu/address/logic/ListElementPointer.java
@@ -1,5 +1,6 @@
 package seedu.address.logic;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.NoSuchElementException;
 
@@ -20,7 +21,7 @@ public class ListElementPointer {
      * The cursor points to the last element in {@code list}.
      */
     public ListElementPointer(List<String> list) {
-        this.list = list;
+        this.list = new ArrayList<>(list);
         index = this.list.size() - 1;
     }
 
--- a/src/test/java/seedu/address/logic/ListElementPointerTest.java
+++ b/src/test/java/seedu/address/logic/ListElementPointerTest.java
@@ -27,6 +27,16 @@ public class ListElementPointerTest {
     }
 
     @Test
+    public void constructor_defensiveCopy_backingListUnmodified() {
+        List<String> list = new ArrayList<>();
+        pointer = new ListElementPointer(list);
+        list.add(FIRST_ELEMENT);
+
+        ListElementPointer emptyPointer = new ListElementPointer(Collections.emptyList());
+        assertEquals(emptyPointer, pointer);
+    }
+
+    @Test
     public void emptyList() {
         pointer = new ListElementPointer(new ArrayList<>());
         assertCurrentFailure();


pyokagan:
@Zhiyuan-Amos btw why does [getHistorySnapshot()](https://github.com/se-edu/addressbook-level4/blob/edc42f0e2579c27c209b1f8f3c0f87ed04e83904/src/main/java/seedu/address/logic/Logic.java#L26) return a ListElementPointer? It assumes that callers would use it as a ListElementPointer, and its even more strange when you see that ListElementPointer takes a `List<String>` in its constructor...

If history is an ordered collection of items, then expose it as a list?

pyokagan:
@damithc Approve?

Zhiyuan-Amos:
> @Zhiyuan-Amos btw why does getHistorySnapshot() return a ListElementPointer? It assumes that callers would use it as a ListElementPointer, and its even more strange when you see that ListElementPointer takes a List<String> in its constructor...
If history is an ordered collection of items, then expose it as a list?

@pyokagan Yup, what you said sounds right. I'm wondering whether we should actually shift the storing of the history of user input in `CommandBox` instead of `Logic`, since `CommandBox` functions as a View-Controller. 

That way, does it make better sense to use a `ListElementPointer` since `CommandBox` will only use its history in the context of a `ListElementPointer`?

damithc:
If we think of _history_ as simply filling the command box with some text when the up/down key is pressed, then it is fine to store that data in the command box itself. i.e. history is local to the command box. Rest of the code doesn't need to know about it. But if we think of history as representing the history of user actions (i.e. it goes beyond just the command box), then it should be in a more central location.

Zhiyuan-Amos:
I think the initial reason why we decided to leave _history_ at a more central location is that it should capture both _commands_ and _user input_. 

However, since we have decided not to do it that way and that we have separate histories (_user input history_ & _command history_), then should the _user input history_ be placed in the _command box_ instead of _logic_?

damithc:
I'm of two minds about it. Our current architecture style follows an n-tier approach where the _model_ is a tier that sits behind the _logic_ tier. Attaching micro-models to various parts of the GUI is a different architecture style altogether. While there is no rule that we have to follow one architecture style exclusively, mixing of styles should be a deliberate and justifiable decision. There is value in giving examples of multiple architecture styles given that this is used for teaching, as long as it is done carefully and not at the expense of simplicity of the design.

pyokagan:
@Zhiyuan-Amos 
> since CommandBox functions as a View-Controller and user input is somewhat like its model,

Ah yes, this makes me notice something. If we are modeling the user's input history as part of the domain, then why isn't the input history stored in the `model` layer?

It seems like a code smell to me when we have to, in addition to the `Model` object, pass separate `CommandHistory` and `UndoRedoStack` objects to the command, which indicates to me that we are storing user state in too many different locations which is unwieldy.

pyokagan:
As an aside, regarding the "history" feature:

imo, I think the feature should be removed. It was sold to me as a "nice to have" feature that would have a negligible impact on code base size since it was thought that it could piggy back on the undo/redo architecture. But now it turns out that it's not the case, and now it's yet another feature in addressbook-level4, which already has the problem of feature/scope creep.

PR #754, opened 2017-12-05
Simplify build.gradle file

### LINKED ISSUE #752: Potential build.gradle cleanups
* We don't use multi-project builds, so it would be better to get rid of the `allprojects` block to make it easier to understand for gradle newcomers.
* Some repositories are not used.
* Would be nice if `allTests`, `guiTests`, `nonGuiTests` and `test` were all integrated with each other. Presently running `./gradlew allTests guiTests test` runs tests 3 times :-P
* `def` could be used in some parts to define variables rather than putting everything in `project.ext`.
* `forkEvery` should be increased to improve test running performance.

pyokagan:
Fixes #752 
```
Simplify the build.gradle file by removing unnecessary/unused stuff, and
standardizing the code style.
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,6 @@
  */
 
 plugins {
-    id 'idea'
     id 'java'
     id 'jacoco'
     id 'checkstyle'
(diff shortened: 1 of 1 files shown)


pyokagan:
[1/8] is mostly whitespace changes so `?w=1` will be useful.

PR #759, opened 2017-12-06
DeveloperGuide: use labeled list for glossary

pyokagan:
```
The glossary uses literal blocks for formatting definitions, which looks
strange because literal blocks are usually used for preformatted text
such as source code.

Instead, let's use a labeled list, which has a formatting that is more
in line with a glossary definition list.

Convert the paragraph anchors into inline anchors[1], so that the items
are joined into one single labeled list with two items, rather than
being chopped into two separate labeled lists with one item each.

[1] See http://asciidoctor.org/docs/user-manual/#anchordef for the
    difference between a paragraph anchor and an inline anchor.
```
---
2c984135-fe9f-4bb8-ac09-730c0ea370e5

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -781,19 +781,11 @@ Use case resumes at step 2.
 [appendix]
 == Glossary
 
-[[mainstream-os]]
-Mainstream OS
-
-....
+[[mainstream-os]] Mainstream OS::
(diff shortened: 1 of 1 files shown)


pyokagan:
@damithc 



> I assume the links to items in this list from elsewhere in the document are still working?

Yup I've tested it.

PR #760, opened 2017-12-06
docs: use full xrefstyle and figure auto-numbering

pyokagan:
```
Switch to `full` xrefstyle so that cross references will display the
section/figure numbers as well, allowing readers to easily find the
referenced section/figure by their number.

Switch to using asciidoctor's auto figure numbering which is less error
prone and doesn't require us to pollute the commit diff when figure
numbers need to be updated.
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ plugins {
     id 'checkstyle'
     id "com.github.kt3k.coveralls" version "2.4.0"
     id "com.github.johnrengelman.shadow" version '1.2.3'
-    id 'org.asciidoctor.convert' version '1.5.3'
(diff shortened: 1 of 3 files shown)


damithc:
Figure captions usually appear below the figure. I wonder if CS2101 side will object to this style of figure captions (i.e. above the figure).

pyokagan:
@damithc The figure captions are rendered below the figure.

pyokagan:
User Guide: https://24-75938115-gh.circle-artifacts.com/0/docs/UserGuide.html

Developer Guide: https://24-75938115-gh.circle-artifacts.com/0/docs/DeveloperGuide.html

PR #763, opened 2017-12-14
AddCommandParser accepts invalid command format #738

### LINKED ISSUE #738: AddCommandParser accepts invalid command format
image

Typing "add **formattingProblem123** n/John Doe p/98765432 e/johnd@example.com a/311, Clementi Ave 2, #02-25 t/friends t/owesMoney" should generate an "**Invalid command format!**" message, since we are not adhering to the correct format of the command.

However, this does not trigger the error message, and the person is successfully added instead (_does not generate any problems though_)

vivekscl:
Fixes #738 
Hi everyone! I'm Vivek here and I decided to work on this issue as a first-time contributor.

There is no check to ensure that the preamble for AddCommand is empty or not. 

Let's add a check to ensure that it is not empty and update the AddCommandParserTest.

### DIFF
--- a/src/test/java/seedu/address/logic/commands/CommandTestUtil.java
+++ b/src/test/java/seedu/address/logic/commands/CommandTestUtil.java
@@ -53,6 +53,9 @@ public class CommandTestUtil {
     public static final String INVALID_ADDRESS_DESC = " " + PREFIX_ADDRESS; // empty string not allowed for addresses
     public static final String INVALID_TAG_DESC = " " + PREFIX_TAG + "hubby*"; // '*' not allowed in tags
 
+    public static final String PREAMBLE_WHITESPACE = "\t  \r  \n";
+    public static final String PREAMBLE_NON_EMPTY = "NonEmptyPreamble";
+
     public static final EditCommand.EditPersonDescriptor DESC_AMY;
     public static final EditCommand.EditPersonDescriptor DESC_BOB;
 
--- a/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
+++ b/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
@@ -14,6 +14,8 @@ import static seedu.address.logic.commands.CommandTestUtil.NAME_DESC_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.NAME_DESC_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.PHONE_DESC_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.PHONE_DESC_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.PREAMBLE_NON_EMPTY;
+import static seedu.address.logic.commands.CommandTestUtil.PREAMBLE_WHITESPACE;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_HUSBAND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_AMY;
@@ -47,6 +49,10 @@ public class AddCommandParserTest {
     public void parse_allFieldsPresent_success() {
         Person expectedPerson = new PersonBuilder().withName(VALID_NAME_BOB).withPhone(VALID_PHONE_BOB)
                 .withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND).build();
+
+        // whitespace only preamble
+        assertParseSuccess(parser, PREAMBLE_WHITESPACE + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
 
         // multiple names - last name accepted
         assertParseSuccess(parser, NAME_DESC_AMY + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
@@ -131,5 +137,10 @@ public class AddCommandParserTest {
         // two invalid values, only first invalid value reported
         assertParseFailure(parser, INVALID_NAME_DESC + PHONE_DESC_BOB + EMAIL_DESC_BOB + INVALID_ADDRESS_DESC,
                 Name.MESSAGE_NAME_CONSTRAINTS);
+
+        // non-empty preamble
+        assertParseFailure(parser, PREAMBLE_NON_EMPTY + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND + TAG_DESC_FRIEND,
+                String.format(MESSAGE_INVALID_COMMAND_FORMAT, AddCommand.MESSAGE_USAGE));
     }
 }
--- a/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
+++ b/src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
@@ -49,28 +49,27 @@ public class AddCommandParserTest {
                 .withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND).build();
 
         // multiple names - last name accepted
-        assertParseSuccess(parser, AddCommand.COMMAND_WORD + NAME_DESC_AMY + NAME_DESC_BOB + PHONE_DESC_BOB
-                + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
+        assertParseSuccess(parser, NAME_DESC_AMY + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
 
         // multiple phones - last phone accepted
-        assertParseSuccess(parser, AddCommand.COMMAND_WORD + NAME_DESC_BOB + PHONE_DESC_AMY + PHONE_DESC_BOB
-                + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
+        assertParseSuccess(parser, NAME_DESC_BOB + PHONE_DESC_AMY + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
 
         // multiple emails - last email accepted
-        assertParseSuccess(parser, AddCommand.COMMAND_WORD + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_AMY
-                + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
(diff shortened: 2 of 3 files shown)


pyokagan:
> By the way, given an input like add n/name t/12345..., will

Very annoyingly, seems that GitHub's rendering of typewriter text drops runs of whitespace. This meant to be:

```
add     n/name t/12345...
```

vivekscl:
Alright, understood. Just one doubt though. Before trying to alter the commit sequence in this PR, I went to try it on a dummy repository, and to rewrite the commit history I need to force push the updated branch. So before performing this change here, I would like to confirm, a force push on this PR won't be destructive or affect anything else since I'm the only one working on this PR right? Thanks! :smile:

damithc:
> So before performing this change here, I would like to confirm, a force push on this PR won't be destructive or affect anything else since I'm the only one working on this PR right?

yes, that's right @vivekscl. It's fine to force-push to your own fork.

vivekscl:
```By the way, given an input like add n/name t/12345..., will
argMultimap.getPreamble() return " " or ""? I didn't check the
source code to find out, but it would be nice to have a test that tests
for the scenarios of argMultimap.getPreamble() returning "" or a
string with purely whitespace.
```
To address this, I added a whitespace preamble check. The argMultimap.getPreamble() returns a trimmed preamble and hence only "" is returned if a string of whitespace is given. Since the whitespace string is returned as an empty string, the check passes.

```// whitespace only preamble
 assertParseSuccess(parser, WHITESPACE_PREAMBLE + NAME_DESC_BOB + PHONE_DESC_BOB
                 + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_FRIEND, new AddCommand(expectedPerson));
```

src/test/java/seedu/address/logic/commands/CommandTestUtil.java
```diff
@@ -53,6 +53,9 @@
     public static final String INVALID_ADDRESS_DESC = " " + PREFIX_ADDRESS; // empty string not allowed for addresses
     public static final String INVALID_TAG_DESC = " " + PREFIX_TAG + "hubby*"; // '*' not allowed in tags
 
+    public static final String WHITESPACE_PREAMBLE = "       ";
+    public static final String NON_EMPTY_PREAMBLE = " " + "FormattingProblem";
```
pyokagan: Strange to see `"FormattingProblem"`. Yes, it's used to test for an invalid command format, but in the context of this constant it's just a non empty preamble.

Why the `" " +` as well? It could just be in the same string?

Maybe just use `" NonEmptyPreamble"`?

src/test/java/seedu/address/logic/commands/CommandTestUtil.java
```diff
@@ -53,6 +53,9 @@
     public static final String INVALID_ADDRESS_DESC = " " + PREFIX_ADDRESS; // empty string not allowed for addresses
     public static final String INVALID_TAG_DESC = " " + PREFIX_TAG + "hubby*"; // '*' not allowed in tags
 
+    public static final String WHITESPACE_PREAMBLE = "       ";
```
pyokagan: Maybe add other common whitespace characters as well, like `'\t'`, `'\r'` and `'\n'`?

vivekscl: Add as in, in the same variable or add as another constant entirely?

pyokagan: In the same variable.

src/test/java/seedu/address/logic/commands/CommandTestUtil.java
```diff
@@ -53,6 +53,9 @@
     public static final String INVALID_ADDRESS_DESC = " " + PREFIX_ADDRESS; // empty string not allowed for addresses
     public static final String INVALID_TAG_DESC = " " + PREFIX_TAG + "hubby*"; // '*' not allowed in tags
 
+    public static final String WHITESPACE_PREAMBLE = "\t  \r  \n";
+    public static final String NON_EMPTY_PREAMBLE = " NonEmptyPreamble";
```
Zhiyuan-Amos: `NON_EMPTY_PREAMBLE` sounds weird, perhaps `PREAMBLE` will do.

pyokagan: purpose of the constant not really has anything to do witg whitespace, but to define any kind of peeamble that is not considered an empty preamble by the parser, inependent of whether it happens to xontain whitespace or not. 

-sent from my phone

eugenepeh: Yes, I agree with pyokagan, the definition of empty is determined by the parser, so I think NON_EMPTY_PREAMBLE is fine

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -87,51 +87,51 @@ public void parse_compulsoryFieldMissing_failure() {
...
-        assertParseFailure(parser, AddCommand.COMMAND_WORD + VALID_NAME_BOB + VALID_PHONE_BOB
+        assertParseFailure(parser, VALID_NAME_BOB + VALID_PHONE_BOB
                 + VALID_EMAIL_BOB + VALID_ADDRESS_BOB, expectedMessage);
     }
 
     @Test
     public void parse_invalidValue_failure() {
         // invalid name
-        assertParseFailure(parser, AddCommand.COMMAND_WORD + INVALID_NAME_DESC + PHONE_DESC_BOB + EMAIL_DESC_BOB
+        assertParseFailure(parser,  INVALID_NAME_DESC + PHONE_DESC_BOB + EMAIL_DESC_BOB
```
pyokagan: two many spaces

vivekscl:
I've rewrapped the lines and made sure that the code that starts on a new line are all aligned properly such that they start on the same column.

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -49,89 +49,87 @@ public void parse_allFieldsPresent_success() {
...
 
         // multiple tags - all accepted
         Person expectedPersonMultipleTags = new PersonBuilder().withName(VALID_NAME_BOB).withPhone(VALID_PHONE_BOB)
-                .withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB)
-                .withTags(VALID_TAG_FRIEND, VALID_TAG_HUSBAND).build();
-        assertParseSuccess(parser, AddCommand.COMMAND_WORD + NAME_DESC_BOB + PHONE_DESC_BOB
-                        + EMAIL_DESC_BOB + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND + TAG_DESC_FRIEND,
-                new AddCommand(expectedPersonMultipleTags));
+                .withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND, VALID_TAG_HUSBAND)
+                .build();
```
pyokagan: please revert lines 69 and 70, they are unrelated changes to the commit (which deals with removing  the COMMAND_WOED from tge asserParse/assertParseFailure statements.) 

we generally don't do style changes unless they happen to be part of the statements which needed to be modified as part of a commit, to not pollute the git-blame view.

vivekscl:
Thanks, @damithc!

1. I agree that `WHITESPACE_PREAMBLE` and `NON_EMPTY_PREAMBLE` are only used once and the alternatives should be implemented instead. Furthermore, when taking into consideration that it is unlikely that they will be used in other classes, it is better to implement one of the alternatives. However, in the event where more commands are added and these variables prove useful, should they be left as constants? 
2. That is a more organized way to name the constants, but would `PREAMBLE_WHITESPACE` or `PREAMBLE_NON_EMPTY` sound weird?

damithc:
1. yes, leaving as constants might be useful if there are similar commands in future. 
2. yes, prefixed names don't read as intuitive as the current ones.

I'm OK with either way. Will leave to @pyokagan to make the call.

pyokagan:
im on holiday :-)

src/test/java/seedu/address/logic/parser/AddCommandParserTest.java
```diff
@@ -48,90 +50,97 @@ public void parse_allFieldsPresent_success() {
...
         // two invalid values, only first invalid value reported
-        assertParseFailure(parser, AddCommand.COMMAND_WORD + INVALID_NAME_DESC + PHONE_DESC_BOB + EMAIL_DESC_BOB
-                + INVALID_ADDRESS_DESC, Name.MESSAGE_NAME_CONSTRAINTS);
+        assertParseFailure(parser, INVALID_NAME_DESC + PHONE_DESC_BOB + EMAIL_DESC_BOB + INVALID_ADDRESS_DESC,
+                Name.MESSAGE_NAME_CONSTRAINTS);
+
+        // non-empty preamble
+        assertParseFailure(parser, NON_EMPTY_PREAMBLE + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND + TAG_DESC_FRIEND, String.format(MESSAGE_INVALID_COMMAND_FORMAT,
+                AddCommand.MESSAGE_USAGE));
```
Zhiyuan-Amos: The line wrapping of `String.format(MESSAGE_INVALID_COMMAND_FORMAT, AddCommand.MESSAGE_USAGE)` can be better. Choose higher-level breaks to lower-level breaks i.e. have a new-line after `TAG_DESC_FRIEND,` so that the entire `String.format(...)` can be on the same line.

Zhiyuan-Amos:
> yes, leaving as constants might be useful if there are similar commands in future.

Can these constants be used for cases such as `DeleteCommand`: "delete    1   " (will pass) & "delete   nonEmptyPreamble 1" (will fail)?

damithc:
Let's leave them as constants in that case.
Let's rename them to be `PREAMBLE_WHITESPACE` or `PREAMBLE_NON_EMPTY`. Although a bit less intuitive, we can get used to that style of naming if we do it consistently. I think the benefit of having them look like they are members of a group outweighs the loss of intuitiveness.

damithc:
> Also, on a side note, since these constants will only be used in the context of XCommandParser, I think we should place it in CommandParserTestUtil instead of CommandTestUtil?

Yes, that's possible too.

src/test/java/seedu/address/logic/commands/CommandTestUtil.java
pyokagan: There's no reason why there needs to be a `' '` in front.

I guess this was probably copied from the `NAME_DESC_AMY`, `NAME_DESC_BOB` etc. but those are conveniences for constructing the input strings for testing. Preamble, is by definition, supposed to only be at the front of the command line input, so there's no need for a leading space.

pyokagan:
@vivekscl Please submit another round with the changes requested by others.

vivekscl:
>Also, on a side note, since these constants will only be used in the context of XCommandParser, I think we should place it in CommandParserTestUtil instead of CommandTestUtil?

>Yes, that's possible too.

Just to confirm, should the constants be placed in `CommandParserTestUtil` instead or left in `CommandTestUtil`?

pyokagan:
I'd say leave it in `CommandTestUtil` for consistency with `NAME_DESC_AMY`, `NAME_DESC_BOB` etc. The argument that they should be placed in `CommandParserTestUtil` applies to them as well, but fixing them is another topic altogether.

pyokagan:
Thanks for your contribution @vivekscl :-)

vivekscl:
No problem @pyokagan! It was a pleasure contributing and learning so much in the process! Now that I have completed a first timer issue, is it alright if I move on to contribute to a `d.Contributors` issue as stated in the [workflow document](https://oss-generic.github.io/process/docs/QuestionsIssuesPrs.html)?

```
After you have managed to get one PR merged, you can gradually move to harder issues, starting with issues labeled d.Contributors.
```

damithc:
> is it alright if I move on to contribute to a d.Contributors issue as stated in the workflow document?

Sure @vivekscl

Issue #653, opened 2017-09-02
UniquePersonList#iterator() should return Iterator<ReadOnlyPerson>


Zhiyuan-Amos:
`UniquePersonList#iterator()` returns `Iterator<Person>`. 

Callers can update the values of the `Person` objects, possibly violating the uniqueness of the `Person` objects in `UniquePersonList`.

Zhiyuan-Amos:
@yamgent issue validation pls ^^

yamgent:
Ok

Issue #651, opened 2017-08-31
ReadOnlyPerson: Change implementation


Zhiyuan-Amos:
`ReadOnlyPerson` can easily be casted back to `Person`, causing it to lose the read-only properties. 

![image](https://user-images.githubusercontent.com/19281514/29931324-7dcc2d18-8ea3-11e7-9558-2869524bc296.png)

Having spent a while looking at Java API, there are multiple occasions when the API returns the original type even though it explicitly states that the returned object is unmodifiable (e.g. in the image above, the return type is `ObservableList` and not something like `ReadOnlyListWrapper`). 

As such, I'm wondering whether we can have `ReadOnlyPerson` to subclass `Person`, and we can override the setter methods to throw a `RunTimeException`. That way, in getter methods like `UniquePersonList#asObservableList()`
```
public UnmodifiableObservableList<Person> asObservableList() {
    return new UnmodifiableObservableList<>(internalList);
}
```
Though the type of the list is `Person`, but more specifically it is of type `ReadOnlyPerson`. Then we can add header comments to mention that both the list and the `Person` objects inside the list are unmodifiable. This resolves the problem mentioned in the first paragraph.

-----------

Alternatively, we can make `Person` immutable by removing the setter methods. We don't use them in the main code base except for `setTags`.

Zhiyuan-Amos:
@yamgent thoughts ^^

damithc:
> As such, I'm wondering whether we can have ReadOnlyPerson to subclass Person, and we can override the setter methods to throw a RunTimeException. That way, in getter methods like UniquePersonList#asObservableList()

That would violate the Liskov Substitution Principle

yamgent:
As Prof mentioned, such change would lead to violation of LSP.

If we want to prevent programmer abuse, maybe such methods should return a defensive copy, so that the original copy is not modified?

Zhiyuan-Amos:
> maybe such methods should return a defensive copy

`UniquePersonList` has to return the list as it is because the UI observes the list. If a defensive copy is being returned, what happens is that in the event that a `Person`'s details are being edited in the original list (though we can't do that now because there's no API exposed by `UniquePersonList`), then the UI won't be notified of the change because it has a defensive copy.

> That would violate the Liskov Substitution Principle

[...]

PR #765, opened 2017-12-20
Make Person immutable #751

### LINKED ISSUE #751: Make Person immutable
Our code base doesn't use any of Person's mutation methods/features, and will be unlikely to do so in the future due to how easy it is to break invariants by mutating an existing object which we do not own.

Making Person immutable will simplify the class hierarchy (no more ReadOnlyPerson) and remove some code we'll never use. Immutability means we won't have to do defensive copying any more, so we'll be kinder to the GC.

Zhiyuan-Amos:
Fixes #751, #653, #651.

```
Person is mutable.

This is bad as:
1. It is easy to break invariants by mutating an existing object which
we do not own.
2. No code uses Person’s setter methods.

By making Person immutable, we can make the following improvements:
1. Simplify the class hierarchy as we no longer require ReadOnlyPerson.
2. Remove the need for defensive copying of Person objects.

Let’s make Person immutable.
```


src/test/java/seedu/address/testutil/PersonBuilder.java
```diff
@@ -42,15 +45,19 @@ public PersonBuilder() {
      * Initializes the PersonBuilder with the data of {@code personToCopy}.
      */
     public PersonBuilder(ReadOnlyPerson personToCopy) {
-        this.person = new Person(personToCopy);
+        name = personToCopy.getName();
+        phone = personToCopy.getPhone();
+        email = personToCopy.getEmail();
+        address = personToCopy.getAddress();
+        tags = personToCopy.getTags();
```
pyokagan: I think we need to do a defensive copy here. `Person#getTags()` returns an unmodifiable set. https://github.com/pyokagan/addressbook-level4/blob/895430f4c9434c980cf6ed148e0ad36e37c5c5c2/src/main/java/seedu/address/model/person/Person.java#L111

From looking at the rest of `PersonBuilder`, it seems the set will not be modified, but it's better to make ownership clear.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -18,12 +17,12 @@
...
 public class Person implements ReadOnlyPerson {
 
-    private ObjectProperty<Name> name;
-    private ObjectProperty<Phone> phone;
-    private ObjectProperty<Email> email;
-    private ObjectProperty<Address> address;
+    private final ObjectProperty<Name> name;
+    private final ObjectProperty<Phone> phone;
+    private final ObjectProperty<Email> email;
+    private final ObjectProperty<Address> address;
```
pyokagan: As long as we are using `ObjectProperty` internally, `Person` is still technically mutable.

This should just be plain `final Name`, `final Phone` etc.

src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -91,7 +90,7 @@ public CommandResult executeUndoableCommand() throws CommandException {
      * Creates and returns a {@code Person} with the details of {@code personToEdit}
      * edited with {@code editPersonDescriptor}.
      */
-    private static Person createEditedPerson(ReadOnlyPerson personToEdit,
+    private static Person createEditedPerson(Person personToEdit,
```
pyokagan: Maybe this line could be re-wrapped.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -38,49 +38,41 @@ public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tag
...
+     * Creates a copy of the given Person.
      */
-    public Person(ReadOnlyPerson source) {
+    public Person(Person source) {
         this(source.getName(), source.getPhone(), source.getEmail(), source.getAddress(),
                 source.getTags());
     }
 
-    @Override
     public ObjectProperty<Name> nameProperty() {
```
pyokagan: There's no point exposing `nameProperty()`, `phoneProperty()` etc. because Person is immutable.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -25,12 +25,12 @@
 
     private final ObservableList<Person> internalList = FXCollections.observableArrayList();
     // used by asObservableList()
-    private final ObservableList<ReadOnlyPerson> mappedList = EasyBind.map(internalList, (person) -> person);
+    private final ObservableList<Person> mappedList = EasyBind.map(internalList, (person) -> person);
```
pyokagan: This may not be required already.

pyokagan:
Developer guide model diagrams also need to be updated in [5/5]

pyokagan:
and all the references to `ReadOnlyPerson` in the developer guide as well.

Zhiyuan-Amos:
Thanks for such a thorough review! :)

> [4/5] The copy constructor of Person can also be removed. Since it is
immutable, there is no point making copies.

I'll do it in [5/5], is that ok? That's because, without the copy constructor in [4/4], I cannot convert a `ReadOnlyPerson` to `Person` easily (many classes are still taking in a `ReadOnlyPerson`, but requires a `Person` for processing some stuffs), and I can only do so by doing `new Person(rop.getName(), rop.getPhone()...)`, which is kinda tedious, especially since I will be removing the construction entirely afterwards.

> Was this done with an automated tool?

There are 4 automated steps required for the entire change, not sure if this is what you're looking for. ><

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
```diff
@@ -36,7 +37,7 @@ public EditPersonDescriptorBuilder(ReadOnlyPerson person) {
         descriptor.setPhone(person.getPhone());
         descriptor.setEmail(person.getEmail());
         descriptor.setAddress(person.getAddress());
-        descriptor.setTags(person.getTags());
+        descriptor.setTags(new HashSet<>(person.getTags()));
```
pyokagan: Hmph, this seems like the wrong layer to be doing defensive copying. Shouldn't `EditPersonDescriptor#setTags` be doing the copying?

src/test/java/seedu/address/testutil/PersonBuilder.java
pyokagan: https://github.com/se-edu/addressbook-level4/pull/765#discussion_r158577449

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -1,41 +1,37 @@
...
 
-import javafx.beans.property.ObjectProperty;
-import javafx.beans.property.SimpleObjectProperty;
 import seedu.address.model.tag.Tag;
 import seedu.address.model.tag.UniqueTagList;
 
 /**
  * Represents a Person in the address book.
- * Guarantees: details are present and not null, field values are validated.
+ * Guarantees: details are present and not null, field values are validated, immutability.
```
pyokagan: `Name`, `Phone`, `Address` etc. all use "immutable", so the same should be followed here as well.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -1,41 +1,37 @@
...
     /**
      * Every field must be present and not null.
      */
     public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tags) {
         requireAllNonNull(name, phone, email, address, tags);
-        this.name = new SimpleObjectProperty<>(name);
-        this.phone = new SimpleObjectProperty<>(phone);
-        this.email = new SimpleObjectProperty<>(email);
-        this.address = new SimpleObjectProperty<>(address);
-        // protect internal tags from changes in the arg list
```
pyokagan: This comment originally came from https://github.com/se-edu/addressbook-level4/commit/bf265bd321dce10ad8d2cfc0203aca8f0d4271c3 (aka the original AB4) so it's better to not remove it.

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -44,26 +43,17 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
...
         id.setText(displayedIndex + ". ");
-        initTags(person);
-        bindListeners(person);
+        initPersonDetails(person);
     }
 
     /**
-     * Binds the individual UI elements to observe their respective {@code Person} properties
-     * so that they will be notified of any changes.
+     * Initialises the {@code Label}s with details of {@code person}.
```
pyokagan: `Initializes`

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -44,26 +43,17 @@ public PersonCard(ReadOnlyPerson person, int displayedIndex) {
...
-        address.textProperty().bind(Bindings.convert(person.addressProperty()));
-        email.textProperty().bind(Bindings.convert(person.emailProperty()));
-        person.tagProperty().addListener((observable, oldValue, newValue) -> {
-            tags.getChildren().clear();
-            person.getTags().forEach(tag -> tags.getChildren().add(new Label(tag.tagName)));
-        });
-    }
-
-    private void initTags(ReadOnlyPerson person) {
+    private void initPersonDetails(ReadOnlyPerson person) {
```
pyokagan: Can just inline the whole method into the constructor. Putting the code in a separate method doesn't add value since it does not introduce a new layer of abstraction.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -95,32 +94,32 @@ public void addPerson(ReadOnlyPerson p) throws DuplicatePersonException {
...
+     * @see #syncMasterTagListWith(Person)
      */
-    public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedReadOnlyPerson)
+    public void updatePerson(Person target, Person editedPerson)
             throws DuplicatePersonException, PersonNotFoundException {
-        requireNonNull(editedReadOnlyPerson);
+        requireNonNull(editedPerson);
 
-        Person editedPerson = syncMasterTagListWith(editedReadOnlyPerson);
+        Person syncedEditedPerson = syncMasterTagListWith(editedPerson);
```
pyokagan: Sneaky change here, should be mentioned in the commit message.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -24,13 +22,11 @@
 public class UniquePersonList implements Iterable<Person> {
 
     private final ObservableList<Person> internalList = FXCollections.observableArrayList();
-    // used by asObservableList()
-    private final ObservableList<ReadOnlyPerson> mappedList = EasyBind.map(internalList, (person) -> person);
```
pyokagan: This can be in its own commit.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -66,16 +54,25 @@ public Address getAddress() {
...
+
+        if (!(other instanceof Person)) {
+            return false;
+        }
+
+        Person otherPerson = (Person) other;
+        return otherPerson.getName().equals(this.getName())
+                && otherPerson.getPhone().equals(this.getPhone())
+                && otherPerson.getEmail().equals(this.getEmail())
+                && otherPerson.getAddress().equals(this.getAddress());
```
pyokagan: The movement of `getAsText()` and `isSameStateAs()` from `ReadOnlyPerson` to `Person` should be in its own commit.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -34,30 +34,18 @@ public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tag
         this.tags = new UniqueTagList(tags);
     }
 
-    /**
-     * Creates a copy of the given ReadOnlyPerson.
-     */
-    public Person(ReadOnlyPerson source) {
```
pyokagan: Removal of copy constructor should be in its own commit.

src/test/java/seedu/address/testutil/EditPersonDescriptorBuilder.java
Zhiyuan-Amos: Ah yes, I mixed this up with your other comment https://github.com/se-edu/addressbook-level4/pull/765#discussion_r158577449. I.e. this change shouldn't be made here. 

On a side note, `EditPersonDescriptorBuilder` & `EditPersonDescriptor` have potential bugs since there's no defensive copying.

Zhiyuan-Amos:
Thanks for your review @pyokagan , sorry for missing out some stuffs in the previous iteration ><

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -157,7 +157,7 @@ public ReadOnlyAddressBook getAddressBook() {
 
         @Override
         public void addPerson(Person person) throws DuplicatePersonException {
-            personsAdded.add(new Person(person));
+            personsAdded.add(person);
```
pyokagan: There should be a non-null assertion here as well to catch errors as soon as possible in tests.

Zhiyuan-Amos:
> No further comments other than the observation that the name syncMasterTagListWith doesn't seem to reflect its true purpose. It sounds like a one-way sync but actually is a two-way sync?

Yup hmm, does `syncWithMasterTagList` sound better?

damithc:
> Yup hmm, does syncWithMasterTagList sound better?

Yes, that's better.

Zhiyuan-Amos:
@damithc for your review in case you forgot, since you reviewed my other PR but not this one :P

PR #768, opened 2017-12-24
README.adoc: add Gitter badge

pyokagan:
```
Let the world know where all the cool people hang out.
```


damithc:
> @damithc are we merging this? :P

We can. Students will have to remove after forking though. I suppose they'll figure it out themselves.

PR #776, opened 2017-12-30
Developer Guide: specify target user profile and value proposition #775

### LINKED ISSUE #775: Developer Guide: specify target user profile and value proposition
One needs to have a clearly defined target user profile and a value
proposition for the app before other requirements are defined. Both
those thing are not specified in the Developer Guide. 

Let' add a sample target user profile and a value proposition to the
DG, just above where user stories are defined.

damithc:
Fixes #775

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -734,6 +734,19 @@ We now have everything set up... but we still can't modify the remarks. Let's fi
 See this https://github.com/se-edu/addressbook-level4/pull/599[PR] for the step-by-step solution.
 
 [appendix]
+== Product Scope
+
+*Target user profile*:
+
+* has a need to manage a significant number of contacts
+* prefer desktop apps over other types
+* can type fast
+* prefers typing over mouse input
+* is reasonably comfortable using CLI apps
+
+*Value proposition*: manage contacts faster than a typical mouse/GUI driven app
+
+[appendix]
 == User Stories
 
 Priorities: High (must have) - `* * \*`, Medium (nice to have) - `* \*`, Low (unlikely to have) - `*`


docs/DeveloperGuide.adoc
```diff
@@ -728,6 +728,10 @@ See this https://github.com/se-edu/addressbook-level4/pull/599[PR] for the step-
 [appendix]
 == User Stories
 
+*Target user profile* : Has a need to manage a significant number of contacts using a desktop app. Can type fast. Prefers typing over mouse input. Experienced with CLI apps.
```
Zhiyuan-Amos: Additional space after `Target user profile`

docs/DeveloperGuide.adoc
```diff
@@ -728,6 +728,10 @@ See this https://github.com/se-edu/addressbook-level4/pull/599[PR] for the step-
 [appendix]
 == User Stories
 
+*Target user profile*: Has a need to manage a significant number of contacts using a desktop app. Can type fast. Prefers typing over mouse input. Experienced with CLI apps.
```
Zhiyuan-Amos: Sorry for not catching this in the earlier iteration: Should we use numbered points over here instead, since there's 4 points?

docs/DeveloperGuide.adoc
```diff
@@ -728,6 +728,16 @@ See this https://github.com/se-edu/addressbook-level4/pull/599[PR] for the step-
 [appendix]
 == User Stories
 
+*Target user profile*:
```
yamgent: Would this be better on a section of its own? While they certainty influence what kind of user stories will be included, they are not really user stories per se.

damithc: Yes indeed. Didn't want to add it as a separate appendix because appendix numbers are used in the module website. Let me take another look to see if it is possible.

damithc: Updated as a separate appendix.

PR #780, opened 2017-12-30
User Guide: add an introduction #779

### LINKED ISSUE #779: Use Guide: add an introduction
Let's add an introduction to the UG which specifieses the following:
* target user profile
* value proposition 

Benefits:
1. the reader gets an early preview of if the app is a good fit for 
   her needs
2. students peer-testing an app based on the UG has easy access to 
   the target user profile and the value proposition

Note: While this information is also present in the Developer Guide,
the target audience for the information (hence, the language used)
is different in the two places.

damithc:
Fixes #779

### DIFF
--- a/docs/UserGuide.adoc
+++ b/docs/UserGuide.adoc
@@ -15,6 +15,10 @@ endif::[]
 
 By: `Team SE-EDU`      Since: `Jun 2016`      Licence: `MIT`
 
+== Introduction
+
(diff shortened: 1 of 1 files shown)


docs/UserGuide.adoc
```diff
@@ -15,6 +15,10 @@ endif::[]
 
 By: `Team SE-EDU`      Since: `Jun 2016`      Licence: `MIT`
 
+== Introduction
+
+AddressBook Level 4 (AB4) is a contact management app for those who *prefer to use a desktop app for managing contacts*. More importantly, AB4 is *optimized for those who prefer to work with a Command Line Interface* (CLI) while still having the benefits of a Graphical User Interface (GUI). If you are a power user who can type fast, AB4 can get your contact management tasks done faster than traditional GUI apps. Interested? Jump to the <<Quick Start>> to get started. Enjoy!
```
Zhiyuan-Amos: > contact management app for those who prefer to use a desktop app for managing contacts

the phrase `contact management app` already implies `managing contacts`, so it's a repetition I think.

> If you are a power user who can type fast

`power user` means `end user expert in a specific computing domain` (did a quick Google), so being able to type fast doesn't quite fall into that category right? ><

Issue #784, opened 2017-12-31
Rebase PRs used in suggested programming tasks


damithc:
PRs given as examples in _Developer Guide: Appendix A: Suggested Programming Tasks to Get Started_ need to be rebased to the latest master and ensure they still work as advertised.

Zhiyuan-Amos:
@yamgent can help me verify if the below checkboxes are correct?

- [ ] Command alias #785
- [ ] Remove tag from Person #790
- [ ] Person's `Tag` with color #798 
- [ ] ResultDisplay with error style #799 
- [ ] StatusBar with total Persons #803 
- [ ] Save alternate location #594
- [ ] Remark

yamgent:
> @yamgent can help me verify if the below checkboxes are correct?

Correct so far :+1:

Zhiyuan-Amos:
This issue will be resolved once #789 is merged. However, if we merge #792 (which fixes the bug for undo/redo commands), I think #599 won't be broken, but it'll be buggy (see #792).

yamgent:
> However, if we merge #792 (which fixes the bug for undo/redo commands), I think #599 won't be broken, but it'll be buggy (see #792).

I will take a look at #599 again when that happens then. Don't think I can do anything else to rectify that problem now.

Zhiyuan-Amos:
@yamgent #599 isn't broken but it's buggy (when you undo/redo on a filtered list for `RemarkCommand`). The fix can be found in #792 [3/6] :)

Zhiyuan-Amos:
I've re-run Travis for the PRs mentioned [above](https://github.com/se-edu/addressbook-level4/issues/784#issuecomment-360106921) and they are all passing :)

Issue #753, opened 2017-12-04
AddressBook needs to track tags properly


pyokagan:
https://github.com/se-edu/addressbook-level4/blob/edc42f0e2579c27c209b1f8f3c0f87ed04e83904/src/main/java/seedu/address/model/AddressBook.java#L89-L91

https://github.com/se-edu/addressbook-level4/blob/edc42f0e2579c27c209b1f8f3c0f87ed04e83904/src/main/java/seedu/address/model/AddressBook.java#L111-L113

There's also the issue that tags are not removed once the last person using that tag is removed from the address book (i.e. the set of tags will just keep growing :-P)

The tag list is also actually not used at all. We should either find some use for it, or remove it.

damithc:
IIRC, the original intention was to have only one instance of each tag as a list of `Tag` objects and all `Person` objects should point to the relevant tags in the list, instead of keeping its own copy. That way, we just need to modify one Tag object when a Tag is edited.

Zhiyuan-Amos:
> That way, we just need to modify one Tag object when a Tag is edited.

`Tag` is immutable though :P Also, we don't have a tag renaming feature.

damithc:
> Tag is immutable though :P Also, we don't have a tag renaming feature
Tag renaming is one of the common extensions students do. Also, another possible extension is the ability to set/change color of a tag. As an OOP solution, we should treat tags as an object that is referenced by multiple other objects rather than a value that is being copied to multiple other objects?

pyokagan:
@damithc 

> As an OOP solution, we should treat tags as an object that is referenced by multiple other objects rather than a value that is being copied to multiple other objects?

I dunno, I wonder if students in the CS2103 level still have trouble with pointers. Even some SE-EDU developers in the past had trouble deciphering the intent of `syncMasterTagListWith`

damithc:
How about we do it the simple way (i.e., no independent tag list) and give the other approach as a 'more OOP' alternative in the dev guide?

Zhiyuan-Amos:
As discussed [here](https://github.com/se-edu/addressbook-level4/pull/794#issuecomment-360468923), we will be removing `UniqueTagList`.

PR #790, opened 2018-01-07
Model: Add deleteTag(Tag)

yamgent:
Part of #784. Fixed the `deleteTag(Tag)` exception by overhauling the design. Also added test.

Proposed commit message:

```
The user may want to remove a particular tag from all persons in the
AddressBook. For example, he may want to delete the tag "colleague"
from all his contacts if he is no longer working at his former company.

No such command exist at the moment, as the model API does not expose a
way to delete a particular tag entirely from the AddressBook (i.e. 
removing the Tag from all Persons and purging the Tag in the 
AddressBook's tag list).

As a step towards allowing the user to remove a particular tag with a
single command, let's implement Model#deleteTag(Tag), so that the
developer dealing with the 'logic' component is able to implement such
a command.
```

_Note:_ For demo purpose, don't merge this to master!

### DIFF
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -2,11 +2,14 @@ package seedu.address.model;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import javafx.collections.ObservableList;
 import seedu.address.model.person.Person;
 import seedu.address.model.person.UniquePersonList;
+import seedu.address.model.tag.Tag;
 
 /**
  * Wraps all data at the address-book level
@@ -91,6 +94,29 @@ public class AddressBook implements ReadOnlyAddressBook {
      */
     public void removePerson(Person key) {
         persons.remove(key);
+    }
+
+    /**
+     * Removes {@code tag} from {@code person} in this {@code AddressBook}.
+     */
+    private void removeTagFromPerson(Tag tag, Person person) {
+        Set<Tag> newTags = new HashSet<>(person.getTags());
+
+        if (!newTags.remove(tag)) {
+            return;
+        }
+
+        Person newPerson =
+                new Person(person.getName(), person.getPhone(), person.getEmail(), person.getAddress(), newTags);
+
+        updatePerson(person, newPerson);
+    }
+
+    /**
+     * Removes {@code tag} from all persons in this {@code AddressBook}.
+     */
+    public void removeTag(Tag tag) {
+        persons.forEach(person -> removeTagFromPerson(tag, person));
     }
 
     //// util methods
--- a/src/main/java/seedu/address/model/ModelManager.java
+++ b/src/main/java/seedu/address/model/ModelManager.java
@@ -13,6 +13,7 @@ import seedu.address.commons.core.ComponentManager;
 import seedu.address.commons.core.LogsCenter;
 import seedu.address.commons.events.model.AddressBookChangedEvent;
 import seedu.address.model.person.Person;
+import seedu.address.model.tag.Tag;
 
 /**
  * Represents the in-memory model of the address book data.
@@ -81,6 +82,11 @@ public class ModelManager extends ComponentManager implements Model {
 
         versionedAddressBook.updatePerson(target, editedPerson);
         indicateAddressBookChanged();
+    }
+
+    @Override
+    public void deleteTag(Tag tag) {
+        versionedAddressBook.removeTag(tag);
     }
 
     //=========== Filtered Person List Accessors =============================================================
--- a/src/test/java/seedu/address/model/AddressBookTest.java
+++ b/src/test/java/seedu/address/model/AddressBookTest.java
@@ -4,8 +4,12 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_UNUSED;
 import static seedu.address.testutil.TypicalPersons.ALICE;
+import static seedu.address.testutil.TypicalPersons.AMY;
+import static seedu.address.testutil.TypicalPersons.BOB;
 import static seedu.address.testutil.TypicalPersons.getTypicalAddressBook;
 
 import java.util.Arrays;
@@ -21,6 +25,8 @@ import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import seedu.address.model.person.Person;
 import seedu.address.model.person.exceptions.DuplicatePersonException;
+import seedu.address.model.tag.Tag;
+import seedu.address.testutil.AddressBookBuilder;
 import seedu.address.testutil.PersonBuilder;
 
 public class AddressBookTest {
@@ -29,6 +35,8 @@ public class AddressBookTest {
     public ExpectedException thrown = ExpectedException.none();
 
     private final AddressBook addressBook = new AddressBook();
+    private final AddressBook addressBookWithBobAndAmy = new AddressBookBuilder().withPerson(BOB)
+            .withPerson(AMY).build();
 
     @Test
     public void constructor() {
@@ -89,6 +97,27 @@ public class AddressBookTest {
     public void getPersonList_modifyList_throwsUnsupportedOperationException() {
         thrown.expect(UnsupportedOperationException.class);
         addressBook.getPersonList().remove(0);
+    }
+
+    @Test
+    public void removeTag_nonExistentTag_addressBookUnchanged() throws Exception {
+        addressBookWithBobAndAmy.removeTag(new Tag(VALID_TAG_UNUSED));
+
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(BOB).withPerson(AMY).build();
+
+        assertEquals(expectedAddressBook, addressBookWithBobAndAmy);
+    }
+
+    @Test
+    public void removeTag_tagUsedByMultiplePersons_tagRemoved() throws Exception {
+        addressBookWithBobAndAmy.removeTag(new Tag(VALID_TAG_FRIEND));
+
+        Person amyWithoutFriendTag = new PersonBuilder(AMY).withTags().build();
+        Person bobWithoutFriendTag = new PersonBuilder(BOB).withTags(VALID_TAG_HUSBAND).build();
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(bobWithoutFriendTag)
+                .withPerson(amyWithoutFriendTag).build();
+
+        assertEquals(expectedAddressBook, addressBookWithBobAndAmy);
     }
 
     /**
--- a/src/test/java/seedu/address/model/ModelManagerTest.java
+++ b/src/test/java/seedu/address/model/ModelManagerTest.java
@@ -1,10 +1,16 @@
 package seedu.address.model;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_FRIEND;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_UNUSED;
 import static seedu.address.model.Model.PREDICATE_SHOW_ALL_PERSONS;
 import static seedu.address.testutil.TypicalPersons.ALICE;
+import static seedu.address.testutil.TypicalPersons.AMY;
 import static seedu.address.testutil.TypicalPersons.BENSON;
+import static seedu.address.testutil.TypicalPersons.BOB;
 
 import java.nio.file.Paths;
 import java.util.Arrays;
@@ -14,7 +20,10 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import seedu.address.model.person.NameContainsKeywordsPredicate;
+import seedu.address.model.person.Person;
+import seedu.address.model.tag.Tag;
 import seedu.address.testutil.AddressBookBuilder;
+import seedu.address.testutil.PersonBuilder;
 
 public class ModelManagerTest {
     @Rule
@@ -43,6 +52,34 @@ public class ModelManagerTest {
     public void getFilteredPersonList_modifyList_throwsUnsupportedOperationException() {
         thrown.expect(UnsupportedOperationException.class);
         modelManager.getFilteredPersonList().remove(0);
+    }
+
+    @Test
+    public void deleteTag_nonExistentTag_modelUnchanged() throws Exception {
+        AddressBook addressBook = new AddressBookBuilder().withPerson(AMY).withPerson(BOB).build();
+        UserPrefs userPrefs = new UserPrefs();
+
+        ModelManager modelManager = new ModelManager(addressBook, userPrefs);
+        modelManager.deleteTag(new Tag(VALID_TAG_UNUSED));
+
+        assertEquals(new ModelManager(addressBook, userPrefs), modelManager);
+    }
+
+    @Test
+    public void deleteTag_tagUsedByMultiplePersons_tagRemoved() throws Exception {
+        AddressBook addressBook = new AddressBookBuilder().withPerson(AMY).withPerson(BOB).build();
+        UserPrefs userPrefs = new UserPrefs();
+
+        ModelManager modelManager = new ModelManager(addressBook, userPrefs);
+        modelManager.deleteTag(new Tag(VALID_TAG_FRIEND));
+
+        ModelManager expectedModelManager = new ModelManager(addressBook, userPrefs);
+        Person amyWithoutFriendTag = new PersonBuilder(AMY).withTags().build();
+        Person bobWithoutFriendTag = new PersonBuilder(BOB).withTags(VALID_TAG_HUSBAND).build();
+        expectedModelManager.updatePerson(AMY, amyWithoutFriendTag);
+        expectedModelManager.updatePerson(BOB, bobWithoutFriendTag);
+
+        assertEquals(expectedModelManager, modelManager);
     }
 
     @Test
--- a/src/main/java/seedu/address/model/Model.java
+++ b/src/main/java/seedu/address/model/Model.java
@@ -4,6 +4,7 @@ import java.util.function.Predicate;
 
 import javafx.collections.ObservableList;
 import seedu.address.model.person.Person;
+import seedu.address.model.tag.Tag;
 
 /**
  * The API of the Model component.
@@ -41,6 +42,9 @@ public interface Model {
      * The person identity of {@code editedPerson} must not be the same as another existing person in the address book.
      */
     void updatePerson(Person target, Person editedPerson);
+
+    /** Removes the given {@code tag} from all {@code Person}s. */
+    void deleteTag(Tag tag);
 
     /** Returns an unmodifiable view of the filtered person list */
     ObservableList<Person> getFilteredPersonList();
--- a/src/test/java/seedu/address/logic/commands/AddCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/AddCommandTest.java
@@ -20,6 +20,7 @@ import seedu.address.model.AddressBook;
 import seedu.address.model.Model;
 import seedu.address.model.ReadOnlyAddressBook;
 import seedu.address.model.person.Person;
+import seedu.address.model.tag.Tag;
 import seedu.address.testutil.PersonBuilder;
 
 public class AddCommandTest {
@@ -150,6 +151,11 @@ public class AddCommandTest {
 
         @Override
         public void commitAddressBook() {
+            throw new AssertionError("This method should not be called.");
+        }
+
+        @Override
+        public void deleteTag(Tag tag) {
             throw new AssertionError("This method should not be called.");
         }
     }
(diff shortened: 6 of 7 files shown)


src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,6 +112,35 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedReadOnlyPer
...
+            if (isTagUsed(tag)) {
+                tagsInUse.add(tag);
+            }
+        });
+
+        tags.setTags(tagsInUse);
+    }
+
+    /**
+     * Returns whether any {@code Person} in this {@code AddressBook} is still using {@code tag}.
```
Zhiyuan-Amos: Returns **true** ~~whether~~ **if** any {@code Person} in ~~this {@code AddressBook}~~ **{@code persons}** is ~~still~~ using {@code tag}.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,6 +112,35 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedReadOnlyPer
...
+        removeUnusedTags();
+    }
+
+    /**
+     * Removes all {@code Tag}s that are not used by any {@code Person} in this {@code AddressBook}.
+     */
+    private void removeUnusedTags() {
+        Set<Tag> tagsInUse = new HashSet<Tag>();
+
+        tags.forEach(tag -> {
```
Zhiyuan-Amos: We should use enhanced for-loop here, since you are doing so as well for the method `isTagUsed`, so we should standardise it (since there's no additional benefit compared to using `forEach` over enhanced for-loop.

Alternatively, you can update the `UniquePersonList` & `UniqueTagList` to return a `Stream<Person>` and `Stream<Tag>` respectively. That will shorten these methods here quite substantially.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,6 +112,35 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedReadOnlyPer
         // This can cause the tags master list to have additional tags that are not tagged to any person
         // in the person list.
         persons.setPerson(target, editedPerson);
+        removeUnusedTags();
```
Zhiyuan-Amos: you spotted a bug here, this should actually be fixed in master right? Wanna raise an issue for this?

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -69,6 +72,16 @@ public void getTagList_modifyList_throwsUnsupportedOperationException() {
         addressBook.getTagList().remove(0);
     }
 
+    @Test
+    public void updatePerson_detailsChanged_personsAndTagsListUpdated() throws Exception {
+        addressBook.addPerson(BOB);
```
Zhiyuan-Amos: I'm wondering if we should use `AddressBookBuilder()` to create an `AddressBook` with `BOB`. Then we call `addressBook.updatePerson(BOB, AMY);` afterwards. 

This is so that we don't need to rely on `AddressBook#addPerson(Person)` to be working correctly, since this method only tests for `updatePerson(Person, Person)`

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,6 +112,35 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedReadOnlyPer
...
+    }
+
+    /**
+     * Removes all {@code Tag}s that are not used by any {@code Person} in this {@code AddressBook}.
+     */
+    private void removeUnusedTags() {
+        Set<Tag> tagsInUse = new HashSet<Tag>();
```
Zhiyuan-Amos: Oh, the word `Tag` in `new HashSet<Tag>` is unnecessary, as reported by my trustworthy IntelliJ :P

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -82,6 +86,29 @@ public void updatePerson_detailsChanged_personsAndTagsListUpdated() throws Excep
...
+
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(AMY).build();
+
+        assertEquals(expectedAddressBook, addressBook);
+    }
+
+    @Test
+    public void removeTagFromPerson_tagUsedByMultiplePersons_personUpdated() throws Exception {
+        addressBook.addPerson(AMY);
+        addressBook.addPerson(BOB);
```
Zhiyuan-Amos: Same comment as mentioned [above](https://github.com/se-edu/addressbook-level4/pull/790#discussion_r160121896).

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -191,6 +191,22 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
         tags.add(t);
     }
 
+    /**
+     * Removes {@code tag} from {@code person} in this {@code AddressBook}.
+     * @throws DuplicatePersonException if removing the tag causes the {@code person} to be equivalent to another
```
Zhiyuan-Amos: The current implementation of `equals` doesn't allow for this exception to happen actually, since it doesn't consider equality of tags. As such, perhaps we shouldn't throw this exception, but do a `try-catch` followed with `throw new AssertionError()`?

I'm not too picky about this though, @damithc what do you think?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -207,6 +207,15 @@ public void removeTagFromPerson(Tag tag, ReadOnlyPerson person) throws Duplicate
         updatePerson(person, newPerson);
     }
 
+    /**
+     * Removes {@code tag} in this {@code AddressBook}.
+     */
+    public void removeTag(Tag tag) throws DuplicatePersonException, PersonNotFoundException {
```
Zhiyuan-Amos: since this method shouldn't be throwing both Exceptions, we should do a `try-catch` followed by `throw new AssertionError()`.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -109,6 +109,34 @@ public void removeTagFromPerson_tagUsedByMultiplePersons_personUpdated() throws
...
+    @Test
+    public void removeTag_nonExistentTag_sameAddressBook() throws Exception {
+        addressBook.addPerson(AMY);
+        addressBook.addPerson(BOB);
+
+        addressBook.removeTag(new Tag(VALID_TAG_UNUSED));
+
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(AMY).withPerson(BOB).build();
+
+        assertEquals(addressBook, expectedAddressBook);
```
Zhiyuan-Amos: wrong order of `assertEquals` :P

src/main/java/seedu/address/model/ModelManager.java
```diff
@@ -81,6 +82,13 @@ public void updatePerson(ReadOnlyPerson target, ReadOnlyPerson editedPerson)
         indicateAddressBookChanged();
     }
 
+    @Override
+    public void deleteTag(Tag tag) throws PersonNotFoundException, DuplicatePersonException {
+        for (ReadOnlyPerson person : addressBook.getPersonList()) {
+            addressBook.removeTagFromPerson(tag, person);
+        }
```
Zhiyuan-Amos: you can just call `addressBook.removeTag(tag);` right? :P That way, you can make `removeTagFromPerson` private instead.

src/test/java/seedu/address/model/ModelManagerTest.java
```diff
@@ -26,6 +35,34 @@ public void getFilteredPersonList_modifyList_throwsUnsupportedOperationException
...
 
+    @Test
+    public void deleteTag_nonExistentTag_sameAddressBook() throws Exception {
+        AddressBook addressBook = new AddressBookBuilder().withPerson(AMY).withPerson(BOB).build();
+        UserPrefs userPrefs = new UserPrefs();
+
+        ModelManager modelManager = new ModelManager(addressBook, userPrefs);
+        modelManager.deleteTag(new Tag(VALID_TAG_UNUSED));
+
+        assertEquals(modelManager, new ModelManager(addressBook, userPrefs));
```
Zhiyuan-Amos: wrong ordering of `assertEquals` :P

src/test/java/seedu/address/model/ModelManagerTest.java
```diff
@@ -26,6 +35,34 @@ public void getFilteredPersonList_modifyList_throwsUnsupportedOperationException
...
+        ModelManager modelManager = new ModelManager(addressBook, userPrefs);
+        modelManager.deleteTag(new Tag(VALID_TAG_FRIEND));
+
+        Person amyWithoutFriendTag = new PersonBuilder(AMY).withTags().build();
+        Person bobWithoutFriendTag = new PersonBuilder(BOB).withTags(VALID_TAG_HUSBAND).build();
+
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(amyWithoutFriendTag)
+                .withPerson(bobWithoutFriendTag).build();
+
+        assertEquals(modelManager, new ModelManager(expectedAddressBook, userPrefs));
```
Zhiyuan-Amos: wrong ordering of `assertEquals` :P

src/test/java/seedu/address/model/ModelManagerTest.java
```diff
@@ -26,6 +35,34 @@ public void getFilteredPersonList_modifyList_throwsUnsupportedOperationException
         modelManager.getFilteredPersonList().remove(0);
     }
 
+    @Test
+    public void deleteTag_nonExistentTag_sameAddressBook() throws Exception {
```
Zhiyuan-Amos: should the last segment of this test be named as `modelUnchanged` instead?

src/main/java/seedu/address/model/AddressBook.java
yamgent: This is already raised in #753 actually.

This commit is a temporary fix, it is not ideal because the complexity will become O(n^2). An overhaul of the tag list is probably required.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -69,6 +72,16 @@ public void getTagList_modifyList_throwsUnsupportedOperationException() {
         addressBook.getTagList().remove(0);
     }
 
+    @Test
+    public void updatePerson_detailsChanged_personsAndTagsListUpdated() throws Exception {
+        AddressBook addressBookWithBob = new AddressBookBuilder().withPerson(BOB).build();
+        addressBookWithBob.updatePerson(BOB, AMY);
```
Zhiyuan-Amos: after the update, this address book is no longer `addressBookWithBob` cos it actually stores `AMY`. Not sure if this may end up being confusing. I can't think of a better naming though. Maybe `addressBookUpdatedToAmy`?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -191,6 +191,25 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
...
+        Set<Tag> newTags = new HashSet<>(person.getTags());
+
+        if (newTags.remove(tag)) {
+            Person newPerson = new Person(person.getName(), person.getPhone(), person.getEmail(), person.getAddress(),
+                    newTags);
+
+            try {
+                updatePerson(person, newPerson);
+            } catch (DuplicatePersonException dpe) {
+                throw new AssertionError("Modifying a person's tags only should not result in a duplicate.");
```
Zhiyuan-Amos: Should we add something like "See Person#equals(Object)" for clarity?

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -116,6 +116,28 @@ public void removeTagFromPerson_tagUsedByMultiplePersons_personUpdated() throws
         assertEquals(expectedAddressBook, addressBookWithBobAndAmy);
     }
 
+    @Test
+    public void removeTag_nonExistentTag_sameAddressBook() throws Exception {
```
Zhiyuan-Amos: should we change `sameAddressBook` -> `addressBookUnchanged`? Sounds a bit more natural.

yamgent:
Resolved conflict as a result of some changes in the codebase.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -182,6 +182,39 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
...
+    /**
+     * Removes {@code tag} from {@code person} in this {@code AddressBook}.
+     * @throws PersonNotFoundException if the {@code person} is not in this {@code AddressBook}.
+     */
+    private void removeTagFromPerson(Tag tag, Person person) throws PersonNotFoundException {
+        Set<Tag> newTags = new HashSet<>(person.getTags());
+
+        if (newTags.remove(tag)) {
+            Person newPerson = new Person(person.getName(), person.getPhone(), person.getEmail(), person.getAddress(),
+                    newTags);
```
Zhiyuan-Amos: Can indent better

```
new Person(
           person.getName()...);
```

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -182,6 +182,39 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
...
+                updatePerson(person, newPerson);
+            } catch (DuplicatePersonException dpe) {
+                throw new AssertionError("Modifying a person's tags only should not result in a duplicate. "
+                        + "See Person#equals(Object).");
+            }
+        }
+    }
+
+    /**
+     * Removes {@code tag} in this {@code AddressBook}.
```
Zhiyuan-Amos: "Removes {@code tag} **from all persons** in this {@code AddressBook}."

src/test/java/seedu/address/model/ModelManagerTest.java
```diff
@@ -26,6 +35,34 @@ public void getFilteredPersonList_modifyList_throwsUnsupportedOperationException
...
+        AddressBook addressBook = new AddressBookBuilder().withPerson(AMY).withPerson(BOB).build();
+        UserPrefs userPrefs = new UserPrefs();
+
+        ModelManager modelManager = new ModelManager(addressBook, userPrefs);
+        modelManager.deleteTag(new Tag(VALID_TAG_FRIEND));
+
+        Person amyWithoutFriendTag = new PersonBuilder(AMY).withTags().build();
+        Person bobWithoutFriendTag = new PersonBuilder(BOB).withTags(VALID_TAG_HUSBAND).build();
+
+        AddressBook expectedAddressBook = new AddressBookBuilder().withPerson(amyWithoutFriendTag)
```
Zhiyuan-Amos: can remove line 59 cos these 3 variable declarations are quite related; preparing for the expectedAB

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -112,6 +112,35 @@ public void updatePerson(Person target, Person editedPerson)
         // This can cause the tags master list to have additional tags that are not tagged to any person
         // in the person list.
         persons.setPerson(target, syncedEditedPerson);
+        removeUnusedTags();
+    }
+
+    /**
+     * Removes all {@code Tag}s that are not used by any {@code Person} in this {@code AddressBook}.
+     */
+    private void removeUnusedTags() {
```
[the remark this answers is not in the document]
yamgent: Was there a reason you used `currentTagSet.retainAll(tagsInPersons);` instead of `tags.setTags(tagsInPersons);`?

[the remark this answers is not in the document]
Zhiyuan-Amos: Was under the impression that `tagsInPersons` may not be pointing to the tags in the master tag list.

But actually they are! Opps. So using `tags.setTags(tagsInPersons)` should work too ^^ Sorry for that.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -165,6 +165,39 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
...
     }
 
+    /**
+     * Removes {@code tag} from {@code person} in this {@code AddressBook}.
+     * @throws PersonNotFoundException if the {@code person} is not in this {@code AddressBook}.
+     */
+    private void removeTagFromPerson(Tag tag, Person person) throws PersonNotFoundException {
+        Set<Tag> newTags = new HashSet<>(person.getTags());
+
+        if (newTags.remove(tag)) {
```
Zhiyuan-Amos: should we reverse the logic so that we don't have that many indentations? i.e.

```
if (!(newTags.remove(tag)) {
    return;
}

Person newPerson = ....
```

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -165,6 +165,39 @@ public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
...
+                        + "See Person#equals(Object).");
+            }
+        }
+    }
+
+    /**
+     * Removes {@code tag} from all persons in this {@code AddressBook}.
+     */
+    public void removeTag(Tag tag) {
+        try {
```
Zhiyuan-Amos: Should we do:

```
for (Person person : persons) {
    try {
        removeTagFromPerson(tag, person);
    } catch...
}
```

Looks more intuitive to me, though I don't quite know how to justify :P

src/main/java/seedu/address/model/AddressBook.java
yamgent: That makes sense, would follow the golden path principle too. Thanks for pointing it out.

src/main/java/seedu/address/model/AddressBook.java
yamgent: > ```
> for (Person person : persons) {
>     try {
>         removeTagFromPerson(tag, person);
>     } catch...
> }
> ```

That would make sense if we need to handle each person differently.

However, I did it this way in the PR because such an exception is supposed to be impossible in the first place, so in order not to clutter the for-loop with unnecessary details, the try-catch is outside instead of inside.

What do you think about this?

Zhiyuan-Amos: Ok, I'm good with this too ^^

yamgent:
> Except that commit 1 isn't required for deleteTag(Tag) right?

Is that a question or a request for change...? :P

The first commit isn't tackling the `deleteTag(Tag)` problem, but is needed as a result of a shortcoming of the codebase.

Zhiyuan-Amos:
> Is that a question or a request for change...? :P
The first commit isn't tackling the deleteTag(Tag) problem, but is needed as a result of a shortcoming of the codebase.

I see :P I'm not sure, cos what's mentioned in the Developer Guide is that they simply need to implement a `deleteTag(Tag)`, so the extra commit may be slightly confusing for the students. ><

yamgent:
I can make a note in the developer guide to explain the extra commit.

yamgent:
> Any updates needed to the dev guide?

Yes, the new SLAP design + test cases should be briefly mentioned in the solution. Also the PR link needs to be updated (this is one of the few PRs, that got rebased before the PR was reopened :P).

yamgent:
* [v6 1/3] was purged because the codebase finally got rid of `syncWithMasterTagList()`, which was what the commit was fixing (hooray 🎉!)
* 20502f52686b29b6b038fd03ff6dce220d26404f: `deleteTag_tagUsedByMultiplePersons_tagRemoved()` had its test logic rewritten because we now have version address books.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -91,6 +94,41 @@ public void removePerson(Person key) throws PersonNotFoundException {
...
+        if (!newTags.remove(tag)) {
+            return;
+        }
+
+        Person newPerson =
+                new Person(person.getName(), person.getPhone(), person.getEmail(), person.getAddress(), newTags);
+
+        try {
+            updatePerson(person, newPerson);
+        } catch (DuplicatePersonException dpe) {
```
Zhiyuan-Amos: `DuplicatePersonException` will be made `RuntimeException` in #896 tho :(

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -91,6 +94,41 @@ public void removePerson(Person key) throws PersonNotFoundException {
...
+
+    /**
+     * Removes {@code tag} from all persons in this {@code AddressBook}.
+     */
+    public void removeTag(Tag tag) {
+        try {
+            for (Person person : persons) {
+                removeTagFromPerson(tag, person);
+            }
+        } catch (PersonNotFoundException pnfe) {
```
Zhiyuan-Amos: Should we catch `PNFE` and `DPE` at the same location (in `removeTagFromPerson`)?

src/main/java/seedu/address/model/AddressBook.java
pyokagan: Eh, I'll say that is a good thing, given the

```java
            throw new AssertionError("Modifying a person's tags only should not result in a duplicate. "
                    + "See Person#equals(Object).");
```

below.

src/main/java/seedu/address/model/AddressBook.java
yamgent: Hmm, it makes sense to handle `PNFE` here, because callers of `removeTagFromPerson` might actually pass in a `person` that does not exist in the address book, so it is a valid exception in that sense. However, handling `DPE` here would be quite strange since callers are not expected to manipulate `Person`s themselves.

yamgent: Nvm I read your comments wrongly, will update it :P.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -93,6 +98,38 @@ public void removePerson(Person key) {
...
+        if (!newTags.remove(tag)) {
+            return;
+        }
+
+        Person newPerson =
+                new Person(person.getName(), person.getPhone(), person.getEmail(), person.getAddress(), newTags);
+
+        try {
+            updatePerson(person, newPerson);
+        } catch (DuplicatePersonException dpe) {
```
pyokagan: Seems like this can be removed?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -93,6 +98,38 @@ public void removePerson(Person key) {
...
+            throw new AssertionError("Modifying a person's tags only should not result in a duplicate. "
+                    + "See Person#equals(Object).");
+        } catch (PersonNotFoundException pnfe) {
+            throw new AssertionError("Original person must be obtained from the address book.");
```
pyokagan: Seems like this can be removed?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -93,6 +98,38 @@ public void removePerson(Person key) {
...
+        }
+    }
+
+    /**
+     * Removes {@code tag} from all persons in this {@code AddressBook}.
+     */
+    public void removeTag(Tag tag) {
+        for (Person person : persons) {
+            removeTagFromPerson(tag, person);
```
Zhiyuan-Amos: Can we do `foreach` here instead?

Issue #757, opened 2017-12-06
Explore adapting AB4 to use a chat interface


damithc:
A CLI interface is very close to a chat interfaces but the latter sounds more 'modern'. Perhaps adapting our command box component to work like a chat interface might make AB4 experience more _marketable_ (e.g. when students apply for internships/jobs)?

pyokagan:
@damithc Now that the chat bot fad (:stuck_out_tongue: ) is over, this seems like it won't really help. Perhaps we should just close this issue?

damithc:
> @damithc Now that the chat bot fad (😛 ) is over, this seems like it won't really help. Perhaps we should just close this issue?

Sure.

PR #793, opened 2018-01-11
BrowserPanel: load dummy page #733

### LINKED ISSUE #733: Some tests can fail on Travis if Google rejects search requests
Some tests depend on being able to load the Google search page. These tests can fail if Google rejects the search request on account of too many search requests being issued from Travis servers. In fact I worry that this might even affect other users on Travis :-p

Zhiyuan-Amos:
Fixes #733 

```
BrowserPanel loads the Google search page whenever a person card
is selected. The tests written for this feature verifies that the
loaded web page is indeed the Google search page with the person's
name.

When many forks of this project are running automated tests, too many
search requests are sent to Google, causing us to exceed the search
service’s rate limits. This is an abuse of Google’s search service.
Also, this causes our tests to fail as the loaded web page will be
redirected to a captcha page.

Let’s update BrowserPanel to load DummySearchPage.html which is hosted
on GitHub Pages. GitHub Pages serve static web pages which requires
much less processing power, thus it is more unlikely for us to exceed
the search service’s rate limits.
```

### DIFF
--- a/docs/DummySearchPage.html
+++ b/docs/DummySearchPage.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Dummy Search Page</title>
+  <script type="text/javascript">
+    function getJsonFromUrl() {
+      var query = location.search.substr(1);
+      var result = {};
+      query.split("&").forEach(function(part) {
+        var item = part.split("=");
+        result[item[0]] = decodeURIComponent(item[1]);
+      });
+      return result;
+    }
+
+    document.addEventListener("DOMContentLoaded", function(event) {
+      document.getElementById("name").innerHTML=getJsonFromUrl().name;
+    });
+  </script>
+</head>
+<body>
+  Hi <span id="name"></span>: This is a placeholder page for se-edu/addressbook-level4.<br>
+  You may update the code to load a page from a real service (e.g., Google search).<br>
+  This dummy page is used here because, given the high number of forks of this repo, loading a page from a real third-party service by default can result in that service taking counter-measures (e.g., redirecting to captcha pages) due to the high number of rapid requests received from a single IP.<br>
+  When you have made the change, please remove:
+  <ol>
+    <li>This file (<code>docs/DummySearchPage.html</code>).</li>
+    <li>Task <code>copyDummySearchPage</code> in both <code>build.gradle</code> and <code>.travis.yml</code>.</li>
+  </ol>
+</body>
+</html>
--- a/src/main/java/seedu/address/ui/BrowserPanel.java
+++ b/src/main/java/seedu/address/ui/BrowserPanel.java
@@ -21,8 +21,8 @@ import seedu.address.model.person.Person;
 public class BrowserPanel extends UiPart<Region> {
 
     public static final String DEFAULT_PAGE = "default.html";
-    public static final String GOOGLE_SEARCH_URL_PREFIX = "https://www.google.com.sg/search?safe=off&q=";
-    public static final String GOOGLE_SEARCH_URL_SUFFIX = "&cad=h";
+    public static final String SEARCH_PAGE_URL =
+            "https://se-edu.github.io/addressbook-level4/DummySearchPage.html?name=";
 
     private static final String FXML = "BrowserPanel.fxml";
 
@@ -42,8 +42,7 @@ public class BrowserPanel extends UiPart<Region> {
     }
 
     private void loadPersonPage(Person person) {
-        loadPage(GOOGLE_SEARCH_URL_PREFIX + person.getName().fullName.replaceAll(" ", "+")
-                + GOOGLE_SEARCH_URL_SUFFIX);
+        loadPage(SEARCH_PAGE_URL + person.getName().fullName);
     }
 
     public void loadPage(String url) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
 
 script: >-
     ./config/travis/run-checks.sh &&
-    travis_retry ./gradlew clean checkstyleMain checkstyleTest headless allTests coverage coveralls asciidoctor
+    travis_retry ./gradlew clean checkstyleMain checkstyleTest headless allTests coverage coveralls asciidoctor copyDummySearchPage
 
 deploy:
   skip_cleanup: true
--- a/build.gradle
+++ b/build.gradle
@@ -196,6 +196,11 @@ task deployOfflineDocs(type: Copy) {
     }
 }
 
+task copyDummySearchPage(type: Copy) {
+    from 'docs/DummySearchPage.html'
+    into "${buildDir}/docs/html5"
+}
+
 deployOfflineDocs.dependsOn asciidoctor
 processResources.dependsOn deployOfflineDocs
 
--- a/src/test/java/seedu/address/ui/BrowserPanelTest.java
+++ b/src/test/java/seedu/address/ui/BrowserPanelTest.java
@@ -5,8 +5,6 @@ import static org.junit.Assert.assertEquals;
 import static seedu.address.testutil.EventsUtil.postNow;
 import static seedu.address.testutil.TypicalPersons.ALICE;
 import static seedu.address.ui.BrowserPanel.DEFAULT_PAGE;
-import static seedu.address.ui.BrowserPanel.GOOGLE_SEARCH_URL_PREFIX;
-import static seedu.address.ui.BrowserPanel.GOOGLE_SEARCH_URL_SUFFIX;
 import static seedu.address.ui.UiPart.FXML_FILE_FOLDER;
 
 import java.net.URL;
@@ -42,8 +40,7 @@ public class BrowserPanelTest extends GuiUnitTest {
 
         // associated web page of a person
         postNow(selectionChangedEventStub);
-        URL expectedPersonUrl = new URL(GOOGLE_SEARCH_URL_PREFIX
-                + ALICE.getName().fullName.replaceAll(" ", "+") + GOOGLE_SEARCH_URL_SUFFIX);
+        URL expectedPersonUrl = new URL(BrowserPanel.SEARCH_PAGE_URL + ALICE.getName().fullName.replaceAll(" ", "%20"));
 
         waitUntilBrowserLoaded(browserPanelHandle);
         assertEquals(expectedPersonUrl, browserPanelHandle.getLoadedUrl());
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -5,8 +5,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static seedu.address.ui.BrowserPanel.DEFAULT_PAGE;
-import static seedu.address.ui.BrowserPanel.GOOGLE_SEARCH_URL_PREFIX;
-import static seedu.address.ui.BrowserPanel.GOOGLE_SEARCH_URL_SUFFIX;
 import static seedu.address.ui.StatusBarFooter.SYNC_STATUS_INITIAL;
 import static seedu.address.ui.StatusBarFooter.SYNC_STATUS_UPDATED;
 import static seedu.address.ui.UiPart.FXML_FILE_FOLDER;
@@ -41,6 +39,7 @@ import seedu.address.logic.commands.SelectCommand;
 import seedu.address.model.AddressBook;
 import seedu.address.model.Model;
 import seedu.address.testutil.TypicalPersons;
+import seedu.address.ui.BrowserPanel;
 import seedu.address.ui.CommandBox;
 
 /**
@@ -215,8 +214,7 @@ public abstract class AddressBookSystemTest {
         String selectedCardName = getPersonListPanel().getHandleToSelectedCard().getName();
         URL expectedUrl;
         try {
-            expectedUrl = new URL(GOOGLE_SEARCH_URL_PREFIX + selectedCardName.replaceAll(" ", "+")
-                    + GOOGLE_SEARCH_URL_SUFFIX);
+            expectedUrl = new URL(BrowserPanel.SEARCH_PAGE_URL + selectedCardName.replaceAll(" ", "%20"));
         } catch (MalformedURLException mue) {
             throw new AssertionError("URL expected to be valid.");
         }


damithc:
Isn't this a case of 'loading the wrong page' (because Google shows an error/verification page) rather than a case of 'loading failed'?

Zhiyuan-Amos:
In our code base, we have 2 kinds of checks:
1. In `BrowserPanelTest`, we verify the actual URL of the webpage.
2. In System Tests, we only verify whether the web page loads or not i.e. it's a more lenient test.

The previous time I noticed, the System Tests failed on Travis CI because the web page failed to load (since it doesn't actually verify the actual URL of the webpage).

eugenepeh:
Was looking at this issue and also the student's [issue](https://github.com/nus-cs2103-AY1718S1/forum/issues/219) before @Zhiyuan-Amos opened the PR. 

In AddressBookSystemTest, there's also this portion where we verify the url, https://github.com/se-edu/addressbook-level4/blob/1bee16fe4bcefbb02c8c87191e136aaadcbe53e8/src/test/java/systemtests/AddressBookSystemTest.java#L213-L225

I assume the student's travis test failed both the url verification for the `BrowserPanelTest` and `assertSelectedCardChanged` in `AddressBookSystemTest`, however, it is also possible to fail the verification test due to time out like what zhiyuan has discovered. I vaguely remembers that happens during my semester, just not a very common issue.

Zhiyuan-Amos:
Thanks @eugenepeh for pointing out that we indeed to verify that the card has changed. ^^ I vaguely recalled that it existed, but I didn't see it gosh x.x

> I assumed that the student's travis test failed both the url verification for the BrowserPanelTest and assertSelectedCardChanged in AddressBookSystemTest, however, it is also possible to fail the verification test due to time out like what zhiyuan has discovered. 

In one of my recent PRs, I noticed that the Travis build failed. I took a look at it and saw that the error was due to a `TimeoutException`, not an `Assertion Error`. Which was why I fixed it in this manner.

Zhiyuan-Amos:
I'm not quite sure what's a complete fool-proof solution for it though, because I'm not sure of how to reproduce this error, and I tried to search up for Google's error / verification page URL, but have been unable to find the URL.

eugenepeh:
@Zhiyuan-Amos hahaha same, i was actually spamming the tests for a few time before you opened up this PR. I think probably can make use of a python script to spam web request till it doesn't return the expected url.

Zhiyuan-Amos:
@eugenepeh XD HAHA I see! Yup, managed to do it on IntelliJ, got this link when I tried to search for Bob Pauline: https://ipv4.google.com/sorry/index?continue=https://www.google.com.sg/search%3Fsafe%3Doff%26q%3DBob%2BPauline%26cad%3Dh&q=EgTepHV2GMiG29IFIhkA8aeDS7NkeYotW6upLds3pZINR6-dGzLZMgFy

@damithc, if we notice that the start of the webpage is `https://ipv4.google.com/sorry`, will that be sufficient?

damithc:
> @damithc, if we notice that the start of the webpage is https://ipv4.google.com/sorry, will that be sufficient?

I guess. May be print out a warning as well?

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -16,6 +16,7 @@
     public static final String BROWSER_ID = "#browser";
 
     private boolean isWebViewLoaded = true;
+    private boolean isLoadingFailed = false;
```
eugenepeh: Perhaps we should use enum for this? To prevent any off-chance that both isWebViewLoaded and isLoadingFailed are true at the same time e.g. worker state runs instantly from READY -> SUCCEEDED / FAILED.

pyokagan:
Eh, Google showing the "sorry" page means it is flagging the IPs for botting, and may be classified as misuse under their terms of service. In turn, this may also be considered misuse under Travis and AppVeyor's ToS as well.

pyokagan:
If we are triggering a 3rd party service's rate limits, it means we are *abusing* their service, and the correct solution would be to lower the request rate.

damithc:
The exceeding of rate limit may be due to IP sharing. i.e., individually each project is within limit but when there are many forks doing the same thing, collectively they exceed the rate limit. Having said that, reducing the number of test cases that trigger the search page can reduce the probability of this occurring.

pyokagan:
> The exceeding of rate limit may be due to IP sharing. i.e., individually each project is within limit but when there are many forks doing the same thing, collectively they exceed the rate limit.

Yes, that is alright if there is a group of human people stuck behind a NAT using Google manually for different intents and purposes. However, in this case, we have 70 groups sending requests to google automatically at the same time *using code provided by us*. Something like our own botnet ;-)

Perhaps we may wish to use our own service which we have control over, instead of a third party service which have their own ToS and we have no control over. Like TEAMMATES, for example. The usage of Google isn't really a requirement since students are expected to modify their code to use another service anyway.

damithc:
That's true. 
1. How about we make the code access a page in their own repo (e.g., set up a fake search page in gh-pages branch). 
2. However, eventually their app will gravitate towards using a popular 3rd party service (google search and google maps are popular choices) and then run into this problem again?

pyokagan:
> However, eventually their app will gravitate towards using a popular 3rd party service (google search and google maps are popular choices) and then run into this problem again?

The numbers for each website would still be much lower than 70 due to projects wanting to differentiate from each other. e.g. I counted 11 for Google Maps, 6 for Google Search in https://github.com/nus-cs2103-AY1718S1/website/blob/master/admin/projectList.md

Plus, the main point is that we are not making automated requests to a third party service *by default*. Teams explicitly opt in to this via their choice of third party service, and so they explicitly opt into bearing the consequences of potential service abuse as well.

damithc:
> Plus, the main point is that we are not making automated requests to a third party service by default. Teams explicitly opt in to this via their choice of third party service, and so they explicitly opt into bearing the consequences of potential service abuse as well.

Alright, let's do it then. Some options:
1. Use a placeholder page in their own repo. In this case the URL needs to be updated after forking or else they'll all spam se-edu/AB4. Another thing to note is that github is frequently a target of DDoS attacks so they too might have sensitive botnet detectors.
2. Host the placeholder page in SoC web server. SoC web servers do occasionally go down but generally stable. 
3. I'd rather not use TEAMMATES because we are billed based on bandwidth used and there is no relation between the two projects.

damithc:
> 2. Host the placeholder page in SoC web server. SoC web servers do occasionally go down but generally stable.

Example [http://www.comp.nus.edu.sg/~seer/search.html](http://www.comp.nus.edu.sg/~seer/search.html)

pyokagan:
> In this case the URL needs to be updated after forking or else they'll all spam se-edu/AB4.

Might still actually be tolerable, as GitHub is just serving static web pages which requires much less processing power. We could try this out first and see if GitHub complains :-)

> Example http://www.comp.nus.edu.sg/~seer/search.html

Yup. We could even add Javascript to make it seem more of a search page. e.g. displaying different content depending on the query in `window.location.search`.

eugenepeh:
Hmm, if DDoS against a webserver is an issue, how about getting rid of BrowserPanel for good and bring in Chat Interface #757?

damithc:
> Hmm, if DDoS against a webserver/botnet detector is an issue, how about getting rid of BrowserPanel for good and bring in Chat Interface #757?

Chat interface is technically a replacement for the status message, not the Browser panel.

If we expect students to integrate the app with third party web apps, it is good to have an example of how to test such integrations.

eugenepeh:
> Chat interface is technically a replacement for the status message, not the Browser panel.
If we expect students to integrate the app with third party web apps, it is good to have an example of how to test such integrations.

Ahh I see, was thinking about Chat interface requires large ui area but forgot about status message box. :P

In such case, I think that `Example http://www.comp.nus.edu.sg/~seer/search.html` is a great idea as well. 

>Use a placeholder page in their own repo. In this case the URL needs to be updated after forking or else they'll all spam se-edu/AB4. Another thing to note is that github is frequently a target of DDoS attacks so they too might have sensitive botnet detectors.

GitHub pages may not be favorable as it will mean more setting up work to do which some students generally don't like due to "non-claimable individual work".

Zhiyuan-Amos:
@eugenepeh sorry I'm not too familiar with Javascript. >< So how should I proceed from here?

eugenepeh:
>@eugenepeh sorry I'm not too familiar with Javascript. >< So how should I proceed from here?

Hmm maybe we go with GET method first? 
E.g. `http://www.comp.nus.edu.sg/~seer/search.html?name=John%20Doe` then we verify that the browserpanel navigates to that page. The browser panel should be able to retain that url even if its not implemented/used yet.

@pyokagan what do you think?

damithc:
If you want to update the dummy search page, just pass me the code.

Zhiyuan-Amos:
Prof, here's the code!

```html
<!doctype html>
<html>
	<head>
		<meta charset="utf-8">
		<title>My test page</title>
		<script type="text/javascript">
		function getJsonFromUrl() {
  			var query = location.search.substr(1);
  			var result = {};
  			query.split("&").forEach(function(part) {
    			var item = part.split("=");
    			result[item[0]] = decodeURIComponent(item[1]);
  			});
  			return result;
		}

		document.addEventListener("DOMContentLoaded", function(event) {
    		document.getElementById("name").innerHTML=getJsonFromUrl().name;
  		});
	</script>
	</head>
	<body>Hi <span id="name"></span>: This is a placeholder page for se-edu/addressbook-level4</body>
</html>
```

pyokagan:
Oh, I'd recommend we try hosting it on GitHub pages first, since we already have an automatic deployment system in place.

Zhiyuan-Amos:
> Oh, I'd recommend we try hosting it on GitHub pages first, since we already have an automatic deployment system in place.

Sorry, how do I do that? >< I tried locally putting the `.html` file into the `docs` folder, but upon pushing it doesn't auto-generate in my Github Pages. But I thought it's weird for me to convert it to `AsciiDoc` also right? :O

Zhiyuan-Amos:
I merged this on my own master branch, as such you can access the GitHub Pages [here](https://zhiyuan-amos.github.io/addressbook-level4/DummySearchPage.html?name=Bob) to verify that it works as intended.

damithc:
Just to confirm, although all forks will deploy their own search page, their app will access upstream se-edu search page by default (because the URL will be hard-coded in the code)?

Zhiyuan-Amos:
> Just to confirm, although all forks will deploy their own search page, their app will access upstream se-edu search page by default (because the URL will be hard-coded in the code)?

Yes. One way to go about this issue is to mention in the Developer Guide since we have a segment informing them to [update documentation to match your fork](https://github.com/se-edu/addressbook-level4/blob/master/docs/DeveloperGuide.adoc#142-updating-documentation-to-match-your-fork).

What Paul suggested [earlier](https://github.com/se-edu/addressbook-level4/pull/793#issuecomment-357411711) was that we could try it out first and see if GitHub complains, since GitHub Pages serves static pages which require less processing power, which perhaps means that it's less likely to cause any issue.

eugenepeh:
> Just to confirm, although all forks will deploy their own search page, their app will access upstream se-edu search page by default (because the URL will be hard-coded in the code)?

Hmm it might be possible to have travis and gradle replace the hard coded string for us, with the help of 
[travis environment variables](https://docs.travis-ci.com/user/environment-variables/) and [gradle replace string](https://stackoverflow.com/questions/33464577/gradle-task-replace-string-in-java-file) if its not desirable to access upstream se-edu search page by default.

docs/DummySearchPage.html
```diff
@@ -0,0 +1,23 @@
...
+  			});
+  			return result;
+		}
+
+		document.addEventListener("DOMContentLoaded", function(event) {
+    		document.getElementById("name").innerHTML=getJsonFromUrl().name;
+  		});
+	</script>
+	</head>
+	<body>Hi <span id="name"></span>: This is a placeholder page for se-edu/addressbook-level4</body>
```
eugenepeh: Since this a dummy, should we add more information telling the students to change their default url and also clean up the gradle task if they plan to use other services?

Zhiyuan-Amos: Hmm is updating the `DeveloperGuide`'s instructions to update the url in `BrowserPanel` to their repo's url sufficient?

Zhiyuan-Amos:
> Hmm 2nd paragraph didn't mention the initial problem we faced

@eugenepeh sorry which initial problem? ><

> Hmm it might be possible to have travis and gradle replace the hard coded string for us, with the help of travis environment variables and gradle replace string if its not desirable to access upstream se-edu search page by default.
Edit: But then again... it will requires the trigger of travis and gradle build so yeh it might be better for them to edit themselves.

Ah I thought this was a good idea too, until you pointed out that tests will still access the .html file in `se-edu` repo when run locally. Only when it's run on Travis then it accesses the .html file in their own repo.

eugenepeh:
> @eugenepeh sorry which initial problem? ><

Uh... the tests failing? ~who cares about google~ :x

Zhiyuan-Amos:
> Other than the indentation of the docs/DummySearchPage.html, everything else seems okay

Totally slipped my mind; was too focused on just to get it to work. Thanks! :)

docs/DummySearchPage.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Dummy Search Page</title>
+  <script type="text/javascript">
+  	function getJsonFromUrl() {
```
yamgent: Need to convert the tab into space

Zhiyuan-Amos: ah missed this one out >< thanks!

Zhiyuan-Amos:
[3/3] Missed out on updating 2nd paragraph of the commit message.

src/main/java/seedu/address/ui/BrowserPanel.java
```diff
@@ -21,8 +21,7 @@
 public class BrowserPanel extends UiPart<Region> {
 
     public static final String DEFAULT_PAGE = "default.html";
-    public static final String GOOGLE_SEARCH_URL_PREFIX = "https://www.google.com.sg/search?safe=off&q=";
-    public static final String GOOGLE_SEARCH_URL_SUFFIX = "&cad=h";
+    public static final String DUMMY_URL = "https://se-edu.github.io/addressbook-level4/DummySearchPage.html?name=";
```
damithc: I don't think this name is appropriate. While it is pointing to a dummy search page at the moment, the intention of the constant is to point to a search service. `SEARCH_PAGE_URL`?

docs/DummySearchPage.html
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    document.addEventListener("DOMContentLoaded", function(event) {
+      document.getElementById("name").innerHTML=getJsonFromUrl().name;
+    });
+  </script>
+</head>
+<body>
+  Hi <span id="name"></span>: This is a placeholder page for se-edu/addressbook-level4.<br>
+  If you are using other services, please remove:
```
damithc: `If you are using other services` -> `When you change the app to load a different page`?

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -56,9 +58,20 @@ public boolean isUrlChanged() {
...
     /**
-     * Returns true if the browser is done loading a page, or if this browser has yet to load any page.
+     * Returns true if:
+     * 1. The browser is done loading a page
+     * 2. The browser has yet to load any page.
+     * 3. The browser tried to load a page and failed.
      */
-    public boolean isLoaded() {
-        return isWebViewLoaded;
+    public boolean isLoadingAttemptCompleted() {
```
damithc: This method name does not match the advertised behavior. May be `isNoLoadingInProgress`?

docs/DeveloperGuide.adoc
```diff
@@ -66,10 +66,12 @@ This project follows https://github.com/oss-generic/process/blob/master/docs/Cod
 
 Optionally, you can follow the <<UsingCheckstyle#, UsingCheckstyle.adoc>> document to configure Intellij to check style-compliance as you write code.
 
-==== Updating documentation to match your fork
+==== Updating documentation and code to match your fork
 
 After forking the repo, links in the documentation will still point to the `se-edu/addressbook-level4` repo. If you plan to develop this as a separate product (i.e. instead of contributing to the `se-edu/addressbook-level4`) , you should replace the URL in the variable `repoURL` in `DeveloperGuide.adoc` and `UserGuide.adoc` with the URL of your fork.
 
+Also, BrowserPanel#DUMMY_URL will still point to `se-edu` repo. You should replace this with the URL of your fork.
```
damithc: Do they need to change it right away? May be OK to keep pointing at se-edu until they replace it with some other thing?

damithc: But we can explain here why we use a dummy search page and that they are expected to replace it with a real service if they plan to use the Browser Panel in their App.

docs/DummySearchPage.html
```diff
@@ -0,0 +1,30 @@
...
+      document.getElementById("name").innerHTML=getJsonFromUrl().name;
+    });
+  </script>
+</head>
+<body>
+  Hi <span id="name"></span>: This is a placeholder page for se-edu/addressbook-level4.<br>
+  This dummy page is used as you are expected to replace this with a real service. When you change the app to load from a page from that service, please remove:
+  <ol>
+    <li>This file (docs/DummySearchPage.html).</li>
+    <li>Task "copyDummySearchPage" in both "build.gradle" and ".travis.yml".</li>
```
damithc: You can probably use the `<code>` html tag here.

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -56,9 +58,20 @@ public boolean isUrlChanged() {
     }
 
     /**
-     * Returns true if the browser is done loading a page, or if this browser has yet to load any page.
+     * Returns true if the browser is currently not loading i.e.:
+     * 1. The browser is done loading a page
+     * 2. The browser has yet to load any page.
+     * 3. The browser tried to load a page and failed.
```
damithc: Item 1 missing full stop.

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -56,9 +58,20 @@ public boolean isUrlChanged() {
...
      */
-    public boolean isLoaded() {
-        return isWebViewLoaded;
+    public boolean isNoLoadingInProgress() {
+        return browserState == BrowserState.LOADED || browserState == BrowserState.READY
+                || browserState == BrowserState.FAILED;
```
damithc: This seems equivalent to `browserState != BrowserState.LOADING`; not sure whether you can use it like that though.

src/test/java/guitests/guihandles/BrowserPanelHandle.java
```diff
@@ -56,9 +58,19 @@ public boolean isUrlChanged() {
     }
 
     /**
-     * Returns true if the browser is done loading a page, or if this browser has yet to load any page.
+     * Returns true if the browser is currently not loading i.e.:
+     * 1. The browser is done loading a page.
+     * 2. The browser has yet to load any page.
+     * 3. The browser tried to load a page and failed.
```
damithc: Now the comment seems unnecessarily descriptive given the simple implementation of the method?

Zhiyuan-Amos:
> Sorry, missed something in the previous review. Also, now that we use a dummy page, should the test pass when there is no Internet connection? Isn't it safer to fail in such a case? If the cause is lack of Internet, repo admin can decide to merge in spite of the failure.

If the connection failed, I'm not sure whether we can tell from the Travis log that the failure is due to no Internet connection. 

Also, even though there's no internet connection and the browser fails to load the page, however the browser's url will be updated to the dummy page's url. Since the test simply verifies that the url is correct (i.e. the url is the same as that of the dummy page. We don't verify the contents of the page), it should pass even though the internet connection fails?

damithc:
> If the connection failed, I'm not sure whether we can tell from the Travis log that the failure is due to no Internet connection.

Is it possible/probable that Travis server loses Internet connection? Did it happen before? If not, we should not create a mechanism that gives a possibly-false positive. If we really want to skip test when there is no Internet, we should have a way to confirm there is no Internet. That is, we should avoid a case where the test passes because it thinks there is no Internet but the page is not loading due to a different bug.

> Also, even though there's no internet connection and the browser fails to load the page, however the browser's url will be updated to the dummy page's url. Since the test simply verifies that the url is correct (i.e. the url is the same as that of the dummy page. We don't verify the contents of the page), it should pass even though the internet connection fails?

Is it possible to check the content of the page too? It's supposed to show the name of the person right? If not, how do we confirm that the the browser component is working? But this can be a separate PR.

What do you guys think?

Zhiyuan-Amos:
> Is it possible/probable that Travis server loses Internet connection? Did it happen before?

I recall seeing it only once happening in my PR. So it's really rare. We can remove it.

> Is it possible to check the content of the page too? It's supposed to show the name of the person right? If not, how do we confirm that the the browser component is working?

Yup we should strengthen the test by verifying that the page shows the name of the person.

PR #794, opened 2018-01-11
AddressBook: Inline setter methods #764

### LINKED ISSUE #764: AddressBook: Make setter methods private
`AddressBook`'s setter methods such as 
`setPerson(List<? extends ReadOnlyPerson>)` and `setTags(Set<Tag>)` have 
public access modifier.

It isn't necessary for these methods to be visible to other classes as:
* These methods are not called by other classes.
* The current implementation of these setter methods do not 
	synchronise the tags in `persons` with the tags in `tags`. 
	Therefore, calling `setPersons(...)` without calling `setTags(...)` may
	 result in some tags  in the new `Person` list not being found in 
	 `tags`, or some tags that do not exist in the new `Person` list are 
	 being found in `tags`.

Let's update the access modifier of these setter methods to be private.

Zhiyuan-Amos:
Fixes #764 

```
setPersons(...) and setTags(...) are only used internally.

It is unnecessary have the logic of setting persons and tags to be
extracted as separate setter methods, since their logic aren’t meant to
be used by outsiders, and does not help to introduce a new level of
abstraction.

Let’s inline these setter methods.
```

### DIFF
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -51,26 +51,18 @@ public class AddressBook implements ReadOnlyAddressBook {
 
     //// list overwrite operations
 
-    public void setPersons(List<Person> persons) throws DuplicatePersonException {
-        this.persons.setPersons(persons);
-    }
-
-    public void setTags(Set<Tag> tags) {
-        this.tags.setTags(tags);
-    }
-
     /**
      * Resets the existing data of this {@code AddressBook} with {@code newData}.
      */
     public void resetData(ReadOnlyAddressBook newData) {
         requireNonNull(newData);
-        setTags(new HashSet<>(newData.getTagList()));
+        tags.setTags(new HashSet<>(newData.getTagList()));
         List<Person> syncedPersonList = newData.getPersonList().stream()
                 .map(this::syncWithMasterTagList)
                 .collect(Collectors.toList());
 
         try {
-            setPersons(syncedPersonList);
+            persons.setPersons(newData.getPersonList());
         } catch (DuplicatePersonException e) {
             throw new AssertionError("AddressBooks should not have duplicate persons");
         }


pyokagan:
In conventional Java API design, if a field `foo` in an object is writable, then it should have `getFoo()` and `setFoo()` methods.

In this case, the `persons` list in the address book is clearly writable since there are publicly exposed `addPerson()`, `updatePerson()`, `removePerson()` methods, so there should be a `setPersons()` method.

pyokagan:
Besides, I think you can make `setPersons()` a useful abstraction by doing something like this?

```diff
diff --git a/src/main/java/seedu/address/model/AddressBook.java b/src/main/java/seedu/address/model/AddressBook.java
index f8d0260..8f5316f 100644
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -52,7 +52,10 @@ public class AddressBook implements ReadOnlyAddressBook {
     //// list overwrite operations
 
     public void setPersons(List<Person> persons) throws DuplicatePersonException {
-        this.persons.setPersons(persons);
+        List<Person> syncedPersonList = persons.stream()
+                .map(this::syncWithMasterTagList)
+                .collect(Collectors.toList());
+        this.persons.setPersons(syncedPersonList);
     }
 
     public void setTags(Set<Tag> tags) {
@@ -65,12 +68,8 @@ public class AddressBook implements ReadOnlyAddressBook {
     public void resetData(ReadOnlyAddressBook newData) {
         requireNonNull(newData);
         setTags(new HashSet<>(newData.getTagList()));
-        List<Person> syncedPersonList = newData.getPersonList().stream()
-                .map(this::syncWithMasterTagList)
-                .collect(Collectors.toList());
-
         try {
-            setPersons(syncedPersonList);
+            setPersons(newData.getPersonList());
         } catch (DuplicatePersonException e) {
             throw new AssertionError("AddressBooks should not have duplicate persons");
         }
```

Zhiyuan-Amos:
@pyokagan Hmm ok! 

1. I noticed that `AddressBook` has `addTag(Tag)`, though that method is only used by tests.. As such, we ignore the presence of `addTag(Tag)` and so there's no need for `setTags`?

2. I should improve the logic of `setPersons` such that the `tags` are updated as well, right?

pyokagan:
@Zhiyuan-Amos 
> I noticed that AddressBook has addTag(Tag), though that method is only used by tests.. As such, we ignore the presence of addTag(Tag) and so there's no need for setTags?

Depends on the design of the tags system. If the list of tags is meant to be writable and independent from the list of persons, then it should have a `setTags`. Otherwise, if the list of tags is dependent on the list of persons (i.e. the list of tags can be computed from the list of persons), then it is a readonly field and should not have a `setTags`.

In this case, the current implementation is somewhat in-between. The list of tags seems to be independent since we have `addTag()`, but it is also dependent on the list of persons[1]. So you could go either way, really, but then I would be lukewarm on that kind of PR because it doesn't really fix anything. A good an useful PR would resolve this discrepancy so it would fall into one of the two cases described above <sup>hint hint #753 ;-)</sup>.

> I should improve the logic of setPersons such that the tags are updated as well, right? Otherwise the internal persons and tags won't tally.

I think `syncWithMasterTagList` will already do that?

[1] If we wanted to implement this behavior, however, I would use two attributes. A writable `explicitTags` list (`getExplicitTags()`, `setExplicitTags()`) that tracks tags explicitly added with `addExplciitTag()`, and a read-only derived `tags` list (`getTags()`) that derives its contents from the union of the tags in the persons list and the `explicitTags` list.

Zhiyuan-Amos:
@pyokagan It seems that from https://github.com/se-edu/addressbook-level4/issues/753#issuecomment-349175934, the list of tags should be dependent on the list of persons. @damithc what do you think prof? :)

damithc:
> @damithc what do you think prof? :)

See https://github.com/se-edu/addressbook-level4/issues/753#issuecomment-360023137

I don't mind going the simpler way (i.e., derive tag list from persons; each person keeps its own tags), but it has obvious shortcomings. 
1. Duplication of tag objects
2. Not possible to have to define a set of tags and use them on persons. This will result in typos in tags going unnoticed (a typoed tag will be treated as a new tag).

We can document the shortcomings and suggest possible improvements in the dev docs.

Zhiyuan-Amos:
@damithc noted!

> derive tag list from persons

I'm wondering whether there's a need for the master tag list now, since the reason why it was there previously was so that all the persons' tags will point to the tags in master tag list. However, that's not the case now.

eugenepeh:
>Duplication of tag objects

That would make the issue mentioned in #809 worse :P

damithc:
> I'm wondering whether there's a need for the master tag list now, since the reason why it was there previously was so that all the persons' tags will point to the tags in master tag list. However, that's not the case now.

I guess no need.

Zhiyuan-Amos:
@Rinder5 do you think that you can work on this issue? I think this seems fairly doable. Few points to note:

1. Paul's [comment](https://github.com/se-edu/addressbook-level4/pull/794#issuecomment-359238361)
1. https://github.com/se-edu/addressbook-level4/issues/753#issuecomment-349175934
1. https://github.com/se-edu/addressbook-level4/issues/764

If you wanna work on this, I'll close my PR.

Rinder5:
Ok I'll work on it. Thanks!

PR #795, opened 2018-01-11
EditPersonDescriptor: update methods to perform defensive copy

Zhiyuan-Amos:
```
getTags() returns the same set of tags stored by this object, while
setTags(Set<Tag>) assigns the given Set<Tag> to its field without
defensive copy.

Callers can modify the state of the Set<Tag> externally, thus breaking
EditPersonDescriptor’s encapsulation.

Let's update getTags() to return an unmodifiable set, and 
setTags(Set<Tag>) to perform a defensive copy of the input before 
assigning it to its field.
```

### DIFF
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -8,6 +8,7 @@ import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 import static seedu.address.model.Model.PREDICATE_SHOW_ALL_PERSONS;
 
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -195,8 +196,13 @@ public class EditCommand extends UndoableCommand {
             this.tags = (tags != null) ? new HashSet<>(tags) : null;
         }
 
+        /**
+         * Returns an unmodifiable tag set, which throws {@code UnsupportedOperationException}
+         * if modification is attempted.
+         * Returns {@code Optional#empty()} if {@code tags} is null.
+         */
         public Optional<Set<Tag>> getTags() {
-            return Optional.ofNullable(tags);
+            return (tags != null) ? Optional.of(Collections.unmodifiableSet(tags)) : Optional.empty();
         }
 
         @Override
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -141,10 +141,10 @@ public class EditCommand extends UndoableCommand {
          * A defensive copy of {@code tags} is used internally.
          */
         public EditPersonDescriptor(EditPersonDescriptor toCopy) {
-            this.name = toCopy.name;
-            this.phone = toCopy.phone;
-            this.email = toCopy.email;
-            this.address = toCopy.address;
+            setName(toCopy.name);
+            setPhone(toCopy.phone);
+            setEmail(toCopy.email);
+            setAddress(toCopy.address);
             setTags(toCopy.tags);
         }
 
--- a/src/main/java/seedu/address/logic/commands/EditCommand.java
+++ b/src/main/java/seedu/address/logic/commands/EditCommand.java
@@ -8,6 +8,7 @@ import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 import static seedu.address.model.Model.PREDICATE_SHOW_ALL_PERSONS;
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -136,11 +137,11 @@ public boolean equals(Object other) {
...
         public EditPersonDescriptor(EditPersonDescriptor toCopy) {
-            this.name = toCopy.name;
-            this.phone = toCopy.phone;
-            this.email = toCopy.email;
-            this.address = toCopy.address;
-            this.tags = toCopy.tags;
+            setName(toCopy.name);
+            setPhone(toCopy.phone);
+            setEmail(toCopy.email);
+            setAddress(toCopy.address);
```
yamgent: Any rationale behind changing these to call `set...()`? Also seems to be better to be a separate commit on its own since this commit is about `setTags()`. :P

Zhiyuan-Amos: It's just for consistency, otherwise we'll have `setTags(toCopy.tags)`, but the rest will remain as `this.name = toCopy.name` ><

Which do you think is better?

yamgent: Using `set...()` for consistency is better, since it allows us to future proof them in case the implementation for them changes (e.g. if `Name` becomes no longer mutable, means that we will have to change here to do defensive copy, etc...).

Zhiyuan-Amos: ok got it! I'll do it in a separate commit. ^^

src/main/java/seedu/address/logic/commands/EditCommand.java
```diff
@@ -183,11 +185,11 @@ public void setAddress(Address address) {
         }
 
         public void setTags(Set<Tag> tags) {
-            this.tags = tags;
+            this.tags = (tags != null) ? new HashSet<>(tags) : null;
         }
 
         public Optional<Set<Tag>> getTags() {
-            return Optional.ofNullable(tags);
+            return (tags != null) ? Optional.of(Collections.unmodifiableSet(tags)) : Optional.empty();
```
Zhiyuan-Amos: @yamgent do you think this parentheses around `tags != null` is necessary? I recall getting corrected by Paul sometime earlier before :P The parentheses for `this.tags = (tags != null) ? new HashSet<>(tags) : null;` few lines above is necessary though right? Otherwise looks quite unreadable imo.

yamgent: > do you think this parentheses around tags != null is necessary?

I don't mind having parentheses around. Since adding parentheses doesn't impede readability, it's fine.

> The parentheses for this.tags = (tags != null) ? new HashSet<>(tags) : null; few lines above is necessary though right? Otherwise looks quite unreadable imo.

Yes.

damithc:
[1/3]
1. commit message: `setTags(Set<Tag>) do not perform ` --> `does not`?
2. commit message title: don't we usually use a lower case letter after the colon?
3. Should we mention the defensive copying in the header comment?

[2/3]
* points 2 and 3 above

[3/3]
* point 2 above

damithc:
The first sentence of the javadoc should be an overview of the method, rather than a mention of its peculiarities :-)

Zhiyuan-Amos:
> The first sentence of the javadoc should be an overview of the method, rather than a mention of its peculiarities :-)

Can't think of a way to phrase it well >< `Sets a copy of {@code tags}?`

damithc:
> Can't think of a way to phrase it well >< Sets a copy of {@code tags}?

Yes, something like that. In cases where the peculiarities needs more explanations, they can be given in the subsequent sentences.
```java
/**
 * Sets the ... of the ... object to given {@code ...}.  A defensive copy of ... is used internally.
 */

damithc:
Looks good. On a related note, when a class doesn't have any set methods, how safe is it to say 'the objects are immutable and therefore no need for defensive copying'? Isn't it possible that setters  are added in future, because when someone looks at the class there is no indication that the class is to remain immutable?

Zhiyuan-Amos:
> Looks good.

Does this mean that I can merge? XD

> Isn't it possible that setters are added in future, because when someone looks at the class there is no indication that the class is to remain immutable?

In our classes we leave header comments to denote that classes are immutable. For e.g. [Name](https://github.com/se-edu/addressbook-level4/blob/master/src/main/java/seedu/address/model/person/Name.java#L8). According to [this answer on SO](https://softwareengineering.stackexchange.com/a/272229), it seems that doing this is sufficient.

We can annotate `@Immutable` in our code base too (import stmt: `import jdk.nashorn.internal.ir.annotations.Immutable;`), but it doesn't do any compilation check to make sure that this class is indeed immutable. I think [Checker Framework](https://checkerframework.org/manual/checker-framework-manual.pdf) may be of some use, but have to take a deeper look into it.

damithc:
> Does this mean that I can merge? XD

Yes, if all relevant immutable classes have such header comments, that should be ok for now.

PR #798, opened 2018-01-19
Ui: Add colors to tag labels

yamgent:
Part of #784. Rebased, and changed the design of determining the colours of the tag.

Propsed commit message:
```
All tags in the address book are displayed with the same color.

It is hard to visually look for contacts sharing the same tags.

Let's add different colors for different tags to make the
differentiation of tags clearer.
```

_Note:_ For demo purpose, don't merge this to master!


Zhiyuan-Amos:
is this done alrd? :P

eugenepeh:
btw for these `don't merge this to master!` PRs, should we just create and assign the corresponding Label to them?

Zhiyuan-Amos:
> should we just create and assign the corresponding Label to them? would be easier to see :P

Yup, thanks for the reminder Eugene ^^

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -47,7 +49,25 @@ public PersonCard(Person person, int displayedIndex) {
...
+
+    /**
+     * Returns the color style for {@code tagName}'s label.
+     */
+    public static String getTagColorStyleFor(String tagName) {
+        return TAG_COLOR_STYLES[Math.abs(tagName.hashCode()) % TAG_COLOR_STYLES.length];
+    }
+
+    /**
+     * Creates the tag labels for {@code person} and attach them to this {@code PersonCard}.
```
Zhiyuan-Amos: Not sure whether it's necessary to have `and attach them to this {@code PersonCard}` since you have already mentioned the word `tag labels`, so it sounds reasonable that it's gonna be on this `PersonCard`

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -36,6 +38,9 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
         assertEquals(expectedPerson.getAddress().value, actualCard.getAddress());
         assertEquals(expectedPerson.getTags().stream().map(tag -> tag.tagName).collect(Collectors.toList()),
                 actualCard.getTags());
+        expectedPerson.getTags().stream().map(tag -> tag.tagName).forEach(tag ->
```
Zhiyuan-Amos: this `mapping` is unnecessary if we store the result of `.collect(Collectors.toList())` on line 39.

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -36,6 +38,9 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
...
+            assertEquals(Arrays.asList("label", PersonCard.getTagColorStyleFor(tag)),
```
Zhiyuan-Amos: We should extract "label" as a constant (it's just a default style class for all `Label` objects right?), otherwise it's unintuitive :P

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -68,4 +68,13 @@ public String getEmail() {
...
                 .collect(Collectors.toList());
     }
+
+    public List<String> getTagStyleClasses(String tag) {
+        return tagLabels
+                .stream()
+                .filter(label -> label.getText().equals(tag))
+                .map(Label::getStyleClass)
+                .findFirst()
+                .orElse(null);
```
Zhiyuan-Amos: we should throw exception instead to make the test fail faster :)

yamgent: What should I throw here? We don't have a `NoSuchTagException` class or something.

Zhiyuan-Amos: `IllegalArgumentException` should do. See [here](https://github.com/se-edu/addressbook-level4/blob/master/src/test/java/seedu/address/logic/parser/CommandParserTestUtil.java#L23).

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -34,8 +38,13 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
         assertEquals(expectedPerson.getPhone().value, actualCard.getPhone());
         assertEquals(expectedPerson.getEmail().value, actualCard.getEmail());
         assertEquals(expectedPerson.getAddress().value, actualCard.getAddress());
-        assertEquals(expectedPerson.getTags().stream().map(tag -> tag.tagName).collect(Collectors.toList()),
-                actualCard.getTags());
+
+        List<String> expectedTags = expectedPerson.getTags().stream().map(tag -> tag.tagName)
+                .collect(Collectors.toList());
```
Zhiyuan-Amos: perhaps we can put a newline before `.map(...)`

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -47,7 +49,25 @@ public PersonCard(Person person, int displayedIndex) {
...
         email.setText(person.getEmail().value);
-        person.getTags().forEach(tag -> tags.getChildren().add(new Label(tag.tagName)));
+        initTags(person);
+    }
+
+    /**
+     * Returns the color style for {@code tagName}'s label.
+     */
+    public static String getTagColorStyleFor(String tagName) {
+        return TAG_COLOR_STYLES[Math.abs(tagName.hashCode()) % TAG_COLOR_STYLES.length];
```
damithc: Give explanatory comments about this calculation?

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -34,8 +38,13 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
...
+        assertEquals(expectedTags, actualCard.getTags());
+        expectedTags.forEach(tag ->
+            assertEquals(Arrays.asList(LABEL_DEFAULT_STYLE, PersonCard.getTagColorStyleFor(tag)),
+                    actualCard.getTagStyleClasses(tag)));
```
damithc: A bit of head scratching needed to figure out what is being verified here. use explanatory variables or comments to make it clearer?

src/main/java/seedu/address/ui/PersonCard.java
```diff
@@ -47,7 +49,25 @@ public PersonCard(Person person, int displayedIndex) {
...
         address.setText(person.getAddress().value);
         email.setText(person.getEmail().value);
-        person.getTags().forEach(tag -> tags.getChildren().add(new Label(tag.tagName)));
+        initTags(person);
+    }
+
+    /**
+     * Returns the color style for {@code tagName}'s label.
+     */
+    public static String getTagColorStyleFor(String tagName) {
```
Zhiyuan-Amos: it seems like this method is `public static` because it's required by the tests? ><

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
Zhiyuan-Amos: Yup, or perhaps we can extract lines 44-47 into a method `assertTagsEqual`.

Also, we need to update `assertCardEquals(PersonCardHandle, PersonCardHandle)` right?
Also, I realised that for the expected result `Arrays.asList(LABEL_DEFAULT_STYLE, PersonCard.getTagColorStyleFor(tag))`, perhaps we shouldn't call `PersonCard#getTagColorStyleFor(Tag)` cos the actual tags are initialised using that method too. So it's like in some sense, "ownself test ownself".

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -34,8 +42,30 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
...
-        assertEquals(expectedPerson.getTags().stream().map(tag -> tag.tagName).collect(Collectors.toList()),
-                actualCard.getTags());
+
+        assertTagsEqual(expectedPerson, actualCard);
+    }
+
+    /**
+     * Returns the color style for {@code tagName}'s label.
+     */
+    private static String getTagColorStyleFor(String tagName) {
```
Zhiyuan-Amos: hmm I'm not too sure about this way of testing though. If we only have a few set values for `tagName`, then maybe we should hardcode:
```java
if (tagName.equals("foo")) {
    return "teal";
} else...
```

Not too sure about it too, I thought it looks kinda weird cos we are calling the exact same code as the main code base, maybe @damithc can comment on this?

damithc: @yamgent what do you think of hardcoding the expected values here? As mentioned, it is a bit odd to use the same code to test the a piece of code.

yamgent: Sure, I can hardcore the expected values. They can seem pretty arbitrary though, but I can add a comment to explain how the values were derived.

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -34,8 +42,51 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
...
+        switch (tagName) {
+        case "classmates":
+            return "teal";
+        case "colleagues":
+            return "yellow";
+        case "family":
+            return "orange";
+        case "friend":
+            return "orange";
```
Zhiyuan-Amos: can we do fallthrough for repeated colors? ^^

yamgent:
Purely just rebasing, technically should not have any change in logic.

src/test/java/seedu/address/ui/testutil/GuiTestAssert.java
```diff
@@ -34,8 +42,53 @@ public static void assertCardDisplaysPerson(Person expectedPerson, PersonCardHan
...
+
+        case "friends":
+            return "brown";
+
+        case "husband":
+            return "grey";
+
+        default:
+            fail(tagName + " does not have a color assigned.");
```
Zhiyuan-Amos: Oh wait, after #894, we don't use `fail` anymore.

PR #802, opened 2018-01-19
Checkstyle: remove unnecessary properties with default values #370

### LINKED ISSUE #370: Checkstyle: remove unnecessary properties with default values
`checkstyle.xml` contains properties with default values.

E.g. "severity" properties with "error" values, where "error" values are the default:
`<property name="severity" value="error"/>`.

These are unnecessary.

vivekscl:
Fixes #370 

On a side note, in `checkstyle.xml` there is a property called `optional` under the module`SuppresionFilter`which has a default value. 

```
<module name="SuppressionFilter">
    <property name="file" value="config/checkstyle/suppressions.xml"/>
    <property name="optional" value="false"/>
</module>
```

However, according to the [Checkstyle documentation](http://checkstyle.sourceforge.net/config_filters.html#SuppressionFilter), this property is included even though it is a default value. Do we remove it or keep it as it is?

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -27,7 +27,6 @@
 
   <module name="SuppressionFilter">
     <property name="file" value="config/checkstyle/suppressions.xml"/>
-    <property name="optional" value="false"/>
   </module>
 
 
@@ -61,7 +60,6 @@
         <property name="specialImportsRegExp" value="^org\."/>
         <property name="thirdPartyPackageRegExp" value="^com\."/>
         <property name="sortImportsInGroupAlphabetically" value="true"/>
-        <property name="separateLineBetweenGroups" value="true"/>
     </module>
 
     <!-- Checks for redundant import statements.
@@ -78,10 +76,7 @@
     -->
     <module name="UnusedImports"/>
 
-    <module name="AvoidStarImport">
-      <property name="allowClassImports" value="false"/>
-      <property name="allowStaticMemberImports" value="false"/>
-    </module>
+    <module name="AvoidStarImport"/>
 
     <!--
     NAMING CHECKS
@@ -108,9 +103,6 @@
     <module name="ConstantName">
       <!-- Validates non-private, static, final fields against the expression "^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$". -->
       <metadata name="altname" value="ConstantName"/>
-      <property name="applyToPublic" value="true"/>
-      <property name="applyToProtected" value="true"/>
-      <property name="applyToPackage" value="true"/>
       <property name="applyToPrivate" value="false"/>
       <message key="name.invalidPattern"
                value="Variable ''{0}'' should be in ALL_CAPS (if it is a constant) or be private (otherwise)."/>
@@ -120,10 +112,6 @@
     <module name="StaticVariableName">
       <!-- Validates static, non-final fields against the supplied expression. -->
       <metadata name="altname" value="StaticVariableName"/>
-      <property name="applyToPublic" value="true"/>
-      <property name="applyToProtected" value="true"/>
-      <property name="applyToPackage" value="true"/>
-      <property name="applyToPrivate" value="true"/>
       <property name="format" value="^[a-z][a-zA-Z0-9]*_?$"/>
       <property name="severity" value="warning"/>
     </module>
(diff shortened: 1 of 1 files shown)


yamgent:
> However, according to the Checkstyle documentation, this property is included even though it is a default value. Do we remove it or keep it as it is?

Can't access the link atm (SourceForge down for me :P).

If it is a default value, I think it is OK to remove it, so long as it doesn't break anything else.

vivekscl:
Oh, it looks like this:
deepinscreenshot_select-area_20180119194220

yamgent:
> Oh, it looks like this:

Ah ok. I have no objections in removing it.

vivekscl:
Oh, and one more doubt, for the module `JavadocMethod` the `scope` property's default value is private but I believe it is there for readability so as to differentiate it from the previous `JavadocMethod` that targets public methods.

deepinscreenshot_select-area_20180119195503

So do leave this as is?

Zhiyuan-Amos:
> So do I leave the second JavadocMethod as is?

Yup, it seems like it's there for clarity purposes.

Zhiyuan-Amos:
> On a side note, in checkstyle.xml there is a property called optional under the moduleSuppresionFilterwhich has a default value.

This PR looks good, you can submit a 2nd commit with these additional changes

eugenepeh:
>On a side note, in checkstyle.xml there is a property called optional under the moduleSuppresionFilterwhich has a default value.
However, according to the Checkstyle documentation, this property is included even though it is a default value. Do we remove it or keep it as it is?

Yes, it should be removed.

Also, there are some other checks that contain default properties, such as
>    <module name="AvoidStarImport">
      <property name="allowClassImports" value="false"/>
      <property name="allowStaticMemberImports" value="false"/>
    </module>

config/checkstyle/checkstyle.xml
```diff
@@ -231,7 +218,6 @@
 
     <module name="UpperEll">
       <!-- Checks that long constants are defined with an upper ell.-->
-      <property name="severity" value="error"/>
```
eugenepeh: For checks such as this, having the only property removed, should just shorten it to 
`<module name="UpperEll" />`
The comments can then be placed on top of the module.

config/checkstyle/checkstyle.xml
```diff
@@ -79,8 +77,6 @@
     <module name="UnusedImports"/>
 
     <module name="AvoidStarImport">
-      <property name="allowClassImports" value="false"/>
-      <property name="allowStaticMemberImports" value="false"/>
     </module>
```
Zhiyuan-Amos: @eugenepeh if this module does not have any property, can we remove this module?

eugenepeh: @Zhiyuan-Amos o_o removing the module = removing this check, so no :P

eugenepeh: @vivekscl The end tag can be removed since this module no longer has any properties.

Zhiyuan-Amos:
@eugenepeh if you're ok, can approve and tag prof? ^^

eugenepeh:
@damithc For your review

damithc:
Are we sure we want to remove these? Don't they help to document the existence of the property and the current value, although it may be the default value? i.e., we can get a more complete picture by reading the file in concern when all the properties are there, rather than rely on checkstyle documentation to find the missing properties. What do you guys think?

eugenepeh:
>
Are we sure we want to remove these? Don't they help to document the existence of the property and the current value, although it may be the default value? i.e., we can get a more complete picture by reading the file in concern when all the properties are there, rather than rely on checkstyle documentation to find the missing properties. What do you guys think?

@damithc , most of these default properties are in consistent with the common coding standard such as java/google, and they will show up as part of the violation message if any of them was triggered during the checks. It is also very unlikely that we will be modifying the default properties, that we're removing, in the near future, so I think its safe for us to remove them. Hence the only default property that I would like to retain is the format property in the `MemberName`.

>
    <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>

vivekscl:
In addition to what @eugenepeh said, I think the default values for certain modules can be inferred by comparing them with other modules of the same nature. For instance, the `ConstantName` module has only the `applyToPrivate`  property set to false while the `MemberName` has no default properties. This way you can infer that `MemberName` has its properties all set to true by default.

vivekscl:
@damithc, may I know the decision of this PR? I believe that the default properties should be removed except for the `format` property as @eugenepeh said but I'm not sure what the final decision is :sweat_smile:.

Issue #749, opened 2017-12-04
Storage layer needs proper error handling


pyokagan:
https://github.com/se-edu/addressbook-level4/blob/edc42f0e2579c27c209b1f8f3c0f87ed04e83904/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java#L53

https://github.com/se-edu/addressbook-level4/blob/edc42f0e2579c27c209b1f8f3c0f87ed04e83904/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java#L67

pyokagan:
There's no reason why `XmlSerializableAddressBook` *must* implement `ReadOnlyAddressBook`.

PR #806, opened 2018-01-20
XmlAdaptedPerson#toModelType(): Enforce checking of required parameters #725

### LINKED ISSUE #725: XmlAdaptedPerson#toModelType(): Enforce checking of required parameters
See https://github.com/nus-cs2103-AY1718S1/forum/issues/137#issue-268097195

yamidark:
Fixes #725

```
XmlAdaptedPerson#toModelType() does not have a non-null check for each
parameter in person before the isValid*(String).

When null value is passed into the parameter's isValid*(String) check,
a NullPointerException is thrown, causing the app to crash.

Let's update XmlAdaptedPerson#toModelType() to have a separate non-null
check for each parameter before passing them into the isValid*(String)
check.
```

### DIFF
--- a/src/main/java/seedu/address/storage/XmlAdaptedPerson.java
+++ b/src/main/java/seedu/address/storage/XmlAdaptedPerson.java
@@ -21,6 +21,8 @@ import seedu.address.model.tag.Tag;
  */
 public class XmlAdaptedPerson {
 
+    public static final String MISSING_FIELD_MESSAGE_FORMAT = "Person's %s field is missing!";
+
     @XmlElement(required = true)
     private String name;
     @XmlElement(required = true)
@@ -79,21 +81,33 @@ public class XmlAdaptedPerson {
             personTags.add(tag.toModelType());
         }
 
+        if (this.name == null) {
+            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Name.class.getSimpleName()));
+        }
         if (!Name.isValidName(this.name)) {
             throw new IllegalValueException(Name.MESSAGE_NAME_CONSTRAINTS);
         }
         final Name name = new Name(this.name);
 
+        if (this.phone == null) {
+            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Phone.class.getSimpleName()));
+        }
         if (!Phone.isValidPhone(this.phone)) {
             throw new IllegalValueException(Phone.MESSAGE_PHONE_CONSTRAINTS);
         }
         final Phone phone = new Phone(this.phone);
 
+        if (this.email == null) {
+            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Email.class.getSimpleName()));
+        }
         if (!Email.isValidEmail(this.email)) {
             throw new IllegalValueException(Email.MESSAGE_EMAIL_CONSTRAINTS);
         }
         final Email email = new Email(this.email);
 
+        if (this.address == null) {
+            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Address.class.getSimpleName()));
+        }
         if (!Address.isValidAddress(this.address)) {
             throw new IllegalValueException(Address.MESSAGE_ADDRESS_CONSTRAINTS);
         }
--- a/src/main/java/seedu/address/storage/XmlAdaptedPerson.java
+++ b/src/main/java/seedu/address/storage/XmlAdaptedPerson.java
@@ -3,6 +3,7 @@ package seedu.address.storage;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.xml.bind.annotation.XmlElement;
@@ -113,10 +114,10 @@ public class XmlAdaptedPerson {
         }
 
         XmlAdaptedPerson otherPerson = (XmlAdaptedPerson) other;
-        return name.equals(otherPerson.name)
-                && phone.equals(otherPerson.phone)
-                && email.equals(otherPerson.email)
-                && address.equals(otherPerson.address)
+        return Objects.equals(name, otherPerson.name)
+                && Objects.equals(phone, otherPerson.phone)
+                && Objects.equals(email, otherPerson.email)
+                && Objects.equals(address, otherPerson.address)
                 && tagged.equals(otherPerson.tagged);
     }
 }
--- a/src/test/java/seedu/address/commons/util/XmlUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/XmlUtilTest.java
@@ -4,14 +4,19 @@ import static org.junit.Assert.assertEquals;
 
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.util.Collections;
+import java.util.List;
 
 import javax.xml.bind.JAXBException;
+import javax.xml.bind.annotation.XmlRootElement;
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import seedu.address.model.AddressBook;
+import seedu.address.storage.XmlAdaptedPerson;
+import seedu.address.storage.XmlAdaptedTag;
 import seedu.address.storage.XmlSerializableAddressBook;
 import seedu.address.testutil.AddressBookBuilder;
 import seedu.address.testutil.PersonBuilder;
@@ -23,7 +28,18 @@ public class XmlUtilTest {
     private static final File EMPTY_FILE = new File(TEST_DATA_FOLDER + "empty.xml");
     private static final File MISSING_FILE = new File(TEST_DATA_FOLDER + "missing.xml");
     private static final File VALID_FILE = new File(TEST_DATA_FOLDER + "validAddressBook.xml");
+    private static final File MISSING_PERSON_FIELD_FILE = new File(TEST_DATA_FOLDER + "missingPersonField.xml");
+    private static final File INVALID_PERSON_FIELD_FILE = new File(TEST_DATA_FOLDER + "invalidPersonField.xml");
+    private static final File VALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "validPerson.xml");
     private static final File TEMP_FILE = new File(TestUtil.getFilePathInSandboxFolder("tempAddressBook.xml"));
+
+    private static final String INVALID_PHONE = "9482asf424";
+
+    private static final String VALID_NAME = "Hans Muster";
+    private static final String VALID_PHONE = "9482424";
+    private static final String VALID_EMAIL = "hans@example";
+    private static final String VALID_ADDRESS = "4th street";
+    private static final List<XmlAdaptedTag> VALID_TAGS = Collections.singletonList(new XmlAdaptedTag("friends"));
 
     @Rule
     public ExpectedException thrown = ExpectedException.none();
@@ -60,6 +76,33 @@ public class XmlUtilTest {
     }
 
     @Test
+    public void xmlAdaptedPersonFromFile_fileWithMissingPersonField_validResult() throws Exception {
+        XmlAdaptedPerson actualPerson = XmlUtil.getDataFromFile(
+                MISSING_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson expectedPerson = new XmlAdaptedPerson(
+                null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(expectedPerson, actualPerson);
+    }
+
+    @Test
+    public void xmlAdaptedPersonFromFile_fileWithInvalidPersonField_validResult() throws Exception {
+        XmlAdaptedPerson actualPerson = XmlUtil.getDataFromFile(
+                INVALID_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson expectedPerson = new XmlAdaptedPerson(
+                VALID_NAME, INVALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(expectedPerson, actualPerson);
+    }
+
+    @Test
+    public void xmlAdaptedPersonFromFile_fileWithValidPerson_validResult() throws Exception {
+        XmlAdaptedPerson actualPerson = XmlUtil.getDataFromFile(
+                VALID_PERSON_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson expectedPerson = new XmlAdaptedPerson(
+                VALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(expectedPerson, actualPerson);
+    }
+
+    @Test
     public void saveDataToFile_nullFile_throwsNullPointerException() throws Exception {
         thrown.expect(NullPointerException.class);
         XmlUtil.saveDataToFile(null, new AddressBook());
@@ -93,4 +136,11 @@ public class XmlUtilTest {
         dataFromFile = XmlUtil.getDataFromFile(TEMP_FILE, XmlSerializableAddressBook.class);
         assertEquals(dataToWrite, dataFromFile);
     }
+
+    /**
+     * Test class annotated with {@code XmlRootElement} to allow unmarshalling of .xml data to {@code XmlAdaptedPerson}
+     * objects.
+     */
+    @XmlRootElement(name = "person")
+    private static class XmlAdaptedPersonWithRootElement extends XmlAdaptedPerson {}
 }
--- a/src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
+++ b/src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
@@ -1,6 +1,7 @@
 package seedu.address.storage;
 
 import static org.junit.Assert.assertEquals;
+import static seedu.address.storage.XmlAdaptedPerson.MISSING_FIELD_MESSAGE_FORMAT;
 import static seedu.address.testutil.TypicalPersons.BENSON;
 
 import java.util.ArrayList;
@@ -10,6 +11,10 @@ import java.util.stream.Collectors;
 import org.junit.Test;
 
 import seedu.address.commons.exceptions.IllegalValueException;
+import seedu.address.model.person.Address;
+import seedu.address.model.person.Email;
+import seedu.address.model.person.Name;
+import seedu.address.model.person.Phone;
 import seedu.address.testutil.Assert;
 
 public class XmlAdaptedPersonTest {
@@ -37,28 +42,60 @@ public class XmlAdaptedPersonTest {
     public void toModelType_invalidName_throwsIllegalValueException() {
         XmlAdaptedPerson person =
                 new XmlAdaptedPerson(INVALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
-        Assert.assertThrows(IllegalValueException.class, person::toModelType);
+        String expectedMessage = Name.MESSAGE_NAME_CONSTRAINTS;
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
+    }
+
+    @Test
+    public void toModelType_nullName_throwsIllegalValueException() {
+        XmlAdaptedPerson person = new XmlAdaptedPerson(null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        String expectedMessage = String.format(MISSING_FIELD_MESSAGE_FORMAT, Name.class.getSimpleName());
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
     }
 
     @Test
     public void toModelType_invalidPhone_throwsIllegalValueException() {
         XmlAdaptedPerson person =
                 new XmlAdaptedPerson(VALID_NAME, INVALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
-        Assert.assertThrows(IllegalValueException.class, person::toModelType);
+        String expectedMessage = Phone.MESSAGE_PHONE_CONSTRAINTS;
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
+    }
+
+    @Test
+    public void toModelType_nullPhone_throwsIllegalValueException() {
+        XmlAdaptedPerson person = new XmlAdaptedPerson(VALID_NAME, null, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        String expectedMessage = String.format(MISSING_FIELD_MESSAGE_FORMAT, Phone.class.getSimpleName());
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
     }
 
     @Test
     public void toModelType_invalidEmail_throwsIllegalValueException() {
         XmlAdaptedPerson person =
                 new XmlAdaptedPerson(VALID_NAME, VALID_PHONE, INVALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
-        Assert.assertThrows(IllegalValueException.class, person::toModelType);
+        String expectedMessage = Email.MESSAGE_EMAIL_CONSTRAINTS;
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
+    }
+
+    @Test
+    public void toModelType_nullEmail_throwsIllegalValueException() {
+        XmlAdaptedPerson person = new XmlAdaptedPerson(VALID_NAME, VALID_PHONE, null, VALID_ADDRESS, VALID_TAGS);
+        String expectedMessage = String.format(MISSING_FIELD_MESSAGE_FORMAT, Email.class.getSimpleName());
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
     }
 
     @Test
     public void toModelType_invalidAddress_throwsIllegalValueException() {
         XmlAdaptedPerson person =
                 new XmlAdaptedPerson(VALID_NAME, VALID_PHONE, VALID_EMAIL, INVALID_ADDRESS, VALID_TAGS);
-        Assert.assertThrows(IllegalValueException.class, person::toModelType);
+        String expectedMessage = Address.MESSAGE_ADDRESS_CONSTRAINTS;
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
+    }
+
+    @Test
+    public void toModelType_nullAddress_throwsIllegalValueException() {
+        XmlAdaptedPerson person = new XmlAdaptedPerson(VALID_NAME, VALID_PHONE, VALID_EMAIL, null, VALID_TAGS);
+        String expectedMessage = String.format(MISSING_FIELD_MESSAGE_FORMAT, Address.class.getSimpleName());
+        Assert.assertThrows(IllegalValueException.class, expectedMessage, person::toModelType);
     }
 
     @Test
--- a/src/test/java/seedu/address/testutil/Assert.java
+++ b/src/test/java/seedu/address/testutil/Assert.java
@@ -11,18 +11,37 @@ public class Assert {
      * Asserts that the {@code callable} throws the {@code expected} Exception.
      */
     public static void assertThrows(Class<? extends Throwable> expected, VoidCallable callable) {
+        assertThrows(expected, null, callable);
+    }
+
+    /**
+     * Asserts that the {@code callable} throws the {@code expectedException} and the {@code expectedMessage}.
+     * If there's no need for the verification of the exception's error message, call
+     * {@code assertThrows(Class<? extends Throwable>, VoidCallable)} instead.
+     * {@see assertThrows(Class<? extends Throwable>, VoidCallable}
+     */
+    public static void assertThrows(Class<? extends Throwable> expectedException, String expectedMessage,
+                                    VoidCallable callable) {
         try {
             callable.call();
         } catch (Throwable actualException) {
-            if (actualException.getClass().isAssignableFrom(expected)) {
+            String errorMessage;
+
+            if (!actualException.getClass().isAssignableFrom(expectedException)) {
+                errorMessage = String.format("Expected exception thrown: %s, actual: %s",
+                        expectedException.getName(), actualException.getClass().getName());
+            } else if (expectedMessage != null && !expectedMessage.equals(actualException.getMessage())) {
+                errorMessage = String.format(
+                        "Expected message thrown: %s, actual: %s", expectedMessage, actualException.getMessage());
+            } else {
                 return;
             }
-            String message = String.format("Expected thrown: %s, actual: %s", expected.getName(),
-                    actualException.getClass().getName());
-            throw new AssertionFailedError(message);
+
+            throw new AssertionFailedError(errorMessage);
         }
-        throw new AssertionFailedError(
-                String.format("Expected %s to be thrown, but nothing was thrown.", expected.getName()));
+
+        throw new AssertionFailedError(String.format(
+                "Expected %s to be thrown, but nothing was thrown.", expectedException.getName()));
     }
 
     /**
--- a/src/test/data/XmlUtilTest/invalidPersonField.xml
+++ b/src/test/data/XmlUtilTest/invalidPersonField.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Person with an invalid phone field -->
+<person>
+    <name>Hans Muster</name>
+    <phone>9482asf424</phone>
+    <email>hans@example</email>
+    <address>4th street</address>
+    <tagged>friends</tagged>
+</person>
--- a/src/test/data/XmlUtilTest/missingPersonField.xml
+++ b/src/test/data/XmlUtilTest/missingPersonField.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Person with missing name field -->
+<person>
+    <phone>9482424</phone>
+    <email>hans@example</email>
+    <address>4th street</address>
+    <tagged>friends</tagged>
+</person>
--- a/src/test/data/XmlUtilTest/validPerson.xml
+++ b/src/test/data/XmlUtilTest/validPerson.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<person>
+    <name>Hans Muster</name>
+    <phone>9482424</phone>
+    <email>hans@example</email>
+    <address>4th street</address>
+    <tagged>friends</tagged>
+</person>


src/test/data/XmlUtilTest/nullValuesAddressBook.xml
```diff
@@ -0,0 +1,58 @@
...
+        <email isPrivate="false">ruth@example.com</email>
+        <address isPrivate="false">81th street</address>
+    </persons>
+    <persons>
+        <name>Lydia Kunz</name>
+        <phone isPrivate="false">9482427</phone>
+        <email isPrivate="false">lydia@example.com</email>
+        <address isPrivate="false">little tokyo</address>
+    </persons>
+    <!-- Missing Phone field -->
```
Zhiyuan-Amos: why do you have 2 test cases for each `Missing X field`?

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +61,15 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void getDataFromFile_nullValuesFile_validResult() throws Exception {
```
Zhiyuan-Amos: why did you test `XmlUtilTest` instead of creating a `XmlSerializableAddressBookTest`? ^^

Also do take note, for the test file, you should have both `null` values and `illegal` values since both of these options will cause `IllegalValueException` to be thrown.

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
```diff
@@ -78,22 +78,22 @@ public Person toModelType() throws IllegalValueException {
             personTags.add(tag.toModelType());
         }
 
-        if (!Name.isValidName(this.name)) {
+        if (this.name == null || !Name.isValidName(this.name)) {
```
Zhiyuan-Amos: @pyokagan I'm wondering if [Name#isValidName(String)](https://github.com/se-edu/addressbook-level4/blob/master/src/main/java/seedu/address/model/person/Name.java#L37) should have considered the null value case as well?

yamgent: I would expect a `null` to be invalid.

In this case it so happens that we reject empty Names, but even for fields that can be blank, I would expect it to be `""` rather than `null`.

src/test/data/XmlUtilTest/nullValuesAddressBook.xml
yamidark: > why do you have 2 test cases for each Missing X field?

Whoops, forgot to add a space between the valid person and the person with `Missing X Field` (there is only 1 test case for each `Missing X Field`) 💦. 

I wanted to test if we can filter out the person with `Missing X Field` in a mixture of both, so I end up putting each `Missing X Field` person in between the valid persons, but seems like its quite confusing. Will separate the valid persons from all the `Missing X Field` persons instead.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: >why did you test XmlUtilTest instead of creating a XmlSerializableAddressBookTest? ^^

Saw that the `getDataFromFile` tests were in `XmlUtilTest` so thought I should put these test there too 😅.

src/test/data/XmlUtilTest/nullValuesAddressBook.xml
Zhiyuan-Amos: You just need 1 valid person right?

src/test/java/seedu/address/commons/util/XmlUtilTest.java
Zhiyuan-Amos: Ok I get it. Yes, you should have this test, but you should also have a test for `XmlSerializableAddressBookTest#getPersonList()` :)

src/test/data/XmlUtilTest/nullValuesAddressBook.xml
yamidark: >You just need 1 valid person right?

Yep, you're right.

src/main/java/seedu/address/model/person/Address.java
```diff
@@ -32,10 +32,10 @@ public Address(String address) {
     }
 
     /**
-     * Returns true if a given string is a valid person email.
+     * Returns true if a given string is not null and is a valid person email.
```
Zhiyuan-Amos: hehe this is for method `isValidAddress`. Not your fault, but we can fix it. Just add in the commit message something like:
```
Also, let's make a minor correction to the header comments of `Address#isValidAddress(String)`
```

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -43,6 +45,10 @@ public XmlSerializableAddressBook(ReadOnlyAddressBook src) {
         tags.addAll(src.getTagList().stream().map(XmlAdaptedTag::new).collect(Collectors.toList()));
     }
 
+    /**
+     * Returns an unmodifiable view of the {@code XmlAdaptedPerson} converted to {@code Person} list.
+     * @return: an unmodifiable view of the person list with no null values or duplicates
```
Zhiyuan-Amos: if you have `Returns` already, then you shouldn't have `@return` too :P Just simply type something like `Guaranteed to have no null values or duplicates in the person list.`

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
```diff
@@ -0,0 +1,26 @@
...
+import seedu.address.commons.util.XmlUtil;
+
+public class XmlSerializableAddressBookTest {
+
+    private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/XmlUtilTest/");
+    private static final File VALID_FILE = new File(TEST_DATA_FOLDER + "validAddressBook.xml");
+    private static final File ILLEGAL_VALUES_FILE = new File(TEST_DATA_FOLDER + "illegalValuesAddressBook.xml");
+
+    @Test
+    public void getPersonList_noNulls() throws Exception {
```
Zhiyuan-Amos: You should have a more robust test by testing for the size of the list as well.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -53,7 +59,8 @@ public XmlSerializableAddressBook(ReadOnlyAddressBook src) {
                 //TODO: better error handling
                 return null;
             }
-        }).collect(Collectors.toCollection(FXCollections::observableArrayList));
+        }).filter(Objects::nonNull).collect(Collectors.toCollection(FXCollections::observableArrayList));
+        assert CollectionUtil.elementsAreUnique(persons);
```
Zhiyuan-Amos: sorry, let's not add this line here, since this PR only concerns itself with removing null values, not ensuring that elements are unique.

Zhiyuan-Amos:
@damithc we have 2 approaches to this PR:
1. Simply remove the entire person if any of the fields are corrupted.
1. Replace the corrupted field with a preset value such as `DATA_MISSING` or something like that.

I'm thinking of using approach 1 since it's simpler. Also, to keep this PR small, I thought that we can include the feature of notifying the user through the `ResultDisplay` box or perhaps through a pop-up if there's corrupted data, in another PR instead.

What are your thoughts prof?

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
pyokagan: @yamgent 
> I would expect a null to be invalid.

Does that mean that `Name.isValidName` should accept null or not?

Anyway, I'll think "no", since the Name object at the model layer does not deal with nulls at all. If there is a null, then it means that there is a programming error somewhere, and `Name.isValidName` should not be silently swallowing that error.

Note that `null` in the storage layer means that the `<name />` tag was not provided in the XML file. In other words, it is a *parsing error* and is a concern that only the storage layer should be dealing with, not the model layer.

pyokagan: The situation may be less clear in this case because @yamidark decided to use the same error code path as `Name.isValidName()`. However, we may wish to have separate error paths -- one to tell the user that the `<name>` tag does not exist, and the `<name>` tag exists but does not have valid data. So, there are two distinct error cases, that should not be folded into one error checking method that only returns `true` or `false`.

eugenepeh:
> Simply remove the entire person if any of the fields are corrupted.

This is not really a good practice though. We can compromise performance but not accuracy. This is as good as telling users, "our app will keep your data 99% of the time, on the 1% chance we deletes everything" which is not cool. We should retain whatever data that is not corrupted, at least in this case we can maintain the stance that "this is not our fault, your file got corrupted somehow" :P

Zhiyuan-Amos:
> This is not really a good practice though. We can compromise performance but not accuracy. This is as good as telling users, "our app will keep your data 99% of the time, on the 1% chance we deletes everything" which is not cool.

@eugenepeh yup I fully agree with you :) But my worry is that this may cause the PR to be pretty big. I'm thinking of this particular corner case, where you have 2 `Person`s with exactly the same details except for the `Name`. 

Then, suppose that both of the `<name>` tags are missing. That way, we can't just assign to both a preset value (e.g. `MISSING_NAME`) cos this will result in duplicate persons. So, there needs to be extra logic to handle this, which I'm unsure whether would bloat up the code base or not.

@pyokagan can I have your thoughts on this as well? Thanks! ^^

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
Zhiyuan-Amos: Ah sounds good. @yamidark take note of the above comment, you can create a new `Exception` (`MissingTagException`)?

This way, `XmlSerializableAddressBook` can handle these exceptions differently.

pyokagan: To clarify: It isn't strictly necessary to create separate exceptions and handle the errors differently (although it would be nice if they at least had different error messages). My above example was just to illustrate that there are two distinct error cases even though they are handled by the same code path.

pyokagan:
> @pyokagan can I have your thoughts on this as well? Thanks! ^^

Note that the issue is just about the *detection* of the error. We already have code for handling storage loading errors in https://github.com/se-edu/addressbook-level4/blob/e7f43d3f05fd8e336c4387e059af92ed50a70512/src/main/java/seedu/address/MainApp.java#L97-L99 . Whether that is the right thing to do or not is completely out of scope for this PR.

pyokagan:
> Whether that is the right thing to do or not is completely out of scope for this PR.

So this PR should be preserving the current behavior of using an empty address book if there is a corrupted file.

yamidark:
>Also, you should have some tags in both the valid and invalid values since you are also testing for:
```
assertEquals(0, dataFromFile.getTagList().size());
```
>This is to show that only tags in the valid example are added, but the tags in the invalid examples aren't added.

@Zhiyuan-Amos If i'm not wrong, `dataFromFile#getTagList()` does not get the tags from each `XmlAdaptedPerson` at
```xml
    <persons>
        <name>Hans Muster</name>
        <phone>93210283</phone>
        <email>charlotte@example.com</email>
        <address>Blk 11 Ang Mo Kio Street 74, #11-04</address>
        <tagged>neighbours</tagged>
    </persons>

```
, but rather in the "master tag list" at 
```xml
    <tags>colleagues</tags>
    <tags>friend</tags>
    <tags>classmates</tags>
    <tags>neighbours</tags>
    <tags>criminal</tags>
    <tags>family</tags>
    <tags>friends</tags>
```
, so I don't think we can test tags from each valid person is saved.

>So this PR should be preserving the current behavior of using an empty address book if there is a corrupted file.

So should I have another `try/catch` block for 
https://github.com/se-edu/addressbook-level4/blob/e7f43d3f05fd8e336c4387e059af92ed50a70512/src/main/java/seedu/address/MainApp.java#L105
to catch an Exception(create a new `CorruptedDataException`?) and set `ModelManager` with an empty `AddressBook`?

pyokagan:
@yamidark 

> So should I have another `try/catch` block for 
https://github.com/se-edu/addressbook-level4/blob/e7f43d3f05fd8e336c4387e059af92ed50a70512/src/main/java/seedu/address/MainApp.java#L105
to catch an Exception(create a new `CorruptedDataException`?) and set `ModelManager` with an empty `AddressBook`?

Maybe use the existing `DataConversionException`? We can't really tell the difference between "data is in an invalid format" and a "corrupted file" right?

pyokagan:
@Zhiyuan-Amos Remember to either put an X or a tick on the PR.

Zhiyuan-Amos:
@pyokagan 

> Note that the issue is just about the detection of the error. We already have code for handling storage loading errors in

https://github.com/se-edu/addressbook-level4/blob/e7f43d3f05fd8e336c4387e059af92ed50a70512/src/main/java/seedu/address/MainApp.java#L97-L99

Hmm if I am not wrong, this error is not caused by having illegal values (i.e. __missing xml tags__ such as `<name>` or an __invalid name__ such as `@@@`). It's caused by this code in `XmlUtil#getDataFromFile(File, Class<T>)`:

```
JAXBContext context = JAXBContext.newInstance(classToConvert);
Unmarshaller um = context.createUnmarshaller();
```

So this code can throw `JAXBException`, which is then rethrown as `DataConversionException` and caught by `MainApp`. 

I.e. it seems that we have no error handling code for __invalid input__ and __missing tags__ currently. The code simply crashes somewhere with `NullPointerException`.

@yamidark 

> @Zhiyuan-Amos If i'm not wrong, dataFromFile#getTagList() does not get the tags from each XmlAdaptedPerson ... in the "master tag list"

Ah I see, ok! Thanks for correcting me. 
Also, as spoken, let's update `XmlSerializableAddressBook#getTagList()` too. ^^

pyokagan:
@Zhiyuan-Amos 

> I.e. it seems that we have no error handling code for invalid input and missing tags currently. The code simply crashes somewhere with NullPointerException.

Yes, hence #749 ;-)

Zhiyuan-Amos:
> Yes, hence #749 ;-)

Ah ok! @yamidark I think over here it's fine to just do a simple fix (i.e. deleting person with invalid / missing fields), then once you're done with this PR, you can work on #749 (if we intend to do something other than just simply deleting the person).

pyokagan:
@Zhiyuan-Amos 
> I think over here it's fine to just do a simple fix (i.e. deleting person with invalid / missing fields

Eh you still need to report errors somehow to the higher layers anyway. It's not a "fix" if you just swallow errors like that.

Zhiyuan-Amos:
> Eh you still need to report errors somehow to the higher layers anyway. It's not a "fix" if you just swallow errors like that.

Yup; just realized that my statement is quite misleading -.- Was comparing btwn setting preset values for persons with corrupted / missing values with this method, so this is definitely easier.

Though now that I think about it, I think an even simpler fix would be to throw the Exception up all the way up to `MainApp`. I.e. `XmlFileStorage#loadDataFromSaveFile(File)` can throw `CorruptedAddressBookException` and it'll be caught by `MainApp` and handled the same way as with the other exceptions?

yamidark:
@Zhiyuan-Amos I think maybe I try to fix #749 first, then later for this PR I can just do the first commit of having the extra `non-null` check for the parameters + regression tests. What do you think about this?

Zhiyuan-Amos:
Also good!

yamidark:
Reworking on this issue since #749 has been resolved.

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -40,27 +40,55 @@ public void toModelType_invalidName_throwsIllegalValueException() {
         Assert.assertThrows(IllegalValueException.class, person::toModelType);
     }
 
+    @Test
+    public void toModelType_nullName_throwsIllegalValueException() {
+        XmlAdaptedPerson person =
+                new XmlAdaptedPerson(null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
```
yamidark: Just noticed that there's no need for line break here =.=. Will resolve this in the next iteration

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
```diff
@@ -20,6 +20,8 @@
  */
 public class XmlAdaptedPerson {
 
+    private static final String MISSING_FIELD_MESSAGE_FORMAT = "Missing %s Field!";
```
Zhiyuan-Amos: We should rephrase this to sound like a sentence: `Person %s field is missing`

yamidark:
@Zhiyuan-Amos @damithc 
Sorry, I actually forgot that I should also add regression tests for `XmlSerializableAddressBookTest` and `XmlAddressBookStorageTest` as well since they were in a separate commit which I removed previously 💦.
Will submit a new iteration again for your review after adding them back.

Zhiyuan-Amos:
> Sorry, I actually forgot that I should also add regression tests for XmlSerializableAddressBookTest and XmlAddressBookStorageTest as well since they were in a separate commit which I removed previously 💦.

Why do you need (regression?) tests for that?

yamidark:
>Why do you need (regression?) tests for that?

To test that reading files with missing values will also throw the `IllegalValueException` to the higher level components.

Zhiyuan-Amos:
> To test that reading files with missing values will also throw the IllegalValueException to the higher level components.

The test in `XmlAdaptedPersonTest` is sufficient, because each test is supposed to cover the code path of the method that it is testing for. E.g. In `XmlSerializableAddressBook`:

https://github.com/se-edu/addressbook-level4/blob/master/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java#L49-L58

We just need a code path to test for 1 kind of wrong tag that will throw `IllegalValueException` at `addressBook.addTag(t.toModelType());`, and 1 kind of wrong person that will throw `IllegalValueException` at `addressBook.addPerson(p.toModelType());`

> While there is test for throws IllegalValueException, there isn't any test to ensure that the correct message is being shown. E.g, phone is null but shows name is empty field instead. This also helps to ensure that the throws clause is being executed at the correct location.

Good catch @eugenepeh. That means we need to write another method in `Assert` class.

yamidark:
@Zhiyuan-Amos @eugenepeh 

>Also, instead of using "Name", would it be better to use Name.class.getSimpleName()?

Sounds good, will try that.

>The test in XmlAdaptedPersonTest is sufficient, because each test is supposed to cover the code path of the method that it is testing for. E.g. In XmlSerializableAddressBook

Hmm, my thinking was that `XmlAdaptedPersonTest` only tests for the `null` values within themselves, but we should also add tests of reading a valid xml format file but it has missing fields (no  `<name> </name>`), testing that doing so will cause the values saved in the `XmlAdaptedPerson` to be `null` and thus throws the `IllegalValueException`.

>That means we need to write another method in Assert class.

Will add that in an earlier commit. Which reminds me, should we instead switch to import the static method `import seedu.address.testutil.Assert.assertThrows` so that we can remove the `Assert` static call in front of each `Assert.assertThrows(IllegalValueException.class, person::toModelType)` in `XmlAdaptedPersonTest`?

Zhiyuan-Amos:
> we should also add tests of reading a valid xml format file but it has missing fields (no <name> </name>)

Hmm I'm thinking that we should test `XmlUtil#getDataFromFile(File, Class<T>)` instead. So we pass in a .xml file with missing field, and we set the class to be `XmlAdaptedPerson`. Then we call `XmlAdaptedPerson#toModelType()` to ensure that `IllegalValueException` is thrown.

> should I overload assertThrows method with an additional String expectedMessage parameter so that methods that also want to check the message thrown can use it?

Yup. This seems to be a more natural way of doing things.

---

On a side note:
https://github.com/se-edu/addressbook-level4/blob/2e7f092c5e58dff587347304f59b0b8286abdaee/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java#L43-L49

Do we have tests to check that duplicate persons will throw `IllegalValueException`? Since it's specified in the header comments.

yamidark:
>We are basically trying to test that when a .xml file with missing field is being converted to an object (whatever object, be it XmlSerializableAddressBook, XmlAdaptedPerson etc), then the field remains missing, and we expect an IllegalValueException to be thrown when toModelType() is called right?
I'm thinking that we should test using XmlUtil#getDataFromFile(File, Class<T>). So we pass in a .xml file with missing field, and we set the class to be XmlAdaptedPerson. Then we call XmlAdaptedPerson#toModelType() to ensure that IllegalValueException is thrown. I.e. this can be written XmlAdaptedPersonTest.
If we have tested for what I've suggested above, then I'm not too sure whether there's a need to write another test in XmlSerializableAddressBook as mentioned above

Noted. Will try that out. Just thought that we should also test reading of files, and what you suggested seems better.

>Do we have tests to check that duplicate persons will throw IllegalValueException? Since it's specified in the header comments.

Don't think there is any 😓. Probably should raise another issue for that.

yamidark:
@Zhiyuan-Amos for your review again.

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -25,6 +25,25 @@ public static void assertThrows(Class<? extends Throwable> expected, VoidCallabl
                 String.format("Expected %s to be thrown, but nothing was thrown.", expected.getName()));
     }
 
+    /**
+     * Asserts that the {@code callable} throws the {@code expectedException} and the {@code expectedMessage}.
+     */
+    public static void assertThrows(Class<? extends Throwable> expectedException, String expectedMessage,
```
Zhiyuan-Amos: After thinking about this for a while, I can't think of a way to do this cleanly without repeating the logic.

1. This way requires us to call `callable.call()` twice, which is pretty weird.
2. We can shift all the logic to this method, so that `assertThrows(Class<? extends Throwable>, VoidCallable)` simply calls `assertThrows(expectedException, null, callable)`. Then, in this method, if we check that `expectedMessage` is null, we will not perform the equality check for the message. This solution seems better though.

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -25,6 +25,25 @@ public static void assertThrows(Class<? extends Throwable> expected, VoidCallabl
...
+                                    VoidCallable callable) {
+        assertThrows(expectedException, callable);
+        try {
+            callable.call();
+        } catch (Throwable actualException) {
+            if (actualException.getMessage().equals(expectedMessage)) {
```
yamgent: Can just use `assertEquals()`.

Zhiyuan-Amos: He copied from Paul's code above :P And I think the reason for not using `assertEquals()` is because you need the if-statement for the early return, otherwise your app will crash.

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
```diff
@@ -18,8 +19,11 @@
 /**
  * JAXB-friendly version of the Person.
  */
+@XmlRootElement(name = "person")
```
Zhiyuan-Amos: Let's not modify the main code base for this. 

Do this in `XmlAdaptedPersonTest` instead:
```
@XmlRootElement(name = "person")
class XmlAdaptedPersonRoot extends XmlAdaptedPerson {}
```

You can rename the class name accordingly.

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -70,4 +113,14 @@ public void toModelType_invalidTags_throwsIllegalValueException() {
         Assert.assertThrows(IllegalValueException.class, person::toModelType);
     }
 
+    /**
+     * Reads the test file where the person has missing name field to ensure name field in {@code XmlAdaptedPerson}
+     * will be null.
+     */
```
Zhiyuan-Amos: Let's not have header comments for this test method because it's self-explanatory.

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -70,4 +113,14 @@ public void toModelType_invalidTags_throwsIllegalValueException() {
...
+    @Test
+    public void getDataFromFile_missingFieldPerson_throwsIllegalValueException() throws Exception {
```
Zhiyuan-Amos: Seems like we are doing an integration test here since we are testing for both `getDataFromFile` and `toModelType`? idk sia

Zhiyuan-Amos: Anws, if you're testing `getDataFromFile`, then this method should be in `XmlUtilTest` :P

yamidark:
2nd para: 
>We are unable to perform stronger tests to verify the correctness of 
the error message thrown.

Would it be better if we add in that `some method/tests throws different instances of the same exception class for different errors, thus there is a need for a stronger test for these method/tests to ensure we are testing the correct exception instance.`?

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
yamidark: Hmm, tried transfering it over to `XmlUtilTest`, but feel weird to have this test here, especially since there isn't a read valid `XmlAdaptedPerson` test.
Actually now that I think about it, doesn't feel necessary to test reading a lone `XmlAdaptedPerson` and have dependcies on `XmlUtil` too. 
I initially thought it may not be obvious that a missing field gives `null` so we should have an additional read file test, but thinking about it again this actually feels quite obvious 😅. So the current tests within `XmlAdaptedPersonTest` should be sufficient as you mentioned.
Should we just remove adding this test? 😅 (Sorry for jumping around here and there)

Zhiyuan-Amos:
> Would it be better if we add in that some method/tests throws different instances of the same exception class for different errors, thus there is a need for a stronger test for these method/tests to ensure we are testing the correct exception instance.?

Also ok, but you have to specify in the first paragraph that `Assert#assertThrows(...)` is a test utility method which helps us to test that certain methods throw Exception... Something like that.  Just be clear ah :)

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
Zhiyuan-Amos: This test seems necessary because:
1. Prevents regression
2. We're testing for the integration between both units `XmlTestUtil` and `XmlAdaptedPerson`. I.e. currently the existing tests only perform unit test, and having unit tests working fine doesn't mean that these classes interact fine when integrated.

So perhaps we can leave this test here, and rename it as `toModelType_missingFieldInXml_throwsIllegalValueException`?

yamidark: >We're testing for the integration between both units XmlTestUtil and XmlAdaptedPerson. I.e. currently the existing tests only perform unit test, and having unit tests working fine doesn't mean that these classes interact fine when integrated.

Yep, understood what you're saying.

>So perhaps we can leave this test here, and rename it as toModelType_missingFieldInXml_throwsIllegalValueException?

Yep sounds good. I should probably also add a valid test `toModelType_validPersonInXml_returnsPerson()` too right?

Zhiyuan-Amos: I think you can write it in `XmlUtilTest`. There's this test over there which seems like an integration test too:

```java
@Test
public void getDataFromFile_validFile_validResult() throws Exception {
        AddressBook dataFromFile = XmlUtil.getDataFromFile(VALID_FILE, XmlSerializableAddressBook.class).toModelType();
        assertEquals(9, dataFromFile.getPersonList().size());
        assertEquals(0, dataFromFile.getTagList().size());
}
```

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +71,28 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void toModelTypeXmlAdaptedPersonFromFile_missingFieldInXml_throwsIllegalValueException() throws Exception {
+        XmlAdaptedPerson dataFromFile = XmlUtil.getDataFromFile(MISSING_FIELD_PERSON_FILE, XmlAdaptedPersonRoot.class);
```
Zhiyuan-Amos: You're supposed to do `XmlAdaptedPersonRoot dataFromFile = ...` :P

src/test/java/seedu/address/storage/XmlAdaptedPersonTest.java
```diff
@@ -69,5 +106,4 @@ public void toModelType_invalidTags_throwsIllegalValueException() {
                 new XmlAdaptedPerson(VALID_NAME, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, invalidTags);
         Assert.assertThrows(IllegalValueException.class, person::toModelType);
     }
-
```
Zhiyuan-Amos: unnecessary delete

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -11,18 +11,33 @@
      * Asserts that the {@code callable} throws the {@code expected} Exception.
      */
     public static void assertThrows(Class<? extends Throwable> expected, VoidCallable callable) {
+        assertThrows(expected, null, callable);
+    }
+
+    /**
+     * Asserts that the {@code callable} throws the {@code expectedException} and the {@code expectedMessage}.
+     */
+    public static void assertThrows(Class<? extends Throwable> expectedException, String expectedMessage,
```
Zhiyuan-Amos: The logic here is quite messy. What do you think of this instead?

```java
        try {
            callable.call();
        } catch (Throwable actualException) { 
            if (actualException.getClass().isAssignableFrom(expectedException) && 
                    (expectedMessage == null || actualException.getMessage().equals(expectedMessage))) {
                    return;
            } else {
                throw new AssertionFailedError(String.format("Expected exception thrown: %s, actual: %s",
                        expectedException.getName(), actualException.getClass().getName()));
            }
        }
        throw new AssertionFailedError(
                String.format("Expected %s to be thrown, but nothing was thrown.", expectedException.getName()));
```

yamgent: Alternative way of doing this:

```java
        try {
            callable.call();
        } catch (Throwable actualException) { 
            String errorMessage;
            
            if (!actualException.getClass().isAssignableFrom(expectedException)) {            
                errorMessage = String.format("Expected exception thrown: %s, actual: %s",
                        expectedException.getName(), actualException.getClass().getName());
            } else if (expectedMessage != null && !actualException.getMessage().equals(expectedMessage)) {
                errorMessage = String.format("Expected message thrown: %s, actual: %s", expectedMessage,
                        actualException.getMessage());                
            } else {
                return;
            }
            
            throw new AssertionFailedError(errorMessage);
        }
        
        throw new AssertionFailedError(
                String.format("Expected %s to be thrown, but nothing was thrown.", expectedException.getName()));
```

src/main/java/seedu/address/storage/XmlAdaptedPerson.java
yamidark: omg im so dumb, completely forgot about this 😱 😱

src/test/data/XmlUtilTest/invalidFieldInXml.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!-- Person with an invalid phone field -->
+<person>
+	<name>Hans Muster</name>
```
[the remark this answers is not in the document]
yamidark: oops, sorry did not set my SublimeText to indent with spaces 😅.
Will make the necessary change.

src/test/data/XmlUtilTest/validPersonInXml.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<person>
+	<name>Benson Meier</name>
```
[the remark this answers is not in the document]
yamidark: same as above

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: Hmm, I was thinking that using `XmlAdaptedPerson` would be clearer.
`XmlAdaptedPersonRoot` is used only for reading the file, but we should be using the `XmlAdaptedPerson` methods for the testing?

src/test/java/seedu/address/testutil/Assert.java
yamidark: yep I prefer @yamgent method more, lesser nested if-else

yamidark: I think @yamgent method is better, since it is more specific on when it is the `expectedMessage` which is different.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
Zhiyuan-Amos: Ok, sounds good. `XmlAdaptedPersonRoot` seems weird though (the word "Root" is... ??? doesn't explain much). We should also write some header comments for the class too.

src/test/java/seedu/address/testutil/Assert.java
Zhiyuan-Amos: Yup, use @yamgent's way ^^

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: >XmlAdaptedPersonRoot seems weird though (the word "Root" is... ??? doesn't explain much).

`XmlAdaptedPersonStub`? Will add the header comments.

Zhiyuan-Amos: That's not how `Stub` is usually used though :P Ah how about `XmlAdaptedPersonWithRootElement`?

yamidark: >That's not how Stub is usually used though :P

Whoops xD.

> Ah how about XmlAdaptedPersonWithRootElement?

Sounds good!

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -11,18 +11,35 @@
...
-            throw new AssertionFailedError(message);
+
+            throw new AssertionFailedError(errorMessage);
         }
-        throw new AssertionFailedError(
-                String.format("Expected %s to be thrown, but nothing was thrown.", expected.getName()));
+
+        throw new AssertionFailedError(String.format("Expected %s to be thrown, but nothing was thrown.",
+                expectedException.getName()));
+
```
Zhiyuan-Amos: whyyyy extra new line

yamidark: 137 chars =O

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -93,4 +129,10 @@ public void saveDataToFile_validFile_dataSaved() throws Exception {
         dataFromFile = XmlUtil.getDataFromFile(TEMP_FILE, XmlSerializableAddressBook.class);
         assertEquals(dataToWrite, dataFromFile);
     }
+
+    /**
+     * Helper class with XmlRootElement for reading of {@code XmlAdaptedPerson} from .xml files
+     */
+    @XmlRootElement(name = "person")
+    static class XmlAdaptedPersonStub extends XmlAdaptedPerson {}
```
Zhiyuan-Amos: class name should be updated to `XmlAdaptedPersonWithRootElement` as discussed previously.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -93,4 +129,10 @@ public void saveDataToFile_validFile_dataSaved() throws Exception {
         dataFromFile = XmlUtil.getDataFromFile(TEMP_FILE, XmlSerializableAddressBook.class);
         assertEquals(dataToWrite, dataFromFile);
     }
+
+    /**
+     * Helper class with XmlRootElement for reading of {@code XmlAdaptedPerson} from .xml files
```
Zhiyuan-Amos: This isn't really a helper class. It's just a `Test class annotated with {@code XmlRootElement} to allow unmarshalling of .xml data to {@code XmlAdaptedPerson} objects`. Something like that.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +71,30 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void toModelTypeXmlAdaptedPersonFromFile_missingPersonFieldFile_throwsIllegalValueException()
```
Zhiyuan-Amos: Actually, I wonder whether we need to call `toModelType` for these tests. Can we create an `XmlAdaptedPerson` object for equality comparison? I.e. if it's missing a `name` field, then we can pass in `null` for the `name` field to the `XmlAdaptedPerson expectedPerson`.

src/test/java/seedu/address/testutil/Assert.java
Zhiyuan-Amos: I'm referring to line 42. :P

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: >Can we create an XmlAdaptedPerson object for equality comparison?

Yep we can, just like the valid test below.

Zhiyuan-Amos: The valid test performs the comparison using `Person` object, not `XmlAdaptedPerson` object.

I'm looking at comparing using `XmlAdaptedPerson` objects.

yamidark: >I'm looking at comparing using XmlAdaptedPerson objects.

We can create the `XmlAdaptedPerson` from a `Person` object directly, or we supply it with `String`s for each parameter.

Zhiyuan-Amos: > I.e. if it's missing a name field, then we can pass in null for the name field to the XmlAdaptedPerson expectedPerson.

Yup that's right, can you try supplying `String`s, and null for the field that's missing?

src/test/java/seedu/address/testutil/Assert.java
eugenepeh: higher level break is preferred over lower level break.

So something like this?
<pre>        
throw new AssertionFailedError(String.format(
        "Expected %s to be thrown, but nothing was thrown.", expectedException.getName()));
</pre>

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,31 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void xmlAdaptedPersonFromFile_missingPersonFieldFile_validResult() throws Exception {
+        XmlAdaptedPerson dataFromFile = XmlUtil.getDataFromFile(
+                MISSING_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson person = new XmlAdaptedPerson(null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(dataFromFile, person);
```
Zhiyuan-Amos: The ordering should be `expectedPerson, actualPerson` right?

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,31 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void xmlAdaptedPersonFromFile_missingPersonFieldFile_validResult() throws Exception {
+        XmlAdaptedPerson dataFromFile = XmlUtil.getDataFromFile(
+                MISSING_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson person = new XmlAdaptedPerson(null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
```
Zhiyuan-Amos: Rename to `expectedPerson`

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,31 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void xmlAdaptedPersonFromFile_missingPersonFieldFile_validResult() throws Exception {
+        XmlAdaptedPerson dataFromFile = XmlUtil.getDataFromFile(
```
Zhiyuan-Amos: Rename to `actualPerson`

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,31 @@ public void getDataFromFile_validFile_validResult() throws Exception {
         assertEquals(0, dataFromFile.getTagList().size());
     }
 
+    @Test
+    public void xmlAdaptedPersonFromFile_missingPersonFieldFile_validResult() throws Exception {
```
Zhiyuan-Amos: Rename the 2nd part of this test name to `..._fileWithMissingPersonField_...` for more clarity.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -23,8 +28,19 @@
...
+    private static final File VALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "validPerson.xml");
     private static final File TEMP_FILE = new File(TestUtil.getFilePathInSandboxFolder("tempAddressBook.xml"));
 
+    private static final String INVALID_PHONE = "9482asf424";
+
+    private static final String VALID_NAME = "Hans Muster";
+    private static final String VALID_PHONE = "9482424";
+    private static final String VALID_EMAIL = "hans@example";
+    private static final String VALID_ADDRESS = "4th street";
+    private static final List<XmlAdaptedTag> VALID_TAGS = Collections.singletonList(new XmlAdaptedTag("friends"));
```
Zhiyuan-Amos: I noticed that you added the tags this iteration. Is there a reason for doing so?

src/test/java/seedu/address/commons/util/XmlUtilTest.java
Zhiyuan-Amos: Make these changes for the tests below as well :)

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -23,8 +28,19 @@
...
     private static final File MISSING_FILE = new File(TEST_DATA_FOLDER + "missing.xml");
     private static final File VALID_FILE = new File(TEST_DATA_FOLDER + "validAddressBook.xml");
+    private static final File MISSING_PERSON_FIELD_FILE = new File(TEST_DATA_FOLDER + "missingPersonField.xml");
+    private static final File INVALID_PERSON_FIELD_FILE = new File(TEST_DATA_FOLDER + "invalidPersonField.xml");
+    private static final File VALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "validPerson.xml");
     private static final File TEMP_FILE = new File(TestUtil.getFilePathInSandboxFolder("tempAddressBook.xml"));
 
+    private static final String INVALID_PHONE = "9482asf424";
+
+    private static final String VALID_NAME = "Hans Muster";
```
Zhiyuan-Amos: Can we use the details of someone from `TypicalPersons` so that we don't have to declare new constants?

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -11,18 +11,34 @@
...
         try {
             callable.call();
         } catch (Throwable actualException) {
-            if (actualException.getClass().isAssignableFrom(expected)) {
+            String errorMessage;
+
+            if (!actualException.getClass().isAssignableFrom(expectedException)) {
+                errorMessage = String.format("Expected exception thrown: %s, actual: %s",
+                        expectedException.getName(), actualException.getClass().getName());
+            } else if (expectedMessage != null && !actualException.getMessage().equals(expectedMessage) ) {
```
Zhiyuan-Amos: `actualException.getMessage()` may return `null`. I think it's safer to do 
`expectedMessage != null && !expectedMessage.equals(actualException.getMessage())`

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: >Can we use the details of someone from TypicalPersons so that we don't have to declare new constants?

it is possible to do so, just that it can be quite ugly since we have to use Strings, thus we would need to call like `BENSON.getName().toString()` for each field, and for the tags `BENSON.getTags().stream().map(XmlAdaptedTag::new).collect(Collectors.toList())`, which was also why I tried declaring constants 1st using the values that were already set in the files.

We could also change the visibility of the constants in `XmlAdaptedPersonTest` to `public` so that `XmlUtilTest` can access too, not too sure about this though.

Zhiyuan-Amos: Maybe we can try having a `XmlAdaptedPersonBuilder`:

```
public XmlAdaptedPersonBuilder(Person person) {
    this.name = person.getName().toString();
    ...
}

public withName(String name) {
    this.name = name;
}
```

If we want a `XmlAdaptedPerson` with a null name, we can do this:
```
XmlAdaptedPersonBuilder builder = new XmlAdaptedPersonBuilder(BENSON);
XmlAdaptedPerson expectedPerson = builder.withName(null).build();

```

yamidark: >Maybe we can try having a XmlAdaptedPersonBuilder

Would have lots of similar code with `PersonBuilder` though.
Maybe we could change `PersonBuilder` to store each field as a `String` instead, then we can  change the methods to `buildPerson` or `buildXmlAdaptedPerson` for what we need respectively.
Also found that `PersonBuilder` does not have assertions that the parameters supplied for each field are non-`null`. Could add those assertion checks in `buildPerson` too.

Zhiyuan-Amos: Nvm, let's just leave it this way then. 👍

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: >I noticed that you added the tags this iteration. Is there a reason for doing so?

Since we are testing reading the fields of the .xml files, I thought it would make sense to test reading of tags too.

Zhiyuan-Amos: Yup, this sounds good. Add that as part of your commit message too :)

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,33 @@ public void getDataFromFile_validFile_validResult() throws Exception {
...
+    public void xmlAdaptedPersonFromFile_fileWithMissingPersonField_validResult() throws Exception {
+        XmlAdaptedPerson actualPerson = XmlUtil.getDataFromFile(
+                MISSING_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson expectedPerson = new XmlAdaptedPerson(
+                null, VALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(expectedPerson, actualPerson);
+    }
+
+    @Test
+    public void xmlAdaptedPersonFromFile_fileWithinvalidPersonField_validResult() throws Exception {
```
Zhiyuan-Amos: wrong casing for `invalid`.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -59,6 +75,33 @@ public void getDataFromFile_validFile_validResult() throws Exception {
...
+        XmlAdaptedPerson actualPerson = XmlUtil.getDataFromFile(
+                INVALID_PERSON_FIELD_FILE, XmlAdaptedPersonWithRootElement.class);
+        XmlAdaptedPerson expectedPerson = new XmlAdaptedPerson(
+                VALID_NAME, INVALID_PHONE, VALID_EMAIL, VALID_ADDRESS, VALID_TAGS);
+        assertEquals(expectedPerson, actualPerson);
+    }
+
+    @Test
+    public void xmlAdaptedPersonFromFile_fileWithvalidPerson_validResult() throws Exception {
```
Zhiyuan-Amos: wrong casing for `valid`

src/test/java/seedu/address/commons/util/XmlUtilTest.java
Zhiyuan-Amos: Did you update the commit message?

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: Sorry, missed that again. Will change it.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
yamidark: I updated the commit message to mention we are adding tests in `XmlUtilTest` for reading of `XmlAdaptedPerson` from .xml files. Did you meant to update that we are adding constants/tags to `XmlUtilTest`?

Zhiyuan-Amos: Ah wait ignore what I said, was a bit confused sorry :<

Zhiyuan-Amos:
@damithc shall we merge this immediately too as this PR fixes one of the issues raised by the students over the past semester?

damithc:
> @damithc shall we merge this immediately too as this PR fixes one of the issues raised by the students over the past semester?

Just checked the CS2103 timeline. level 4 is released next week, not this week. So it is fine if we merge by  this Friday.

src/test/java/seedu/address/testutil/Assert.java
```diff
@@ -11,18 +11,34 @@
      * Asserts that the {@code callable} throws the {@code expected} Exception.
      */
     public static void assertThrows(Class<? extends Throwable> expected, VoidCallable callable) {
+        assertThrows(expected, null, callable);
+    }
+
+    /**
+     * Asserts that the {@code callable} throws the {@code expectedException} and the {@code expectedMessage}.
```
yamgent: Should we mention that `expectedMessage` is optional?

yamidark: Hmm, we probably should add that `expectedMessage` is `null` if we do not need the additional message verification.

yamidark: `@param expectedMessage is optional and should be passed as null if the message verification is not needed.`

Would something like this be good?

Zhiyuan-Amos: The reason why I thought that it wasn't necessary is cos I would expect callers to see the overloaded method `assertThrows(Class<? extends Throwable>, VoidCallable)` and call that instead if they do not wish to perform the message verification.

As such, I'm thinking maybe we don't need to do `@param expectedMessage is optional...`, but we can do something like:

```
If there's no need for the verification of the exception's error message, call 
{@code assertThrows(Class<? extends Throwable>, VoidCallable)} instead.
@see assertThrows(Class<? extends Throwable>, VoidCallable)
```

Zhiyuan-Amos: On a side note, there's a small issue with this code: We cannot test that the `actualException` is null or not :P

yamgent: > I'm wondering whether this is ok with you @yamgent, or should we update this method to be able to test for null error messages?

If we do not need such functionalities, it should be OK to omit it by YAGNI.

Although I can see an issue of whether `null` meant no message or don't change message. This would have been an excellent place to use `Optional`s, but by convention we shouldn't use `Optional` on parameters, so... :/

PR #811, opened 2018-01-24
JUnit: Update to use latest version #707

### LINKED ISSUE #707: JUnit: Update to use latest version
We are currently using JUnit version 4.12. We can upgrade it to JUnit 5.0.0.

Rinder5:
fixes #707 (eventually, right now it just starts to fix it)

This is a start towards upgrading towards JUnit5. It does not changes any tests, but it makes it such that Gradle is able to read JUnit5 tests and execute them.

One thing to keep in consideration if we want to change the tests to be JUnit5 tests:
Right now there is no easy way to do the temporary folder rule, it might be messy to create our own version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,8 +38,13 @@ jacocoTestReport {
     }
 }
 
+test {
+    useJUnitPlatform()
+}
+
 dependencies {
     String testFxVersion = '4.0.12-alpha'
+    String jUnitVersion = '5.1.0'
 
     compile group: 'org.controlsfx', name: 'controlsfx', version: '8.40.11'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.7.0'
@@ -59,8 +64,12 @@ dependencies {
         exclude group: 'junit', module: 'junit'
     }
 
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: jUnitVersion
+
     testRuntime group: 'org.testfx', name: 'testfx-internal-java9', version: testFxVersion
     testRuntime group: 'org.testfx', name: 'openjfx-monocle', version: 'jdk-9+181'
+    testRuntimeOnly group:'org.junit.vintage', name:'junit-vintage-engine', version: jUnitVersion
+    testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: jUnitVersion
 }
 
 shadowJar {


yamgent:
> Starts fixing #707

I don't think GitHub recognizes `fixing`. :P You need to use `Fix #707` or `Fixes #707` so that GitHub will automatically close the issue when the PR is completed (see [this doc](https://help.github.com/articles/closing-issues-using-keywords/)).

Zhiyuan-Amos:
As spoken, let's split the commit up and write commit messages so that noobs like me know what's going on ^^

Rinder5:
Holding off on this for now. Gradle 4.6 includes JUnit5 support, so this may be unnecessary.

Zhiyuan-Amos:
Hmm this PR can still update to use JUnit 5 right? I.e. we don't need to close this PR :P

Zhiyuan-Amos:
@Rinder5 you can continue on this PR since #837 has been merged :)

Rinder5:
Currently this just enables JUnit5 functionality. It uses the vintage engine bundled with JUnit5 to run our JUnit4 tests. Should I go about changing all the tests to JUnit5 versions?

build.gradle
```diff
@@ -38,8 +38,13 @@ jacocoTestReport {
     }
 }
 
+test {
+    useJUnitPlatform()
+}
```
Zhiyuan-Amos: This is necessary for JUnit 5?

build.gradle
```diff
@@ -51,6 +56,9 @@ dependencies {
     compile group: 'javax.activation', name: 'activation', version: '1.1.1'
 
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testRuntime 'org.junit.vintage:junit-vintage-engine:4.12.3'
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: jUnitJuniperVersion
+    testRuntime group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: jUnitJuniperVersion
```
Zhiyuan-Amos: Why aren't we using the latest version for these 3 new imports `5.1.0`?

build.gradle
```diff
@@ -51,6 +56,9 @@ dependencies {
     compile group: 'javax.activation', name: 'activation', version: '1.1.1'
 
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testRuntime 'org.junit.vintage:junit-vintage-engine:4.12.3'
```
Zhiyuan-Amos: let's standardize the import format i.e. `group: ... name:... version:...`

build.gradle
Zhiyuan-Amos: Do explain why some import use `testImplementation` while some imports use `testRuntime` :)

build.gradle
Rinder5: Yea. Without it, gradle doesn't check for JUnit5 tests. Running the test manually with intellij works without it though.

Zhiyuan-Amos:
@Rinder5 I think for this PR, you can simply update it to JUnit 5 without making other changes in the code base. :) Is it possible for you to do this soon (as I have another PR that's dependent on your PR :P)

Rinder5:
I don't think there's anything specific to the upgrade that JUnit5 can remove over JUnit4. I did notice that my tests still ran normally without textfx-legacy though. And it is depreciated. Is that related?

Zhiyuan-Amos:
> I don't think there's anything specific to the upgrade that JUnit5 can remove over JUnit4. I did notice that my tests still ran normally without textfx-legacy though. And it is depreciated. Is that related?

Ah ok. Since we're only upgrading JUnit here, let's not do anything related to TestFX :) You can make that change in a separate PR though

Rinder5:
am I allowed to just make the change or do i need to create a new issue (probably will end up as first timers and probably won't be touched for a while =P)?

Zhiyuan-Amos:
You can make the change, but you need to find out what `testfx-legacy` does, and how we aren't using the API etc anymore that's why we can remove it. :p so you need to do some investigation actually.

Zhiyuan-Amos:
Also do clean up your [comment](https://github.com/se-edu/addressbook-level4/pull/811#issue-164927310) above. :)

Zhiyuan-Amos:
Remember to request for review too :p

> Changing the normal tests - including before and after segments- should be okay, and there is a workaround for the expected exception rule.

you can remove this since this works fine :p

build.gradle
```diff
@@ -51,6 +56,9 @@ dependencies {
     compile group: 'javax.activation', name: 'activation', version: '1.1.1'
 
     testCompile group: 'junit', name: 'junit', version: '4.12'
+    testRuntimeOnly group:'org.junit.vintage', name:'junit-vintage-engine', version: jUnitVersion
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: jUnitVersion
+    testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: jUnitVersion
```
Zhiyuan-Amos: Let's group `testRuntimeOnly` with `testRuntime` by placing them together.

yamgent:
> You can take a look at PersonListPanelTest#performanceTest and simplify using this method (I got it from JUnit 5 User Guide):
> 
> @Test
> void timeoutExceededWithPreemptiveTermination() {
>     // The following assertion fails with an error message similar to:
>     // execution timed out after 10 ms
>     assertTimeoutPreemptively(ofMillis(10), () -> {
>         // Simulate task that takes more than 10 ms.
>         Thread.sleep(100);
>     });
> }
> This PR is of lower importance compared to the other PRs that you are doing, so if you have sufficient time then you can explore more about JUnit 5 and make additional changes to the codebase :)
> 
> @yamgent look at this beautiful code in JUnit 5 which would have saved us so much time if we knew about it earlier :P

@Zhiyuan-Amos, I was wondering whether `assertTimeoutPreemptively()` can be used to solve the issue of unnecessarily testing the XML loading sequence. Apologies in advance if this is mentioned elsewhere already.

I am just considering the benefits of bringing in over JUnit5 now, because should we merge this PR, we will have to live with the fact that we are running both JUnit4 and JUnit5 together, which seems pretty messy to me.

I can understand that some tests are more difficult to port over from JUnit4 to JUnit5 (which will take a bigger PR to fix), but I feel that there still needs to be a strong justification of adding JUnit5 into the mix instead of retaining the status quo of just using only JUnit4, especially if JUnit5 features are not extremely necessary for our purposes right now.

Zhiyuan-Amos:
@yamgent I thought of this issue [here](https://github.com/se-edu/addressbook-level4/pull/863#comment-376525173).

If this change is possible, then upgrading to Junit5 is good cos currently the test is quite complex now as we have 3 timeouts and 1 log to ensure that we can find out where's the error. Compared to using `assertTimeoutPreemptively()`, which is just 1 timeout. 

Also, we can keep track of the things to change in an issue such as upgrading the rules, upgrading temporary folder etc. So at least we know when we can remove junit4 support.

yamgent:
> Also, we can keep track of the things to change in an issue such as upgrading the rules, upgrading temporary folder etc. So at least we know when we can remove junit4 support.

Thanks for the update. Should we jot them down in the commit message as well? As right now it just mentions that there's a few differences between JUnit4 and JUnit5, but doesn't really mention what the differences are, which will be a critical explanation of why we are not updating the tests in the commit itself.

Zhiyuan-Amos:
@yamgent yup you are right, that's a good point :+1: @Rinder5 do update according to the comments made above :p

Zhiyuan-Amos:
@yamgent ok with you? :) 

@Rinder5 Do raise the issues using a task list like:

- [ ] this

Zhiyuan-Amos:
@Rinder5 can you change the PR status to toMerge and do a last call request on Gitter? :)

Rinder5:
@Zhiyuan-Amos I don't have permissions to change the label

Zhiyuan-Amos:
Oh opps Prof haven't reviewed yet. @damithc for your review :)

Rinder5:
Rebased it since master was updated. should I resubmit a canihasreview?

Zhiyuan-Amos:
> Rebased it since master was updated. should I resubmit a canihasreview?

Yup! Remember to do a last call for review on Gitter as well.

PR #815, opened 2018-01-30
HelpWindow: May interfere with UI Components while active #727

### LINKED ISSUE #727: HelpWindow: May interfere with UI Components while active
Discovered this while testing other applications, the commands fail to show if the help window is open in the background.
Refer to: https://github.com/CS2103AUG2017-W14-B4/main/issues/106

`exit` command is blocked from executing while help window is open too.

Steps to reproduce:
1. Open help window
2a. Enter any command
Command result messages fail to display to user.
Use case end.

2b. Type `exit`
Application will not exit until help window is closed
Use case end

Normally closing the application by using the mouse will still close the application

Edit by Zhi Yuan: See https://github.com/nus-cs2103-AY1718S1/forum/issues/181#issuecomment-341664086 as well

vivekscl:
Fixes #727 

On a side note, should this TODO be converted into a separate issue or is it already covered by #551?
```
/**
 * TODO: This test is incomplete as it is missing test cases.
 */
public class HelpCommandSystemTest extends AddressBookSystemTest {
```

### DIFF
--- a/src/test/java/systemtests/HelpCommandSystemTest.java
+++ b/src/test/java/systemtests/HelpCommandSystemTest.java
@@ -1,16 +1,24 @@
 package systemtests;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
+import static seedu.address.testutil.TypicalIndexes.INDEX_FIRST_PERSON;
+import static seedu.address.ui.testutil.GuiTestAssert.assertListMatching;
 
 import org.junit.Test;
 
 import guitests.GuiRobot;
 import guitests.guihandles.HelpWindowHandle;
+import seedu.address.logic.commands.DeleteCommand;
 import seedu.address.logic.commands.HelpCommand;
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.ui.BrowserPanel;
+import seedu.address.ui.StatusBarFooter;
 
 /**
- * TODO: This test is incomplete as it is missing test cases.
+ * A system test class for the help window, which contains interaction with other UI components.
  */
 public class HelpCommandSystemTest extends AddressBookSystemTest {
     private static final String ERROR_MESSAGE = "ATTENTION!!!! : On some computers, this test may fail when run on "
@@ -46,6 +54,23 @@ public class HelpCommandSystemTest extends AddressBookSystemTest {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        // open help window and give it focus
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+
+        // assert that while the help window is open the UI updates correctly for a command execution
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertEquals("", getCommandBox().getInput());
+        assertCommandBoxShowsDefaultStyle();
+        assertNotEquals(HelpCommand.SHOWING_HELP_MESSAGE, getResultDisplay().getText());
+        assertNotEquals(BrowserPanel.DEFAULT_PAGE, getBrowserPanel().getLoadedUrl());
+        assertListMatching(getPersonListPanel(), getModel().getFilteredPersonList());
+
+        // assert that the status bar too is updated correctly while the help window is open
+        // note: the select command tested above does not update the status bar
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
     /**
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -59,6 +59,6 @@ public class HelpWindow extends UiPart<Stage> {
      */
     public void show() {
         logger.fine("Showing help page about the application.");
-        getRoot().showAndWait();
+        getRoot().show();
     }
 }


pyokagan:
> According to the JavaFX documentation for Stage, showAndWait() should 
be used if there is a need to block the caller until the modal stage is 
closed. In this case, the caller should not be blocked as other UI 
components have to function properly even after the help window is 
opened.

I'm wondering though -- usually when the event loop is blocked (e.g. you put an infinite while loop into an event handler), the whole application would not respond at all.

The situation here is more peculiar though, since the UI components do seem to respond to input. e.g. The command box actually does allow us to input commands, and in fact does run them.

So, I'm wondering if there is a more in-depth explanation for these particularities.

(As for why I am requesting this, to quote one of the [TEAMMATES principles](https://github.com/TEAMMATES/teammates/blob/master/docs/overview.md): "**We know what we are doing**: For us, it is not enough to know something is broken, we should also know why it is broken. It is not enough to get something working, we should know how we got it to work.") ;-)

vivekscl:
According to the JavaFX documentation on `Stage#showAndWait()`:
Shows this stage and waits for it to be hidden (closed) before returning to the caller. This method temporarily **blocks processing of the current event**, and starts a **nested event loop to handle other events**. This method must be called on the FX Application thread.

The current event in this case would be the command box area since that's where the event was called from, only that area is blocked while the other events work fine as the other events are in a nested event loop. While the command box is blocked, the commands run fine and UI changes accordingly everywhere except at the command box.

On the other hand, the `Stage#show()` only sets the visibility of the stage in question to true and doesn't block any events, so it is the preferred method in this case.

Zhiyuan-Amos:
> On a side note, should this TODO be converted into a separate issue or is it already covered by #551?

Yup. That's why it remains unchecked in #551. :p

The commit message can be clearer. In the 2nd paragraph, you should state what you have mentioned [above](https://github.com/se-edu/addressbook-level4/pull/815#issuecomment-361820886) - Only when you type `help` in Command Box then the Command Box will not execute any commands because the processing of the command is blocked, whereas if you press `F1` etc, the execution of further commands aren't blocked. So, you need to be more detailed in your commit message. You can take a look at Paul's commit messages [here](https://github.com/se-edu/addressbook-level4/commit/3148428e3ec2bb2313c68e3599b89d40734da252) to see how to structure longer commit messages.

> @Zhiyuan-Amos Do you think we should add a regression test for this?

Yup I think we can add a test in `HelpCommandSystemTest`, such that when you execute `help` command, you are still able to execute another simple command such as `ExitCommand`

pyokagan:
@vivekscl 
> The current event in this case would be the command box area since that's where the event was called from, only that area is blocked while the other events work fine as the other events are in a nested event loop. While the command box is blocked, the commands run fine and UI changes accordingly everywhere except at the command box.

Hmm? According to my testing, other UI elements do not seem to update. e.g. try adding a person while the help window is opened. The status bar should update to reflect the fact that the address book was modified, but it does not (!?)

Zhiyuan-Amos:
> Hmm? According to my testing, other UI elements do not seem to update. e.g. try adding a person while the help window is opened. The status bar should update to reflect the fact that the address book was modified, but it does not (!?)

Yup just tested too. 

1. If you use `F1`, command execution will update *all* UI components.
2. If you execute `help`, command execution will only update `PersonListPanel`. It seems that `StatusBarFooter`, `ResultDisplay` and `BrowserPanel` do not update because additional events are raised. Whereas `PersonListPanel` is updated because it is being binded to the `Model`; don't think there's any events raised to update the `PersonListPanel`.

pyokagan:
Hint: The use of `showAndWait()`, by itself, isn't what is causing the problem (i.e. A vanilla JavaFX project will likely not have this problem), so the commit message is inaccurate. As the JavaFX mentions, showAndWait() starts a nested event loop precisely to allow other events to run *including the same event type of the caller that was blocked*. Hence, why the command box still allows us to run commands when the help window is open.

What actually happens is the use of `showAndWait()` *plus* some other factor in our code base. Of course, if we have factor A and B which causes the problem, if we remove factor A of course the problem disappears. But factor B still remains, and may combine with another factor to cause us trouble again in the future, which is why it needs to be properly investigated and mentioned in the commit message rather than being handwaved away.

vivekscl:
Since it seems like the investigation will take a while, I'll post my progress here in the meantime.
After tracing the program execution after typing the `help` command, it gets stuck at `Dispatcher.java` where it enters the `dispatchEvent` method located beside the black arrow. After that, it enters a nested event loop in `Stage#showAndWait()`. As such it doesn't return back and the while loop doesn't terminate. If the while loop doesn't terminate, the `dispatching.remove()` method isn't called resulting in the if condition, `!dispatching.get()` to be false. If that is false, subsequent events will not be able to enter the while loop and be dispatched to their respective subscribers. Furthermore, the`queueForThread` keeps increasing in size as it just collects the subsequent events as seen by the "size = 3" note at the side after attempting to delete a person. After closing the help window, the execution continues from the while loop, causing a mass execution of all the events that were in the queue.

deepinscreenshot_select-area_20180201195757

I'm still finding that factor B. Hopefully I'll arrive at it soon :smile:

vivekscl:
@Zhiyuan-Amos Would it be better if the new test that was added was changed to a private method instead?
On a side note, this test fails for `Stage#showAndWait()` but passes for `Stage#show()`.

Zhiyuan-Amos:
> Would it be better if the new test that was added was changed to a private method instead?

@vivekscl yes, you should put it in a separate method. The test should have more rigorous checks though i.e. verify that other components are updated too.

Also, take a look at your commit message on GitHub, there's some minor mistakes :P 

Lastly:
> At this point, it iterates through the subscribers and dispatches the event to each of 
them. However, as Stage#showAndWait() temporarily blocks processing of 
the current event and starts a nested event loop to handle other events, 
the iteration of the subscribers does not terminate. As a result, 
subsequent events are unable to enter the iteration and be dispatched 
to their respective subscribers as these events are held up in the 
queue.

This part seems a bit confusing. ><

pyokagan:
@vivekscl From your commit message:
> However, as Stage#showAndWait() temporarily blocks processing of 
> the current event and starts a nested event loop to handle other events, 
> the iteration of the subscribers does not terminate. As a result, 
> subsequent events are unable to enter the iteration and be dispatched 
> to their respective subscribers as these events are held up in the 
> queue

So, the nested event loop started by `Stage#showAndWait()` does not bother dispatching our events such as `ShowHelpRequestEvent`, `NewResultAvailableEvent` etc. However, it must be dispatching some kind of events, otherwise our UI would be completely frozen and we would not, for example, be able to press enter in the command box and run the command (Callbacks like `CommandBox#handleCommandInputChanged()` still get called). So, why does `Stage#showAndWait()` choose to dispatch events to callbacks such as `CommandBox#handleCommandInputChanged()`, and not the subscribers of `ShowHelpRequestEvent`, `NewResultAvailableEvent` etc.? In other words, what are those "other events" mentioned in your commit message, and what makes them so special compared to `ShowHelpRequestEvent`, `NewResultAvailableEvent` etc.?

pyokagan:
Yes, as @Zhiyuan-Amos mentioned, the commit message (2nd and 3rd paragraphs) contains implementation details which are not important at the level of abstraction that we are looking at this issue from, and thus could be made more succinct.

Also, in the commit message:
> https://docs.oracle.com/javacard/3.0.5/api/javacard/framework/service/Dispatcher.html

!???

vivekscl:
@pyokagan 
>So, why does Stage#showAndWait() choose to dispatch events to callbacks such as CommandBox#handleCommandInputChanged(), and not the subscribers of ShowHelpRequestEvent, NewResultAvailableEvent

Only those with the events `@Subscribe` annotation are part of the subscriber queue that is being held up.The `CommandBox#handleCommandInputChanged()` does not use the eventBus to post events and as such, they do not enter the subscriber queue. If they do not enter the queue, they won't be held up and will continue to work.

deepinscreenshot_select-area_20180202230727


deepinscreenshot_select-area_20180202230752


>Also, in the commit message:
https://docs.oracle.com/javacard/3.0.5/api/javacard/framework/service/Dispatcher.html
!???
 
Oh I thought that might be a useful reference if needed, but maybe it isn't since the implementation details are not that important :sweat_smile:.

pyokagan:
@vivekscl 
> The CommandBox#handleCommandInputChanged() does not use the eventBus to post events and as such, they do not enter the subscriber queue. If they do not enter the queue, they won't be held up and will continue to work.

Yes, that's the important point: We have *two* event systems in the code base -- the JavaFX event system, and the Guava EventBus system. `ShowHelpRequestEvent`, `NewResultAvailableEvent` etc are *EventBus events*, whereas `CommandBox#handleCommandInputChanged()` is called on a *JavaFX event*. These systems are not aware of each other, so it makes sense that when `Stage#showAndWait()` starts a nested event loop, it will only dispatch *JavaFX events*.

So, your commit message will need to be very clear whether it is talking about *EventBus events* or *JavaFX events*.

The second piece of the puzzle (Factor B) is that the default `Dispatcher` used by EventBus, `Dispatcher.perThreadDispatchQueue()`, *does not allow nested callback invocations*, and instead chooses to queue them to be invoked after the current invoked callback completes. This could be fixed by using the `Dispatcher.immediate()` dispatcher, for example.

> Oh I thought that might be a useful reference if needed, but maybe it isn't since the implementation details are not that important 😅.

No, the documentation linked is not Guava docs. I think you are looking for https://github.com/google/guava/blob/master/guava/src/com/google/common/eventbus/Dispatcher.java

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +73,24 @@ private void assertHelpWindowNotOpen() {
...
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box displays the success message of executing {@code ListCommand}.<br>
+     * 4. {@code Model}, {@code Storage} and {@code PersonListPanel} equal to the corresponding components in
+     * the current model.<br>
+     * 5. Browser url and selected card remain unchanged.<br>
+     * 6. Status bar remains unchanged.<br>
+     * Verifications 1, 3 and 4 are performed by
+     * {@code AddressBookSystemTest#assertApplicationDisplaysExpected(String, String, Model)}.<br>
+     */
+    private void assertListCommandSuccess(String command, Model expectedModel, String expectedResultMessage) {
```
Zhiyuan-Amos: For a more thorough testing, we should ensure that `BrowserPanel`, `StatusBar` and `PersonListPanel` are updated too. `ListCommand` doesn't actually verify that `PersonListPanel` is updated because it wasn't filtered beforehand :P Perhaps we need to execute multiple commands, such as `Delete 1` and `Select 1`.

We don't need to verify the `model` though. We just need to verify that the UI components are updated. 

For e.g.
```
assertEquals("", commandBoxHandle.getInput());
assertFalse(resultDisplayHandle.getText().isEmpty());
assertFalse(browserPanelHandle.getLoadedUrl().equals(BrowserPanel#DEFAULT_PAGE));
...
```

vivekscl:
>There's no need for line break for the hyperlink of Guava Dispatcher.

@Zhiyuan-Amos but shouldn't the commit message wrap at 72 characters? Without the line break, it exceeds 72 characters. Or does this not apply to hyperlinks?

>4th para: ~~To fix this~~... it is sufficient

There's a `let's replace....` after `To fix this` so replacing it with `it is sufficient` would sound off right?

Zhiyuan-Amos:
> There's a let's replace.... after To fix this so replacing it with it is sufficient would sound off right?

Sorry, I meant:

4th para: ~~to fix this~~
5th para: ... in this case as it is ~~suffice~~ sufficient to show...

vivekscl:
@Zhiyuan-Amos and leave the hyperlink as is or change it? :sweat_smile:

Zhiyuan-Amos:
I don't think we wrap because if you wrap it, the hyperlink can't be clicked. If I'm not wrong, the line break after 72 characters is for [readability](https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting), but we don't need to read the html link. Sorry for missing this out!

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +52,11 @@ public void openHelpWindow() {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        //use command box to open help window and then assert that the UI has been updated
+        executeCommand(HelpCommand.COMMAND_WORD);
```
yamgent: We already executed a `help` command earlier, do we need to execute it again?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
...
+     * After execution of multiple commands, asserts that the,<br>
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. BrowserPanel does not show the default page.<br>
+     * 5. Status bar sync status changes.<br>
+     * 6.{@code PersonListPanel} equals to the corresponding components in
+     * the current model.<br>
+     */
+    private void assertUiUpdated() {
```
yamgent: The method name `assertUiUpdated()` doesn't sound right to me. :P Maybe `assertUiUpdatesWhileHelpWindowOpened()`?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
...
 
+    /**
+     * After execution of multiple commands, asserts that the,<br>
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. BrowserPanel does not show the default page.<br>
+     * 5. Status bar sync status changes.<br>
+     * 6.{@code PersonListPanel} equals to the corresponding components in
+     * the current model.<br>
```
Zhiyuan-Amos: is there a reason for wrapping it this way?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
...
+     */
+    private void assertUiUpdated() {
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
```
Zhiyuan-Amos: we should leave comments why we execute both of these commands: to ensure that all ui parts update.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
...
     }
 
+    /**
+     * After execution of multiple commands, asserts that the,<br>
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. BrowserPanel does not show the default page.<br>
+     * 5. Status bar sync status changes.<br>
+     * 6.{@code PersonListPanel} equals to the corresponding components in
```
Zhiyuan-Amos: is there a reason why you use {@code PersonListPanel}, but not this format for the rest of the components? :P

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
...
+     * the current model.<br>
+     */
+    private void assertUiUpdated() {
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertEquals("", getCommandBox().getInput());
+        assertCommandBoxShowsDefaultStyle();
+        assertFalse(getResultDisplay().getText().isEmpty());
+        assertFalse(getBrowserPanel().getLoadedUrl().equals(BrowserPanel.DEFAULT_PAGE));
+        assertStatusBarUnchangedExceptSyncStatus();
```
Zhiyuan-Amos: hmm we should check that `getSyncStatus()` does not return `SYNC_STATUS_INITIAL (Not updated yet in this session)`

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +77,25 @@ private void assertHelpWindowNotOpen() {
         assertFalse(ERROR_MESSAGE, HelpWindowHandle.isWindowPresent());
     }
 
+    /**
+     * After execution of multiple commands, asserts that the,<br>
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. BrowserPanel does not show the default page.<br>
```
Zhiyuan-Amos: should have a space: Browser panel.

src/test/java/systemtests/HelpCommandSystemTest.java
vivekscl: The help window from the previous execution of the `help` command was closed due to `assertHelpWindowOpen()`. So we need to execute it again to open it and test whether the UI has been updated right?

src/test/java/systemtests/HelpCommandSystemTest.java
vivekscl: Oh, my bad. I'll change it back to `Person list panel` instead

src/test/java/systemtests/HelpCommandSystemTest.java
vivekscl: Nope! Will fix it to be on the same line.

src/test/java/systemtests/HelpCommandSystemTest.java
vivekscl: But isn't done by `assertStatusBarUnchangedExceptSyncStatus()`?

Zhiyuan-Amos: We are doing a looser check for all the tests here because we just need to ensure that the UI components are updated.

`assertStatusBarUnchangedExceptSyncStatus()` looks too complex; why should we concern about which part of StatusBar is changed? As long as it doesn't display the original value, it's fine.

This helps us to lower the dependencies we have with all these commands. Suppose `SelectCommand` now updates the `saveLocation` too (for some weird reason). Then, this test will fail.

src/test/java/systemtests/HelpCommandSystemTest.java
yamgent: Ah okay, my bad, I didn't notice that.

Zhiyuan-Amos: @yamgent actually this means we shouldn't have designed the API this way because there's a secret behavior that we can't tell from reading the method name `assertHelpWindowOpen` :P

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +78,24 @@ private void assertHelpWindowNotOpen() {
         assertFalse(ERROR_MESSAGE, HelpWindowHandle.isWindowPresent());
     }
 
+    /**
+     * Executes multiple commands to ensure that all UI components update and then asserts that the,<br>
```
Zhiyuan-Amos: Executes multiple commands to ensure that all UI components update and then asserts... -> Executes multiple commands to update all UI components. To verify that all UI components are updated, we assert that...

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +78,24 @@ private void assertHelpWindowNotOpen() {
...
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. Browser panel does not show the default page.<br>
+     * 5. Status bar updates.<br>
+     * 6. Person list panel equals to the corresponding components in the current model.<br>
```
Zhiyuan-Amos: Not sure whether there's a need for line break at the last line :P

src/test/java/systemtests/HelpCommandSystemTest.java
yamgent: @Zhiyuan-Amos You are right. :P We can rectify that in a different PR.

vivekscl:
@damithc for your review.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,11 @@ public void openHelpWindow() {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        //use command box to open help window and then assert that the UI has been updated
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+        assertUiUpdatesWhileHelpWindowOpened();
```
damithc: Not sure if we should extract this as a method; seems like suddenly we are going for a higher level abstraction (compared to the rest of the method) at the last step.

vivekscl: Do you mean we should extract these 3 lines into one method?

Zhiyuan-Amos: I think prof meant we should inline everything :P

vivekscl: Oh, so take out all the code in `assertUiUpdatesWhileHelpWindowOpened()` and place it outside?

Zhiyuan-Amos: Yup, but now having taken a 2nd look, @damithc I think it's gonna look quite messy if we inline everything here :O

vivekscl: I do agree with @Zhiyuan-Amos, it would become really cluttered.

damithc: If you must extract out lines, it makes more sense to extract out lines 93-98 as they are all parts of one bigger thing.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,11 @@ public void openHelpWindow() {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        //use command box to open help window and then assert that the UI has been updated
```
damithc: This comment is missing something. Why should the UI be updated when the help window is open?

vivekscl: Should it be:
`use command box to open help window and assert that other UI components update so as to verify that the help window does not block subsequent events`

yamgent: "Use command box to open help window" doesn't seem to be require mentioning :P.

Maybe just `assert that an open help window does not block subsequent UI events`?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,13 @@ public void openHelpWindow() {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        //assert that an open help window does not block subsequent UI events
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
```
Zhiyuan-Amos: there should be comments to mention that the execution of these 2 commands are to ensure that all UI components are updated. So maybe something like:

```
//assert that an open help window does not block subsequent UI events.
//the execution of SelectCommand and DeleteCommand should update all UI components.
```

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -66,4 +81,22 @@ private void assertHelpWindowNotOpen() {
...
+    /**
+     * To verify that all UI components are updated, we assert that the,<br>
+     * 1. Command box displays an empty string.<br>
+     * 2. Command box has the default style class.<br>
+     * 3. Result display box is not empty.<br>
+     * 4. Browser panel does not show the default page.<br>
+     * 5. Status bar updates.<br>
+     * 6. Person list panel equals to the corresponding components in the current model.
+     */
+    private void assertUiUpdatesWhileHelpWindowOpened() {
```
damithc: I still can't get my head around this abstraction. For example, the code has nothing to do with the Help window, yet the method name mentions Help window.

On the other hand, I can live with a method that executes a single command and ensures the UI is updated as expected by the command. That method makes sense on it's own. The code can read like this:
```java
// open help window and give it focus
executeCommand(HelpCommand.COMMAND_WORD);
getMainWindowHandle().focus();

//ensure the UI responds to command executions while the help window is open
veryUiUpdatesForCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased())
veryUiUpdatesForCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased())
```
Would that be better?

Also, why do we execute two commands here? Isn't one enough?

Zhiyuan-Amos: We need 2 commands because the `SelectCommand` doesn't update the `StatusBarFooter`, but it updates the `BrowserPanel`, while the `DeleteCommand` doesn't update the `BrowserPanel`, but it updates the `StatusBarFooter`. We don't have any commands that updates all UI components upon execution.

How about if we extract everything into a method `openHelpWindowAndAssertsUiUpdates`?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,14 @@ public void openHelpWindow() {
...
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        //assert that an open help window does not block subsequent UI events.
+        //the execution of SelectCommand and DeleteCommand should update all UI components.
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertUiUpdatesWhileHelpWindowOpened();
```
damithc: Something of this sort?

```java
// assert that while the help Window is open the UI updates correctly for a command execution
executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
assertEquals("", getCommandBox().getInput());
assertCommandBoxShowsDefaultStyle();
assertFalse(getResultDisplay().getText().isEmpty());
assertFalse(getBrowserPanel().getLoadedUrl().equals(BrowserPanel.DEFAULT_PAGE));
assertListMatching(getPersonListPanel(), getModel().getFilteredPersonList());

// assert that status bar too is updated correctly while the help window is open
// note: the select command tested above does not update the status bar
executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
assertFalse(getStatusBarFooter().getSyncStatus().equals(StatusBarFooter.SYNC_STATUS_INITIAL));
```

I'm favoring the code to be here because it's easier to tell the true story that way. Extracting it to a method gives a superficial neatness but doesn't help the reader.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -1,13 +1,20 @@
...
 import guitests.GuiRobot;
 import guitests.guihandles.HelpWindowHandle;
+import seedu.address.logic.commands.DeleteCommand;
 import seedu.address.logic.commands.HelpCommand;
+import seedu.address.logic.commands.SelectCommand;
+import seedu.address.ui.BrowserPanel;
+import seedu.address.ui.StatusBarFooter;
 
 /**
  * TODO: This test is incomplete as it is missing test cases.
```
eugenepeh: Should the header comment be updated?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,24 @@ public void openHelpWindow() {
...
+        assertCommandBoxShowsDefaultStyle();
+        assertFalse(getResultDisplay().getText().isEmpty());
+        assertFalse(getBrowserPanel().getLoadedUrl().equals(BrowserPanel.DEFAULT_PAGE));
+        assertListMatching(getPersonListPanel(), getModel().getFilteredPersonList());
+
+        // assert that status bar too is updated correctly while the help window is open
+        // note: the select command tested above does not update the status bar
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertFalse(getStatusBarFooter().getSyncStatus().equals(StatusBarFooter.SYNC_STATUS_INITIAL));
+
```
Zhiyuan-Amos: unnecessary new line here

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,24 @@ public void openHelpWindow() {
         //use command box
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
+
+        // open help window and give it focus
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+
+        // assert that while the help Window is open the UI updates correctly for a command execution
```
Zhiyuan-Amos: small 'w' for `Window` :P

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: Seems like it. 👍

vivekscl: Aren't there more test cases to be added later on, making this test still incomplete?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +53,24 @@ public void openHelpWindow() {
...
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertEquals("", getCommandBox().getInput());
+        assertCommandBoxShowsDefaultStyle();
+        assertFalse(getResultDisplay().getText().isEmpty());
```
yamgent: Is it sufficient to check that the `ResultDisplay` is not empty? I don't think the `ResultDisplay` before the `select` command was empty as well, am I right?

src/test/java/systemtests/HelpCommandSystemTest.java
eugenepeh: I think previously it is incomplete in the sense where no interaction between others components were added. Since now there is something, we can fill up the class header comments and other TODO should be within the class itself.

src/test/java/systemtests/HelpCommandSystemTest.java
vivekscl: Ah that makes sense. I guess it would be better to check if the text in the `ResultDisplay` is equal to the result of the `select` command instead.

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: Yes, what Eugene says seems right: Previously we only executed `HelpCommand` on its own in isolation from other commands and we did not verify how the other components interacted with `HelpCommand`.

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: I would prefer to verify that it's not the `HelpCommand`'s result message. That way, we have slightly lesser dependency because we do not need to care about what is the result message of `SelectCommand`.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +54,25 @@ public void openHelpWindow() {
...
+        assertNotEquals(getResultDisplay().getText(), HelpCommand.SHOWING_HELP_MESSAGE);
+        assertFalse(getBrowserPanel().getLoadedUrl().equals(BrowserPanel.DEFAULT_PAGE));
+        assertListMatching(getPersonListPanel(), getModel().getFilteredPersonList());
+
+        // assert that the status bar too is updated correctly while the help window is open
+        // note: the select command tested above does not update the status bar
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertFalse(getStatusBarFooter().getSyncStatus().equals(StatusBarFooter.SYNC_STATUS_INITIAL));
+
+        //TODO: More tests to be added.
```
Zhiyuan-Amos: I wonder what other meaningful tests can we write though... :X

vivekscl: I think we can still test with other commands. For instance, after exiting the AB4, the help window still remains open. We can still test for that case.
```java
executeCommand(HelpCommand.COMMAND_WORD);
executeCommand(ExitCommand.COMMAND_WORD);
assertHelpWindowOpen();
```

Zhiyuan-Amos: LOL that's an interesting interaction. This doesn't seem like what we want though. Can you raise an issue for it?

vivekscl: >LOL that's an interesting interaction

Hahaha, I thought the same too. Wasn't sure if it was intended or bug at the time though. But yes, I'll raise an issue for it:+1:.

>Can you fix that after this PR? ^^

Fix the help window issue or the additional test cases? :sweat_smile:

vivekscl: >executeCommand(HelpCommand.COMMAND_WORD);
executeCommand(ExitCommand.COMMAND_WORD);
assertHelpWindowOpen();

I did verify multiple times, the test mentioned above, actually passed in my case. So maybe its an operating system thing? I'm currently running Linux.

Zhiyuan-Amos: Hmm yup perhaps. I'm on MacOS. :)

vivekscl: So it does close but only when it is not in fullscreen mode. Since my default resolution was fullscreen I kept verifying it that it didn't close. But once I changed the resolution to something smaller, the help window closes as expected.

So please do ignore my previous comment :sweat_smile:.

yamgent:
```java
//TODO: More tests to be added.
```

As per discussion above we could create issues for it. The comment doesn't seem to be really informative otherwise :P (and I am of the view that TODOs comments should only exist in the repo if there's a serious flaw in the codebase).

Zhiyuan-Amos:
@yamgent for your approval :)

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +54,23 @@ public void openHelpWindow() {
...
+
+        // open help window and give it focus
+        executeCommand(HelpCommand.COMMAND_WORD);
+        getMainWindowHandle().focus();
+
+        // assert that while the help window is open the UI updates correctly for a command execution
+        executeCommand(SelectCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertEquals("", getCommandBox().getInput());
+        assertCommandBoxShowsDefaultStyle();
+        assertNotEquals(getResultDisplay().getText(), HelpCommand.SHOWING_HELP_MESSAGE);
```
pyokagan: Should the first argument be `HelpCommand.SHOWING_HELP_MESSAGE` instead, since `getResultDisplay().getText()` is the "actual" value that we are testing?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +54,23 @@ public void openHelpWindow() {
...
+        assertFalse(getBrowserPanel().getLoadedUrl().equals(BrowserPanel.DEFAULT_PAGE));
```
pyokagan: Why not use `assertNotEquals()`?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -46,6 +54,23 @@ public void openHelpWindow() {
...
+        assertListMatching(getPersonListPanel(), getModel().getFilteredPersonList());
+
+        // assert that the status bar too is updated correctly while the help window is open
+        // note: the select command tested above does not update the status bar
+        executeCommand(DeleteCommand.COMMAND_WORD + " " + INDEX_FIRST_PERSON.getOneBased());
+        assertFalse(getStatusBarFooter().getSyncStatus().equals(StatusBarFooter.SYNC_STATUS_INITIAL));
```
pyokagan: Why not use `assertNotEquals()`?

damithc:
@Zhiyuan-Amos let's merge this as we have enough reviews and we need to release level 4 for this sem soon.

PR #818, opened 2018-02-04
PersonListPanel#personListView: revert type change #796

### LINKED ISSUE #796: Performance regression when the model filter is modified
Bisected to 75cbfa3804ff365646e903514282d9313da4300e 

Reproduction:

1. Create 9999 persons in the address book
2. Launch AB-4, which will show the unfiltered list by default.
3. Run: find X, where X is the exact name of a person, such that the list only displays 1 person.
4. AB-4 takes a long time to run the command, and the UI completely freezes for around 25s which is way above a usable response time. Prior to the aforementioned commit it is fast enough.

A few guesses:

1. EasyBind.map is not implemented in a performant way, or
2. 75cbfa3804ff365646e903514282d9313da4300e has the effect of creating 9999 scene nodes inside the list box. JavaFX does not perform well with huge scene graphs, especially when lots of scene nodes are removed at once.

Quick script to generate a 9999 person addressbook:
```python
print('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>')
print('<addressbook>')
for i in range(9999):
    print('<persons>')
    print('<name>{}</name>'.format(''.join('ABCDEFGHIJ'[int(x)] for x in str(i))))
    print('<phone>123</phone>')
    print('<email>a@a</email>')
    print('<address>a</address>')
    print('</persons>')
print('</addressbook>')
```

Zhiyuan-Amos:
Fixes #796.

```
personListView is of type ListView<PersonCard>. We use EasyBind library
to map the Person objects to PersonCard objects. The mapping / applying
of Function<? super F, ? extends E> occurs only when either get(int) or
getRemoved() methods are called[1]; the creation of PersonCard objects
occur only when either of these methods are called.

The ClearCommand deletes all Person objects. When we have large number
of persons and getRemoved() is called, all the corresponding cards
will be created prior to them being removed. As creation of PersonCard
objects are costly, the app slows down tremendously.

Let's revert ListView<PersonCard> to ListView<Person>.
PersonListViewCell will create the PersonCard object. According to
javadoc for Cell that "it is not feasible to create an actual Cell for
every single item in the control. We represent extremely large data
sets using only very few Cells. Each Cell is "recycled", or reused."[2]
A Cell is created for every row in the ListView that is displayed on
the scene i.e. that is visible to the user, and additional Cells may be
created for efficient scroll handling.[3] This implies that there will
only be about 10 Cells at any point in time since the default screen
size is only large enough to display 10 Cells. Therefore, there will
only be about 10 PersonCard objects at any point. Creation of 10
PersonCard objects is performed in negligible time, thus the app no
longer slows down. This commit is performed by reverting from
75cbfa3.

Also, since PersonCard objects are only created when the card is
displayed on the list view, let's teach
GuiTestAssert#assertListMatching(PersonListPanelHandle, Person...) and
AddressBookSystemTest#assertSelectedCardChange(Index) to navigate to
the card that they are verifying before performing the verification.

[1]: https://github.com/TomasMikula/EasyBind/blob/master/src/main/java/org/fxmisc/easybind/MappedList.java
[2]: https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Cell.html
[3]: https://stackoverflow.com/questions/18159303/listview-in-javafx-adds-multiple-cells
```

### DIFF
--- a/src/test/java/seedu/address/ui/PersonListPanelTest.java
+++ b/src/test/java/seedu/address/ui/PersonListPanelTest.java
@@ -1,21 +1,31 @@
 package seedu.address.ui;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 import static seedu.address.testutil.EventsUtil.postNow;
 import static seedu.address.testutil.TypicalIndexes.INDEX_SECOND_PERSON;
 import static seedu.address.testutil.TypicalPersons.getTypicalPersons;
 import static seedu.address.ui.testutil.GuiTestAssert.assertCardDisplaysPerson;
 import static seedu.address.ui.testutil.GuiTestAssert.assertCardEquals;
 
-import org.junit.Before;
+import java.io.File;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
+
 import org.junit.Test;
+
+import com.google.common.base.Stopwatch;
 
 import guitests.guihandles.PersonCardHandle;
 import guitests.guihandles.PersonListPanelHandle;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
+import seedu.address.commons.core.LogsCenter;
 import seedu.address.commons.events.ui.JumpToListRequestEvent;
+import seedu.address.commons.util.FileUtil;
+import seedu.address.commons.util.XmlUtil;
 import seedu.address.model.person.Person;
+import seedu.address.storage.XmlSerializableAddressBook;
 
 public class PersonListPanelTest extends GuiUnitTest {
     private static final ObservableList<Person> TYPICAL_PERSONS =
@@ -23,19 +33,20 @@ public class PersonListPanelTest extends GuiUnitTest {
 
     private static final JumpToListRequestEvent JUMP_TO_SECOND_EVENT = new JumpToListRequestEvent(INDEX_SECOND_PERSON);
 
+    private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/sandbox/");
+
+    private static final Logger logger = LogsCenter.getLogger(PersonListPanelTest.class);
+
+    private static final int TEST_TIMEOUT = 8000;
+    private static final int CARD_CREATION_AND_DELETION_TIMEOUT = 2500;
+    private static final int LIST_CREATION_TIMEOUT = TEST_TIMEOUT - CARD_CREATION_AND_DELETION_TIMEOUT;
+
     private PersonListPanelHandle personListPanelHandle;
-
-    @Before
-    public void setUp() {
-        PersonListPanel personListPanel = new PersonListPanel(TYPICAL_PERSONS);
-        uiPartRule.setUiPart(personListPanel);
-
-        personListPanelHandle = new PersonListPanelHandle(getChildNode(personListPanel.getRoot(),
-                PersonListPanelHandle.PERSON_LIST_VIEW_ID));
-    }
 
     @Test
     public void display() {
+        initUi(TYPICAL_PERSONS);
+
         for (int i = 0; i < TYPICAL_PERSONS.size(); i++) {
             personListPanelHandle.navigateToCard(TYPICAL_PERSONS.get(i));
             Person expectedPerson = TYPICAL_PERSONS.get(i);
@@ -48,11 +59,100 @@ public class PersonListPanelTest extends GuiUnitTest {
 
     @Test
     public void handleJumpToListRequestEvent() {
+        initUi(TYPICAL_PERSONS);
         postNow(JUMP_TO_SECOND_EVENT);
         guiRobot.pauseForHuman();
 
-        PersonCardHandle expectedCard = personListPanelHandle.getPersonCardHandle(INDEX_SECOND_PERSON.getZeroBased());
-        PersonCardHandle selectedCard = personListPanelHandle.getHandleToSelectedCard();
-        assertCardEquals(expectedCard, selectedCard);
+        PersonCardHandle expectedPerson = personListPanelHandle.getPersonCardHandle(INDEX_SECOND_PERSON.getZeroBased());
+        PersonCardHandle selectedPerson = personListPanelHandle.getHandleToSelectedCard();
+        assertCardEquals(expectedPerson, selectedPerson);
+    }
+
+    /**
+     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
+     * to execute.
+     * <ol>
+     *     <li>If {@code TestTimedOutException} is thrown before the logs are printed, then the failure occurs during
+     *         the preparation of the test (creating the xml file containing the large number of persons).</li>
+     *     <li>If {@code TestTimedOutException} is thrown after the logs are printed, then the failure occurs during
+     *         the creation and deletion of the person cards.</li>
+     * </ol>
+     */
+    // The timeout makes the test fail fast. Without it, the test can stall for a long time if
+    // guiRobot#interact(Runnable) takes very long to complete execution because guiRobot#interact(Runnable) only
+    // returns after the Runnable completes execution.
+    @Test(timeout = TEST_TIMEOUT)
+    public void performanceTest() throws Exception {
+        ObservableList<Person> backingList = createBackingList(10000);
+
+        Stopwatch stopwatch = Stopwatch.createStarted();
+        initUi(backingList);
+        guiRobot.interact(backingList::clear);
+        if (stopwatch.elapsed(TimeUnit.MILLISECONDS) >= CARD_CREATION_AND_DELETION_TIMEOUT) {
+            fail("Creation and deletion of person cards exceeded time limit");
+        }
+    }
+
+    /**
+     * Returns a list of persons containing {@code personCount} persons that is used to populate the
+     * {@code PersonListPanel}. Logs the time taken for method execution if it finishes before
+     * {@code LIST_CREATION_TIMEOUT}.
+     *
+     * @throws AssertionError if this method takes too long to execute.
+     */
+    private ObservableList<Person> createBackingList(int personCount) throws Exception {
+        Stopwatch stopwatch = Stopwatch.createStarted();
+
+        File xmlFile = createXmlFileWithPersons(personCount);
+        XmlSerializableAddressBook xmlAddressBook =
+                XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
+        ObservableList<Person> personList =
+                FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
+
+        long createListTime = stopwatch.elapsed(TimeUnit.MILLISECONDS);
+        String createListMessage = "List creation took: " + createListTime + "ms. ";
+        if (createListTime >= LIST_CREATION_TIMEOUT) {
+            fail(createListMessage + "Time limit exceeded.");
+        } else {
+            logger.info(createListMessage + "Continuing test.");
+        }
+
+        return personList;
+    }
+
+    /**
+     * Returns a .xml file containing {@code personCount} persons. This file will be deleted when the JVM terminates.
+     */
+    private File createXmlFileWithPersons(int personCount) throws Exception {
+        StringBuilder builder = new StringBuilder();
+        builder.append("<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?>\n");
+        builder.append("<addressbook>\n");
+        for (int i = 0; i < personCount; i++) {
+            builder.append("<persons>\n");
+            builder.append("<name>a</name>\n");
+            builder.append("<phone>").append(i).append("00</phone>\n");
+            builder.append("<email>a@aa</email>\n");
+            builder.append("<address>a</address>\n");
+            builder.append("</persons>\n");
+        }
+        builder.append("</addressbook>\n");
+
+        File manyPersonsFile = new File(TEST_DATA_FOLDER + "manyPersons.xml");
+        FileUtil.createFile(manyPersonsFile);
+        FileUtil.writeToFile(manyPersonsFile, builder.toString());
+        manyPersonsFile.deleteOnExit();
+        return manyPersonsFile;
+    }
+
+    /**
+     * Initializes {@code personListPanelHandle} with a {@code PersonListPanel} backed by {@code backingList}.
+     * Also shows the {@code Stage} that displays only {@code PersonListPanel}.
+     */
+    private void initUi(ObservableList<Person> backingList) {
+        PersonListPanel personListPanel = new PersonListPanel(backingList);
+        uiPartRule.setUiPart(personListPanel);
+
+        personListPanelHandle = new PersonListPanelHandle(getChildNode(personListPanel.getRoot(),
+                PersonListPanelHandle.PERSON_LIST_VIEW_ID));
     }
 }
--- a/src/test/java/systemtests/AddressBookSystemTest.java
+++ b/src/test/java/systemtests/AddressBookSystemTest.java
@@ -211,6 +211,7 @@ public abstract class AddressBookSystemTest {
      * @see PersonListPanelHandle#isSelectedPersonCardChanged()
      */
     protected void assertSelectedCardChanged(Index expectedSelectedCardIndex) {
+        getPersonListPanel().navigateToCard(getPersonListPanel().getSelectedCardIndex());
         String selectedCardName = getPersonListPanel().getHandleToSelectedCard().getName();
         URL expectedUrl;
         try {
--- a/README.adoc
+++ b/README.adoc
@@ -35,6 +35,6 @@ endif::[]
 
 * Some parts of this sample application were inspired by the excellent http://code.makery.ch/library/javafx-8-tutorial/[Java FX tutorial] by
 _Marco Jakob_.
-* Libraries used: https://github.com/TomasMikula/EasyBind[EasyBind], https://github.com/TestFX/TestFX[TextFX], https://bitbucket.org/controlsfx/controlsfx/[ControlsFX], https://github.com/FasterXML/jackson[Jackson], https://github.com/google/guava[Guava], https://github.com/junit-team/junit4[JUnit4]
+* Libraries used: https://github.com/TestFX/TestFX[TextFX], https://bitbucket.org/controlsfx/controlsfx/[ControlsFX], https://github.com/FasterXML/jackson[Jackson], https://github.com/google/guava[Guava], https://github.com/junit-team/junit4[JUnit4]
 
 == Licence : link:LICENSE[MIT]
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,6 @@ jacocoTestReport {
 dependencies {
     String testFxVersion = '4.0.12-alpha'
 
-    compile group: 'org.fxmisc.easybind', name: 'easybind', version: '1.0.3'
     compile group: 'org.controlsfx', name: 'controlsfx', version: '8.40.11'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.7.0'
     compile group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.7.4'
--- a/src/main/java/seedu/address/commons/events/ui/PersonPanelSelectionChangedEvent.java
+++ b/src/main/java/seedu/address/commons/events/ui/PersonPanelSelectionChangedEvent.java
@@ -1,7 +1,7 @@
 package seedu.address.commons.events.ui;
 
 import seedu.address.commons.events.BaseEvent;
-import seedu.address.ui.PersonCard;
+import seedu.address.model.person.Person;
 
 /**
  * Represents a selection change in the Person List Panel
@@ -9,9 +9,9 @@ import seedu.address.ui.PersonCard;
 public class PersonPanelSelectionChangedEvent extends BaseEvent {
 
 
-    private final PersonCard newSelection;
+    private final Person newSelection;
 
-    public PersonPanelSelectionChangedEvent(PersonCard newSelection) {
+    public PersonPanelSelectionChangedEvent(Person newSelection) {
         this.newSelection = newSelection;
     }
 
@@ -20,7 +20,7 @@ public class PersonPanelSelectionChangedEvent extends BaseEvent {
         return this.getClass().getSimpleName();
     }
 
-    public PersonCard getNewSelection() {
+    public Person getNewSelection() {
         return newSelection;
     }
 }
--- a/src/main/java/seedu/address/ui/BrowserPanel.java
+++ b/src/main/java/seedu/address/ui/BrowserPanel.java
@@ -67,6 +67,6 @@ public class BrowserPanel extends UiPart<Region> {
     @Subscribe
     private void handlePersonPanelSelectionChangedEvent(PersonPanelSelectionChangedEvent event) {
         logger.info(LogsCenter.getEventHandlingLogMessage(event));
-        loadPersonPage(event.getNewSelection().person);
+        loadPersonPage(event.getNewSelection());
     }
 }
--- a/src/main/java/seedu/address/ui/PersonListPanel.java
+++ b/src/main/java/seedu/address/ui/PersonListPanel.java
@@ -2,8 +2,6 @@ package seedu.address.ui;
 
 import java.util.logging.Logger;
 
-import org.fxmisc.easybind.EasyBind;
-
 import com.google.common.eventbus.Subscribe;
 
 import javafx.application.Platform;
@@ -25,7 +23,7 @@ public class PersonListPanel extends UiPart<Region> {
     private final Logger logger = LogsCenter.getLogger(PersonListPanel.class);
 
     @FXML
-    private ListView<PersonCard> personListView;
+    private ListView<Person> personListView;
 
     public PersonListPanel(ObservableList<Person> personList) {
         super(FXML);
@@ -34,9 +32,7 @@ public class PersonListPanel extends UiPart<Region> {
     }
 
     private void setConnections(ObservableList<Person> personList) {
-        ObservableList<PersonCard> mappedList = EasyBind.map(
-                personList, (person) -> new PersonCard(person, personList.indexOf(person) + 1));
-        personListView.setItems(mappedList);
+        personListView.setItems(personList);
         personListView.setCellFactory(listView -> new PersonListViewCell());
         setEventHandlerForSelectionChangeEvent();
     }
@@ -68,19 +64,18 @@ public class PersonListPanel extends UiPart<Region> {
     }
 
     /**
-     * Custom {@code ListCell} that displays the graphics of a {@code PersonCard}.
+     * Custom {@code ListCell} that displays the graphics of a {@code Person} using a {@code PersonCard}.
      */
-    class PersonListViewCell extends ListCell<PersonCard> {
-
+    class PersonListViewCell extends ListCell<Person> {
         @Override
-        protected void updateItem(PersonCard person, boolean empty) {
+        protected void updateItem(Person person, boolean empty) {
             super.updateItem(person, empty);
 
             if (empty || person == null) {
                 setGraphic(null);
                 setText(null);
             } else {
-                setGraphic(person.getRoot());
+                setGraphic(new PersonCard(person, getIndex() + 1).getRoot());
             }
         }
     }
--- a/src/test/java/guitests/guihandles/PersonCardHandle.java
+++ b/src/test/java/guitests/guihandles/PersonCardHandle.java
@@ -6,6 +6,7 @@ import java.util.stream.Collectors;
 import javafx.scene.Node;
 import javafx.scene.control.Label;
 import javafx.scene.layout.Region;
+import seedu.address.model.person.Person;
 
 /**
  * Provides a handle to a person card in the person list panel.
@@ -67,5 +68,15 @@ public class PersonCardHandle extends NodeHandle<Node> {
                 .stream()
                 .map(Label::getText)
                 .collect(Collectors.toList());
+    }
+
+    /**
+     * Returns true if this handle contains {@code person}.
+     */
+    public boolean equals(Person person) {
+        return getName().equals(person.getName().fullName)
+                && getAddress().equals(person.getAddress().value)
+                && getPhone().equals(person.getPhone().value)
+                && getEmail().equals(person.getEmail().value);
     }
 }
--- a/src/test/java/guitests/guihandles/PersonListPanelHandle.java
+++ b/src/test/java/guitests/guihandles/PersonListPanelHandle.java
@@ -2,20 +2,23 @@ package guitests.guihandles;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
+import javafx.scene.Node;
 import javafx.scene.control.ListView;
 import seedu.address.model.person.Person;
-import seedu.address.ui.PersonCard;
 
 /**
  * Provides a handle for {@code PersonListPanel} containing the list of {@code PersonCard}.
  */
-public class PersonListPanelHandle extends NodeHandle<ListView<PersonCard>> {
+public class PersonListPanelHandle extends NodeHandle<ListView<Person>> {
     public static final String PERSON_LIST_VIEW_ID = "#personListView";
 
-    private Optional<PersonCard> lastRememberedSelectedPersonCard;
+    private static final String CARD_PANE_ID = "#cardPane";
 
-    public PersonListPanelHandle(ListView<PersonCard> personListPanelNode) {
+    private Optional<Person> lastRememberedSelectedPersonCard;
+
+    public PersonListPanelHandle(ListView<Person> personListPanelNode) {
         super(personListPanelNode);
     }
 
@@ -23,15 +26,20 @@ public class PersonListPanelHandle extends NodeHandle<ListView<PersonCard>> {
      * Returns a handle to the selected {@code PersonCardHandle}.
      * A maximum of 1 item can be selected at any time.
      * @throws AssertionError if no card is selected, or more than 1 card is selected.
+     * @throws IllegalStateException if the selected card is currently not in the scene graph.
      */
     public PersonCardHandle getHandleToSelectedCard() {
-        List<PersonCard> personList = getRootNode().getSelectionModel().getSelectedItems();
+        List<Person> selectedPersonList = getRootNode().getSelectionModel().getSelectedItems();
 
-        if (personList.size() != 1) {
+        if (selectedPersonList.size() != 1) {
             throw new AssertionError("Person list size expected 1.");
         }
 
-        return new PersonCardHandle(personList.get(0).getRoot());
+        return getAllCardNodes().stream()
+                .map(PersonCardHandle::new)
+                .filter(handle -> handle.equals(selectedPersonList.get(0)))
+                .findFirst()
+                .orElseThrow(IllegalStateException::new);
     }
 
     /**
@@ -45,7 +53,7 @@ public class PersonListPanelHandle extends NodeHandle<ListView<PersonCard>> {
      * Returns true if a card is currently selected.
      */
     public boolean isAnyCardSelected() {
-        List<PersonCard> selectedCardsList = getRootNode().getSelectionModel().getSelectedItems();
+        List<Person> selectedCardsList = getRootNode().getSelectionModel().getSelectedItems();
 
         if (selectedCardsList.size() > 1) {
             throw new AssertionError("Card list size expected 0 or 1.");
@@ -58,35 +66,28 @@ public class PersonListPanelHandle extends NodeHandle<ListView<PersonCard>> {
      * Navigates the listview to display {@code person}.
      */
     public void navigateToCard(Person person) {
-        List<PersonCard> cards = getRootNode().getItems();
-        Optional<PersonCard> matchingCard = cards.stream().filter(card -> card.person.equals(person)).findFirst();
-
-        if (!matchingCard.isPresent()) {
+        if (!getRootNode().getItems().contains(person)) {
             throw new IllegalArgumentException("Person does not exist.");
         }
 
         guiRobot.interact(() -> {
-            getRootNode().scrollTo(matchingCard.get());
+            getRootNode().scrollTo(person);
         });
         guiRobot.pauseForHuman();
     }
 
     /**
-     * Returns the person card handle of a person associated with the {@code index} in the list.
+     * Navigates the listview to {@code index}.
      */
-    public PersonCardHandle getPersonCardHandle(int index) {
-        return getPersonCardHandle(getRootNode().getItems().get(index).person);
-    }
+    public void navigateToCard(int index) {
+        if (index < 0 || index >= getRootNode().getItems().size()) {
+            throw new IllegalArgumentException("Index is out of bounds.");
+        }
 
-    /**
-     * Returns the {@code PersonCardHandle} of the specified {@code person} in the list.
-     */
-    public PersonCardHandle getPersonCardHandle(Person person) {
-        Optional<PersonCardHandle> handle = getRootNode().getItems().stream()
-                .filter(card -> card.person.equals(person))
-                .map(card -> new PersonCardHandle(card.getRoot()))
-                .findFirst();
-        return handle.orElseThrow(() -> new IllegalArgumentException("Person does not exist."));
+        guiRobot.interact(() -> {
+            getRootNode().scrollTo(index);
+        });
+        guiRobot.pauseForHuman();
     }
 
     /**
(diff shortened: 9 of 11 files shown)


src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -39,10 +36,10 @@ public void display() {
         for (int i = 0; i < TYPICAL_PERSONS.size(); i++) {
             personListPanelHandle.navigateToCard(TYPICAL_PERSONS.get(i));
             Person expectedPerson = TYPICAL_PERSONS.get(i);
-            PersonCardHandle actualCard = personListPanelHandle.getPersonCardHandle(i);
+            Person actualPerson = personListPanelHandle.getPerson(i);
 
-            assertCardDisplaysPerson(expectedPerson, actualCard);
-            assertEquals(Integer.toString(i + 1) + ". ", actualCard.getId());
+            assertEquals(expectedPerson, actualPerson);
+            // Because we do not have the actual card, we cannot compare the index on the card
```
Zhiyuan-Amos: This is a side effect of changing from `ListView<PersonCard>` to `ListView<Person>`; we cannot get the actual `PersonCard` and the corresponding `PersonCardHandle` :/ Not quite sure how to fix this though. Prior to 75cbfa3804ff365646e903514282d9313da4300e, the following code is being used to retrieve the `PersonCard` nodes.

```
private Set<Node> getAllCardNodes() {
    return guiRobot.lookup(CARD_PANE_ID).queryAll();
}
```

However, this code doesn't really get all the card nodes, because the `PersonCard` objects are only created when required; if we have 9999 `PersonCard` objects and the `ListView` currently displays from indices 1 to 5, then only the first 5 `PersonCard` objects are created. As such, we cannot get the 1000th `PersonCard` node, for example. So in this case, there's dependency on line 37 (`personListPanelHandle.navigateToCard(TYPICAL_PERSONS.get(i))`): By navigating to the card, you ensure that the `PersonCard` object is created, so you are able to find it using `getAllCardNodes()`.

Not sure if there's a better way to do this.

Zhiyuan-Amos:
Ah, forgot to add regression tests.

yamgent:
When I first saw the number of line changes, I had a shock of my life. :P

I am not sure whether we should explore generating `9999persons.xml` on the fly locally, rather than uploading the big file to the repository...

yamgent:
Btw is the revert based on a certain commit? (i.e. you did a `git revert <sha>`)

Zhiyuan-Amos:
> When I first saw the number of line changes, I had a shock of my life. :P
I am not sure whether we should explore generating 9999persons.xml on the fly locally, rather than uploading the big file to the repository...

Oh yes you are right. In DeveloperGuide we have this:

> Open a console and run the command gradlew processResources (Mac/Linux: ./gradlew processResources). It should finish with the BUILD SUCCESSFUL message.
This will generate all resources required by the application and tests.

So we should generate the .xml file as part of `processResources`?

> Btw is the revert based on a certain commit? (i.e. you did a git revert <sha>)

Yup it's based on a certain commit. Do I need to mention it as part of commit message? Cos there were additional changes apart from just reverting.

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: @yamgent need your thought on this too :)

yamgent:
> Yup it's based on a certain commit. Do I need to mention it as part of commit message? Cos there were additional changes apart from just reverting.

As spoken, yes, you might also want to mention the additional changes made after reverting the commit, so that we can verify and understand your flow.

Zhiyuan-Amos:
> I am not sure whether we should explore generating 9999persons.xml on the fly locally, rather than uploading the big file to the repository...

Since it's 1MB, I wonder if it's considered as too big and whether we can just upload it to GitHub :P

src/test/java/seedu/address/ui/PersonListPanelTest.java
yamgent: As spoken, I guess the dependency is necessary, otherwise the person card verification is nonsense. :P

pyokagan:
@yamgent 
> I am not sure whether we should explore generating 9999persons.xml on the fly locally, rather than uploading the big file to the repository...

Yes, imagine what happens when we add a new field to the `Person` class and need to update all 9999 entries in the test file. :-)

src/test/data/PersonListPanelTest/generate9999persons.py
```diff
@@ -0,0 +1,12 @@
...
+f.write('<?xml version="1.0" encoding="UTF-8" standalone="yes"?>\n')
+f.write('<addressbook>\n')
+for i in range(9999):
+    f.write('<persons>\n')
+    f.write('<name>{}</name>\n'.format(''.join('ABCDEFGHIJ'[int(x)] for x in str(i))))
+    f.write('<phone>123</phone>\n')
+    f.write('<email>a@a</email>\n')
+    f.write('<address>a</address>\n')
+    f.write('</persons>\n')
+f.write('</addressbook>\n')
```
eugenepeh: shouldn't this file deserve its own commit prior to this commit? 
That way reviewer can compare to the performance before and after as well.

Zhiyuan-Amos: Oh I thought that what was mentioned in the [issue](https://github.com/se-edu/addressbook-level4/issues/796#issue-288127194) is sufficient; the issue describes the problem while the PR resolves the problem? ^^

Zhiyuan-Amos:
> Here I abit don't get it. Doesn't the selection of the card triggers the selected event?

Yup the selection of the card will automatically post a card selected event. There's a difference between `navigate` and `select`: navigation causes your list view to shift to the card, while select... selects the card. XD

Zhiyuan-Amos:
@eugenepeh I edited your commit message suggestion slightly, ok with you?

> The ClearCommand deletes all Person objects. When we have large number of persons and getRemoved() is called, all the corresponding cards ~~were~~ **will be** created prior to the**m being removed** ~~deletion~~. As creation of PersonCard objects are costly, the app slows down tremendously.

eugenepeh:
> Yup the selection of the card will automatically post a card selected event. There's a difference between navigate and select: navigation causes your list view to shift to the card, while select... selects the card. XD

In that case, do we still need the selected event? o_o

src/test/data/PersonListPanelTest/generate9999persons.py
eugenepeh: Hmm okay... then I will have to copy this file to the previous commit to see the improvement >_>

Zhiyuan-Amos:
> In that case, do we still need the selection change event? o_o

Oh yes, an event is posted when a card is selected is because of this code in `PersonListPanel`:

https://github.com/se-edu/addressbook-level4/blob/eaa0cfbee95d15b88f550122d0c9f115f1c9a551/src/main/java/seedu/address/ui/PersonListPanel.java#L44-L52

The reason why I didn't want `navigateToCard(Person)` to select the card is that in [2/3], this method will be called at the [start of the system tests](https://github.com/se-edu/addressbook-level4/pull/818/commits/d258903371094d7df8084eb22f6dd3a409c4e18f#diff-16e8193ce6874fb434b8aa49d25422f0R47) to navigate through all the cards (see paragraph 4 of [2/3]). If I select the card during navigation, then it causes the `BrowserPanel` to load someone's page, which causes the default state of the app to be modified. 

So that's why I need commit [1/3]. However, I chose not to put the above reason as part of commit message in [1/3] because there's no need to do so, since there's sufficient reasons for removing it on its own already.

eugenepeh:
>The reason why I didn't want navigateToCard(Person) to select the card is that in [2/3], this method will be called at the start of the system tests to navigate through all the cards (see paragraph 4 of [2/3]). If I select the card during navigation, then it causes the BrowserPanel to load someone's page, which causes the default state of the app to be modified.

Hmm okay... but in that case, the selectionchanged will be redundant now isn't it? Is there any other methods that cause selectionchanged other than user click on the card itself?

Zhiyuan-Amos:
> Hmm okay... but in that case, the selectionchanged will be redundant now isn't it? Is there any other methods that cause selectionchanged other than user click on the card itself?

As in whether we need this class `PersonPanelSelectionChangedEvent`? We still need it because `BrowserPanel` catches this event to load the person :P 

The following code is called when user executes `SelectCommand` :P

https://github.com/se-edu/addressbook-level4/blob/eaa0cfbee95d15b88f550122d0c9f115f1c9a551/src/main/java/seedu/address/ui/PersonListPanel.java#L57-L62

yamgent:
Sorry I missed out on this PR, go ahead and make the changes first.

Zhiyuan-Amos:
@eugenepeh 

> [1/3] Here I abit don't get it. Doesn't the selection of the card triggers the selected event?

Should I still update [1/3] commit message? ^^

pyokagan:
Don't add a new dependency on Python, just write the file in the test's setup fixture.

eugenepeh:
>@eugenepeh
[1/3] Here I abit don't get it. Doesn't the selection of the card triggers the selected event?
Should I still update [1/3] commit message? ^^

Nope its fine. Just wanted to double check that its not the only function trigger the selection changed event.

Zhiyuan-Amos:
> Don't add a new dependency on Python, just write the file in the test's setup fixture.

Sorry Paul, can you elaborate further? >< Sorry first time working on this kinda thing.

pyokagan:
@Zhiyuan-Amos Basically, don't add a python script to the repository, otherwise developers need to install Python to run tests.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -1,41 +1,44 @@
...
+import seedu.address.storage.XmlSerializableAddressBook;
 
 public class PersonListPanelTest extends GuiUnitTest {
     private static final ObservableList<Person> TYPICAL_PERSONS =
             FXCollections.observableList(getTypicalPersons());
 
     private static final JumpToListRequestEvent JUMP_TO_SECOND_EVENT = new JumpToListRequestEvent(INDEX_SECOND_PERSON);
 
-    private PersonListPanelHandle personListPanelHandle;
+    private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/PersonListPanelTest/");
```
yamgent: Should the files be stored under `src/test/data/sandbox/` instead, especially since this is auto-generated?

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +51,63 @@ public void display() {
...
+        FileUtil.createFile(manyPersonsFile);
+        FileUtil.writeToFile(manyPersonsFile, builder.toString());
+        manyPersonsFile.deleteOnExit();
+        return manyPersonsFile;
+    }
+
+    /**
+     * Initializes {@code personListPanelHandle} with a {@code PersonListPanel} backed by {@code backingList}.
+     */
+    private void setUp(ObservableList<Person> backingList) {
```
yamgent: Since you are modifying the flow of the test, you might want to call this something else rather than `setUp()` to avoid confusing it with the JUnit's terminology.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +51,63 @@ public void display() {
...
+    public void performanceTest() throws Exception {
+        File xmlFile = createManyPersonsXmlFile();
+        XmlSerializableAddressBook xmlAddressBook =
+                XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
+        ObservableList<Person> personList =
+                FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
+
+        Stopwatch stopwatch = Stopwatch.createStarted();
+        setUp(personList);
+        guiRobot.interact(personList::clear);
```
yamgent: Just a comment, if we did not apply the fix, then the test will just continue to run even if it has already exceeded 2 seconds (in fact, when I tried this on `master`, the window just froze, and I waited patiently for ~30 seconds before manually aborting it).

Is there a way to fail faster? (i.e. timeout on 2 seconds, rather than waiting for `personList::clear` to finish.)

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: Yup seems right! Should I have the file deleted after the test ends? That's cos I realized my sandbox folder isn't empty; there's no auto-deletion of these sandbox files.

I'm thinking that it's fine to not auto-delete now since the other sandbox files aren't deleted as well.

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: is `initUiPart` ok?

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: Hmm, 1 way I can think of is multithreading: 1 thread on the timer, 1 thread on clearing the list. So once the timer is up, it will check whether the list is empty. If not, fail the test. Sounds ok?

src/test/java/seedu/address/ui/PersonListPanelTest.java
yamgent: We gitignored files inside sandbox so we theoretically can do whatever with it. However, I am inclined to delete this file because this file can become quite big, and is pretty useless after the test... :P

src/test/java/seedu/address/ui/PersonListPanelTest.java
yamgent: Yep sounds good

src/test/java/seedu/address/ui/PersonListPanelTest.java
yamgent: > Hmm, 1 way I can think of is multithreading: 1 thread on the timer, 1 thread on clearing the list. So once the timer is up, it will check whether the list is empty. If not, fail the test. Sounds ok?

You can give it a try, I seem to recall writing a similar method to test methods with a timeout stopwatch though, was that no longer in the codebase or...?

Zhiyuan-Amos: Yup, we have `GuiRobot#waitForEvent(BooleanSupplier, int)` implemented by you :P However, currently, whenever this method is called, the concurrency is automatically handled. For e.g. when we are waiting for the browser to load, we simply call this method because the loading of the browser is automatically handled on another thread, so we can call `waitForEvent(...)` and let it block the current thread. (see `WebViewUtil#waitUntilBrowserLoaded(BrowserPanelHandle)`)

In this case, we can't do this:
```
guiRobot.waitForEvent(personList::isEmpty, 2000); // BLOCK :( so this definitely fails
initUi(personList);
guiRobot.interact(personList::clear);
```
nor this:
```
initUi(personList);
guiRobot.interact(personList::clear); // method only returns after personList is cleared
guiRobot.waitForEvent(personList::isEmpty, 2000); // this method definitely passes
```

So we have to manually perform the multi-threading.

yamgent:
Sorry this might take a while to review! Will come back asap!

Zhiyuan-Amos:
@yamgent nps, thanks! :) 👍

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +47,66 @@ public void display() {
...
+    @Test
+    public void performanceTest() throws Exception {
+        File xmlFile = createManyPersonsXmlFile();
+        XmlSerializableAddressBook xmlAddressBook =
+                XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
+        ObservableList<Person> personList =
+                FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
+
+        // Initializes the PersonListPanel with 10000 persons and deletes all the persons
+        new Thread(() -> {
```
yamgent: I tried this and it still didn't work. Not sure if there's any better way of fixing this issue though... :(

Here's what I did to test whether it works or not:
1. Ensure that the PR is being checkout on your repo.
1. Copy `PersonListPanelTest.java` to a temporary place on your computer.
1. Now checkout `master`.
1. Add `PersonListPanelTest.java` back into the your repo, effectively inserting the new tests in.
1. Run `performanceTest()`

Expected: Automatic failure after two seconds.
Actual: Requires manual termination.

Zhiyuan-Amos: Ah, I tried with 500ms & 1000ms and they work. As such, I thought it should work with 2000ms too. But it turns out that there's a bug.
```
new Thread(() -> {
    initUi(personList);
    guiRobot.interact(personList::clear);
}).start();

guiRobot.waitForEvent(personList::isEmpty, 2000);
```

On master branch:
1. `personList` is actually cleared almost instantaneously since it's just deleting `Person` objects. It takes about 1300ms to delete, so when I set 1000ms as the timeout, the test fails and is terminated early (i.e. there's no issue with the concurrency). However, when I set 2000ms as the timeout, `guiRobot.waitForEvent(personList::isEmpty, 2000)` passes, so the test has already exited. The bug is here: I used the wrong condition `personList::isEmpty`.
2. The `ListView<PersonCard>` is also cleared very quickly.
3. When `personList::clear` is called, the code below is triggered which performs the mapping of **all 10000** `Person`s to `PersonCard`s:

https://github.com/TomasMikula/EasyBind/blob/409202070c279956f22ea83c8fb46db87575231f/src/main/java/org/fxmisc/easybind/MappedList.java#L78-L85

This is the segment that causes the lag. As such, it seems that there's no way for us to terminate earlier because there's no condition on our end that we can use (e.g. `personList::isEmpty` & `listView::isEmpty` don't work because the time taken to clear of these lists doesn't reflect anything about the creation of the 10000 `PersonCard`s).

Ways to resolve:

1. We can use a JUnit annotation to resolve this: `@Test(timeout=5000)` [see documentation](https://github.com/junit-team/junit4/wiki/timeout-for-tests). The main issue with this is that "The timeout specified in the Timeout rule applies to the entire test fixture"; since we load up the .xml file etc as well, we aren't able to specifically test how long `personList::clear` takes. 

1. Fork EasyBind and update the code above to not return anything when the objects are removed :P Not sure if this breaks anything though since it violates the contract of the class implementing this method.

yamgent: > We can use a JUnit annotation to resolve this: `@Test(timeout=5500)` see documentation. The main issue with this is that "The timeout specified in the Timeout rule applies to the entire test fixture"; since we load up the .xml file etc as well, we aren't able to specifically test how long personList::clear takes. Currently, our test only requires 2000ms, but if we are using timeout then we will need to up the time to 5500ms (which may not be that reflective of how long personList::clear actually takes).

As spoken in person, let's try this one. This is probably the most tactical way of solving this issue without accidentally complicating the codebase further. :P Any further design issues pertaining to the GUI (like how the `ListView` is holding `Person`s instead of `PersonCard`s) could be raised in the repo issue tracker and we can work it out further.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +47,60 @@ public void display() {
...
+            builder.append("<persons>\n");
+            builder.append("<name>a</name>\n");
+            builder.append("<phone>").append(i).append("00</phone>\n");
+            builder.append("<email>a@aa</email>\n");
+            builder.append("<address>a</address>\n");
+            builder.append("</persons>\n");
+        }
+        builder.append("</addressbook>\n");
+
+        File manyPersonsFile = new File(TEST_DATA_FOLDER + "9999persons.xml");
```
eugenepeh: uh header comments says 10000 person but the file name is 9999persons? how about just renaming it to manypersons.xml?

src/test/java/systemtests/AddressBookSystemTest.java
```diff
@@ -211,6 +211,7 @@ protected void assertSelectedCardDeselected() {
      * @see PersonListPanelHandle#isSelectedPersonCardChanged()
      */
     protected void assertSelectedCardChanged(Index expectedSelectedCardIndex) {
+        getPersonListPanel().navigateToCard(getPersonListPanel().getSelectedCardIndex());
```
eugenepeh: Need some clarification here cus it looks a bit weird - 
What happens if we don't navigate to the card? 
Isn't the `select` function suppose to navigates and selects the card already? 
In that case, do we still need to navigate to card again in the assertion? 
Since unable to perform any operation without navigating to it should automatically deem as failed.

Zhiyuan-Amos: `SelectCommandSystemTest` fails because as part of test verification, `AddressBookSystemTest#assertApplicationDisplaysExpected(String, String Model)` is called which verifies all the cards are displaying the correct persons. This will then cause the navigation to go off thus causing the test to fail :(

And yes, I agree we need some verification here. Is the below suggestion ok?
```
// navigates to the selected card to satisfy {@code PersonListPanelHandle#getHandleToSelectedCard()}'s precondition.
getPersonListPanel().navigateToCard(getPersonListPanel().getSelectedCardIndex());

String selectedCardName = getPersonListPanel().getHandleToSelectedCard().getName();
...
```

eugenepeh: perhaps include that in as well? 
`as AddressBookSystemTest#assertApplicationDisplaysExpected(String, String Model) is called to verify all the cards are displaying the correct persons, which causes the navigation to go off.`
since it's not that explicit... but adding that all make it seems like a bit too long for a comment... *smells like a design problem* >_<

Zhiyuan-Amos: Yes, this is a design problem. Was just discussing with @yamgent, by right `ListView` should be of type `PersonCard` instead of type `Person` because `ListView` belongs to the `View` component of MVC, and so it shouldn't need to know about the `Model` i.e. `Person`. If we have `ListView<PersonCard>`, then we can retrieve the `PersonCardHandle` wherever it is; we don't have to perform a prior navigation.

We have thought of a few alternative ways to allow `ListView` to remain as type `PersonCard`, but none of them are easy to implement..... :(

eugenepeh: In that case, @pyokagan do you approve of such a solution?

pyokagan: @Zhiyuan-Amos 
> by right ListView should be of type PersonCard instead of type Person because ListView belongs to the View component of MVC, and so it shouldn't need to know about the Model i.e. Person.

The dependencies between the components in an MVC architecture are as follows: (from https://martinfowler.com/eaaDev/uiArchs.html#ModelViewController)

mvc-deps

In other words, the view component has a dependency on the model. And really, it doesn't make sense to not have the view depend on the model -- what else would the view display?

Zhiyuan-Amos: > In other words, the view component has a dependency on the model. And really, it doesn't make sense to not have the view depend on the model -- what else would the view display?

Yup, the View depends on the Model, and so the `PersonCard` depends on the `Person`. Sorry that I wasn't entirely clear in my above comment.

I'm primarily concerned about the `ListCell` creating the `PersonCard` object because it seems more of the role of a controller to manipulate the Model (`Person`) into a View model (`PersonCard`) and hand it to the `ListCell` to display.

Also, I thought that the `ListView<Person>` can be somewhat misleading because it seems like we are displaying a list of persons on the UI when we are actually displaying a list of person cards.

pyokagan: @Zhiyuan-Amos 
> I'm primarily concerned about the ListCell creating the PersonCard object because it seems more of the role of a controller to manipulate the Model (Person) into a View model (PersonCard) and hand it to the ListCell to display.

The role of data flow for the controller and view is different.

Controller: User input -> domain objects
View: Domain objects -> User interface

Given that a `PersonCard`, a user interface object, is created from a domain object (`Person`), it belongs in the "view".

Also, `PersonCard` is not a "View-Model" since it contains actual JavaFX UI controls (which is also why it takes so long to be constructed). Mixing it up with the "View-Model" concept will just bring back the original problem. Besides, we don't use a "Model-ViewModel" architecture.

pyokagan: As for the original problem: it makes sense to *scroll*[1] to a `PersonCard` in order to check its contents. In a true end-to-end test, we should only be testing what is visible on the screen, anyway.

Footnote [1]: Not sure why `PersonListPanel#navigateToCard()` needs to select the card as well. If its technically possible, it would be good to have a method that just *scrolls* to the card without selecting it.

Zhiyuan-Amos: @pyokagan thanks for your insight! :)

eugenepeh: > As for the original problem: it makes sense to scroll[1] to a PersonCard in order to check its contents. In a true end-to-end test, we should only be testing what is visible on the screen, anyway.

I agree that we should be testing what is visible on the screen, which is what prompt my initial concern that `if we have already selected (which includes navigating to it), why do we navigate to it again to assert selection change?`. It is like filling up the condition for the app functionality in order to pass the test case.

Can we swap the order of `assertApplicationDisplaysExpected` and `assertSelectedCardChanged`?

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: @yamgent I'm thinking if we can use both `Timeout` and `Stopwatch`:
1. Have the `Timeout` for the purpose of **failing the test automatically**.
1. The `Stopwatch` tests specifically for the **init & clearing of the person list**.

```java
@Test(timeout = 5500)
public void performanceTest() throws Exception {
    File xmlFile = createManyPersonsXmlFile();
    XmlSerializableAddressBook xmlAddressBook =
            XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
    ObservableList<Person> personList =
            FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());

    Stopwatch stopwatch = Stopwatch.createStarted();
    initUi(personList);
    guiRobot.interact(personList::clear);
    if (stopwatch.elapsed(TimeUnit.MILLISECONDS) >= 2000) {
        fail(); 
    }
}
```

src/test/java/systemtests/AddressBookSystemTest.java
pyokagan: @eugenepeh 
> if we have already selected (which includes navigating to it), why do we navigate to it again to assert selection change? It is like filling up the condition for the app functionality in order to pass the test case.

This makes sense for the `SelectCommand` test, but not for other tests. I agree that there's room for making `SelectCommandSystemTest` more specific  (e.g. `assertCardIsVisible(Person)`), but `assertSelectedCardChanged` is used by other system tests as well.

Selection does not imply the selected item is visible. The scroll position of a ListView and its selection are completely independent things. `SelectCommand` is specifically implemented to modify both the scroll position and the selected item, but other commands like `DeleteCommand` may modify the selected item but may not modify the scroll position. For those commands, explicitly scrolling to the card to make it visible is needed to ensure the test is not flaky.

pyokagan: > Not sure why PersonListPanel#navigateToCard() needs to select the card as well. If its technically possible, it would be good to have a method that just scrolls to the card without selecting it.

Just realized this was already done in this PR :-P

eugenepeh: >SelectCommandSystemTest fails because as part of test verification, AddressBookSystemTest#assertApplicationDisplaysExpected(String, String Model) is called which verifies all the cards are displaying the correct persons. This will then cause the navigation to go off thus causing the test to fail :(

@Zhiyuan-Amos  is `SelectCommandSystemTest` the only one that was failed prior to this change?

Zhiyuan-Amos: @eugenepeh if you remove this line, only `SelectCommandSystemTest` fails. However, other parts of the code can possibly fail too:

https://github.com/se-edu/addressbook-level4/blob/529d75ffaf497d650fc077251bb3eb0fe976c5e8/src/test/java/systemtests/DeleteCommandSystemTest.java#L78-L87

At the time of execution of the above code, only 3 cards remain. As such, the navigation doesn't go off (because there's no scrolling required). If there are more cards, the test will fail too.

---

I realized that `EditCommandSystemTest` is supposed to fail too. However, it doesn't. As I looked further into this, I realized:
1. When you execute `FindCommand` which displays 0 / 1 persons, followed by `ListCommand`, and you call `navigateToCard` for all cards, somehow **all the card nodes** will be created after the navigation is completed (the card nodes are created as you navigate through the cards). I've tested on `TypicalPersons` where there are 7 persons and the screen size is only big enough for 4 persons. As long as we don't do any further operations on the filtered list, then all 7 card nodes remain. As such, we don't need to perform navigation before verifying the selected card.
1. Following the above procedure but `FindCommand` displays 2 persons, then there are only 6 person card nodes (instead of 7).
1. Following the above procedure but `FindCommand` displays 3 persons, then there are only 5 person card nodes (instead of 7).

I'm not quite sure why the behavior is this way though.... x.x

pyokagan: @Zhiyuan-Amos 
> I'm not quite sure why the behavior is this way though.... x.x

I think you need more persons than 7. [14 works for me](https://gist.github.com/pyokagan/4d25589cf7bb6896acefcc27f03bfb92)

This demonstrates another point, though: Just because a card node is inside the scene graph does not necessarily mean that it is visible. So, `getVisibleCardNodes()` is slightly misnamed, it does return all visible card nodes, but it also returns extra card nodes that may have been scrolled off-view. iow, we can't only use it for testing if a command explicitly scrolled the list view such that a person card is visible, we'll also need to compare its on-screen bounding box with the ListView's.

Zhiyuan-Amos: > I think you need more persons than 7. 14 works for me

Yup! 👍 

> This demonstrates another point, though: Just because a card node is inside the scene graph does not necessarily mean that it is visible. So, getVisibleCardNodes() is slightly misnamed, it does return all visible card nodes, but it also returns extra card nodes that may have been scrolled off-view.

Noted, will rename it.

> iow, we can't only use it for testing if a command explicitly scrolled the list view such that a person card is visible, we'll also need to compare its on-screen bounding box with the ListView's.

Yes, shall I raise this as an issue since the code in master also doesn't verify that the list view has been scrolled?

src/test/java/seedu/address/ui/PersonListPanelTest.java
damithc: Making the number of persons a parameter makes the code more self-explanatory? At the moment `1000` is a magic number embedded in the code/comment.

Zhiyuan-Amos:
@yamgent for your review since you requested changes previously :)

Zhiyuan-Amos:
> Making the number of persons a parameter makes the code more self-explanatory? At the moment 1000 is a magic number embedded in the code/comment.

Ah okay I misread the comment; I thought prof meant renaming "manyPersonsFile" -> "fileWith10000Persons"... ><

What should I name the constant though? `NUM_PERSONS` sounds weird, `LARGE_NUMBER` sounds weird too.

damithc:
> What should I name the constant though? NUM_PERSONS sounds weird, LARGE_NUMBER sounds weird too.

No need for a constant. I meant like this:
```java
private File createManyPersonsXmlFile(int personCount) throws Exception {
...
```

Zhiyuan-Amos:
@damithc ah ok! I initially decided against that because there's this prompt from IntelliJ:

image

Will make the change.

Zhiyuan-Amos:
@eugenepeh @yamgent thanks for your approvals! :) I've added an additional commit as I thought that this would make the test clearer:

1. The current test tests for the timing of both creating the xml file, the time taken to `initUI` and to clear the list of persons.
1. However, we don't actually want to test the timing of creating the xml file, since this test is specifically testing that the `PersonListPanel` doesn't take too long to start up and to be cleared.
1. As such, I decided to use a `StopWatch` to time the starting up and clearing timing. Then, the `Timeout(5500)` will be for the purpose of ending the test early as per Wang Leng's [comment](https://github.com/se-edu/addressbook-level4/pull/818#discussion_r168083994) above.

Does this look ok?

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -56,6 +60,10 @@ public void handleJumpToListRequestEvent() {
         assertCardEquals(expectedPerson, selectedPerson);
     }
 
+    /*
+     * Causes test to fail-fast; guiRobot#interact(Runnable) only returns after the Runnable completes execution.
+     * If guiRobot#interact(Runnable) takes very long to complete execution, the test will stall for a long time.
+     */
```
yamgent: I think it needs to be more specific in terms of what you are talking about (at first read, it wasn't immediately obvious that you were talking about the `@Test(timeout = 5500)`; I thought it was `performanceTest()` 's javadoc).

Zhiyuan-Amos: Yup, the missing `*` is to show that this is a code block rather than a javadoc. How should I make this clearer? ><

Zhiyuan-Amos:
> But as I said, certainty an improvement because previously it was just totally indistinguishable:

Thanks for your thoughts @yamgent, the table is good 👍

src/test/java/seedu/address/ui/PersonListPanelTest.java
yamgent: Perhaps you can add:

```
**The timeout specified for the following test** causes it to fail fast; ...
... the test will stall for a long time **without the timeout.**
```

Also maybe the use of `//` is better than `/*`? Just a suggestion, thought that would make it more obvious than observing one missing `*`. :P

yamgent:
> Hmm I wonder if I should log after the stopwatch starts. I should log how long the xml creation took, and whether it's lesser than 5500 - 2000 = 3500:
> 
> Timeout thrown, log not printed == stuck in xml.
> Timeout thrown, log printed == stuck in card.

Ah just saw the edits. 

I think logging can be useful, although logging is not enabled by default when running on Travis or terminal without additional parameters, but up to the developer to enable it, so you will have to document that down and hope that they read it if they have issues with this test.

Alternatively, maybe consider timing the loading of xml as well?

```java
Stopwatch stopwatch1 = ...;

// xml loading code...

if (stopwatch1.elapsed(TimeUnit.MILLISECONDS) >= 3000) {
    fail("Xml took too long to load.");
}

Stopwatch stopwatch2 = ...;

// deleting card code...

if (stopwatch2.elapsed(TimeUnit.MILLISECONDS) >= 2000) {
    fail("Card deleting took too long to load.");
}
```

Zhiyuan-Amos:
> Alternatively, maybe consider timing the loading of xml as well?

I think this doesn't work. Suppose we hit the `Timeout`, it may be because:
1. Xml creation took >= `Timeout`. This happens before the xml loading ends, so the `fail(...)` didn't happen.
2. Card addition and deletion >= `Timeout`.

So we still can't tell the difference :P As such, we need to merge our solutions by doing both timing the xml & logging the `else statement` i.e. if creation of xml passed, we need to log it. :P

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +55,96 @@ public void display() {
...
+        initUi(backingList);
+        guiRobot.interact(backingList::clear);
+        if (stopwatch.elapsed(TimeUnit.MILLISECONDS) >= 2000) {
+            fail("Creation and deletion of person cards exceeded time limit");
+        }
+    }
+
+    /**
+     * Returns a list of persons containing {@code personCount} persons that is used to populate the
+     * {@code PersonListPanel}.
```
yamgent: The javadocs here didn't mention about the logging.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +55,96 @@ public void display() {
...
+
+        File xmlFile = createXmlFileWithPersons(personCount);
+        XmlSerializableAddressBook xmlAddressBook =
+                XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
+        ObservableList<Person> personList =
+                FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
+
+        long createListTime = stopwatch.elapsed(TimeUnit.MILLISECONDS);
+        String createListMessage = "List creation took: " + createListTime + "ms. ";
+        if (createListTime >= 4000) {
```
yamgent: I wonder whether the proliferation of magic numbers should be a cause of concern, since it is not immediately obvious in this method scope why it is given 4000ms.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +55,96 @@ public void display() {
...
+    /**
+     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
+     * to execute.
+     * 1. If {@code TestTimedOutException} is thrown before the logs are printed, then the failure occurs during the
+     * preparation of the test (creating the xml file containing the large number of persons).
+     * 2. If {@code TestTimedOutException} is thrown after the logs are printed, then the failure occurs during the
+     * creation and deletion of the person cards.
+     */
+    // Causes test to fail-fast; guiRobot#interact(Runnable) only returns after the Runnable completes execution.
+    // If guiRobot#interact(Runnable) takes very long to complete execution, the test will stall for a long time.
```
Zhiyuan-Amos: I should add {@code } here

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: Hmm, so I should have a:
```
private static final int TEST_TIMEOUT = 6000;
private static final int CARD_CREATION_AND_DELETION_TIMEOUT = 2000;
private static final int LIST_CREATION_TIMEOUT = TEST_TIMEOUT - CARD_CREATION_AND_DELETION_TIMEOUT;
```

Does this look ok?

yamgent: Yes.

Zhiyuan-Amos:
Edit: I increased the test timings as the test was sometimes failing on Travis.

Zhiyuan-Amos:
@damithc for your review again as I've made quite a bit of changes since your previous approval.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +59,97 @@ public void display() {
...
+
+    /**
+     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
+     * to execute.
+     * 1. If {@code TestTimedOutException} is thrown before the logs are printed, then the failure occurs during the
+     * preparation of the test (creating the xml file containing the large number of persons).
+     * 2. If {@code TestTimedOutException} is thrown after the logs are printed, then the failure occurs during the
+     * creation and deletion of the person cards.
+     */
+    // Causes test to fail-fast; guiRobot#interact(Runnable) only returns after the Runnable completes execution.
```
damithc: The comment can be clearer: what causes test to fail fast? the timeout? e.g.,
>The timeout makes the test fail fast. Without it, the test can stall for a long time because ...

damithc: Also, I'm not sure if javadoc tools allows normal comments between javadoc and the method.

Zhiyuan-Amos: > Also, I'm not sure if javadoc tools allows normal comments between javadoc and the method.

Yup, verified that having normal comments between javadoc and the method won't interfere with the generation of the javadoc.

image

Also, @damithc I've updated the header comments with the html tags to format it to look as above, otherwise it will be in a single paragraph. Is it ok?

damithc: Looks good 👍

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -48,11 +59,104 @@ public void display() {
...
+    }
+
+    /**
+     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
+     * to execute.
+     * <ol>
+     *     <li>
+     *         If {@code TestTimedOutException} is thrown before the logs are printed, then the failure occurs during
+     *         the preparation of the test (creating the xml file containing the large number of persons).
+     *     </li>
```
damithc: Like this will work too (fewer lines needed):
```html
*     <li>If {@code TestTimedOutException} is thrown before the logs are printed, then the failure occurs during
*         the preparation of the test (creating the xml file containing the large number of persons).</li>
```

Zhiyuan-Amos:
Changelog: Rebased off master after merging #837, otherwise Travis fails.

PR #820, opened 2018-02-06
Email regex validation does not accept `+` #743

### LINKED ISSUE #743: Email regex validation does not accept `+`
It was posted as one of our bugs that the email address validation is not entirely correct, as certain characters such as "+" are not accepted, even though they are valid. More details can be found [here](https://github.com/nus-cs2103-AY1718S1/pe-results/issues/42) .

vivekscl:
Fixes #743 

Do note that the domain name has to be 2 characters long so that it does not start and end with dots or hyphens.
 
On a side note, the updated email regex just allows the use of more characters to solve the issue. Should a more restrictive validation be used instead?

For instance this regex,
```
^[\\w!#$%&’*+/=?`{|}~^-]+(?:\\.[\\w!#$%&’*+/=?`{|}~^-]+)*@(?:[a-zA-Z0-9-]+\\.)+[a-zA-Z]{2,6}$
```
sets a predefined length for the top-level domain name (which would be **edu** in someone@u.nus.edu) and restricts trailing and consecutive dots.

### DIFF
--- a/src/main/java/seedu/address/model/person/Email.java
+++ b/src/main/java/seedu/address/model/person/Email.java
@@ -4,14 +4,28 @@ import static java.util.Objects.requireNonNull;
 import static seedu.address.commons.util.AppUtil.checkArgument;
 
 /**
- * Represents a Person's phone number in the address book.
+ * Represents a Person's email in the address book.
  * Guarantees: immutable; is valid as declared in {@link #isValidEmail(String)}
  */
 public class Email {
 
-    public static final String MESSAGE_EMAIL_CONSTRAINTS =
-            "Person emails should be 2 alphanumeric/period strings separated by '@'";
-    public static final String EMAIL_VALIDATION_REGEX = "[\\w\\.]+@[\\w\\.]+";
+    private static  final String SPECIAL_CHARACTERS = "!#$%&'*+/=?`{|}~^.-";
+    public static final String MESSAGE_EMAIL_CONSTRAINTS = "Person emails should be of the format local-part@domain "
+            + "and adhere to the following constraints:\n"
+            + "1. The local-part should only contain alphanumeric characters and these special characters, excluding "
+            + "the parentheses, (" + SPECIAL_CHARACTERS + ") .\n"
+            + "2. This is followed by a '@' and then a domain name. "
+            + "The domain name must:\n"
+            + "    - be at least 2 characters long\n"
+            + "    - start and end with alphanumeric characters\n"
+            + "    - consist of alphanumeric characters, a period or a hyphen for the characters in between, if any.";
+    // alphanumeric and special characters
+    private static final String LOCAL_PART_REGEX = "^[\\w" + SPECIAL_CHARACTERS + "]+";
+    private static final String DOMAIN_FIRST_CHARACTER_REGEX = "[^\\W_]"; // alphanumeric characters except underscore
+    private static final String DOMAIN_MIDDLE_REGEX = "[a-zA-Z0-9.-]*"; // alphanumeric, period and hyphen
+    private static final String DOMAIN_LAST_CHARACTER_REGEX = "[^\\W_]$";
+    public static final String EMAIL_VALIDATION_REGEX = LOCAL_PART_REGEX + "@"
+            + DOMAIN_FIRST_CHARACTER_REGEX + DOMAIN_MIDDLE_REGEX + DOMAIN_LAST_CHARACTER_REGEX;
 
     public final String value;
 
--- a/src/test/java/seedu/address/model/person/EmailTest.java
+++ b/src/test/java/seedu/address/model/person/EmailTest.java
@@ -35,22 +35,28 @@ public class EmailTest {
         assertFalse(Email.isValidEmail("peterjack@")); // missing domain name
 
         // invalid parts
-        assertFalse(Email.isValidEmail("-@example.com")); // invalid local part
         assertFalse(Email.isValidEmail("peterjack@-")); // invalid domain name
+        assertFalse(Email.isValidEmail("peterjack@exam_ple.com")); // underscore in domain name
         assertFalse(Email.isValidEmail("peter jack@example.com")); // spaces in local part
         assertFalse(Email.isValidEmail("peterjack@exam ple.com")); // spaces in domain name
+        assertFalse(Email.isValidEmail(" peterjack@example.com")); // leading space
+        assertFalse(Email.isValidEmail("peterjack@example.com ")); // trailing space
         assertFalse(Email.isValidEmail("peterjack@@example.com")); // double '@' symbol
         assertFalse(Email.isValidEmail("peter@jack@example.com")); // '@' symbol in local part
         assertFalse(Email.isValidEmail("peterjack@example@com")); // '@' symbol in domain name
+        assertFalse(Email.isValidEmail("peterjack@.example.com")); // domain name starts with a period
+        assertFalse(Email.isValidEmail("peterjack@example.com.")); // domain name ends with a period
+        assertFalse(Email.isValidEmail("peterjack@-example.com")); // domain name starts with a hyphen
+        assertFalse(Email.isValidEmail("peterjack@example.com-")); // domain name ends with a hyphen
 
         // valid email
         assertTrue(Email.isValidEmail("PeterJack_1190@example.com"));
-        assertTrue(Email.isValidEmail("a@b"));  // minimal
+        assertTrue(Email.isValidEmail("a@bc"));  // minimal
         assertTrue(Email.isValidEmail("test@localhost"));   // alphabets only
+        assertTrue(Email.isValidEmail("!#$%&'*+/=?`{|}~^.-@example.org")); // special characters local part
         assertTrue(Email.isValidEmail("123@145"));  // numeric local part and domain name
-        assertTrue(Email.isValidEmail("a1@example1.com"));  // mixture of alphanumeric and dot characters
-        assertTrue(Email.isValidEmail("_user_@_e_x_a_m_p_l_e_.com_"));    // underscores
-        assertTrue(Email.isValidEmail("peter_jack@very_very_very_long_example.com"));   // long domain name
+        assertTrue(Email.isValidEmail("a1+be!@example1.com")); // mixture of alphanumeric and special characters
+        assertTrue(Email.isValidEmail("peter_jack@very-very-very-long-example.com"));   // long domain name
         assertTrue(Email.isValidEmail("if.you.dream.it_you.can.do.it@example.com"));    // long local part
     }
 }


Zhiyuan-Amos:
Just a quick thought, is the regex too confusing for students to understand? :O @eugenepeh any thoughts? :)

eugenepeh:
> Just a quick thought, is the regex too confusing for students to understand? :O @eugenepeh any thoughts? :)

Not just for students, its a bit too confusing for me to understand too hahahaha, I will need to refer to https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html. 

Plus I am not sure if regex was taught in the current batch students' syllabus. I wonder if we should make a regex util or something that breaks down and combine regex string for readability.

Zhiyuan-Amos:
Yup @vivekscl you will need to split up the regex into smaller parts, maybe something like:

```java
String DOMAIN_REGEX = ...
String LOCAL_PART_REGEX = ...
String EMAIL_VALIDATION_REGEX = DOMAIN REGEX + "@" + LOCAL_PART_REGEX;
```

Feel free to split it up further as you deem fit :)

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,17 @@
...
-    public static final String MESSAGE_EMAIL_CONSTRAINTS =
-            "Person emails should be 2 alphanumeric/period strings separated by '@'";
-    public static final String EMAIL_VALIDATION_REGEX = "[\\w\\.]+@[\\w\\.]+";
+    public static final String MESSAGE_EMAIL_CONSTRAINTS = "Person emails should follow the following constraints:\n"
+            + "The local part should only contain alphanumeric characters and the following special characters "
+            + "excluding the parentheses, (!#$%&'*+/=?`{|}~^.-) . This is followed by a '@' and then a domain name.\n "
+            + "The domain name must be at least 2 characters long and must start and end with alphanumeric characters. "
+            + "The characters in between can be alphanumeric characters, a period or a hyphen.";
+    private static final String LOCAL_PART_REGEX = "^[\\w!#$%&'*+/=?`{|}~^.-]+"; // alphanumeric and special characters
+    private static final String DOMAIN_START_REGEX = "[^\\W_]"; // alphanumeric characters except underscore
```
Zhiyuan-Amos: we should somehow specify that this is the first character of the domain portion. Maybe `DOMAIN_FIRST_CHARACTER_REGEX`? Same for `DOMAIN_END_REGEX`

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,31 @@ public void isValidEmail() {
...
+        assertFalse(Email.isValidEmail(" peterjack@example.com")); // leading space
+        assertFalse(Email.isValidEmail("peterjack@example.com ")); // trailing space
         assertFalse(Email.isValidEmail("peterjack@@example.com")); // double '@' symbol
         assertFalse(Email.isValidEmail("peter@jack@example.com")); // '@' symbol in local part
         assertFalse(Email.isValidEmail("peterjack@example@com")); // '@' symbol in domain name
+        assertFalse(Email.isValidEmail("peterjack@.example.com")); // domain name starts with a dot
+        assertFalse(Email.isValidEmail("peterjack@example.com.")); // domain name ends with a dot
+        assertFalse(Email.isValidEmail("peterjack@-example.com")); // domain name starts with a hyphen
+        assertFalse(Email.isValidEmail("peterjack@example.com-")); // domain name ends with a hyphen
+
```
Zhiyuan-Amos: extra new line

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,31 @@ public void isValidEmail() {
...
-        assertTrue(Email.isValidEmail("if.you.dream.it_you.can.do.it@example.com"));    // long local part
+        assertTrue(Email.isValidEmail("a@bc")); // minimal
+        assertTrue(Email.isValidEmail("test@localhost")); // alphabets only
+        assertTrue(Email.isValidEmail("#!$%&'*+-/=?^_`{}|~@example.org")); // special characters local part
+        assertTrue(Email.isValidEmail("123@145")); // numeric local part and domain name
+        assertTrue(Email.isValidEmail("a1@example1.com")); // mixture of alphanumeric and dot characters
+        assertTrue(Email.isValidEmail("a1+be!@example1.com")); // mixture of alphanumeric and special characters
+        assertTrue(Email.isValidEmail("_user_@example.com")); // underscores
+        assertTrue(Email.isValidEmail("peter_jack@very-very-very-long-example.com")); // long domain name
+        assertTrue(Email.isValidEmail("if.you.dream.it_you.can.do.it@example.com")); // long local part
```
Zhiyuan-Amos: there's additional changes here; you shouldn't shift the placement of the comments.

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,31 @@ public void isValidEmail() {
...
-        assertTrue(Email.isValidEmail("a1@example1.com"));  // mixture of alphanumeric and dot characters
-        assertTrue(Email.isValidEmail("_user_@_e_x_a_m_p_l_e_.com_"));    // underscores
-        assertTrue(Email.isValidEmail("peter_jack@very_very_very_long_example.com"));   // long domain name
-        assertTrue(Email.isValidEmail("if.you.dream.it_you.can.do.it@example.com"));    // long local part
+        assertTrue(Email.isValidEmail("a@bc")); // minimal
+        assertTrue(Email.isValidEmail("test@localhost")); // alphabets only
+        assertTrue(Email.isValidEmail("#!$%&'*+-/=?^_`{}|~@example.org")); // special characters local part
+        assertTrue(Email.isValidEmail("123@145")); // numeric local part and domain name
+        assertTrue(Email.isValidEmail("a1@example1.com")); // mixture of alphanumeric and dot characters
+        assertTrue(Email.isValidEmail("a1+be!@example1.com")); // mixture of alphanumeric and special characters
```
Zhiyuan-Amos: line 59 and 60 can be collapsed as 1 test.

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,31 @@ public void isValidEmail() {
...
-        assertTrue(Email.isValidEmail("a@b"));  // minimal
-        assertTrue(Email.isValidEmail("test@localhost"));   // alphabets only
-        assertTrue(Email.isValidEmail("123@145"));  // numeric local part and domain name
-        assertTrue(Email.isValidEmail("a1@example1.com"));  // mixture of alphanumeric and dot characters
-        assertTrue(Email.isValidEmail("_user_@_e_x_a_m_p_l_e_.com_"));    // underscores
-        assertTrue(Email.isValidEmail("peter_jack@very_very_very_long_example.com"));   // long domain name
-        assertTrue(Email.isValidEmail("if.you.dream.it_you.can.do.it@example.com"));    // long local part
+        assertTrue(Email.isValidEmail("a@bc")); // minimal
+        assertTrue(Email.isValidEmail("test@localhost")); // alphabets only
+        assertTrue(Email.isValidEmail("#!$%&'*+-/=?^_`{}|~@example.org")); // special characters local part
```
Zhiyuan-Amos: Can we have the same order as `LOCAL_PART_REGEX` i.e. `!#$%&'*+/=?`{|}~^.-` so that it's easier to see that all the special characters are included.

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,31 @@ public void isValidEmail() {
...
+        assertTrue(Email.isValidEmail("123@145")); // numeric local part and domain name
+        assertTrue(Email.isValidEmail("a1@example1.com")); // mixture of alphanumeric and dot characters
+        assertTrue(Email.isValidEmail("a1+be!@example1.com")); // mixture of alphanumeric and special characters
+        assertTrue(Email.isValidEmail("_user_@example.com")); // underscores
```
Zhiyuan-Amos: This is already tested as part of `// special characters local part`

src/test/java/seedu/address/model/person/EmailTest.java
```diff
@@ -35,22 +35,28 @@ public void isValidEmail() {
         assertFalse(Email.isValidEmail("peterjack@")); // missing domain name
 
         // invalid parts
-        assertFalse(Email.isValidEmail("-@example.com")); // invalid local part
         assertFalse(Email.isValidEmail("peterjack@-")); // invalid domain name
+        assertFalse(Email.isValidEmail("peterjack@exam_ple.com")); // underscores in domain name
```
Zhiyuan-Amos: underscore~~s~~

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,17 @@
  */
 public class Email {
 
-    public static final String MESSAGE_EMAIL_CONSTRAINTS =
-            "Person emails should be 2 alphanumeric/period strings separated by '@'";
-    public static final String EMAIL_VALIDATION_REGEX = "[\\w\\.]+@[\\w\\.]+";
+    public static final String MESSAGE_EMAIL_CONSTRAINTS = "Person emails should follow the following constraints:\n"
```
Zhiyuan-Amos: image

There seems to be an additional spacing (as highlighted in blue).

vivekscl:
For the link, I believed that it was best to get into the details immediately, but then again, the overview does explain the RFC 5322 format. Will change it accordingly then.

vivekscl:
Oh and for the email constraints message, when in full screen, there is no need to scroll up, down, left or right. But it seems in other laptops it's too long. I'll try to make it consistent with the other constraint messages in AB4 and try to make it more succinct then.

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,20 @@
...
+            + "The local part should only contain alphanumeric characters and these special characters, excluding the "
+            + "parentheses, (!#$%&'*+/=?`{|}~^.-) .\n"
+            + "This is followed by a '@' and then a domain name. "
+            + "The domain name must:\n"
+            + "- be at least 2 characters long\n"
+            + "- start and end with alphanumeric characters\n"
+            + "- be alphanumeric characters, a period or a hyphen for the characters in between, if any.";
```
Zhiyuan-Amos: This can be rephrased: `Any characters in between (if any) must be alphanumeric...` to have a closer link between the previous point and this point.

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,20 @@
  */
 public class Email {
 
-    public static final String MESSAGE_EMAIL_CONSTRAINTS =
-            "Person emails should be 2 alphanumeric/period strings separated by '@'";
-    public static final String EMAIL_VALIDATION_REGEX = "[\\w\\.]+@[\\w\\.]+";
+    public static final String MESSAGE_EMAIL_CONSTRAINTS = "Person emails should follow the following constraints:\n"
+            + "The local part should only contain alphanumeric characters and these special characters, excluding the "
+            + "parentheses, (!#$%&'*+/=?`{|}~^.-) .\n"
+            + "This is followed by a '@' and then a domain name. "
```
Zhiyuan-Amos: We should number: 1. 2.

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,20 @@
...
+            + "The domain name must:\n"
+            + "- be at least 2 characters long\n"
```
Zhiyuan-Amos: should we add a space here?

vivekscl: Yes, an indentation would make it more clear.

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,20 @@
...
-            "Person emails should be 2 alphanumeric/period strings separated by '@'";
-    public static final String EMAIL_VALIDATION_REGEX = "[\\w\\.]+@[\\w\\.]+";
+    public static final String MESSAGE_EMAIL_CONSTRAINTS = "Person emails should follow the following constraints:\n"
+            + "1. The local part should only contain alphanumeric characters and these special characters, excluding "
+            + "the parentheses, (!#$%&'*+/=?`{|}~^.-) .\n"
+            + "2. This is followed by a '@' and then a domain name. "
+            + "The domain name must:\n"
+            + "    - be at least 2 characters long\n"
+            + "    - start and end with alphanumeric characters\n"
+            + "    - consist of alphanumeric characters, a period or a hyphen for characters in between, if any.";
```
damithc: I don't think we can expect end user to know what `local` and `domain name` means. If we are following a standard described elsewhere, we can point to that page.

damithc: Yes to both

vivekscl: But if we add the `local-part@domain` to the message, there is no need for the link right? Since in this case, we would have made it clear what `local` and `domain` means

damithc: Yes.

src/main/java/seedu/address/model/person/Email.java
```diff
@@ -9,9 +9,20 @@
...
+    private static final String LOCAL_PART_REGEX = "^[\\w!#$%&'*+/=?`{|}~^.-]+"; // alphanumeric and special characters
```
damithc: The same set of symbols is repeated in the error message. May be extract into one constant?

vivekscl:
The comment says `Represents a Person's phone number...`. I'll change it to email in this PR itself as it seems too trivial to make it into a new issue. 
deepinscreenshot_select-area_20180209000916

Issue #822, opened 2018-02-08
typicalPersonsAddressBook.xml: remove @code tag for header comment


yamidark:
`@code` tag is used for javadocs and not .xml files.

yamgent:
For clarification, the offending line is this:

https://github.com/se-edu/addressbook-level4/blob/95611eb437b8e84e51f4b0555fa483fbad05a16b/src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml#L3

PR #792, opened 2018-01-10
Issue with undo/redo and modifications to filtered lists #737

yamidark:
Fixes #737 

EditCommand and DeleteCommand stores the target index of the person to edit
or delete respectively.

As a result, redoing of these index-based commands will be executed on the incorrect
person due to different indexing in differing views.

Let's update EditCommand and DeleteCommand to find and store the personToEdit
or personToDelete respectively before it's first execution. This ensures that future
execution of the same command will be on the correct person.

Issue #658, opened 2017-09-06
AddressBook#AddressBook(ReadOnlyAddressBook): Update method call


Zhiyuan-Amos:
`AddressBook#AddressBook(ReadOnlyAddressBook)`: Update method call

This constructor calls `AddressBook#resetData(ReadOnlyAddressBook)`. The
header comments of that method states that it resets the existing data
of the `AddressBook` with the data of the `ReadOnlyAddressBook`.

This is misleading as the constructor does not have any data to be
reset.

Perhaps we can fix this by renaming 
`AddressBook#resetData(ReadOnlyAddressBook)` and update the header
comments accordingly.

PR #825, opened 2018-02-09
AddressBook: remove master tag list #753

### LINKED ISSUE #659: AddressBook: Simplify resetData(ReadOnlyAddressBook)
`resetData(ReadOnlyAddressBook)` calls 
`syncMasterTagListWith(UniquePersonList)` to synchronize the tags in the
`Person` objects to point to a `Tag` object in the master list.

This is unnecessary if we ensure that this invariant is maintained in
`AddressBook`. That is, we can assume that any `AddressBook` that is
passed in as a parameter to `resetData(ReadOnlyAddressBook)` is already
rightly synchronized.

Let's teach `AddressBook` to maintain this invariant, and let's remove
the call to `syncMasterTagListWith(UniquePersonList)`.

### LINKED ISSUE #753: AddressBook needs to track tags properly

### LINKED ISSUE #764: AddressBook: Make setter methods private

Rinder5:
fixes #753 
update 1 : Also resolves #764 (The second issue of synchronizing tags is addressed, the first issue of public setter methods should not be changed as discussed in #794) 
update 2: Also resolves #659 

```
We keep track of a master tag list which holds all the tags in
AddressBook.

The tag list is not implemented correctly. The original intention was 
for tags in Person to point to the tags in the master tag list. 
However, implementing it in this way may be too complex for new 
developers to grasp, as discussed in #753. Furthermore, from the 
discussion in #794, the tag list in its current state is not needed.

Let's remove the master tag list in AddressBook.

Furthermore, without the master tag list in AddressBook,
UniqueTagList is no longer necessary.

Let's remove UniqueTagList.
```


Zhiyuan-Amos:
I'll just leave a quick comment: Why isn't `UniqueTagList` class deleted? :P

yamidark:
https://github.com/Rinder5/addressbook-level4/blob/3cb4c70eaf25462f91550e84730d1216ac91f37c/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java#L22-L23
should be removed too.

Rinder5:
> I'll just leave a quick comment: Why isn't UniqueTagList class deleted? :P 
UniqueTagList is actually just a list of tags which can't have duplicates in them.
The Person class uses it to store the tags for it's object.
The issue we have is with the specific UniqueTagList in AddressBook, which is updated but isn't used.

> Also, do take note of your commit messages to follow the standard format.
Noted. I tried to make it follow the format but I guess I messed up somewhere.

> should be removed too I think.
My bad, I'll make the change.

Zhiyuan-Amos:
Remember to update Developer Guide too since `AddressBook` no longer has a `UniqueTagList`.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -36,7 +29,6 @@
      */
     {
         persons = new UniquePersonList();
-        tags = new UniqueTagList();
```
Zhiyuan-Amos: Update header comments of line 37 `Creates an AddressBook using the Persons and Tags in the {@code toBeCopied}`

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -55,18 +47,12 @@ public void setPersons(List<Person> persons) throws DuplicatePersonException {
...
-
     /**
      * Resets the existing data of this {@code AddressBook} with {@code newData}.
      */
     public void resetData(ReadOnlyAddressBook newData) {
         requireNonNull(newData);
-        setTags(new HashSet<>(newData.getTagList()));
         List<Person> syncedPersonList = newData.getPersonList().stream()
-                .map(this::syncWithMasterTagList)
                 .collect(Collectors.toList());
```
Zhiyuan-Amos: There's no need to stream anymore LOL. `newData.getPersonList()` will do.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -86,11 +72,7 @@ public void resetData(ReadOnlyAddressBook newData) {
      * @throws DuplicatePersonException if an equivalent person already exists.
      */
     public void addPerson(Person p) throws DuplicatePersonException {
-        Person person = syncWithMasterTagList(p);
```
Zhiyuan-Amos: update header comments.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -100,39 +82,12 @@ public void addPerson(Person p) throws DuplicatePersonException {
      * @throws DuplicatePersonException if updating the person's details causes the person to be equivalent to
      *      another existing person in the list.
      * @throws PersonNotFoundException if {@code target} could not be found in the list.
-     *
-     * @see #syncWithMasterTagList(Person)
```
Zhiyuan-Amos: update header comments

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -166,22 +115,16 @@ public String toString() {
...
-                && this.persons.equals(((AddressBook) other).persons)
-                && this.tags.equalsOrderInsensitive(((AddressBook) other).tags));
+                && this.persons.equals(((AddressBook) other).persons));
     }
 
     @Override
     public int hashCode() {
         // use this method for custom fields hashing instead of implementing your own
-        return Objects.hash(persons, tags);
+        return Objects.hash(persons);
```
Zhiyuan-Amos: We can do `return persons.hashCode()` now since it's only 1 object.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -37,7 +37,6 @@ public XmlSerializableAddressBook() {
     public XmlSerializableAddressBook(ReadOnlyAddressBook src) {
         this();
         persons.addAll(src.getPersonList().stream().map(XmlAdaptedPerson::new).collect(Collectors.toList()));
-        tags.addAll(src.getTagList().stream().map(XmlAdaptedTag::new).collect(Collectors.toList()));
```
Zhiyuan-Amos: Since you are updating `XmlSerializableAddressBook` in [3/3], then it seems that the changes here in [1/3] for this file should be done on [3/3] instead?

src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml
```diff
@@ -1,5 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
Zhiyuan-Amos: From what I mentioned above, then this shouldn't be removed here, but in [3/3]

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -77,18 +70,12 @@ public void getTagList_modifyList_throwsUnsupportedOperationException() {
 
         AddressBookStub(Collection<Person> persons, Collection<? extends Tag> tags) {
             this.persons.setAll(persons);
-            this.tags.setAll(tags);
```
Zhiyuan-Amos: you should remove line 69 too.

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
```diff
@@ -40,12 +39,4 @@ public void toModelType_invalidPersonFile_throwsIllegalValueException() throws E
         thrown.expect(IllegalValueException.class);
         dataFromFile.toModelType();
     }
-
-    @Test
-    public void toModelType_invalidTagFile_throwsIllegalValueException() throws Exception {
```
Zhiyuan-Amos: Same as above: XmlSerializableAddressBook should be edited in [3/3]

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -64,6 +61,6 @@ public boolean equals(Object other) {
         }
 
         XmlSerializableAddressBook otherAb = (XmlSerializableAddressBook) other;
-        return persons.equals(otherAb.persons) && tags.equals(otherAb.tags);
+        return persons.equals(otherAb.persons);
```
Zhiyuan-Amos: Lines 63 & 64 can be simplified into 1 line.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -19,16 +19,13 @@
...
-    @XmlElement
-    private List<XmlAdaptedTag> tags;
 
     /**
      * Creates an empty XmlSerializableAddressBook.
      * This empty constructor is required for marshalling.
      */
     public XmlSerializableAddressBook() {
         persons = new ArrayList<>();
-        tags = new ArrayList<>();
```
Zhiyuan-Amos: `toModelType()`'s header comments need to be updated.

Zhiyuan-Amos:
@yamidark & @vivekscl this PR I will leave to both of you to review since you already have some experience :) Do take note of:

1. Commit messages
2. Documentation such as Header comments, Developer Guide etc, and unused methods are deleted.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
yamidark: I think the change is made here due to `ReadOnlyAddressBook` no longer having the `getTagList()` method.

Zhiyuan-Amos: @yamidark can you help me take a look whether we can reorganize commits to make it tidier e.g.
1. Remove tags from `XmlAddressBook` and remove things from related .xml classes
2. Remove `UniqueTagList` from `AddressBook`.

I'm not sure whether this works, can you help me to investigate? Otherwise @Rinder5 can ownself investigate also :P

yamidark: Yea, I think it would be better to remove the tag list from `XmlSerializableAddressBook` first, then we have a stronger justification to removing the master tag list in `AddressBook` since it really won't be used anywhere anymore. Just remember to update the commit message for [3/3].

Another option would be to derive the tag list from all the persons in the `src` in [1/3], but this is too troublesome and don't think this is necessary though...

src/test/java/seedu/address/model/AddressBookTest.java
vivekscl: After removing line 69, don't forget to remove the `Collection<? extends Tag> tags` parameter in line 71 and the `newTags` argument in line 51 and 52.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -48,9 +47,6 @@ public XmlSerializableAddressBook(ReadOnlyAddressBook src) {
      */
```
vivekscl: Update header comment to remove `{@code XmlAdaptedTag}`.

vivekscl:
For the developer guide, I think Figure 9 needs to be updated after removal of the `UniqueTagList`.

Rinder5:
I'm not too sure how to handle a part of the developer guide.
In the section A.1. Improving each component, there is a section talking about the bug with the master tag list, which includes a PR:

> The current codebase has a flaw in tags management. Tags no longer in use by anyone may still exist on the AddressBook. This may cause some tests to fail. See issue #753 for more information about this flaw. 
> The solution PR has a temporary fix for the flaw mentioned above in its first commit.

Should I create a new PR with the same solution but without the bug fix?

yamgent:
> Should I create a new PR with the same solution but without the bug fix?

I will have to update that PR to remove the bug fix then. I think it is going to break anyway if the master list is going to be removed.

Zhiyuan-Amos:
> Should I create a new PR with the same solution but without the bug fix?

> I will have to update that PR to remove the bug fix then.

@yamgent @Rinder5 just to make sure we are on the same page: We don't intend to merge this PR in before releasing to the students, so there's no need to update the PR any time soon ^^

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -55,19 +45,12 @@ public void setPersons(List<Person> persons) throws DuplicatePersonException {
...
     public void resetData(ReadOnlyAddressBook newData) {
         requireNonNull(newData);
-        setTags(new HashSet<>(newData.getTagList()));
-        List<Person> syncedPersonList = newData.getPersonList().stream()
-                .map(this::syncWithMasterTagList)
-                .collect(Collectors.toList());
+        List<Person> syncedPersonList = newData.getPersonList();
 
         try {
             setPersons(syncedPersonList);
```
yamidark: As mentioned by @Zhiyuan-Amos, we can just directly use `newData.getPersonList()` in `setPersons`. Rather, naming the person list as `syncedPersonList` is also misleading since we are no longer syncing with the master tag list anymore.

docs/DeveloperGuide.adoc
```diff
@@ -208,6 +208,17 @@ The `Model`,
...
+
+===== Aspect: Implementation of `UniqueTagList`
+
+* **Alternative 1 (current choice):** Each Person manages their own `UniqueTagList`
+** Pros: Easier for new developers to understand.
+** Cons: Tags are duplicated many times.
+* **Alternative 2:** AddressBook has one `UniqueTagList`, each person has a list of tags which points to objects in this list.
+** Pros: Easy to edit tags across the board.
+** Cons: Hard to sync tags with master tag list when adding, editing, or removing people and tags.
+
```
yamidark: Don't think this should be put here, but at the Implementation section (3.). We could add a new section where we first explain the current implementation of `UniqueTagList`, and propose the other implementation with some of the code examples previously used to explain how it could be done.

@Zhiyuan-Amos @vivekscl your thoughts on this?

Zhiyuan-Amos: > Don't think this should be put here, but at the Implementation section (3.).

Yup, agreed. 

This segment definitely has to be more descriptive; I don't think that students will understand what it means that `AddressBook has one UniqueTagList, each person has a list of tags which points to objects in this list.` We may need to create diagrams for to illustrate the pointers (see the segment of Design Considerations for Undo/Redo).

Great work on mentioning this; I didn't think of adding this in initially.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -56,7 +56,6 @@ public void getDataFromFile_emptyFile_dataFormatMismatchException() throws Excep
     public void getDataFromFile_validFile_validResult() throws Exception {
         AddressBook dataFromFile = XmlUtil.getDataFromFile(VALID_FILE, XmlSerializableAddressBook.class).toModelType();
         assertEquals(9, dataFromFile.getPersonList().size());
-        assertEquals(0, dataFromFile.getTagList().size());
```
Zhiyuan-Amos: This change seems weird because an `AddressBook` still has a `UniqueTagList`, but we aren't testing for it.

I think we should change this test a little:
1. Instead of calling `toModelType()`, let's use `AddressBookBuilder` to create another `AddressBook`, then use it to construct an `expectedXmlAddressBook` object.
2. Let's use `typicalPersonAddressBook.xml` so that we can call `AddressBookBuilder#addPerson(Person)` more easily.
3. use `assertEquals(expectedXmlAddressBook, dataFromFile)`.

Zhiyuan-Amos: Though I wonder whether this should be changed in this commit because:
1. The test still passes.
1. `VALID_FILE` is not changed in this commit, so it's hard to understand why we are changing this test.
1. As mentioned above, this test can be better coded. 

I think we should change this in an earlier commit.

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -166,22 +109,16 @@ public String toString() {
...
         return other == this // short circuit if same object
                 || (other instanceof AddressBook // instanceof handles nulls
-                && this.persons.equals(((AddressBook) other).persons)
-                && this.tags.equalsOrderInsensitive(((AddressBook) other).tags));
+                && this.persons.equals(((AddressBook) other).persons));
     }
 
     @Override
     public int hashCode() {
         // use this method for custom fields hashing instead of implementing your own
```
Zhiyuan-Amos: this comment can be removed since it was referring to `Objects#hash(...)` which has been removed.

docs/DeveloperGuide.adoc
Zhiyuan-Amos: Perhaps @vivekscl can give better comments on this since he has been working on documentation for quite a bit :P

Zhiyuan-Amos:
[3/4] after removing all these methods, it seems like there's no need for UniqueTagList anymore. Let's remove this class woo.

You can do a next iteration first since there's still many changes to be made.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
```diff
@@ -67,7 +60,6 @@ public boolean equals(Object other) {
             return false;
         }
 
-        XmlSerializableAddressBook otherAb = (XmlSerializableAddressBook) other;
-        return persons.equals(otherAb.persons) && tags.equals(otherAb.tags);
+        return persons.equals(((XmlSerializableAddressBook) other).persons);
```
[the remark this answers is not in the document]
yamidark: >This can be shortened. See AddCommand#equals(Object)

Should we be changing the format of `XmlSerializableAddressBook#equals(Object)` to the `AddCommand#equals(Object)` style just because it only compares 1 field now, since there are other classes such as `XmlAdaptedTag` and `Prefix` that only compares 1 field and has this same style? 

Maybe we can set a standard when to use between the 2 styles and raise an issue to update the `equals(Object)` method in all classes as mentioned by @pyokagan [previously](https://github.com/se-edu/addressbook-level4/pull/792#discussion_r164665468)?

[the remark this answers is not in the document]
Zhiyuan-Amos: Ok you are right; let's not make too many changes in this commit then, we can leave it this way. 👍

docs/DeveloperGuide.adoc
vivekscl: Adding this to section 3 would indeed be a better idea. I don't think it should be in the Model section since it gives an overview of how it the Model component works and hence explaining implementation here would be a jump from a high-level explanation to a low-level one.   

>This segment definitely has to be more descriptive

Agreed. It should be more detailed as students would find it hard to understand how this actually works

On a side note, since we would end up removing the `UniqueTagList` as mentioned below, is this section still required?

Zhiyuan-Amos: > since we would end up removing the UniqueTagList as mentioned below, is this section still required?

Yes you are right, it seems that this section isn't required. The rationale should be described in the commit message and hyperlinked to the relevant discussion threads.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
Rinder5: sorry, do you mean earlier commit as in a different pr or as in a new section before I change the xml to remove tags?

Zhiyuan-Amos: Do it as the first commit.

Rinder5:
Sorry, still have things I need to add. Need to change the image and need to fix the test. Other than that, comments are welcome =)

Zhiyuan-Amos:
@Rinder5 you can make the fixes first :)

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -70,9 +78,19 @@ public void getDataFromFile_emptyFile_dataFormatMismatchException() throws Excep
 
     @Test
     public void getDataFromFile_validFile_validResult() throws Exception {
-        AddressBook dataFromFile = XmlUtil.getDataFromFile(VALID_FILE, XmlSerializableAddressBook.class).toModelType();
-        assertEquals(9, dataFromFile.getPersonList().size());
-        assertEquals(0, dataFromFile.getTagList().size());
+        AddressBook dataFromFile = XmlUtil
+                .getDataFromFile(TYPICAL_PERSON_FILE, XmlSerializableAddressBook.class)
+                .toModelType();
```
yamidark: Unnecessary line break here.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -20,8 +22,7 @@
     private final Email email;
     private final Address address;
 
-    private final UniqueTagList tags;
-
+    private final ObservableList<Tag> tagList = FXCollections.observableArrayList();
```
yamidark: Unnecessary removal of line break.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -32,7 +33,7 @@ public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tag
         this.email = email;
         this.address = address;
         // protect internal tags from changes in the arg list
-        this.tags = new UniqueTagList(tags);
+        this.tagList.addAll(tags);
```
yamidark: No need for `this`.

We should also add an `assert` that each element in the tag is unique.

Zhiyuan-Amos:
👍 good & thorough review @yamidark. I won't comment for now; leave it to you guys to sort things out first :P

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -70,9 +78,19 @@ public void getDataFromFile_emptyFile_dataFormatMismatchException() throws Excep
...
+        AddressBook expectedXmlAddressBook = new AddressBookBuilder()
+                .withPerson(ALICE)
+                .withPerson(BENSON)
+                .withPerson(CARL)
+                .withPerson(DANIEL)
+                .withPerson(ELLE)
+                .withPerson(FIONA)
+                .withPerson(GEORGE)
+                .build();
+        assertEquals(expectedXmlAddressBook, dataFromFile);
```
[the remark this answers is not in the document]
vivekscl: Hmm. Since we're testing `XmlUtil` we should have the necessary test data files under the `XmlUtil` folder. But then again, there is a duplication of files if we do that. 

But if we try to use files from the `XmlSerializableAddressBookTest` folder, we would have to create a second `TEST_DATA _FOLDER` variable solely for the `typicalPersonsAddressBook.xml` file.

Since file duplication is not good practice, we should avoid it by importing the file from `XmlSerializableAddressBookTest` folder instead?

src/main/java/seedu/address/model/person/Person.java
vivekscl: But since the tags are in a set, wouldn't that ensure that all the tags are unique? So there is no need to `assert` that each element in tags is unique?

yamidark: >But since the tags are in a set, wouldn't that ensure that all the tags are unique? So there is no need to assert that each element in tags is unique?

Hmm, what you're mentioning is right. Just that previously `UniqueTagList` had the assertion test for all unique tags so I felt it should be brought over here too. 
Maybe its because we were adding into a `List` which could allow duplicates, so the assertion is just extra precaution like most other assertions we have :O.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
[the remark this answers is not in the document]
yamidark: >Since file duplication is not good practice, we should avoid it by importing the file from XmlSerializableAddressBookTest folder instead?

Hmm, don't think we should import from `XmlSerializableAddressBookTest` for a test in `XmlUtilTest`. Rather, maybe we can create a new parent folder `XmlTest` and store each tests folders there, then we can also have common .xml files such as `typicalPersonsAddressBook.xml` without a specific folder in `XmlTest`. 

Your thoughts on this?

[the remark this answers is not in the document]
vivekscl: >Rather, maybe we can create a new parent folder XmlTest and store each tests folders there, then we can also have common .xml files such as typicalPersonsAddressBook.xml without a specific folder in XmlTest.

That's a much better idea :+1: .

src/main/java/seedu/address/model/person/Person.java
vivekscl: Oh, if it's for extra precaution and it's been approved of before, then I guess an assert would be required.

@Zhiyuan-Amos your thoughts?

Zhiyuan-Amos: Taking a look at the big picture, there's no need for us to use `ObservableList` to store the tags. Yes, they are used in `UniqueTagList`, but now, the observable property is not needed and there's no reason to use `List` over a `Set` here.

As such, if we are using a `Set` to store the tags, then it seems redundant to have the assertion that the tags in the `Set` are unique.

Rinder5:
@Zhiyuan-Amos @yamidark @vivekscl Thanks for all the feedback! I made the changes requested.

One thing I noticed is that the TypicalPersons xml file wasn't exactly correct. By default, the PersonBuilder seems to add the tag "friends" to everything if none is chosen, so by right everyone should get tagged with friends in the xml file. I added it in to make the tests in XmlUtil pass, but it's a bit awkward. 

If I just add it in by itself, it would be weird for me to edit a seemingly random file used by XmlSerializableAddressBookTest since it's not really within the scope of this pr. Otherwise, when I add it in to the first commit here for XmlUtil, it feels like I am doing too much, because I am changing the tests in xmlutil and fixing the typicalpersons xml file (both the tagging and @code tag for #822).

Any thoughts and ideas?

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -56,7 +57,7 @@ public Address getAddress() {
      * if modification is attempted.
      */
     public Set<Tag> getTags() {
-        return Collections.unmodifiableSet(tags.toSet());
+        return Collections.unmodifiableSet(new HashSet<>(tagList));
```
Zhiyuan-Amos: Is there a reason for using a copy constructor here?

Rinder5: I guess not. actually since now it's a set, we can just return the tagList itself, unless we want it to still be unmodifiable

Zhiyuan-Amos: Let's retain it to be unmodifiable, but let's remove the defensive copying.

1. Both achieve similar purposes: they allow Person to be immutable.
2. Unmodifiable has an additional property of throwing exception when the set is modified. We keep this property since there is no reason now where we need to get the Set of Person's tags and update it.

Zhiyuan-Amos:
@Rinder5 sorry for changing my mind, but I think that we shouldn't squash the commits. The commit organization should look something like:

1. Address Book: remove master tag list
2. Xml...# remove master tag list

I was thinking, suppose we dont have this feature in our code base yet. So, to create this feature, we will be doing:

1. Xml... Add master tag list
2. Address book: Add master tag list

As such, when we are removing this feature, we remove it in the opposite manner.
As such, apologies for the earlier suggestion to swap the commits, my bad :(

Rinder5:
No worries! Thanks for all the help =)

Zhiyuan-Amos:
Ah, so sorry for the extra work you have to do :( thanks!

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -20,8 +20,8 @@
     private final Email email;
     private final Address address;
 
-    private final UniqueTagList tags;
 
+    private final Set<Tag> tagList = FXCollections.observableSet();
```
Zhiyuan-Amos: there's no need to rename this. Also, there's no need for `observableSet()`. As such, you shouldn't initialize the tags here.

docs/DeveloperGuide.adoc
```diff
@@ -553,7 +553,7 @@ Do take a look at <<Design-Model>> before attempting to modify the `Model` compo
 ** Add a new API method `deleteTag(Tag)` in link:{repoURL}/src/main/java/seedu/address/model/ModelManager.java[`ModelManager`]. Your link:{repoURL}/src/main/java/seedu/address/model/ModelManager.java[`ModelManager`] should call `AddressBook#removeTag(Tag)`.
 ** Add new tests for each of the new public methods that you have added.
 ** See this https://github.com/se-edu/addressbook-level4/pull/790[PR] for the full solution.
-*** The current codebase has a flaw in tags management. Tags no longer in use by anyone may still exist on the link:{repoURL}/src/main/java/seedu/address/model/AddressBook.java[`AddressBook`]. This may cause some tests to fail. See issue  https://github.com/se-edu/addressbook-level4/issues/753[`#753`] for more information about this flaw.
+*** The previous codebase had a flaw in tags management. Tags no longer in use by anyone may still exist on the link:{repoURL}/src/main/java/seedu/address/model/AddressBook.java[`AddressBook`]. This may cause some tests to fail. See issue  https://github.com/se-edu/addressbook-level4/issues/753[`#753`] for more information about this flaw.
```
Zhiyuan-Amos: Hmm, once we've fixed this, there's no need for us to document this segment anymore.

src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml
Zhiyuan-Amos: it isn't clear why this is in this commit. Should it be in the first commit instead?

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -55,22 +45,14 @@ public void setPersons(List<Person> persons) throws DuplicatePersonException {
...
     }
 
-    public void setTags(Set<Tag> tags) {
-        this.tags.setTags(tags);
-    }
-
     /**
      * Resets the existing data of this {@code AddressBook} with {@code newData}.
      */
     public void resetData(ReadOnlyAddressBook newData) {
```
Zhiyuan-Amos: this method is now entirely similar with `setPersons(List<Person> persons)`.

src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
Zhiyuan-Amos: This line doesn't need to be removed here. It should be removed as part of the 2nd commit.

Zhiyuan-Amos:
@vivekscl @yamidark do take a look and get a feel of consistently doing code reviews for a PR :P

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -36,13 +27,12 @@
      */
     {
         persons = new UniquePersonList();
-        tags = new UniqueTagList();
```
vivekscl: In this commit, the master tag list in `AddressBook` is being removed. The removal of `UniqueTagList` is done in the 3rd commit, so shouldn't this be in the 3rd commit?

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -20,8 +20,8 @@
     private final Email email;
     private final Address address;
 
-    private final UniqueTagList tags;
```
vivekscl: Too many whitespaces.

src/main/java/seedu/address/model/AddressBook.java
Zhiyuan-Amos: The idea behind this commit organization is: Suppose you are trying to create `UniqueTagList` feature for the app. Your commit organization will be something like:

1. Create UniqueTagList
2. Teach XmlSerializableAddressBook to read and write tags from and to UniqueTagList
3. Teach AddressBook to read from XmlSerializableAddressBook

As such, when we are tearing down this feature, we work from the opposite direction (from 3 to 1). Though I agree with you that the 3rd commit message has to be better.

> Many methods in UniqueTagList such as mergeFrom(UniqueTagList) and
equalsOrderInsensitive(UniqueTagList) are unused. The methods in use are equivalent to that of a Set.

> There's no reason to use UniqueTagList over a Set.

> Let's remove UniqueTagList, and replace usage of UniqueTagList with Set.

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
yamidark: Since `tags` in `XmlSerializableAddressBook` still exist in [1/4], this test should be removed in [2/4] instead.

src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml
yamidark: I believe this should be removed in this commit, since the test `toModelType_invalidTagFile_throwsIllegalValueException()` that was removed in [1/4] should be removed in this commit instead along with this file.

src/test/java/seedu/address/model/UniqueTagListTest.java
```diff
@@ -1,19 +0,0 @@
...
-import org.junit.rules.ExpectedException;
-
-import seedu.address.model.tag.UniqueTagList;
-
-public class UniqueTagListTest {
-    @Rule
-    public ExpectedException thrown = ExpectedException.none();
-
-    @Test
-    public void asObservableList_modifyList_throwsUnsupportedOperationException() {
```
yamidark: Should we transfer this test over to another test class under `seedu.address.model.person` as `TagsTest`?

Zhiyuan-Amos: I think it should be `PersonTest`, whereby we test for `getTags()`

Zhiyuan-Amos:
If we're removing `resetData(ReadOnlyAddressBook)`, then we are resolving #658 and #659 too.

Zhiyuan-Amos:
I also noticed that `AddressBook` class is kinda useless, since `AddressBook` methods are all 1-liners; they just call the methods in `UniquePersonList`.

yamgent:
> I also noticed that AddressBook class is kinda useless, since AddressBook methods are all 1-liners; they just call the methods in UniquePersonList.

I would think that CS2103T would appreciate having `AddressBook` already in the codebase when they inherit it, rather than having to start refactoring and restoring back `AddressBook` should we decide to remove it. 

Perhaps we should try making it more useful by adding a feature or two? (E.g. Making tags re-nameable, or the capability of conjuring all `Person` s with a particular tag).

Zhiyuan-Amos:
> I would think that CS2103T would appreciate having AddressBook already in the codebase when they inherit it, rather than having to start refactoring and restoring back AddressBook should we decide to remove it.

Ah I see. I get it that if we have an `AddressBook`, it's easier to add additional stuffs alongside with `UniquePersonList`. However, I'm wondering whether students actually add things alongside with `UniquePersonList` :P 

> Perhaps we should try making it more useful by adding a feature or two? (E.g. Making tags re-nameable, or the capability of obtaining all Person s with a particular tag).

Yup, but unfortunately I think we don't have time to do it now. But this is a good suggestion.

@damithc, do you have opinions on this?

damithc:
> @damithc, do you have opinions on this?

Let's keep it for now. Note that we also want to mention the original design in the developer guide as a 'more OOP' solution. Easier to do that if the `AddressBook` class remains.

docs/DeveloperGuide.adoc
Rinder5: I'm worried that if I remove it, the solution PR might not make sense, because the first commit deals with the issue. I'll remove it for now though.

src/test/java/seedu/address/model/UniqueTagListTest.java
Rinder5: I don't think we currently have a PersonTest, should I make one?

src/main/java/seedu/address/model/AddressBook.java
Rinder5: it takes in a different parameter and deals with the thrown exception though. should i overload setpersons and deal with the exception in where they call this?

src/test/java/seedu/address/model/UniqueTagListTest.java
Zhiyuan-Amos: yup! You don't need to write tests to verify everything in `Person`; `getTags()` will do.

docs/DeveloperGuide.adoc
Zhiyuan-Amos: Ah I see. Don't worry about that, the solution PR will be rebased by the next semester :) So, we can remove this for now.

src/main/java/seedu/address/model/AddressBook.java
Zhiyuan-Amos: Hmm you're right. Let's leave it here then so that students can simply add code to reset `AddressBook` after they have added more fields into `AddressBook`.

Zhiyuan-Amos:
Let's make these changes in the first commit. 

1. `addTag(Tag)` currently does nothing which is an unintended behavior (you would expect `addTag(Tag)` to add the `tag` to the master tag list). So, let's remove the method altogether and update `XmlSerializableAddressBook` to remove the `addTag(Tag)` call. This means we need to remove a test too.

1. `getTagList()` is not functioning properly too. So, we just write simple code to get all the tags from all the `persons`.

1. Following the above changes, we can remove `tags` from `AddressBook` in the first commit.

```diff
diff --git a/src/main/java/seedu/address/model/AddressBook.java b/src/main/java/seedu/address/model/AddressBook.java
index 0415d42..a9b7661 100644
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -2,8 +2,13 @@ package seedu.address.model;
 
 import static java.util.Objects.requireNonNull;
 
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import seedu.address.model.person.Person;
 import seedu.address.model.person.UniquePersonList;
@@ -19,7 +24,6 @@ import seedu.address.model.tag.UniqueTagList;
 public class AddressBook implements ReadOnlyAddressBook {
 
     private final UniquePersonList persons;
-    private final UniqueTagList tags;
 
     /*
      * The 'unusual' code block below is an non-static initialization block, sometimes used to avoid duplication
@@ -30,7 +34,6 @@ public class AddressBook implements ReadOnlyAddressBook {
      */
     {
         persons = new UniquePersonList();
-        tags = new UniqueTagList();
     }
 
     public AddressBook() {}
@@ -95,11 +98,6 @@ public class AddressBook implements ReadOnlyAddressBook {
         persons.remove(key);
     }
 
-    //// tag-level operations
-
-    public void addTag(Tag t) throws UniqueTagList.DuplicateTagException {
-    }
-
     //// util methods
 
     @Override
@@ -115,7 +113,10 @@ public class AddressBook implements ReadOnlyAddressBook {
 
     @Override
     public ObservableList<Tag> getTagList() {
-        return tags.asObservableList();
+        Set<Tag> tagSet = new HashSet<>();
+        persons.forEach(person -> tagSet.addAll(person.getTags()));
+        ObservableList<Tag> tagList = FXCollections.observableArrayList(tagSet);
+        return FXCollections.unmodifiableObservableList(tagList);
     }
 
     @Override
diff --git a/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java b/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
index dc82089..03ab4fd 100644
--- a/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
+++ b/src/main/java/seedu/address/storage/XmlSerializableAddressBook.java
@@ -48,9 +48,6 @@ public class XmlSerializableAddressBook {
      */
     public AddressBook toModelType() throws IllegalValueException {
         AddressBook addressBook = new AddressBook();
-        for (XmlAdaptedTag t : tags) {
-            addressBook.addTag(t.toModelType());
-        }
         for (XmlAdaptedPerson p : persons) {
             addressBook.addPerson(p.toModelType());
         }
diff --git a/src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml b/src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml
deleted file mode 100644
index 5fa697c..0000000
--- a/src/test/data/XmlSerializableAddressBookTest/invalidTagAddressBook.xml
+++ /dev/null
@@ -1,5 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<addressbook>
-    <!-- Invalid tag -->
-    <tags>frie!nds</tags>
-</addressbook>
diff --git a/src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java b/src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
index cf405c9..9985299 100644
--- a/src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
+++ b/src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
@@ -19,7 +19,6 @@ public class XmlSerializableAddressBookTest {
     private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/XmlSerializableAddressBookTest/");
     private static final File TYPICAL_PERSONS_FILE = new File(TEST_DATA_FOLDER + "typicalPersonsAddressBook.xml");
     private static final File INVALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "invalidPersonAddressBook.xml");
-    private static final File INVALID_TAG_FILE = new File(TEST_DATA_FOLDER + "invalidTagAddressBook.xml");
 
     @Rule
     public ExpectedException thrown = ExpectedException.none();
@@ -40,12 +39,4 @@ public class XmlSerializableAddressBookTest {
         thrown.expect(IllegalValueException.class);
         dataFromFile.toModelType();
     }
-
-    @Test
-    public void toModelType_invalidTagFile_throwsIllegalValueException() throws Exception {
-        XmlSerializableAddressBook dataFromFile = XmlUtil.getDataFromFile(INVALID_TAG_FILE,
-                XmlSerializableAddressBook.class);
-        thrown.expect(IllegalValueException.class);
-        dataFromFile.toModelType();
-    }
 }
```

I'll stop reviewing here first. :) Just make the above changes and request for a review again.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -1,19 +1,20 @@
...
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
-import seedu.address.model.tag.UniqueTagList;
+import seedu.address.testutil.PersonBuilder;
 
-public class UniqueTagListTest {
+
```
Zhiyuan-Amos: extra new line here.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -31,8 +29,7 @@ public Person(Name name, Phone phone, Email email, Address address, Set<Tag> tag
         this.phone = phone;
         this.email = email;
         this.address = address;
-        // protect internal tags from changes in the arg list
-        this.tags = new UniqueTagList(tags);
+        this.tags = tags;
```
Zhiyuan-Amos: we need a defensive copy here, otherwise the caller can externally modify the `tags` in this object.
Alternatively, we can do an `addAll(Collection)` just as the code in `UniqueTagList`'s constructor.

Zhiyuan-Amos:
Also do take note to include what prof mentioned [above](https://github.com/se-edu/addressbook-level4/pull/825#issuecomment-375554311).

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -20,8 +19,7 @@
     private final Email email;
     private final Address address;
 
-    private final UniqueTagList tags;
-
+    private final Set<Tag> tags;
     /**
```
yamidark: Should have a new line here to separate the variable from the comments.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -62,22 +59,15 @@ public void getPersonList_modifyList_throwsUnsupportedOperationException() {
         addressBook.getPersonList().remove(0);
     }
 
-    @Test
-    public void getTagList_modifyList_throwsUnsupportedOperationException() {
```
Zhiyuan-Amos: This test can remain in commit 1 and only deleted in future commits, since getTagList() has such a behavior now right?

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -29,7 +28,6 @@
     @Test
     public void constructor() {
         assertEquals(Collections.emptyList(), addressBook.getPersonList());
-        assertEquals(Collections.emptyList(), addressBook.getTagList());
```
Zhiyuan-Amos: This doesn't have to be removed in commit 1

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
```diff
@@ -39,5 +40,4 @@ public void toModelType_invalidPersonFile_throwsIllegalValueException() throws E
         thrown.expect(IllegalValueException.class);
         dataFromFile.toModelType();
     }
-
```
Zhiyuan-Amos: Unnecessary change

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
```diff
@@ -19,6 +19,7 @@
     private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/XmlSerializableAddressBookTest/");
     private static final File TYPICAL_PERSONS_FILE = new File(TEST_DATA_FOLDER + "typicalPersonsAddressBook.xml");
     private static final File INVALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "invalidPersonAddressBook.xml");
+    private static final File INVALID_TAG_FILE = new File(TEST_DATA_FOLDER + "invalidTagAddressBook.xml");
```
Zhiyuan-Amos: Why are we adding this back in commit 2?

Zhiyuan-Amos:
TypicalPersonsAddressBook.xml has a "master tag list", which should be removed in commit 2 too right?

docs/DeveloperGuide.adoc
```diff
@@ -202,6 +202,9 @@ The `Model`,
 * exposes an unmodifiable `ObservableList<Person>` that can be 'observed' e.g. the UI can be bound to this list so that the UI automatically updates when the data in the list change.
 * does not depend on any of the other three components.
 
+[NOTE]
+As a more OOP model, we can store a Tag list in Address Book, which Persons can reference. This would allow Address Book to only require one Tag object per unique Tag, instead of each Person needing their own Tag object.
```
Zhiyuan-Amos: I think we should use images created using Powerpoint to help students to visualise this :p

src/main/java/seedu/address/commons/events/model/AddressBookChangedEvent.java
```diff
@@ -14,6 +14,6 @@ public AddressBookChangedEvent(ReadOnlyAddressBook data) {
 
     @Override
     public String toString() {
-        return "number of persons " + data.getPersonList().size() + ", number of tags " + data.getTagList().size();
+        return "number of persons " + data.getPersonList().size();
```
yamidark: Since in [1/4] we still have `AddressBook#getTagList()`, this should be removed in [2/4] instead.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -64,7 +63,6 @@ public void getPersonList_modifyList_throwsUnsupportedOperationException() {
      */
     private static class AddressBookStub implements ReadOnlyAddressBook {
         private final ObservableList<Person> persons = FXCollections.observableArrayList();
-        private final ObservableList<Tag> tags = FXCollections.observableArrayList();
```
yamidark: Since [1/4] should be removing the master tag list, maybe we can remove this in [1/4], and replace the `AddressBookStub#getTagList()` method to directly return `FXCollections.observableArrayList()` instead?

Zhiyuan-Amos: If we do so, the `AddressBookStub#getTagList()` will be returning wrong values.

I think it's fine to leave it here since this is a test class. A `Stub` is meant to have a simpler implementation of the actual class that it is stubbing for, so instead of having the logic of gathering all the `Person`s tags and then returning it (see [1/4], `AddressBook#getTagList()`), we let `AddressBookStub` simply store the tags in a field, and we trust that the tester will pass in the correct tags.

src/test/java/seedu/address/model/AddressBookTest.java
```diff
@@ -69,15 +67,14 @@ public void getTagList_modifyList_throwsUnsupportedOperationException() {
...
+     * A stub ReadOnlyAddressBook whose persons list can violate interface constraints.
      */
     private static class AddressBookStub implements ReadOnlyAddressBook {
         private final ObservableList<Person> persons = FXCollections.observableArrayList();
         private final ObservableList<Tag> tags = FXCollections.observableArrayList();
 
-        AddressBookStub(Collection<Person> persons, Collection<? extends Tag> tags) {
+        AddressBookStub(Collection<Person> persons) {
             this.persons.setAll(persons);
-            this.tags.setAll(tags);
```
Zhiyuan-Amos: Following what I mentioned [earlier](https://github.com/se-edu/addressbook-level4/pull/825#discussion_r177418990), you should make changes to this file only in the 2nd commit. Otherwise, your `tags` is not functioning correctly (it doesn't store anything even when `persons` have tags).

Zhiyuan-Amos: @Rinder5 you didn't do this in v.12 :p What I'm saying is that this file shouldn't be updated at all in commit 1, only in commit 2, otherwise `AddressBookStub` isn't functioning properly.

Rinder5: oh my bad, was reading that wrongly >.<

docs/DeveloperGuide.adoc
```diff
@@ -202,6 +202,11 @@ The `Model`,
 * exposes an unmodifiable `ObservableList<Person>` that can be 'observed' e.g. the UI can be bound to this list so that the UI automatically updates when the data in the list change.
 * does not depend on any of the other three components.
 
+[NOTE]
+As a more OOP model, we can store a Tag list in Address Book, which Persons can reference. This would allow Address Book to only require one Tag object per unique Tag, instead of each Person needing their own Tag object. An example of how such a model may look like is given below.
+
+image::ModelClassBetterOopDiagram.png[width="800"]
```
damithc: Is it possible to make this image appear inside the note?

Rinder5:
@damithc I was able to add the picture to the note, but now the bottom line for the note does not appear. Thoughts?

damithc:
> @damithc I was able to add the picture to the note, but now the bottom line for the note does not appear. Thoughts?

Not sure. Haven't used asciidoc myself much. Anyway, this is how it it looks now. What line is missing?
image

Rinder5:
@damithc It might be rendering it differently for different settings.

This is what I see now:

image

Earlier, the red circled line was missing when I checked

Zhiyuan-Amos:
@Rinder5 please do a last call for reviews for your PR :) finally your PRs are getting merged! :P

Zhiyuan-Amos:
@Rinder5 your [post](https://github.com/se-edu/addressbook-level4/pull/825#issue-168198136) isn't updated yet, since #822 isn't resolved here :p

PR #832, opened 2018-02-16
HelpWindow: help window stays open even after exiting by clicking the close button #826

### LINKED ISSUE #826: HelpWindow: help window stays open even after exiting by clicking the close button
This should not be the case as closing the app should close every other window related to it.

Edit: Likely to happen on certain operating systems only. So far, it seems to be happening on Linux.

yamidark:
Fixes #826 
```
The MainWindow does not have an event handler to handle external
requests (such as the close button or through the taskbar) to close
the MainWindow.

This may cause the app to not close properly since MainApp#stop() is
not called as the ExitAppRequestEvent is not raised. MainApp#stop() is
the method used to call Platform#exit() and System#exit(int) which
closes our app and all remaining open windows. Thus, if MainApp#stop()
is not called, the app may continue to run even after the MainWindow is
closed, causing existing HelpWindows to remain open even when the user
expects the app to stop.

According to the life-cycle section in the documentation on JavaFX
Application[1], the app will only finish and call MainApp#stop() when
the app either calls Platform#exit() or when the last window has been
closed. Thus, should there be any HelpWindows still opened, the app
will continue running, causing those HelpWindows to remain open.

Let's add an event handler for the onCloseRequest event in the root of
MainWindow.fxml. This event will call MainWindow#handleExit(), which
raises an ExitAppRequestEvent, when there is an external request to
close the app.

EmptyMainWindowHandle does not contain a close button, thus to test if
an ExitAppRequestEvent was raised, we fire a
WindowEvent#WINDOW_CLOSE_REQUEST event which will trigger the
onCloseRequest event.

[1] JavaFx documentation for Application:
https://docs.oracle.com/javafx/2/api/javafx/application/Application.html
```

### DIFF
--- a/src/main/resources/view/MainWindow.fxml
+++ b/src/main/resources/view/MainWindow.fxml
@@ -12,7 +12,7 @@
 <?import javafx.scene.layout.VBox?>
 
 <fx:root type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1"
-         minWidth="450" minHeight="600">
+         minWidth="450" minHeight="600" onCloseRequest="#handleExit">
   <icons>
     <Image url="@/images/address_book_32.png" />
   </icons>
--- a/src/test/java/seedu/address/ui/MainWindowCloseTest.java
+++ b/src/test/java/seedu/address/ui/MainWindowCloseTest.java
@@ -1,0 +1,83 @@
+package seedu.address.ui;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.testfx.api.FxToolkit;
+
+import guitests.guihandles.StageHandle;
+import javafx.stage.Stage;
+import javafx.stage.WindowEvent;
+import seedu.address.commons.core.Config;
+import seedu.address.commons.events.ui.ExitAppRequestEvent;
+import seedu.address.logic.LogicManager;
+import seedu.address.model.ModelManager;
+import seedu.address.model.UserPrefs;
+import seedu.address.ui.testutil.EventsCollectorRule;
+
+/**
+ * Contains tests for closing of the {@code MainWindow}.
+ */
+public class MainWindowCloseTest extends GuiUnitTest {
+    @Rule
+    public final EventsCollectorRule eventsCollectorRule = new EventsCollectorRule();
+
+    private MainWindow mainWindow;
+    private EmptyMainWindowHandle mainWindowHandle;
+    private Stage stage;
+
+    @Before
+    public void setUp() throws Exception {
+        FxToolkit.setupStage(stage -> {
+            this.stage = stage;
+            mainWindow = new MainWindow(stage, new Config(), new UserPrefs(), new LogicManager(new ModelManager()));
+            mainWindowHandle = new EmptyMainWindowHandle(stage);
+
+            stage.setScene(mainWindow.getRoot().getScene());
+            mainWindowHandle.focus();
+        });
+        FxToolkit.showStage();
+    }
+
+    @Test
+    public void close_menuBarExitButton_exitAppRequestEventPosted() {
+        mainWindowHandle.clickOnMenuExitButton();
+        assertTrue(eventsCollectorRule.eventsCollector.getMostRecent() instanceof ExitAppRequestEvent);
+        assertTrue(eventsCollectorRule.eventsCollector.getSize() == 1);
+    }
+
+    @Test
+    public void close_externalRequest_exitAppRequestEventPosted() {
+        mainWindowHandle.closeMainWindowExternally();
+        assertTrue(eventsCollectorRule.eventsCollector.getMostRecent() instanceof ExitAppRequestEvent);
+        assertTrue(eventsCollectorRule.eventsCollector.getSize() == 1);
+    }
+
+    /**
+     * A handle for an empty {@code MainWindow}. The components in {@code MainWindow} are not initialized.
+     */
+    private class EmptyMainWindowHandle extends StageHandle {
+
+        private EmptyMainWindowHandle(Stage stage) {
+            super(stage);
+        }
+
+        /**
+         * Closes the {@code MainWindow} by clicking on the menu bar's exit button.
+         */
+        private void clickOnMenuExitButton() {
+            guiRobot.clickOn("File");
+            guiRobot.clickOn("Exit");
+        }
+
+        /**
+         * Closes the {@code MainWindow} through an external request {@code MainWindow} (e.g pressing the 'X' button on
+         * the {@code MainWindow} or closing the app through the taskbar).
+         */
+        private void closeMainWindowExternally() {
+            guiRobot.interact(() -> stage.fireEvent(new WindowEvent(stage, WindowEvent.WINDOW_CLOSE_REQUEST)));
+        }
+    }
+}


yamidark:
>The title of the commit message could give a better overview instead? Like for example, MainWindow: add an event handler for the close button?

Kind of followed the style I have been using, where the commit title just states what I'm doing, and the other parts of the message explains what it is and why we need it.

>any other external method sounds a bit unclear to me.

We technically can close the `MainWindow` via other means like closing it in the taskbar below which also result in the bug. Not too sure how to phrase this better though, maybe we can just give the 1 example I mentioned earlier?

>This may cause the app to not close properly
To be more specific, only the HelpWindow doesn't close, right?

The app itself also doesn't stop running which is the reason why the `HelpWindow` does not close. 
You can try just clicking the close button with `HelpWindow` open, and the logger on the console does not state that `MainApp#stop()` is called, only after the `HelpWindow` is closed does `Mainapp#stop()` get called.

>The 4th paragraph seems to have gone into too much detail such as explaining what handleExit() does

Hmmm, yep probably don't need to go into detail what `handleExit()` does.

vivekscl:
>Kind of followed the style I have been using, where the commit title just states what I'm doing, and the other parts of the message explains what it is and why we need it.

Oh, in that case, it should be fine :+1: 

>We technically can close the MainWindow via other means like closing it in the taskbar below which also result in the bug. Not too sure how to phrase this better though, maybe we can just give the 1 example I mentioned earlier?

You could add the example and phrase it like, `There is no event handler for closing the MainWindow via external methods such as the close button or the taskbar`.

>The app itself also doesn't stop running which is the reason why the HelpWindow does not close.
You can try just clicking the close button with HelpWindow open, and the logger on the console does not state that MainApp#stop() is called, only after the HelpWindow is closed does Mainapp#stop() get called.

Oh, didn't notice that, my bad :sweat_smile:

src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -112,6 +113,16 @@ private void setAccelerator(MenuItem menuItem, KeyCombination keyCombination) {
         });
     }
 
+    /**
+     * Sets an event handler for the app to stop when the {@code MainWindow}'s close button is pressed.
+     */
+    private void setCloseButtonEvent() {
+        getPrimaryStage().setOnCloseRequest(event -> {
```
yamgent: If some other parts of the application calls `getPrimaryStage().close();`, it does not trigger `OnCloseRequest`. Perhaps we should use `setOnHiding` instead? ([Source](https://stackoverflow.com/questions/44548460/javafx-stage-close-event-handler))

yamgent:
Also I wonder whether we should perhaps use `fxml` to do our deed instead of writing some more code...

```diff
diff --git a/src/main/resources/view/MainWindow.fxml b/src/main/resources/view/MainWindow.fxml
index 1dadb95b..daf386d8 100644
--- a/src/main/resources/view/MainWindow.fxml
+++ b/src/main/resources/view/MainWindow.fxml
@@ -12,7 +12,7 @@
 <?import javafx.scene.layout.VBox?>

 <fx:root type="javafx.stage.Stage" xmlns="http://javafx.com/javafx/8" xmlns:fx="http://javafx.com/fxml/1"
-         minWidth="450" minHeight="600">
+         minWidth="450" minHeight="600" onHiding="#handleExit">
   <icons>
     <Image url="@/images/address_book_32.png" />
   </icons>
```

src/main/java/seedu/address/ui/MainWindow.java
yamidark: >If some other parts of the application calls getPrimaryStage().close();, it does not trigger OnCloseRequest. Perhaps we should use setOnHiding instead?

I wanted to differentiate between closing the app internally (via `ExitCommand` or the exit menu button) and externally (close button or from taskbar). 
If we use `setOnHiding`, `handleExit()` actually gets called twice if we close it internally since `setOnHiding` would call `handleExit()` for the 2nd time after`MainApp.stop()` calls `mainWindow.hide()` in `ui.stop()`.

yamidark:
>Also I wonder whether we should perhaps use fxml to do our deed instead of writing some more code...

Wanted to do that too, but didn't know if it could be done and my `MainWindow.fxml` doesn't seem to open in Scenebuilder. Yep, doing this in the `.fxml` would definitely be better.

yamgent:
> my MainWindow.fxml doesn't seem to open in Scenebuilder

Ah okay. IIRC SceneBuilder has never been working even when I was a cs2103t student, so don't worry about it. :P SceneBuilder has strict requirements on how we structure our UI codes (e.g. controllers) and widgets (e.g. placeholders), so loading problems are probably not unheard of.

Anyway, you can also spot possible attributes in the [JavaFX docs under Property Summary](https://docs.oracle.com/javase/8/javafx/api/javafx/stage/Window.html#property.summary) as well. `onCloseRequest="#handleExit"` is probably what you are looking for.

Zhiyuan-Amos:
Can you call some method in `guiRobot` to move the cursor to click on the `x` button to close the window? That way we can write regression tests for it.

yamidark:
>Can you call some method in guiRobot to move the cursor to click on the x button to close the window? That way we can write regression tests for it.

Couldn't find any method to move it to the x button, tried manually shifting the cursor there by using `getWidth()` , but same problem as closing the window via `ExitCommand` or the menu item, the test class terminates after it is closed, so I don't think this can be tested :X.

Zhiyuan-Amos:
>  but same problem as closing the window via ExitCommand or the menu item, the test class terminates after it is closed

Take a look at the [2nd answer](https://stackoverflow.com/questions/309396/java-how-to-test-methods-that-call-system-exit) and [here](http://neverfear.org/blog/view/157/Testing_code_that_calls_System_exit_in_Java) as well. 

It seems like the main thing we are testing for is that whenever we click on the `x` button, `System#exit(int)` is called (if I'm reading correctly, the reason why `HelpWindow` closes when we click on `x` button is because `System#exit(int)` is called right?).

yamidark:
>Take a look at the 2nd answer and here as well.

Probably should have searched for this `System.exit()` test :O.

Anyway, I tried this extending of the `SecurityManager` but the test now hangs instead of terminating after we try any methods to exit the app.
Seems like [this is because we call `Platform.exit()` first](https://github.com/TestFX/TestFX/issues/247).
Can't seem to find a way catch and stop this `Platform.exit()`, so for this test should we override `MainApp.stop()` in `TestApp` to not call `Platform.exit()` instead?

Zhiyuan-Amos:
> Can't seem to find a way catch and stop this Platform.exit(), so for this test should we override MainApp.stop() in TestApp to not call Platform.exit() instead?

I don't think we can override `TestApp` for this test alone right? I.e. this update will affect all the other tests as well.

Can we try [System Rules](https://github.com/stefanbirkner/system-rules) (also in the link I posted above)?

yamidark:
>Can we try System Rules (also in the link I posted above)?

System Rules needs at least [JUnit 4.9](http://stefanbirkner.github.io/system-rules/) while ours is 4.12(?) :X.

yamidark:
Just tested the `System Rules`, since the `ExpectedSystemExit` also uses an extended [SecurityManager](https://github.com/stefanbirkner/system-rules/blob/master/src/main/java/org/junit/contrib/java/lang/system/internal/NoExitSecurityManager.java), the same problem occurs and the test hangs.

Zhiyuan-Amos:
Hmm ok, document this down as part of the commit message then ^^ 

> This may cause the app to not close properly since it will not raise a
ExitAppRequestEvent, in turn MainApp#stop() is not called and the app
may continue to run.

Commit message needs to be clearer
1. `MainApp#stop()` calls `System#exit()` (or is it `Platform#exit()`?) which then closes all the remaining windows.
1. App continues to run -> Causing existing windows to remain open.

Zhiyuan-Amos:
> Can't seem to find a way catch and stop this Platform.exit(), so for this test should we override MainApp.stop() in TestApp to not call Platform.exit() instead?

If you override the method in `TestApp`, then you aren't testing the actual class `MainApp` isn't it :P I.e. you have no guarantee whether `MainApp` is calling `System#exit(int)` or not.

Can we do unit test for `MainWindow`, particularly for the method `#handleExit()`? We just check that whenever you press `x` or press the `Exit` button, this `ExitAppRequestEvent` is raised.

Ideally, we should have an integration test between `MainWindow` & `MainApp`. However, that doesn't seem possible in this scenario.

yamidark:
>Showing the link to the API of SecurityManager without explaining in the commit message isn't helpful :P

Hmm, so I should explain how we can use `SecurityManager` to catch this `System.exit(int)`, and maybe instead of the API link which isn't really useful I can show the link to the stackoverflow post?

>No need for line break for website links, otherwise the hyperlink is unclickable.

Whoops, didn't notice that. Will update in next iteration.

>If you override the method in TestApp, then you aren't testing the actual class MainApp isn't it :P I.e. you have no guarantee whether MainApp is calling System#exit(int) or not.

Now that you mention this, yep, dunno what I was thinking xD.

>Can we do unit test for MainWindow, particularly for the method #handleExit()? We just check that whenever you press x or press the Exit button, this ExitAppRequestEvent is raised.

Will check to see how we can add this test. Should this be done in this PR though?

Zhiyuan-Amos:
> Hmm, so I should explain how we can use SecurityManager to catch this System.exit(int), and maybe instead of the API link which isn't really useful I can show the link to the stackoverflow post?

The link to the SO post is more helpful. 

> Can we do unit test for MainWindow, particularly for the method #handleExit()? We just check that whenever you press x or press the Exit button, this ExitAppRequestEvent is raised.

> Will check to see how we can add this test. Should this be done in this PR though?

Yup it should be done in this PR since this PR updates `MainWindow.fxml` `onCloseRequest="#handleExit"`, so you can write unit tests for this addition.

Zhiyuan-Amos:
> Can we do unit test for MainWindow, particularly for the method #handleExit()? We just check that whenever you press x or press the Exit button, this ExitAppRequestEvent is raised.

> Yup it should be done in this PR since this PR updates MainWindow.fxml onCloseRequest="#handleExit", so you can write unit tests for this addition.

As discussed verbally, my bad, we can't do that because it depends on who catches the `ExitAppRequestEvent` first. If `MainWindow` catches it first, then the app closes. If the test catches it first, then the test passes.

Zhiyuan-Amos:
@yamidark I tried this:

```java
public class MainWindowTest extends GuiUnitTest {
    private MainWindow mainWindow;
    private MainWindowHandleTest mainWindowHandle;

    @Before
    public void setUp() throws Exception {
        FxToolkit.setupStage((stage) -> {
            mainWindow =
                    new MainWindow(stage, new Config(), new UserPrefs(), new LogicManager(new ModelManager()));
            stage.setScene(mainWindow.getRoot().getScene());
            mainWindowHandle = new MainWindowHandleTest(stage);
            mainWindowHandle.focus();
        });
        FxToolkit.showStage();
    }

    @Test
    public void temp() {
        // exit program and catch ExitAppRequestEvent
    }

    private class MainWindowHandleTest extends StageHandle {
        public MainWindowHandleTest(Stage stage) {
            super(stage);
        }
    }
}
```

Seems to be able to work. What do you think?

yamidark:
```java
@Before
    public void setUp() throws Exception {
        FxToolkit.setupStage(stage -> {
            mainWindow =
                    new MainWindow(stage, new Config(), new UserPrefs(), new LogicManager(new ModelManager()));
            stage.setScene(mainWindow.getRoot().getScene());
            mainWindowHandle = new MainWindowHandleTest(stage);
            mainWindowHandle.focus();
        });
    }

    @Test
    public void temp() throws Exception {
        FxToolkit.setupStage(stage ->
                stage.fireEvent(new WindowEvent(mainWindow.getPrimaryStage(), WindowEvent.WINDOW_CLOSE_REQUEST)));
        assertTrue(eventsCollectorRule.eventsCollector.getMostRecent() instanceof ExitAppRequestEvent);
        assertTrue(eventsCollectorRule.eventsCollector.getSize() == 1);
    }
```
@Zhiyuan-Amos How does this look?

Zhiyuan-Amos:
This part looks weird though :P You shouldn't setup a new stage but use `mainWindow.getPrimaryStage()` right?

```
FxToolkit.setupStage(stage ->
    stage.fireEvent(new WindowEvent(mainWindow.getPrimaryStage(), WindowEvent.WINDOW_CLOSE_REQUEST)));
```

Zhiyuan-Amos:
Perhaps we should leave a comment that WINDOW_CLOSE_REQUEST triggers the .fxml function.

yamidark:
>You shouldn't setup a new stage but use mainWindow.getPrimaryStage() right?

The `fireEvent(Event)` method needs to be called in the `FX user thread` which is why I used the `FxToolkit`.

>Perhaps we should leave a comment that WINDOW_CLOSE_REQUEST triggers the .fxml function.

Yea, will mention the `MainWindow` for the test does not have a `X` button too, and this `WINDOW_CLOSE_REQUEST` is used to simulate any external request to close the `MainWindow` as mentioned in earlier parts of the commit message.

Zhiyuan-Amos:
> The fireEvent(Event) method needs to be called in the FX user thread which is why I used the FxToolkit.

You can run this on the FX Thread by calling `guiRobot#interact(() -> ...)`

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private class MainWindowHandleUnit extends StageHandle {
+        private final Stage stage;
+
+        private MainWindowHandleUnit(Stage stage) {
+            super(stage);
+            this.stage = stage;
+        }
+
+        /**
+         * Closes the {@code MainWindow} using it's menu bar.
```
Zhiyuan-Amos: its

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,79 @@
...
+                    new MainWindow(stage, new Config(), new UserPrefs(), new LogicManager(new ModelManager()));
+            stage.setScene(mainWindow.getRoot().getScene());
+            mainWindowHandle = new MainWindowHandleUnit(stage);
+            mainWindowHandle.focus();
+        });
+        FxToolkit.showStage();
+    }
+
+    @Test
+    public void menuBar_exit_success() {
```
Zhiyuan-Amos: We are testing exit through menu bar, and the end result we are verifying for is `ExitAppRequestEvent` posted. So it should be:

`exit_menuBarExitButton_ExitAppRequestEventPosted`

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,79 @@
...
+        /**
+         * Closes the {@code MainWindow} using it's menu bar.
+         */
+        private void clickOnMenuExitButton() {
+            guiRobot.clickOn("File");
+            guiRobot.clickOn("Exit");
+        }
+
+        /**
+         * Simulate an external request to close the {@code MainWindow}.
```
Zhiyuan-Amos: It's unclear what "external request" means.

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * {@code MainWindow} unit test class without any other components to test closing of the {@code MainWindow}.
+     */
+    private class MainWindowHandleUnit extends StageHandle {
+        private final Stage stage;
+
+        private MainWindowHandleUnit(Stage stage) {
+            super(stage);
+            this.stage = stage;
```
Zhiyuan-Amos: this is a hackish way of doing things :P We can remove the need for this by storing `stage` as a variable in the test class alongside with `mainWindow` and `mainWindowHandle`.

src/test/java/seedu/address/ui/MainWindowTest.java
yamidark: Hmm, should I give examples like in this commit message (using the 'X' button or closing from taskbar?

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public final EventsCollectorRule eventsCollectorRule = new EventsCollectorRule();
+
+    private MainWindow mainWindow;
+    private MainWindowHandleUnit mainWindowHandle;
+
+    @Before
+    public void setUp() throws Exception {
+        FxToolkit.setupStage(stage -> {
+            mainWindow =
+                    new MainWindow(stage, new Config(), new UserPrefs(), new LogicManager(new ModelManager()));
```
yamidark: @Zhiyuan-Amos Actually, should I extract this out as a method in the test class as `prepareMainWindow(Stage)`, which should make the test cleaner and clearer?

src/test/java/seedu/address/ui/MainWindowTest.java
Zhiyuan-Amos: It's better to leave it in the code so that future developers know what is this piece of code doing :P

src/test/java/seedu/address/ui/MainWindowTest.java
Zhiyuan-Amos: Yup, looks good!

Zhiyuan-Amos:
@vivekscl can take a look? ^^

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,83 @@
...
+import javafx.stage.WindowEvent;
+import seedu.address.commons.core.Config;
+import seedu.address.commons.events.ui.ExitAppRequestEvent;
+import seedu.address.logic.LogicManager;
+import seedu.address.model.ModelManager;
+import seedu.address.model.UserPrefs;
+import seedu.address.ui.testutil.EventsCollectorRule;
+
+/**
+ * Contains test for closing of the {@code MainWindow}.
```
damithc: This comment sounds like there is only one test? 
Also, the class name doesn't indicate that the scope is limited to testing the 'close' feature only.

src/test/java/seedu/address/ui/MainWindowTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    public void close_externalRequest_exitAppRequestEventPosted() {
+        mainWindowHandle.closeMainWindowExternally();
+        assertTrue(eventsCollectorRule.eventsCollector.getMostRecent() instanceof ExitAppRequestEvent);
+        assertTrue(eventsCollectorRule.eventsCollector.getSize() == 1);
+    }
+
+    /**
+     * Provides a handle for {@code MainWindow}. The components in {@code MainWindow} are not initialized.
+     */
```
damithc: `Provides` is redundant?

Zhiyuan-Amos:
@yamidark update the merge commit message ya :)

PR #836, opened 2018-02-28
UniquePersonList, AddressBook: update return type for remove and removePerson #830

### LINKED ISSUE #830: UniquePersonList#remove(Person): update return value
`remove(Person)` returns true if the person passed into this method can be found in the internal list, and false otherwise. `remove(Person)` also throws `PersonNotFoundException` if the person cannot be found in the internal list.

There's no need for this method to return a boolean value to signal whether the person can be found in the internal list, since this method already throws `PersonNotFoundException` when the person cannot be found.

Let's update this method's return value to `void`.

nicholaschuayunzhi:
fixes #830 
```
AddressBook#removePerson(Person) returns true if the person passed into
this method can be found in the internal list. 
UniquePersonList#remove(Person) returns true if the person passed into
this method can be found in the internal list, and false otherwise.
They both also throw PersonNotFoundException if a person cannot be
found.

Returning a boolean is not required as the thrown exception would
indicate that the person cannot be found.

Let's update the return type for AddressBook#removePerson(Person) and 
UniquePersonList#remove(Person) to void.
```

### DIFF
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -139,12 +139,8 @@ public class AddressBook implements ReadOnlyAddressBook {
      * Removes {@code key} from this {@code AddressBook}.
      * @throws PersonNotFoundException if the {@code key} is not in this {@code AddressBook}.
      */
-    public boolean removePerson(Person key) throws PersonNotFoundException {
-        if (persons.remove(key)) {
-            return true;
-        } else {
-            throw new PersonNotFoundException();
-        }
+    public void removePerson(Person key) throws PersonNotFoundException {
+        persons.remove(key);
     }
 
     //// tag-level operations
--- a/src/main/java/seedu/address/model/person/UniquePersonList.java
+++ b/src/main/java/seedu/address/model/person/UniquePersonList.java
@@ -72,13 +72,11 @@ public class UniquePersonList implements Iterable<Person> {
      *
      * @throws PersonNotFoundException if no such person could be found in the list.
      */
-    public boolean remove(Person toRemove) throws PersonNotFoundException {
+    public void remove(Person toRemove) throws PersonNotFoundException {
         requireNonNull(toRemove);
-        final boolean personFoundAndDeleted = internalList.remove(toRemove);
-        if (!personFoundAndDeleted) {
+        if (!internalList.remove(toRemove)) {
             throw new PersonNotFoundException();
         }
-        return personFoundAndDeleted;
     }
 
     public void setPersons(UniquePersonList replacement) {


src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -72,13 +72,12 @@ public void setPerson(Person target, Person editedPerson)
      *
      * @throws PersonNotFoundException if no such person could be found in the list.
      */
-    public boolean remove(Person toRemove) throws PersonNotFoundException {
+    public void remove(Person toRemove) throws PersonNotFoundException {
         requireNonNull(toRemove);
         final boolean personFoundAndDeleted = internalList.remove(toRemove);
```
yamidark: Since we are no longer returning a `boolean`, so there is no need to store the result.
We can also just follow `add` and `setPerson` style to be clearer:
```Java
if (!contains(toRemove)) {
    throw new PersonNotFoundException();
}
internalList.remove(toRemove);
```

src/main/java/seedu/address/model/AddressBook.java
```diff
@@ -140,11 +140,8 @@ private Person syncWithMasterTagList(Person person) {
      * @throws PersonNotFoundException if the {@code key} is not in this {@code AddressBook}.
      */
     public boolean removePerson(Person key) throws PersonNotFoundException {
-        if (persons.remove(key)) {
-            return true;
-        } else {
-            throw new PersonNotFoundException();
-        }
+        persons.remove(key);
+        return true;
```
yamidark: Maybe we can shift this commit to after updating `AddressBook.removePerson(Person)`, so that we have a stronger justification to update the return type in `UniquePersonList.remove(Person)` as it won't be used anywhere anymore.

Alternatively, we can just squash the 2 commits together and update the commit message accordingly since technically both commits are doing the same thing.

@Zhiyuan-Amos @vivekscl  your opinion on this?

src/main/java/seedu/address/model/person/UniquePersonList.java
Zhiyuan-Amos: Since `remove(Person)` returns true/false depending on whether the person exists, I think there's no need to check using `contains(Person)`. 

I understand that you're trying to be clearer, but because `List#remove(Object)` is a common method, so it's clear to readers already what `!internalList.remove(toRemove)` is intended to do :)

src/main/java/seedu/address/model/AddressBook.java
Zhiyuan-Amos: > Maybe we can shift this commit to after updating AddressBook.removePerson(Person), so that we have a stronger justification to update the return type in UniquePersonList.remove(Person) as it won't be used anywhere anymore.

Sounds good. In short:
commit 1 - update address book
commit 2 - update uniquepersonlist

nicholaschuayunzhi:
Thank you @yamidark and @Zhiyuan-Amos for the comments and review! I have updated the PR with the changes with requested.

Zhiyuan-Amos:
This looks good! Just a minor nit before it's good to go (wait for @yamidark to review first as well):

> AddressBook: update removePerson return type

Update commit message to `AddressBook#removePerson: update return type` :p Same for `UniquePersonList`.

yamidark:
Just minor nits in the commit message:

**AddressBook**#removePerson(Person) returns true if the person passed into
this method can be found in **the** internal list, and throws a
PersonNotFoundException otherwise.

Zhiyuan-Amos:
> AddressBook#removePerson(Person) returns true if the person passed into
this method can be found in the internal list, and throws a
PersonNotFoundException otherwise.

Take note to do so for `UniquePersonList` as well :) Also, let's wait for prof to review first before you make the changes.

@damithc for your review.

damithc:
> @damithc Prof, for your review.

Looks like no one else has approved the PR yet?

Zhiyuan-Amos:
@nicholaschuayunzhi take note that you need to come up with a merge commit message as well. You can take a look at previous PRs for reference (e.g. [here](https://github.com/se-edu/addressbook-level4/pull/806)). :)

Zhiyuan-Amos:
@nicholaschuayunzhi thanks for your contribution! :)

Issue #706, opened 2017-09-28
Gradle: update to latest version


Zhiyuan-Amos:
We are using a much older version of Gradle (2.12). 
Lets:

- [x] Update Gradle to the latest version
- [x] Find out what are the changes that have been made since then, and so tidy up our `build.gradle` file as well.

Issue #681, opened 2017-09-19
Upgrade to Java 9


Zhiyuan-Amos:
Java 9 ~~will be released on 21st September~~ is released! 

Some changes that we can look into:
1. Convert method calls like `Arrays#asList(...)` to use `List#of(...)`.  https://dzone.com/articles/java-9-factory-methods-to-create-immutable-collect
1. Look into using `Runtime.Version`. https://dzone.com/articles/process-api-and-version-in-java-9

Zhiyuan-Amos:
@damithc Can we upgrade the code base to Java 9? This fixes #374 too.

damithc:
I guess we can, as long as we do that quickly, before students start on level 4.

Zhiyuan-Amos:
@yamidark take a look at what I've done so far for [Java 9](https://github.com/Zhiyuan-Amos/addressbook-level4/tree/java9) on my branch.

damithc:
It's too late for this sem but we have to make the project Java 9 compatible by next sem. I thought there is no hurry because Java is normally backward compatible but apparently AB4 runs into issues in Java 9 environments.

Zhiyuan-Amos:
Ah I see; I thought students were still using Java 8. I'll look into it.

damithc:
> I thought students were still using Java 8.

CS2030 uses Java 9; that's why some students are already on Java 9 when they come to CS2103. In any case, we should ensure _compatibility_ with the latest public version as soon as it is out (but we can wait a while before we make the latest version a  _prerequisite_)

pyokagan:
@Zhiyuan-Amos 
> Convert method calls like Arrays#asList(...) to use List#of(...). https://dzone.com/articles/java-9-factory-methods-to-create-immutable-collect
> Look into using Runtime.Version. https://dzone.com/articles/process-api-and-version-in-java-9

Can split this into separate issues? I guess they will be lower priority as well (and maybe first-timers?)

I think this issue, as it is, has been resolved and can be closed.

Zhiyuan-Amos:
@pyokagan noted.

Closing this issue, for other updates that we can make to our codebase, see #871 and #872.

Issue #817, opened 2018-02-03
Preparation for Java 9: replace JAXB


Zhiyuan-Amos:
Required for #681.

As part of upgrading to use Java 9, we are no longer able to use `javax.xml`. The module `java.xml` is also [deprecated and will be removed in future Java versions](https://docs.oracle.com/javase/9/docs/api/java.xml.bind-summary.html#packages.summary).

A quick fix in this [SO post](https://stackoverflow.com/a/43574427/8828382) is to specify the following command-line option: `--add-modules java.xml.bind`. However: 
1. This requires additional configuration by all the developers on their IDE to include this option.
2. Additional configuration on CI.
3. Once this module is deprecated and removed, this solution will no longer work.

A better solution proposed by the SO post above is to include the dependency in the Gradle script (the given solution is for Maven though):
```
<!-- Java 8 = JAX-B Version 2.2.8 -->
<dependency>
    <groupId>javax.xml.bind</groupId>
    <artifactId>jaxb-api</artifactId>
    <version>2.2.8</version>
</dependency>
```

However, I thought that this will bloat up `build.gradle` file, and it may perhaps be better for us to shift from using `JAXB` to some other libraries instead. From this [guide](http://www.baeldung.com/java-xml), we have a few choices such as `XStream` and `Simple XML`.

If we are doing the earlier solutions proposed above, then we can make the changes in the same PR that resolves #681. If we are using other libraries instead, then we can have a separate PR to resolve that.

Zhiyuan-Amos:
@eugenepeh can I seek your help for this issue validation? ^^

yamgent:
> include the dependency in the Gradle script (the given solution is for Maven though):

Shouldn't be an issue, afaik they are usually interchangeable.

> However, I thought that this will bloat up build.gradle file, and it may perhaps be better for us to shift from using JAXB to some other libraries instead. From this guide, we have a few choices such as XStream and Simple XML.

But using the other libraries would also 'bloat' up `build.gradle`, no? :P

eugenepeh:
Issue verified, it should be fixed if it is deprecated and will be removed in the future version. 
However, since it does not really affect the migration to java 9, and like what @yamgent mentioned: 
>But using the other libraries would also 'bloat' up build.gradle, no? :P

I would think that this is currently a low priority.

Zhiyuan-Amos:
@yamgent @eugenepeh thanks for both of your thoughts! 👍 :)

[...]

PR #837, opened 2018-03-01
build.gradle: update to build project using Java 9 #374

### LINKED ISSUE #374: HelpWindow: Gibberish text (Mac only)
screen shot 2017-03-31 at 12 47 49 am

Tested on java version "1.8.0_121" and Mac OS version "10.12.4"

It turns out to be a bug in JavaFx WebView on Mac, it tries to load some fonts locally and is disallowed (explained in [this StackOveflow answer](http://stackoverflow.com/a/42375360/3522482)). Also as pointed out in the related [open JDK issue](https://bugs.openjdk.java.net/browse/JDK-8175802?focusedCommentId=14055132&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14055132), this is fixed in Java 9 early access version (9-ea+148)

Some possible walk-arounds:

1. change GitHub Pages theme. For example, [MERLOT THEME](https://github.com/pages-themes/merlot) works.
1. programatically [insert another `<style>` tag into the `WebView` html](http://stackoverflow.com/questions/41952734/javafx-webview-font-issue-on-mac#comment71578136_41952734) to change the font used.
1. revert [this PR](https://github.com/se-edu/addressbook-level4/pull/259)

There is still possibility that if we download open sans font as a `.tff` file and add to the app, then load it in JavaFX at runtime, it will display the font correctly, but this approach is relatively more complicated and not yet tried out.

Zhiyuan-Amos:
Fixes #374, part of #681 and #706.

```
build.gradle: upgrade to use Java 9

The project currently runs on Java 8.

We should ensure compatibility with the latest public version as soon
as it is out.

Let's upgrade the project to run on Java 9.
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,7 +63,7 @@ shadowJar {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '2.12'
+    gradleVersion = '4.6'
 }
 
 task coverage(type: JacocoReport) {
--- a/build.gradle
+++ b/build.gradle
@@ -18,8 +18,8 @@ plugins {
 // Specifies the entry point of the application
 mainClassName = 'seedu.address.MainApp'
 
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_1_9
+targetCompatibility = JavaVersion.VERSION_1_9
 
 repositories {
     mavenCentral()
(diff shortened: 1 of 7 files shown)


damithc:
This fix require everyone to install Java 9?

Zhiyuan-Amos:
@damithc 

> In any case, we should ensure compatibility with the latest public version as soon as it is out (but we can wait a while before we make the latest version a prerequisite)

Ah sorry, I just understood what you meant here. Will make the changes accordingly.

Zhiyuan-Amos:
@damithc I'm not sure if we can make it compatible with both Java 8 and Java 9.

On this branch, Java 9 is supported. When running on Java 8, TestFX crashes because we have jdk9 features as dependencies which are required to support the tests when running in Java 9.

On master branch, Java 9 is unsupported because Gradle hasn't imported the jdk9 features as dependencies for tests. 

I've verified the issue from [AddressBook-Lvl2](https://github.com/nus-cs2103-AY1718S2/forum/issues/12#issuecomment-369554375) that students do not have to delete Java 9 just to run the tests successfully.

damithc:
I see. In that case I guess there's no choice but to do a full upgrade to Java 9.

Zhiyuan-Amos:
> I see. In that case I guess there's no choice but to do a full upgrade to Java 9.

Are we intending to rush this PR? An alternative solution is that when students face issues concerning Java 9 compatibility, I can respond to them.

damithc:
> Are we intending to rush this PR?

I guess no need.

build.gradle
```diff
@@ -46,14 +46,21 @@ dependencies {
...
+    testCompile group: 'org.testfx', name: 'testfx-core', version: '4.0.12-alpha', {
+        exclude group: 'org.testfx', module: 'testfx-internal-java8'
+    }
     testCompile group: 'org.testfx', name: 'testfx-junit', version: testFxVersion
     testCompile group: 'org.testfx', name: 'testfx-legacy', version: testFxVersion, {
         exclude group: 'junit', module: 'junit'
     }
-    testCompile group: 'org.testfx', name: 'openjfx-monocle', version: '1.8.0_20'
+    testRuntime 'org.testfx:testfx-internal-java9:4.0.11-alpha'
+    testRuntime 'org.testfx:openjfx-monocle:jdk-9+181'
```
Zhiyuan-Amos: Follow `testCompile` style.

damithc:
Shouldn't there be some updates to docs (i.e., prerequisites)?

Zhiyuan-Amos:
@pyokagan @yamgent @eugenepeh tagging all of you in to take a look at this small PR if you happen to be free.

Zhiyuan-Amos:
@eugenepeh thanks for the review! :)

Zhiyuan-Amos:
> What do these properties do in our project? Is there a substitute for these options? Should we be using the substitutes (if any) for our project?

Good catch. I've updated the code slightly accordingly and updated the commit message.

build.gradle
```diff
@@ -39,21 +39,29 @@ jacocoTestReport {
...
 
     compile group: 'org.fxmisc.easybind', name: 'easybind', version: '1.0.3'
     compile group: 'org.controlsfx', name: 'controlsfx', version: '8.40.11'
     compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.7.0'
     compile group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-jsr310', version: '2.7.4'
     compile group: 'com.google.guava', name: 'guava', version: '19.0'
+    compile group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.8'
+    compile group: 'com.sun.xml.bind', name: 'jaxb-impl', version: '2.3.0'
+    compile group: 'com.sun.xml.bind', name: 'jaxb-core', version: '2.3.0'
+    compile group: 'javax.activation', name: 'activation', version: '1.1.1'
```
yamgent: Who is the provider of these packages?

Also, I wonder whether it is worth to seek out alternative libraries so that we don't get a impression that we are trying to circumvent the modules system... (don't need to be in this PR)

Zhiyuan-Amos: > Who is the provider of these packages?

[Java EE](https://docs.oracle.com/javaee/7/api/) for the `javax.xml.bind`, which internally uses the `com.sun.xml.bind` (that's why we have to import them as well).

> Also, I wonder whether it is worth to seek out alternative libraries so that we don't get a impression that we are trying to circumvent the modules system... (don't need to be in this PR)

Yup, we actually discussed that before in #817. :P

yamgent: Ah, had a slight feeling that it was discussed before but couldn't find the exact issue, thanks for pointing it out.

Zhiyuan-Amos:
@Rinder5 @yamidark @vivekscl as our project has been upgraded to use Java 9 with the merging of this PR, do update your local machine's `$JAVA_HOME` to reflect this change as well, otherwise you can't compile the code locally. See this [link](http://www.baeldung.com/java-home-on-windows-7-8-10-mac-os-x-linux) and ping me up if you need any help.

PR #844, opened 2018-03-09
Add different levels of comparison to equality methods of Person class #742

### LINKED ISSUE #742: Add different levels of comparison to equality methods of Person class
Not sure if this is intended or not.
The address book does not show the duplicate person error and allows us to add duplicates (readding the same person) as long as we change any of their other attributes when adding them.
duplicates

For instance,
add n/Johnny Eco Ela Doe Li p/98765437 e/johnd@example.com a/311, Clementi Ave 2, #02-25 t/friends t/owesMoney

followed by,
add n/Johnny Eco Ela Doe Li p/98765433 e/johnd@example.com a/311, Clementi Ave 2, #02-25 t/friends t/owesMoney

vivekscl:
Fixes #742 

Some changes to note:
In `AddCommandSystemTest`, line 120 was changed from `assertCommandSuccess(CARL)` to `assertCommandSuccess(AMY)`  to test for duplicate persons in the next few lines as the duplicate person tests require the default person to be used, which is either AMY or BOB.

While porting from [se-edu/addressbook-level2#122](https://github.com/se-edu/addressbook-level2/issues/122), I realized only name and phone were checked. Since email is a unique identifier for a person as well, I added that to the equality method.

### DIFF
--- a/src/main/java/seedu/address/model/person/Person.java
+++ b/src/main/java/seedu/address/model/person/Person.java
@@ -15,11 +15,13 @@ import seedu.address.model.tag.Tag;
  */
 public class Person {
 
+    // Identity fields
     private final Name name;
     private final Phone phone;
     private final Email email;
-    private final Address address;
 
+    // Data fields
+    private final Address address;
     private final Set<Tag> tags = new HashSet<>();
 
     /**
@@ -58,6 +60,24 @@ public class Person {
         return Collections.unmodifiableSet(tags);
     }
 
+    /**
+     * Returns true if both persons of the same name have at least one other identity field that is the same.
+     * This defines a weaker notion of equality between two persons.
+     */
+    public boolean isSamePerson(Person otherPerson) {
+        if (otherPerson == this) {
+            return true;
+        }
+
+        return otherPerson != null
+                && otherPerson.getName().equals(this.getName())
+                && (otherPerson.getPhone().equals(this.getPhone()) || otherPerson.getEmail().equals(this.getEmail()));
+    }
+
+    /**
+     * Returns true if both persons have the same identity and data fields.
+     * This defines a stronger notion of equality between two persons.
+     */
     @Override
     public boolean equals(Object other) {
         if (other == this) {
--- a/src/main/java/seedu/address/model/person/Person.java
+++ b/src/main/java/seedu/address/model/person/Person.java
@@ -92,7 +92,8 @@ public class Person {
         return otherPerson.getName().equals(this.getName())
                 && otherPerson.getPhone().equals(this.getPhone())
                 && otherPerson.getEmail().equals(this.getEmail())
-                && otherPerson.getAddress().equals(this.getAddress());
+                && otherPerson.getAddress().equals(this.getAddress())
+                && otherPerson.getTags().equals(this.getTags());
     }
 
     @Override
--- a/src/main/java/seedu/address/model/person/UniquePersonList.java
+++ b/src/main/java/seedu/address/model/person/UniquePersonList.java
@@ -8,17 +8,19 @@ import java.util.List;
 
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
-import seedu.address.commons.util.CollectionUtil;
 import seedu.address.model.person.exceptions.DuplicatePersonException;
 import seedu.address.model.person.exceptions.PersonNotFoundException;
 
 /**
  * A list of persons that enforces uniqueness between its elements and does not allow nulls.
+ * A person is considered unique by comparing using {@code Person#isSamePerson(Person)}. As such, adding and updating of
+ * persons uses Person#isSamePerson(Person) for equality so as to ensure that the person being added or updated is
+ * unique in terms of identity in the UniquePersonList. However, the removal of a person uses Person#equals(Object) so
+ * as to ensure that the person with exactly the same fields will be removed.
  *
  * Supports a minimal set of list operations.
  *
- * @see Person#equals(Object)
- * @see CollectionUtil#elementsAreUnique(Collection)
+ * @see Person#isSamePerson(Person)
  */
 public class UniquePersonList implements Iterable<Person> {
 
@@ -29,7 +31,7 @@ public class UniquePersonList implements Iterable<Person> {
      */
     public boolean contains(Person toCheck) {
         requireNonNull(toCheck);
-        return internalList.contains(toCheck);
+        return internalList.stream().anyMatch(toCheck::isSamePerson);
     }
 
     /**
@@ -60,7 +62,7 @@ public class UniquePersonList implements Iterable<Person> {
             throw new PersonNotFoundException();
         }
 
-        if (!target.equals(editedPerson) && internalList.contains(editedPerson)) {
+        if (!target.isSamePerson(editedPerson) && contains(editedPerson)) {
             throw new DuplicatePersonException();
         }
 
@@ -85,7 +87,7 @@ public class UniquePersonList implements Iterable<Person> {
 
     public void setPersons(List<Person> persons) throws DuplicatePersonException {
         requireAllNonNull(persons);
-        if (!CollectionUtil.elementsAreUnique(persons)) {
+        if (!personsAreUnique(persons)) {
             throw new DuplicatePersonException();
         }
 
@@ -114,5 +116,19 @@ public class UniquePersonList implements Iterable<Person> {
     @Override
     public int hashCode() {
         return internalList.hashCode();
+    }
+
+    /**
+     * Returns true if {@code persons} contains only unique persons.
+     */
+    private boolean personsAreUnique(List<Person> persons) {
+        for (int i = 0; i < persons.size() - 1; i++) {
+            for (int j = i + 1; j < persons.size(); j++) {
+                if (persons.get(i).isSamePerson(persons.get(j))) {
+                    return false;
+                }
+            }
+        }
+        return true;
     }
 }
--- a/src/main/java/seedu/address/model/person/exceptions/DuplicatePersonException.java
+++ b/src/main/java/seedu/address/model/person/exceptions/DuplicatePersonException.java
@@ -3,7 +3,8 @@ package seedu.address.model.person.exceptions;
 import seedu.address.commons.exceptions.DuplicateDataException;
 
 /**
- * Signals that the operation will result in duplicate Person objects.
+ * Signals that the operation will result in duplicate Persons (Persons are considered duplicates if they have the same
+ * identity).
  */
 public class DuplicatePersonException extends DuplicateDataException {
     public DuplicatePersonException() {
--- a/src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
+++ b/src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
@@ -27,6 +27,7 @@
         <phone>87652533</phone>
         <email>cornelia@example.com</email>
         <address>10th street</address>
+        <tagged>friends</tagged>
     </persons>
     <persons>
         <name>Elle Meyer</name>
--- a/src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
+++ b/src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
@@ -3,7 +3,7 @@
 <addressbook>
     <persons>
         <name>Alice Pauline</name>
-        <phone>85355255</phone>
+        <phone>94351253</phone>
         <email>alice@example.com</email>
         <address>123, Jurong West Ave 6, #08-111</address>
         <tagged>friends</tagged>
--- a/src/test/java/guitests/guihandles/PersonCardHandle.java
+++ b/src/test/java/guitests/guihandles/PersonCardHandle.java
@@ -3,6 +3,8 @@ package guitests.guihandles;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import com.google.common.collect.ImmutableMultiset;
+
 import javafx.scene.Node;
 import javafx.scene.control.Label;
 import javafx.scene.layout.Region;
@@ -77,6 +79,9 @@ public class PersonCardHandle extends NodeHandle<Node> {
         return getName().equals(person.getName().fullName)
                 && getAddress().equals(person.getAddress().value)
                 && getPhone().equals(person.getPhone().value)
-                && getEmail().equals(person.getEmail().value);
+                && getEmail().equals(person.getEmail().value)
+                && ImmutableMultiset.copyOf(getTags()).equals(ImmutableMultiset.copyOf(person.getTags().stream()
+                        .map(tag -> tag.tagName)
+                        .collect(Collectors.toList())));
     }
 }
--- a/src/test/java/seedu/address/model/person/PersonTest.java
+++ b/src/test/java/seedu/address/model/person/PersonTest.java
@@ -91,5 +91,9 @@ public class PersonTest {
         // different address -> returns false
         editedAlice = new PersonBuilder(ALICE).withAddress(VALID_ADDRESS_BOB).build();
         assertFalse(ALICE.equals(editedAlice));
+
+        // different tags -> returns false
+        editedAlice = new PersonBuilder(ALICE).withTags(VALID_TAG_HUSBAND).build();
+        assertFalse(ALICE.equals(editedAlice));
     }
 }
--- a/src/test/java/seedu/address/model/person/PersonTest.java
+++ b/src/test/java/seedu/address/model/person/PersonTest.java
@@ -1,5 +1,15 @@
 package seedu.address.model.person;
 
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
+import static seedu.address.testutil.TypicalPersons.ALICE;
+import static seedu.address.testutil.TypicalPersons.BOB;
+
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -15,5 +25,71 @@ public class PersonTest {
         Person person = new PersonBuilder().build();
         thrown.expect(UnsupportedOperationException.class);
         person.getTags().remove(0);
+    }
+
+    @Test
+    public void isSamePerson() {
+        // same object -> returns true
+        assertTrue(ALICE.isSamePerson(ALICE));
+
+        // null -> returns false
+        assertFalse(ALICE.isSamePerson(null));
+
+        // different phone and email -> returns false
+        Person editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB).build();
+        assertFalse(ALICE.isSamePerson(editedAlice));
+
+        // different name -> returns false
+        editedAlice = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
+        assertFalse(ALICE.isSamePerson(editedAlice));
+
+        // same name, same phone, different attributes -> returns true
+        editedAlice = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+
+        // same name, same email, different attributes -> returns true
+        editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withAddress(VALID_ADDRESS_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+
+        // same name, same phone, same email, different attributes -> returns true
+        editedAlice = new PersonBuilder(ALICE).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+    }
+
+    @Test
+    public void equals() {
+        // same values -> returns true
+        Person aliceCopy = new PersonBuilder(ALICE).build();
+        assertTrue(ALICE.equals(aliceCopy));
+
+        // same object -> returns true
+        assertTrue(ALICE.equals(ALICE));
+
+        // null -> returns false
+        assertFalse(ALICE.equals(null));
+
+        // different type -> returns false
+        assertFalse(ALICE.equals(5));
+
+        // different person -> returns false
+        assertFalse(ALICE.equals(BOB));
+
+        // different name -> returns false
+        Person editedAlice = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
+        assertFalse(ALICE.equals(editedAlice));
+
+        // different phone -> returns false
+        editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).build();
+        assertFalse(ALICE.equals(editedAlice));
+
+        // different email -> returns false
+        editedAlice = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).build();
+        assertFalse(ALICE.equals(editedAlice));
+
+        // different address -> returns false
+        editedAlice = new PersonBuilder(ALICE).withAddress(VALID_ADDRESS_BOB).build();
+        assertFalse(ALICE.equals(editedAlice));
     }
 }
--- a/src/test/java/seedu/address/ui/PersonListPanelTest.java
+++ b/src/test/java/seedu/address/ui/PersonListPanelTest.java
@@ -94,8 +94,8 @@ public class PersonListPanelTest extends GuiUnitTest {
         builder.append("<addressbook>\n");
         for (int i = 0; i < personCount; i++) {
             builder.append("<persons>\n");
-            builder.append("<name>a</name>\n");
-            builder.append("<phone>").append(i).append("00</phone>\n");
+            builder.append("<name>").append(i).append("a</name>\n");
+            builder.append("<phone>000</phone>\n");
             builder.append("<email>a@aa</email>\n");
             builder.append("<address>a</address>\n");
             builder.append("</persons>\n");
--- a/src/test/java/systemtests/AddCommandSystemTest.java
+++ b/src/test/java/systemtests/AddCommandSystemTest.java
@@ -20,7 +20,6 @@ import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_BOB;
-import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
@@ -86,25 +85,11 @@ public class AddCommandSystemTest extends AddressBookSystemTest {
                 + TAG_DESC_FRIEND;
         assertCommandSuccess(command, toAdd);
 
-        /* Case: add a person with all fields same as another person in the address book except phone -> added */
-        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_AMY)
-                .withAddress(VALID_ADDRESS_AMY).withTags(VALID_TAG_FRIEND).build();
-        command = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_BOB + EMAIL_DESC_AMY + ADDRESS_DESC_AMY
-                + TAG_DESC_FRIEND;
-        assertCommandSuccess(command, toAdd);
-
-        /* Case: add a person with all fields same as another person in the address book except email -> added */
-        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_BOB)
-                .withAddress(VALID_ADDRESS_AMY).withTags(VALID_TAG_FRIEND).build();
-        command = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_BOB + ADDRESS_DESC_AMY
-                + TAG_DESC_FRIEND;
-        assertCommandSuccess(command, toAdd);
-
-        /* Case: add a person with all fields same as another person in the address book except address -> added */
-        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY)
-                .withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND).build();
-        command = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_AMY + ADDRESS_DESC_BOB
-                + TAG_DESC_FRIEND;
+        /* Case: add a person with all fields same as another person in the address book except phone and email
+         * -> added
+         */
+        toAdd = new PersonBuilder(AMY).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB).build();
+        command = PersonUtil.getAddCommand(toAdd);
         assertCommandSuccess(command, toAdd);
 
         /* Case: add to empty address book -> added */
@@ -136,6 +121,21 @@ public class AddCommandSystemTest extends AddressBookSystemTest {
 
         /* Case: add a duplicate person -> rejected */
         command = PersonUtil.getAddCommand(HOON);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a duplicate person except with different phone -> rejected */
+        toAdd = new PersonBuilder(HOON).withPhone(VALID_PHONE_BOB).build();
+        command = PersonUtil.getAddCommand(toAdd);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a duplicate person except with different email -> rejected */
+        toAdd = new PersonBuilder(HOON).withEmail(VALID_EMAIL_BOB).build();
+        command = PersonUtil.getAddCommand(toAdd);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a duplicate person except with different address -> rejected */
+        toAdd = new PersonBuilder(HOON).withAddress(VALID_ADDRESS_BOB).build();
+        command = PersonUtil.getAddCommand(toAdd);
         assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
 
         /* Case: add a duplicate person except with different tags -> rejected */
--- a/src/test/java/systemtests/EditCommandSystemTest.java
+++ b/src/test/java/systemtests/EditCommandSystemTest.java
@@ -1,6 +1,7 @@
 package systemtests;
 
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 import static seedu.address.logic.commands.CommandTestUtil.ADDRESS_DESC_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.ADDRESS_DESC_BOB;
@@ -18,14 +19,17 @@ import static seedu.address.logic.commands.CommandTestUtil.PHONE_DESC_BOB;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.TAG_DESC_HUSBAND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_ADDRESS_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_EMAIL_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_NAME_BOB;
+import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_AMY;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_PHONE_BOB;
-import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_FRIEND;
 import static seedu.address.logic.commands.CommandTestUtil.VALID_TAG_HUSBAND;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 import static seedu.address.model.Model.PREDICATE_SHOW_ALL_PERSONS;
 import static seedu.address.testutil.TypicalIndexes.INDEX_FIRST_PERSON;
+import static seedu.address.testutil.TypicalIndexes.INDEX_SECOND_PERSON;
 import static seedu.address.testutil.TypicalPersons.AMY;
 import static seedu.address.testutil.TypicalPersons.BOB;
 import static seedu.address.testutil.TypicalPersons.KEYWORD_MATCHING_MEIER;
@@ -84,16 +88,28 @@ public class EditCommandSystemTest extends AddressBookSystemTest {
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandSuccess(command, index, BOB);
 
-        /* Case: edit some fields -> edited */
-        index = INDEX_FIRST_PERSON;
-        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + TAG_DESC_FRIEND;
-        Person personToEdit = getModel().getFilteredPersonList().get(index.getZeroBased());
-        editedPerson = new PersonBuilder(personToEdit).withTags(VALID_TAG_FRIEND).build();
+        /* Case: edit a person with new values same as another person's values but with different name -> edited */
+        assertTrue(getModel().getAddressBook().getPersonList().contains(BOB));
+        index = INDEX_SECOND_PERSON;
+        assertNotEquals(getModel().getFilteredPersonList().get(index.getZeroBased()), BOB);
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_AMY + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        editedPerson = new PersonBuilder(BOB).withName(VALID_NAME_AMY).build();
+        assertCommandSuccess(command, index, editedPerson);
+
+        /* Case: edit a person with new values same as another person's values but with different phone and email
+         * -> edited
+         */
+        index = INDEX_SECOND_PERSON;
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_AMY + EMAIL_DESC_AMY
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        editedPerson = new PersonBuilder(BOB).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY).build();
         assertCommandSuccess(command, index, editedPerson);
 
         /* Case: clear tags -> cleared */
         index = INDEX_FIRST_PERSON;
         command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + " " + PREFIX_TAG.getPrefix();
+        Person personToEdit = getModel().getFilteredPersonList().get(index.getZeroBased());
         editedPerson = new PersonBuilder(personToEdit).withTags().build();
         assertCommandSuccess(command, index, editedPerson);
 
@@ -185,6 +201,21 @@ public class EditCommandSystemTest extends AddressBookSystemTest {
         /* Case: edit a person with new values same as another person's values but with different tags -> rejected */
         command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
                 + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND;
+        assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: edit a person with new values same as another person's values but with different address -> rejected */
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
+                + ADDRESS_DESC_AMY + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: edit a person with new values same as another person's values but with different phone -> rejected */
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_AMY + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: edit a person with new values same as another person's values but with different email -> rejected */
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_AMY
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
     }
 
--- a/src/main/java/seedu/address/commons/util/CollectionUtil.java
+++ b/src/main/java/seedu/address/commons/util/CollectionUtil.java
@@ -32,11 +32,4 @@ public class CollectionUtil {
     public static boolean isAnyNonNull(Object... items) {
         return items != null && Arrays.stream(items).anyMatch(Objects::nonNull);
     }
-
-    /**
-     * Returns true if every element in a collection are unique by {@link Object#equals(Object)}.
-     */
-    public static boolean elementsAreUnique(Collection<?> items) {
-        return items.size() == items.stream().distinct().count();
-    }
 }
--- a/src/main/java/seedu/address/model/AddressBook.java
+++ b/src/main/java/seedu/address/model/AddressBook.java
@@ -12,7 +12,7 @@ import seedu.address.model.person.exceptions.PersonNotFoundException;
 
 /**
  * Wraps all data at the address-book level
- * Duplicates are not allowed (by .equals comparison)
+ * Duplicates are not allowed (by .isSamePerson comparison)
  */
 public class AddressBook implements ReadOnlyAddressBook {
 
--- a/src/test/java/seedu/address/commons/util/CollectionUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/CollectionUtilTest.java
@@ -82,31 +82,6 @@ public class CollectionUtilTest {
         assertTrue(CollectionUtil.isAnyNonNull(new Object(), null));
     }
 
-    @Test
-    public void elementsAreUnique() {
-        // empty list
-        assertAreUnique();
-
-        // only one object
-        assertAreUnique((Object) null);
-        assertAreUnique(1);
-        assertAreUnique("");
-        assertAreUnique("abc");
-
-        // all objects unique
-        assertAreUnique("abc", "ab", "a");
-        assertAreUnique(1, 2);
-
-        // some identical objects
-        assertNotUnique("abc", "abc");
-        assertNotUnique("abc", "", "abc", "ABC");
-        assertNotUnique("", "abc", "a", "abc");
-        assertNotUnique(1, Integer.valueOf(1));
-        assertNotUnique(null, 1, Integer.valueOf(1));
-        assertNotUnique(null, null);
-        assertNotUnique(null, "a", "b", null);
-    }
-
     /**
      * Asserts that {@code CollectionUtil#requireAllNonNull(Object...)} throw {@code NullPointerException}
      * if {@code objects} or any element of {@code objects} is null.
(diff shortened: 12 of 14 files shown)


src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +59,17 @@ public Address getAddress() {
         return Collections.unmodifiableSet(tags.toSet());
     }
 
+    /**
+     * Returns true if both persons have the same identity fields (name and (phone or email)).
+     */
```
yamidark: Could add the {@code name} here too. 
Can add that this method defines a weaker notion of equality to check that the 2 persons are the same here instead of in `UniquePersonList#contains` and `UniquePersonList#add`.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +59,17 @@ public Address getAddress() {
...
+    public boolean isSamePerson(Person otherPerson) {
+        return (otherPerson == this)
```
yamidark: Can remove this line since we are only checking the fields as stated in comments.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -25,17 +25,20 @@
     private final ObservableList<Person> internalList = FXCollections.observableArrayList();
 
     /**
-     * Returns true if the list contains an equivalent person as the given argument.
+     * Checks if the list contains an equivalent person as the given argument.
```
yamidark: Don't think this change is necessary.

src/main/java/seedu/address/model/person/Person.java
Zhiyuan-Amos: The double parentheses looks weird.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +59,17 @@ public Address getAddress() {
...
+                || (otherPerson != null
+                    && otherPerson.getName().equals(this.getName())
+                    && (otherPerson.getPhone().equals(this.getPhone())
+                        || otherPerson.getEmail().equals(this.getEmail())));
```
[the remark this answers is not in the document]
Zhiyuan-Amos: Leaving `this` here helps in the readability even though the compiler doesn't complain about it.
Also, the `equals#(Object)` method in `Person` also uses `this`, so we can follow it as well.

src/main/java/seedu/address/model/person/Person.java
Zhiyuan-Amos: I am conflicted about this, in the sense that this seems that this works as an early return; if both references are referring to the same object, then they are definitely the same person.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -25,17 +25,20 @@
...
         requireNonNull(toCheck);
-        return internalList.contains(toCheck);
+        return internalList.stream().anyMatch(p -> p.isSamePerson(toCheck));
     }
 
     /**
      * Adds a person to the list.
      *
      * @throws DuplicatePersonException if the person to add is a duplicate of an existing person in the list.
+     * The @link{ReadOnlyPerson#isSamePerson} method is used for this comparison, which defines a weaker notion of
```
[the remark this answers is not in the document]
Zhiyuan-Amos: hello hello we no longer have `ReadOnlyPerson` :P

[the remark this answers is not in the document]
yamidark: Whoops missed that 1 xD

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -25,17 +25,20 @@
     private final ObservableList<Person> internalList = FXCollections.observableArrayList();
 
     /**
-     * Returns true if the list contains an equivalent person as the given argument.
+     * Checks if the list contains an equivalent person as the given argument.
+     * The {@link Person#isSamePerson} method is used for this comparison, which defines a weaker notion of equality.
```
Zhiyuan-Amos: ~~The~~ `Person#isSamePerson(Person)` ~~method~~

Zhiyuan-Amos: Though I wonder if we should declare this in the class header instead, maybe something like `A person is considered unique by comparing using Person#isSamePerson(Person)`.

src/main/java/seedu/address/model/person/Person.java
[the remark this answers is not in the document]
yamidark: >Also, the equals#(Object) method in Person also uses this, so we can follow it as well.

While I do agree having `this` helps in readability, the use of `this` for `equals#(Object) is not standardized throughout the codebase (`EditCommand` doesn't use it), so I thought for new instances we should follow the standard and remove it.

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -130,14 +117,32 @@ public void add() throws Exception {
...
+        command = PersonUtil.getAddCommand(toAdd);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a person with all fields same as another person in the address book except email -> rejected */
+        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_BOB)
+                .withAddress(VALID_ADDRESS_AMY).withTags(VALID_TAG_FRIEND).build();
+        command = PersonUtil.getAddCommand(toAdd);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a person with all fields same as another person in the address book except address -> rejected */
```
Zhiyuan-Amos: the comments can follow that of the test case for tags: Case: `add a duplicate person except with different X -> rejected`

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -130,14 +117,32 @@ public void add() throws Exception {
...
+        assertCommandSuccess(AMY);
 
         /* ----------------------------------- Perform invalid add operations --------------------------------------- */
 
         /* Case: add a duplicate person -> rejected */
         command = PersonUtil.getAddCommand(HOON);
         assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
 
+        /* Case: add a person with all fields same as another person in the address book except phone -> rejected */
+        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_AMY)
```
Zhiyuan-Amos: I wonder if we should create a constant `AMY` so that we can do `new PersonBuilder(AMY).withPhone(VALID_PHONE_BOB)` instead.

This means all the tests need to be changed though LOL. Maybe we can raise an issue for this.

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -130,14 +117,32 @@ public void add() throws Exception {
 
         /* Case: selects first card in the person list, add a person -> added, card selection remains unchanged */
         selectPerson(Index.fromOneBased(1));
-        assertCommandSuccess(CARL);
+        assertCommandSuccess(AMY);
```
Zhiyuan-Amos: why does this test change?

vivekscl: Oh, that was explained in the PR message:
```
In AddCommandSystemTest, line 120 was changed from assertCommandSuccess(CARL) to
assertCommandSuccess(AMY) to test for duplicate persons in the next few lines as 
the duplicate person tests require the default person to be used, which is either 
AMY or BOB.
```

src/main/java/seedu/address/model/person/UniquePersonList.java
vivekscl: >Though I wonder if we should declare this in the class header instead, maybe something like A person is considered unique by comparing using Person#isSamePerson(Person).

I think that's a good idea. Since this line is repeated in two areas, we can consolidate it at the header instead.

src/test/java/systemtests/AddCommandSystemTest.java
vivekscl: I'll raise a new issue for it soon then :+1: .

vivekscl:
>Also, as per the change in AB-Level2, we need to update Person#equals(Object) to verify tags as well?

Oh right. I'll do this in a new commit then.

vivekscl:
@Zhiyuan-Amos, for the verifying of tags, it seems like it won't be possible to add this until #845 is merged. This is because in `XmlSerializableAddressBookTest#toModelType_typicalPersonsFile_success` the `addressBookFromFile` contains many contacts without tags while the persons in `TypicalPersons` are built with a default tag, `friends`. So in this case, I would have to add the `friends` tag to every person in `typicalPersonsAddressBook.xml` for the test case to pass, but that would rather unintuitive. So in that case, I'll just fix the current issues first and then after #845 is merged I'll add the verification for tags?

Zhiyuan-Amos:
@vivekscl Yup, sounds good! :)

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -14,6 +14,7 @@
 
 /**
  * A list of persons that enforces uniqueness between its elements and does not allow nulls.
+ * A person is considered unique by comparing using {@link Person#isSamePerson}.
```
Zhiyuan-Amos: `Person#isSamePerson(Person)`

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +61,22 @@ public Address getAddress() {
         return Collections.unmodifiableSet(tags.toSet());
     }
 
+    /**
+     * Returns true if both persons have the same identity fields.
+     * This defines a weaker notion of equality between two persons.
```
Zhiyuan-Amos: This javadoc doesn't really quite tell the user that 2 persons are considered as same if they have the same name and **either** same phone or email. It may seem that all 3 need to be the same actually.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -29,7 +30,7 @@
      */
     public boolean contains(Person toCheck) {
         requireNonNull(toCheck);
-        return internalList.contains(toCheck);
+        return internalList.stream().anyMatch(p -> p.isSamePerson(toCheck));
```
Zhiyuan-Amos: can we do `anyMatch(toCheck::isSamePerson)`?

src/test/java/systemtests/AddCommandSystemTest.java
Zhiyuan-Amos: Remember to raise an issue :P

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -14,6 +14,7 @@
 
 /**
  * A list of persons that enforces uniqueness between its elements and does not allow nulls.
+ * A person is considered unique by comparing using Person#isSamePerson(Person).
```
Zhiyuan-Amos: I think we need a `{@code Person...}` :p

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -130,14 +117,32 @@ public void add() throws Exception {
...
 
         /* ----------------------------------- Perform invalid add operations --------------------------------------- */
 
         /* Case: add a duplicate person -> rejected */
         command = PersonUtil.getAddCommand(HOON);
```
yamidark: Maybe we can change this to `AMY` to be more consistent with the other duplicate tests below?

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -130,14 +117,32 @@ public void add() throws Exception {
...
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: add a duplicate person except with different address -> rejected */
+        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY)
+                .withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND).build();
+        command = PersonUtil.getAddCommand(toAdd);
+        assertCommandFailure(command, AddCommand.MESSAGE_DUPLICATE_PERSON);
+
         /* Case: add a duplicate person except with different tags -> rejected */
         // "friends" is an existing tag used in the default model, see TypicalPersons#ALICE
```
yamidark: Should we change the different tag test to follow the same format as above (the other tag `VALID_TAG_HUSBAND` works for this test as well)?

yamidark:
As mentioned in #855, we can use the other `PersonBuilder` constructor to write a more concise form for the duplicate tests here, which should also be much clearer to users what we are testing as well.

src/test/java/systemtests/AddCommandSystemTest.java
Zhiyuan-Amos: > As mentioned in #855, we can use the other PersonBuilder constructor to write a more concise form for the duplicate tests here, which should also be much clearer to users what we are testing as well.

I think we should change all the tests below to use `HOON` actually. Since we don't have an `AMY` constant but we have a `HOON` constant, we can use the method that you suggested.

vivekscl:
.As mentioned in #855, we can use the other PersonBuilder constructor to write a more concise form for the duplicate tests here, which should also be much clearer to users new developers what we are testing as well.

But since this refactoring is unrelated to this PR, we should do that in a separate PR?

src/test/java/systemtests/AddCommandSystemTest.java
vivekscl: Hmm, I'm not sure what this means :sweat_smile:

yamidark: Like this?
`toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY).withAddress(VALID_ADDRESS_AMY).withTags(VALID_TAG_HUSBAND).build();`

src/test/java/systemtests/AddCommandSystemTest.java
yamidark: Since we are adding the new tests here, we can change it here since `PersonBuilder` already has the other constructor which allows for the concise form.

yamidark:
Just noticed we should change the header comments for `AddressBook` too:
https://github.com/se-edu/addressbook-level4/blob/2f4405c75cd21111952565a9706a9793b475c41e/src/main/java/seedu/address/model/AddressBook.java#L21-L24

since we are using different comparison for `Persons`.

src/test/java/systemtests/AddCommandSystemTest.java
vivekscl: I did rewrite it to follow the above format, but I'm thinking, is it supposed to be done in this PR? :thinking:

src/main/java/seedu/address/model/person/Person.java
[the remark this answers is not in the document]
eugenepeh: > While I do agree having this helps in readability, the use of this for equals#(Object) is not standardized throughout the codebase (EditCommand doesn't use it), so I thought for new instances we should follow the standard and remove it.

I think in this case, we can leave it as it is, following the readability practice. At the same time, can probably raise an issue to change all the existing difference in standard.

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +61,22 @@ public Address getAddress() {
...
     }
 
+    /**
+     * Returns true if both persons of the same name have at least one other identity field that is the same.
+     * This defines a weaker notion of equality between two persons.
+     */
+    public boolean isSamePerson(Person otherPerson) {
+        return (otherPerson == this)
+                || (otherPerson != null
+                    && otherPerson.getName().equals(this.getName())
```
eugenepeh: isn't line wrap suppose to have an indentation of 8 spaces?

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -82,28 +81,14 @@ public void add() throws Exception {
...
 
-        /* Case: add a person with all fields same as another person in the address book except address -> added */
-        toAdd = new PersonBuilder().withName(VALID_NAME_AMY).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY)
-                .withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_FRIEND).build();
-        command = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_AMY + ADDRESS_DESC_BOB
+        /* Case: add a person with all fields same as another person in the address book except phone and email
+         * -> added
+         */
+        toAdd = new PersonBuilder(AMY).withPhone(VALID_PHONE_BOB).withEmail(VALID_EMAIL_BOB).build();
+        command = AddCommand.COMMAND_WORD + NAME_DESC_AMY + PHONE_DESC_BOB + EMAIL_DESC_BOB + ADDRESS_DESC_AMY
```
Zhiyuan-Amos: Can we use `PersonUtil.getAddCommand(toAdd)` here?

src/test/java/systemtests/AddCommandSystemTest.java
Zhiyuan-Amos: Sorry can someone explain this issue to me? I don't quite get what's being discussed here.

yamidark: I suggested that we should change the old different tags test to follow the same format as the other different field tests above.
However since it is not a newly added test we are not sure if we should change the format in this PR, but perhaps when resolving #855 instead.

Zhiyuan-Amos: Ah ok! :) Yup let's not change the format here.

src/main/java/seedu/address/model/person/Person.java
vivekscl: You mean this line should be indented 4 spaces to the left? In this case, it's indented this way because there is a nested condition, and the extra indentation is to clearly show the nesting.

eugenepeh: No, I mean it should be indented another 4 spaces to the right. This seems to be the way we do it according to our [coding convention](https://oss-generic.github.io/process/codingStandards/CodingStandard-Java.html#layout), example no. 3.

eugenepeh: So probably something like this?
<pre>
return (otherPerson == this) || (otherPerson != null
        && otherPerson.getName().equals(this.getName())
        && (otherPerson.getPhone().equals(this.getPhone())
                || otherPerson.getEmail().equals(this.getEmail())));
</pre>

I think we do not have such occurrence in our AB before (or rather most `.equal` puts the same amount of indentation despite different abstraction level). @Zhiyuan-Amos , any comment on this?

Zhiyuan-Amos: Now that @eugenepeh mentioned, I think this looks kinda confusing because there's too many parentheses. As such, I'm wondering whether we should actually consider doing it the lengthy way for the purpose of code readability:

```java
public boolean isSamePerson(Person otherPerson) {
    if (otherPerson == this) {
        return true;
    }

    if (otherPerson == null) {
        return false;
    }

    return otherPerson.getName().equals(this.getName())
            && (otherPerson.getPhone().equals(this.getPhone()) || otherPerson.getEmail().equals(this.getEmail()));
}
```

vivekscl: The if statement looks alot better to read. On a side note, what do you think about this version?
```java
boolean sameObjectAndNotNull = otherPerson == this && otherPerson != null;

return sameObjectAndNotNull || otherPerson.getName().equals(this.getName())
        && (otherPerson.getPhone().equals(this.getPhone()) || otherPerson.getEmail().equals(this.getEmail()));
```

Zhiyuan-Amos: I think the logic isn't easy to catch. That's because the `NotNull` corresponds to the later portion (see code below) to ensure that we do not get a NPE, whereas in your code, the link between `NotNull` and the code below is not linked.

```java
return otherPerson.getName().equals(this.getName())
            && (otherPerson.getPhone().equals(this.getPhone()) || otherPerson.getEmail().equals(this.getEmail()));
```

vivekscl: Oh, I understand now. I'll go with the if statement suggestion then :+1: .

Zhiyuan-Amos:
First time trying to comment on phone, seems like it doesn't work out quite well. You can see them if you click into the link to the file. Sorry about that!

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -131,7 +131,7 @@ private File createXmlFileWithPersons(int personCount) throws Exception {
             builder.append("<persons>\n");
             builder.append("<name>a</name>\n");
             builder.append("<phone>").append(i).append("00</phone>\n");
-            builder.append("<email>a@aa</email>\n");
+            builder.append("<email>").append(i).append("a@aa</email>\n");
```
Zhiyuan-Amos: Can we append i to name instead? Then we can remove the appending of i to phone. That way, we may not have to increase the test timeout.

src/test/java/systemtests/AddCommandSystemTest.java
```diff
@@ -80,31 +75,15 @@ public void add() throws Exception {
         assertCommandSuccess(command, model, expectedResultMessage);
 
         /* Case: add a person with all fields same as another person in the address book except name -> added */
-        toAdd = new PersonBuilder().withName(VALID_NAME_BOB).withPhone(VALID_PHONE_AMY).withEmail(VALID_EMAIL_AMY)
-                .withAddress(VALID_ADDRESS_AMY).withTags(VALID_TAG_FRIEND).build();
-        command = AddCommand.COMMAND_WORD + NAME_DESC_BOB + PHONE_DESC_AMY + EMAIL_DESC_AMY + ADDRESS_DESC_AMY
-                + TAG_DESC_FRIEND;
+        toAdd = new PersonBuilder(AMY).withName(VALID_NAME_BOB).build();
+        command = PersonUtil.getAddCommand(toAdd);
```
Zhiyuan-Amos: There's no reason to change these lines right? I.e. the tests don't fail if these lines don't change.

yamidark:
Should we update the `.equals` method in `PersonCardHandle` to include checking of `tags` too:
https://github.com/se-edu/addressbook-level4/blob/a4ef5229e29bb1da91d462a650ee82c243ab1b7c/src/test/java/guitests/guihandles/PersonCardHandle.java#L73-L81

Zhiyuan-Amos:
> Should we update the .equals method in PersonCardHandle to include checking of tags too:

@yamidark this is a good catch. We should consider having a reference to `Person` inside the `PersonCardHandle` actually so that we can simplify the `equals` method in `PersonCardHandle`, and also so that we won't forget to make changes in the event we update `Person#equals(Object)`

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+        // same object -> returns true
+        assertTrue(person1.isSamePerson(person1));
+
+        // null -> returns false
+        assertFalse(person1.isSamePerson(null));
+
+        // different person -> returns false
+        assertFalse(person1.isSamePerson(person2));
+
+        // same attributes except for name -> returns false
```
Zhiyuan-Amos: I took a look at `EditPersonDescriptorTest`, can we shorten this to `different name -> returns false`?

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static seedu.address.testutil.TypicalPersons.BOB;
+
+import org.junit.Test;
+
+import seedu.address.testutil.PersonBuilder;
+
+public class PersonTest {
+
+    private Person person1 = ALICE;
+    private Person person2 = BOB;
```
Zhiyuan-Amos: You can just call ALICE and BOB in the tests :P you don't have to assign them haha.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertTrue(person1.isSamePerson(person1));
+
+        // null -> returns false
+        assertFalse(person1.isSamePerson(null));
+
+        // different person -> returns false
+        assertFalse(person1.isSamePerson(person2));
+
+        // same attributes except for name -> returns false
+        Person person3 = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
```
Zhiyuan-Amos: Let's name this `editedAlice` as `person3` is quite undescriptive.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+        person3 = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).build();
+        assertTrue(person1.isSamePerson(person3));
+
+        // same attributes except for email -> returns true
+        person3 = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).build();
+        assertTrue(person1.isSamePerson(person3));
+    }
+
+    @Test
+    public void equals() {
```
Zhiyuan-Amos: You can add the different tags test first and call `assertTrue` for 2 person with different tags.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+        Person person3 = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
+        assertFalse(person1.isSamePerson(person3));
+
+        // same attributes except for phone -> returns true
+        person3 = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).build();
+        assertTrue(person1.isSamePerson(person3));
+
+        // same attributes except for email -> returns true
+        person3 = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).build();
+        assertTrue(person1.isSamePerson(person3));
```
Zhiyuan-Amos: You should consider:
1. `same name, same phone, different attributes -> pass`
1. `same name, same email, different attributes -> pass`

Instead of:
1. `same attributes except for phone -> pass`
1. `same attributes except for email -> pass`

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    public void isSamePerson() {
+        // same object -> returns true
+        assertTrue(person1.isSamePerson(person1));
+
+        // null -> returns false
+        assertFalse(person1.isSamePerson(null));
+
+        // different person -> returns false
```
Zhiyuan-Amos: This test seems to be not tight enough. For failure cases, we should do something like:
1. Same attributes except phone and email.
1. Same attributes except name (which you've done below alrd).

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertFalse(person1.isSamePerson(person2));
+
+        // same attributes except for name -> returns false
+        Person person3 = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
+        assertFalse(person1.isSamePerson(person3));
+
+        // same attributes except for phone -> returns true
+        person3 = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).build();
+        assertTrue(person1.isSamePerson(person3));
```
yamidark: We should also add a test for `same name, different phone and email -> false` right?

Zhiyuan-Amos: This is same as what I commented [earlier](https://github.com/se-edu/addressbook-level4/pull/844#discussion_r176763721) right? :P

yamidark: whoops, missed that....

vivekscl:
There seems to be a problem with Travis in installing Java 9. It worked just yesterday and I've been restarting the build for quite some time to no avail. So it seems like I'll have to wait before I request a review.

Zhiyuan-Amos:
Does CanIHasReview blocks you from requesting for reviews? If not, you can just request :)

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,80 @@
...
+        editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withAddress(VALID_ADDRESS_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+    }
+
+    @Test
+    public void equals() {
+        // same object -> returns true
+        assertTrue(ALICE.equals(ALICE));
+
```
yamidark: I think we should also have a check for all same attributes, but different object.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,84 @@
...
+        Person aliceCopy = new PersonBuilder(ALICE).build();
+        assertTrue(ALICE.equals(aliceCopy));
+
+        // same object -> returns true
+        assertTrue(ALICE.equals(ALICE));
+
+        // null -> returns false
+        assertFalse(ALICE.equals(null));
+
+        // different types -> returns false
```
Zhiyuan-Amos: Should be singular `type`? :p

Zhiyuan-Amos: @vivekscl let's make this change and I'll have prof to review.

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,84 @@
...
+        assertFalse(ALICE.isSamePerson(editedAlice));
+
+        // different name -> returns false
+        editedAlice = new PersonBuilder(ALICE).withName(VALID_NAME_BOB).build();
+        assertFalse(ALICE.isSamePerson(editedAlice));
+
+        // same name, same phone, different attributes -> returns true
+        editedAlice = new PersonBuilder(ALICE).withEmail(VALID_EMAIL_BOB).withAddress(VALID_ADDRESS_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
```
damithc: why no case for `same name, same email, same phone`?

vivekscl: Ahh, my bad, didn't consider that case :sweat_smile: .

Zhiyuan-Amos: I thought that because we are doing white-box testing, so there's no need to verify for `same name, same email and same phone`; `same name, same email` and `same name, same phone` will do.

damithc: Unit tests are cheap; we an do both white and black-box testing (I guess that's more like greybox testing :-p)

src/test/java/seedu/address/model/person/PersonTest.java
```diff
@@ -0,0 +1,91 @@
...
+        editedAlice = new PersonBuilder(ALICE).withPhone(VALID_PHONE_BOB).withAddress(VALID_ADDRESS_BOB)
+                .withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+
+        // same name, same phone, same email, different attributes -> returns true
+        editedAlice = new PersonBuilder(ALICE).withAddress(VALID_ADDRESS_BOB).withTags(VALID_TAG_HUSBAND).build();
+        assertTrue(ALICE.isSamePerson(editedAlice));
+
+        // different person -> returns false
+        assertFalse(ALICE.isSamePerson(BOB));
```
Zhiyuan-Amos: You added this in the latest revision. Y tho, since `different person` is defined already in `different phone and email` and `different name`.

vivekscl: But to cover all cases, shouldn't we test for the case when all of these are false at the same time? The test cases you mentioned are testing them separately right?

Zhiyuan-Amos: Yup, but using your line of argument then we can also test for:

1. Same name, same email, same phone, same address, different tags
1. Same name, same email, same phone, same tags, different address
1. Everything different except address
1. Everything different except tags

And so on.

As such, I'm suggesting that we only test corner cases (e.g. Different name, because that's a key condition in the code), rather than testing all permutations.

vivekscl: Ahh, understood! Thanks for the clarification :smile: .

vivekscl:
Sure! But on a side note, since I have yet to add the equality for `tags` after #845 is merged, do I call for last reviews now or once I add the equality for the `tags`?

Zhiyuan-Amos:
We can do the last call for reviews after #845 is merged. :) I forgot about that ><

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -59,6 +61,25 @@ public Address getAddress() {
...
+     */
+    public boolean isSamePerson(Person otherPerson) {
+        if (otherPerson == this) {
+            return true;
+        }
+
+        return otherPerson != null
+                && otherPerson.getName().equals(this.getName())
+                && (otherPerson.getPhone().equals(this.getPhone()) || otherPerson.getEmail().equals(this.getEmail()));
+
```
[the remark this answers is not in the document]
vivekscl: Oh, seems like I missed that. Will fix it along with the adding of equality for tags after #845 is merged. :+1:

Zhiyuan-Amos:
@vivekscl you can continue on this PR now :)

Zhiyuan-Amos:
Well you can post it on the [issue tracker](https://github.com/CanIHasReview/CanIHasReview/issues) :P

src/main/java/seedu/address/model/person/Person.java
```diff
@@ -72,7 +93,8 @@ public boolean equals(Object other) {
         return otherPerson.getName().equals(this.getName())
                 && otherPerson.getPhone().equals(this.getPhone())
                 && otherPerson.getEmail().equals(this.getEmail())
-                && otherPerson.getAddress().equals(this.getAddress());
+                && otherPerson.getAddress().equals(this.getAddress())
+                && otherPerson.getTags().equals(this.getTags());
```
Zhiyuan-Amos: you should add this as a separate commit, since it's not exactly related to the addition of `isSamePerson(Person)`. Same for `PersonCardHandle`.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -77,6 +77,10 @@ public boolean equals(Person person) {
         return getName().equals(person.getName().fullName)
                 && getAddress().equals(person.getAddress().value)
                 && getPhone().equals(person.getPhone().value)
-                && getEmail().equals(person.getEmail().value);
+                && getEmail().equals(person.getEmail().value)
+                && person.getTags().stream()
+                .map(tag -> tag.tagName)
+                .collect(Collectors.toList())
+                .containsAll(getTags());
```
Zhiyuan-Amos: this `containsAll` only ensure that `person.getTags()` contains all the tags in `getTags()` but not the other way round; you've not ensured that `getTags()` contains all the tags in `person.getTags()` :p

See if there's better [solutions](https://stackoverflow.com/a/22619319/8828382).

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -181,11 +187,6 @@ public void edit() throws Exception {
         command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_BOB
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
-
-        /* Case: edit a person with new values same as another person's values but with different tags -> rejected */
```
Zhiyuan-Amos: Why is this test edited?

vivekscl:
Oh because this test passes now since different tags mean they are different persons.

src/test/java/guitests/guihandles/PersonCardHandle.java
vivekscl: In this case, would it be better to implement our own version of `isEqualCollection` in `CollectionsUtil` or just import the method from Apache commons?

Zhiyuan-Amos: Easier to import, don't have to write tests for the method :) if you implement your own isEqualCollection, you need to write tests for it.

vivekscl:
There still seems to be an error with `CanIHasReview` for this PR. I'll request for it the old fashion way then. 
@Zhiyuan-Amos Ready for Review.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -77,6 +79,9 @@ public boolean equals(Person person) {
         return getName().equals(person.getName().fullName)
                 && getAddress().equals(person.getAddress().value)
                 && getPhone().equals(person.getPhone().value)
-                && getEmail().equals(person.getEmail().value);
+                && getEmail().equals(person.getEmail().value)
+                && CollectionUtils.isEqualCollection(getTags(), person.getTags().stream()
+                                                                                .map(tag -> tag.tagName)
+                                                                                .collect(Collectors.toList()));
```
yamidark: Actually, shouldn't we just change the `getTags()` to give a `Set` instead? Then we can just use `.equals` and won't need to import another library.

vivekscl: I've tried changing it to return a `Set<String>` and it does work. And since `getTags()` is only used in `GuiAssert` we only have to change it to test for a `Set` instead of a `List`. And it doesn't seem to violate anything. It seems like a simpler solution, so I think we should do this too.

Zhiyuan-Amos: Great work guys :) When returning tags as a `Set`, we should ensure the cardinality is the same between the `List` and the `Set`.

Zhiyuan-Amos: Hang on, `getTags()` shouldn't be returning a `Set` because a `PersonCard` can have multiple tags of the same value. :P

So, `getTags()` still returns a `List`, but you can convert it to a `Set` after ensuring the cardinality is the same between the `List` and the `Set`.

vivekscl: >after ensuring the cardinality is the same between the List and the Set

Hmm, by cardinality do you mean the length of the `List` and `Set`? So only if the length is the same we should convert the `List` into a `Set`? But wouldn't that mean it's better just to convert the `person.getTags()` to a `List` and then use the library that was imported?

Zhiyuan-Amos: haha yup you're right.

vivekscl: Alright then, I'll just continue with the current progress and modify the readme file :+1: .

Zhiyuan-Amos:
@damithc gentle reminder for your [review](https://github.com/se-edu/addressbook-level4/pull/844#pullrequestreview-110237525) :)

src/test/java/guitests/guihandles/PersonCardHandle.java
damithc: Does this comply with your indentation rules? I thought we use 8-spaces?

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -77,6 +79,9 @@ public boolean equals(Person person) {
         return getName().equals(person.getName().fullName)
                 && getAddress().equals(person.getAddress().value)
                 && getPhone().equals(person.getPhone().value)
-                && getEmail().equals(person.getEmail().value);
+                && getEmail().equals(person.getEmail().value)
+                && CollectionUtils.isEqualCollection(getTags(), person.getTags().stream()
```
yamgent: According to [Guava's documentation](https://github.com/google/guava/wiki/ApacheCommonCollectionsEquivalents), it looks like you can achieve the same using the `ImmutableMultiset` code:

> If both are `Set`s or `Multiset`s, use `equals()`; otherwise **`ImmutableMultiset.copyOf(a).equals(ImmutableMultiset.copyOf(b)`**

Was this tried before? Seems like we can avoid importing a new library if we use this instead.

vivekscl: I just tried this. All tests pass. So I guess it's better to use this then! :smile:

yamidark:
Just thought of this, we should update the header comments of `DuplicatePersonException` as well:
https://github.com/se-edu/addressbook-level4/blob/ae383756eaef6702531f5258fe501b7c0e2c831d/src/main/java/seedu/address/model/person/exceptions/DuplicatePersonException.java#L5-L8

Since the header comments mention `duplicate Person objects` which implies the stronger `.equals` method instead, we should change this to something like `duplicate person being added` to imply we are using the weaker `isSamePerson(Person)` instead?

Also, `UniquePersonList#setPerson(Person, Person)` should be updated as well:
https://github.com/se-edu/addressbook-level4/blob/ae383756eaef6702531f5258fe501b7c0e2c831d/src/main/java/seedu/address/model/person/UniquePersonList.java#L62-L65

we should call `(!target.isSamePerson(editedPerson) && internalList.contains(editedPerson))` instead, else we could throw `DuplicatePersonException()` incorrectly, for situations such as when we only want to change the data fields of the person.

vivekscl:
**To be added to the header comments of UniquePersonList**
Adding and updating of `Persons` uses `Person#isSamePerson(Person)` for equality so as to ensure that the person being added or updated is unique in terms of identity with respect to the `UniquePersonList`. However, the removal of a `Person` uses `Person#equals(Person)` so as to ensure that the exact person will be removed.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -5,20 +5,23 @@
...
-import seedu.address.commons.util.CollectionUtil;
 import seedu.address.model.person.exceptions.DuplicatePersonException;
 import seedu.address.model.person.exceptions.PersonNotFoundException;
 
 /**
  * A list of persons that enforces uniqueness between its elements and does not allow nulls.
+ * A person is considered unique by comparing using {@code Person#isSamePerson(Person)}. As such, adding and updating of
+ * Persons uses Person#isSamePerson(Person) for equality so as to ensure that the person being added or updated is
+ * unique in terms of identity in the UniquePersonList. However, the removal of a Person uses Person#equals(Object) so
+ * as to ensure that the Person with exactly the same fields will be removed.
```
Zhiyuan-Amos: @damithc are you okay with this?

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -115,4 +118,13 @@ public boolean equals(Object other) {
     public int hashCode() {
         return internalList.hashCode();
     }
+
+    /**
+     * Returns true of the given Persons list contains only unique elements and false otherwise.
+     */
+    private boolean elementsAreUnique(List<Person> persons) {
```
Zhiyuan-Amos: This should be renamed to `personsAreUnique`.

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -115,4 +118,13 @@ public boolean equals(Object other) {
     public int hashCode() {
         return internalList.hashCode();
     }
+
+    /**
+     * Returns true of the given Persons list contains only unique elements and false otherwise.
```
Zhiyuan-Amos: Returns true if {@code persons} contains only unique persons.

src/main/java/seedu/address/model/person/exceptions/DuplicatePersonException.java
```diff
@@ -3,7 +3,7 @@
 import seedu.address.commons.exceptions.DuplicateDataException;
 
 /**
- * Signals that the operation will result in duplicate Person objects.
+ * Signals that the operation will result in duplicate Persons meaning, two Persons with the same identity.
```
Zhiyuan-Amos: The phrasing is weird. :P 

Signals that the operation will result in duplicate Persons (Persons are considered duplicates if they have the same identity).

src/main/java/seedu/address/model/person/UniquePersonList.java
```diff
@@ -115,4 +118,13 @@ public boolean equals(Object other) {
...
+     */
+    private boolean elementsAreUnique(List<Person> persons) {
+        TreeSet<Person> personsTreeSet = new TreeSet<>((p1, p2) -> p1.isSamePerson(p2) ? 0 : 1);
+        personsTreeSet.addAll(persons);
+        return persons.size() == personsTreeSet.size();
```
Zhiyuan-Amos: Hmm I realized this way of implementing looks weird :P Let's use the traditional loops to solve this.

src/main/java/seedu/address/model/person/UniquePersonList.java
damithc: Sounds OK @Zhiyuan-Amos . Do you anticipate any issues with it?

Zhiyuan-Amos: Sounds ok to me too, clearing this decision with you since there's different behavior for `add / set` vs `delete`. :)

@vivekscl let's standardize by using lower-case `p` for `Persons`.

src/main/java/seedu/address/model/person/UniquePersonList.java
vivekscl: After trying to use traditional loops for this, it seems like we could use streams instead. But the stream does look rather long. So is it better to stick with traditional loops instead since it's easier to understand?

Zhiyuan-Amos: You can do both and add them on top of existing commits then tag me for a look :)

vivekscl: Oh this is how it looks like
Streams:
deepinscreenshot_select-area_20180413162906

For Loops:
deepinscreenshot_select-area_20180413162929

Zhiyuan-Amos: @yamgent any thoughts on this? I personally think nested streams are kinda hard to understand :P

yamgent: The stream implementation looks pretty odd on first glance. :P I think the traditional way looks better.

It is a shame that stream's `distinct()` doesn't allow you to use methods other than `Object.equals()` for comparison, I would have preferred that if it was possible.

vivekscl:
@yamidark for `AddressBookTest`, you mean instead of using `ALICE` twice, we should use two persons  with only the identity fields being equal?

vivekscl:
>This test can be updated to edit a person with same data fields but different identity fields.

I did try adding this test, but now the first person has different identity fields and all the other tests below have to change as well. 

After this change, I've tried to change the second person instead of the first but then there's no `ADDRESS_DESC_CARL`. Another workaround would then be to change the tests below, but that's out of the scope of this PR right?

vivekscl:
>Also, should we consider a case for:
Case: edit a person with new values same as another person's values but with different name -> rejected?

@Zhiyuan-Amos isn't this case actually a valid one? Since if the name is different, it is already assumed to be another person altogther?

Zhiyuan-Amos:
> isn't this case actually a valid one? Since if the name is different, it is already assumed to be another person altogther?

Oh yes, what am I thinking. Sorry!

vivekscl:
@Zhiyuan-Amos I realized there's no need to add a private method, `editFirstPersonToBob` as after the 3rd commit, it's enough to change the `personToEdit` to the first person in the list and the other tests that follow would still pass. Furthermore, the tests further down below, namely the identity field tests, just require that `BOB` be **on the list** and **not the first person on the list**.

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -186,6 +193,16 @@ public void edit() throws Exception {
...
+
+        /* Case: edit a person with new values same as another person's values but with different phone -> rejected */
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_AMY + EMAIL_DESC_BOB
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
+
+        /* Case: edit a person with new values same as another person's values but with different email -> rejected */
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_BOB + PHONE_DESC_BOB + EMAIL_DESC_AMY
+                + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
+        assertCommandFailure(command, EditCommand.MESSAGE_DUPLICATE_PERSON);
```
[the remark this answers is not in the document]
vivekscl: Why would different address be rejected? It's not an identity field right?

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -84,16 +86,21 @@ public void edit() throws Exception {
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandSuccess(command, index, BOB);
 
-        /* Case: edit some fields -> edited */
+        /* Case: edit a person with new values same as another person's data fields but different identity fields
+         * -> edited
+         */
```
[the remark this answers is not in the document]
vivekscl: Sounds good to me :+1:

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -97,10 +97,16 @@ public void edit() throws Exception {
                 .withEmail(VALID_EMAIL_AMY).build();
         assertCommandSuccess(command, index, editedPerson);
 
+        /* Case: edit a person with new values same as another person's values but with different tags -> edited */
```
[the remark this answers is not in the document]
vivekscl: Ahh, that is true. Will make the changes accordingly :+1:

src/test/java/systemtests/EditCommandSystemTest.java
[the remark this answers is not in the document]
yamidark: >Oh right you mean, different address but everything else the same right?

Yep, same as **another** person, but different address

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -97,10 +97,16 @@ public void edit() throws Exception {
...
+        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + NAME_DESC_AMY + PHONE_DESC_AMY + EMAIL_DESC_AMY
+                + ADDRESS_DESC_BOB + TAG_DESC_HUSBAND;
+        personToEdit = getModel().getFilteredPersonList().get(index.getZeroBased());
+        editedPerson = new PersonBuilder(personToEdit).withTags(VALID_TAG_HUSBAND).build();
+        assertCommandSuccess(command, index, editedPerson);
+
         /* Case: clear tags -> cleared */
         index = INDEX_FIRST_PERSON;
         command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + " " + PREFIX_TAG.getPrefix();
-        personToEdit = getModel().getFilteredPersonList().get(index.getZeroBased());
```
Zhiyuan-Amos: This line shouldn't be deleted right even though the test passes? Cos the `personToEdit` in the previous test (line 103) and the `personToEdit` now are different already.

vivekscl: Oh, I deleted because originally before this PR it wasn't there. My bad! Will fix it accordingly :+1:

src/test/java/systemtests/EditCommandSystemTest.java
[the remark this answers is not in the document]
yamidark: In this case, we are changing only the tags of the person himself, thus the editedPerson is comparing with it's old self.
Actually, I don't think this test is even necessary, since we have the `/* Case: edit a person with new values same as existing values -> edited */` test above which checks that we can edit a person to be equal to himself.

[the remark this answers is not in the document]
vivekscl: Oh, but this check is needed since the tags have been added to the equality method right?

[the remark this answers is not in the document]
vivekscl: Oh right, it uses the identity equality method now. Just tested it, and it is rejected as expected. I'll just leave the test alone then.

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -84,16 +86,28 @@ public void edit() throws Exception {
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandSuccess(command, index, BOB);
 
-        /* Case: edit some fields -> edited */
+        /* Case: edit a person with new values same as another person's values but with different name -> edited */
```
Zhiyuan-Amos: At this stage of the test, we don't have a person `AMY`, and the first person is `BOB` right? As such, we are simply editing `BOB` to have the name of `AMY`, while other details remain as `BOB`'s.

So, this test case isn't actually doing what we want it to do :P We want to test that we have `BOB` and `FOO`, then we edit `FOO` to have the name of `AMY`, while other details remain as `BOB`'s, and to verify that the edit is successful right?

vivekscl: In that case, we should be edit using `INDEX_SECOND_PERSON` instead of `INDEX_FIRST_PERSON`?

vivekscl: This is how the code would like after the changes:
deepinscreenshot_select-area_20180418144130

The change was also done to the `same values, different phone and email` test since that seemed to exhibit the fallacy you mentioned.

Zhiyuan-Amos: Yup looks good!

src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
```diff
@@ -27,6 +27,7 @@
         <phone>87652533</phone>
         <email>cornelia@example.com</email>
         <address>10th street</address>
+        <tagged>friends</tagged>
```
yamidark: Shouldn't this be in [3/3]?

vivekscl: Ahh, just tested it. It should be in [3/3]. Great catch!

src/test/java/systemtests/EditCommandSystemTest.java
```diff
@@ -84,16 +87,26 @@ public void edit() throws Exception {
                 + ADDRESS_DESC_BOB + TAG_DESC_FRIEND + TAG_DESC_HUSBAND;
         assertCommandSuccess(command, index, BOB);
 
-        /* Case: edit some fields -> edited */
-        index = INDEX_FIRST_PERSON;
-        command = EditCommand.COMMAND_WORD + " " + index.getOneBased() + TAG_DESC_FRIEND;
-        Person personToEdit = getModel().getFilteredPersonList().get(index.getZeroBased());
-        editedPerson = new PersonBuilder(personToEdit).withTags(VALID_TAG_FRIEND).build();
+        /* Case: edit a person with new values same as another person's values but with different name -> edited */
+        index = INDEX_SECOND_PERSON;
```
Zhiyuan-Amos: hang on, over here we should do this:

https://github.com/se-edu/addressbook-level4/blob/69b40e4b1a5db3552b3570228fb7e81eb6860a42/src/test/java/systemtests/EditCommandSystemTest.java#L178-L180

This ensures that the preconditions of the test are met :)

vivekscl: Should we use `assertNotEquals` instead?

Zhiyuan-Amos: ok!

vivekscl: And this should be down for the test below as well?

Zhiyuan-Amos: It should be done before the test `/* Case: edit a person with new values same as another person's values but with different tags -> rejected */`, but since it's not done, then nvm :P

vivekscl:
@damithc for your review :smile:

yamidark:
@vivekscl Remember to do last calls for review in gitter.

vivekscl:
@Zhiyuan-Amos is this ready for merging? :smile: 
Here's the merge commit message:
```
Person class contains only one equality method Person#equals(Object) 
which checks if all fields of the Person except tags are identical.

Two Persons of the same name that have at least one other equivalent 
identity field (phone or email) but different data fields (address and/
or tags), is considered different even though they are actually the same 
person.

Let's add another equality method, Person#isSamePerson(Person), for 
identity fields in Person class and use this equality check when adding
or editing a person.
```

PR #845, opened 2018-03-11
TypicalPersons: Unintuitive interaction with PersonBuilder #685

### LINKED ISSUE #685: TypicalPersons: Unintuitive interaction with PersonBuilder
In `TypicalPersons`, we see this line of code: 
```
HOON = new PersonBuilder().withName("Hoon Meier").withPhone("8482424")
                .withEmail("stefan@example.com").withAddress("little india").build();
```

On first sight, this looks like `Hoon Meier` doesn't have tags. However, he / she actually has `friends` tag because of the way `PersonBuilder` gives default values.

This seems unintuitive.

Rinder5:
resolves #685 

```
If no tag is set when constructing a new Person using PersonBuilder, 
the tag "friends" is added to the Person by default.

This should not happen as tags are optional. As a result, all Persons 
in TypicalPersons have the "friends" tag while this isn't the case for 
the Persons in typicalPersonsAddressBook.xml, which is unintuitive.

Let's replace the default tag in PersonBuilder with an empty set of 
tags.
```

### DIFF
--- a/src/test/java/seedu/address/logic/parser/AddressBookParserTest.java
+++ b/src/test/java/seedu/address/logic/parser/AddressBookParserTest.java
@@ -66,7 +66,7 @@ public class AddressBookParserTest {
         Person person = new PersonBuilder().build();
         EditPersonDescriptor descriptor = new EditPersonDescriptorBuilder(person).build();
         EditCommand command = (EditCommand) parser.parseCommand(EditCommand.COMMAND_WORD + " "
-                + INDEX_FIRST_PERSON.getOneBased() + " " + PersonUtil.getPersonDetails(person));
+                + INDEX_FIRST_PERSON.getOneBased() + " " + PersonUtil.getEditPersonDescriptorDetails(descriptor));
         assertEquals(new EditCommand(INDEX_FIRST_PERSON, descriptor), command);
     }
 
--- a/src/test/java/seedu/address/testutil/PersonUtil.java
+++ b/src/test/java/seedu/address/testutil/PersonUtil.java
@@ -6,8 +6,12 @@ import static seedu.address.logic.parser.CliSyntax.PREFIX_NAME;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_PHONE;
 import static seedu.address.logic.parser.CliSyntax.PREFIX_TAG;
 
+import java.util.Set;
+
 import seedu.address.logic.commands.AddCommand;
+import seedu.address.logic.commands.EditCommand.EditPersonDescriptor;
 import seedu.address.model.person.Person;
+import seedu.address.model.tag.Tag;
 
 /**
  * A utility class for Person.
@@ -33,6 +37,26 @@ public class PersonUtil {
         person.getTags().stream().forEach(
             s -> sb.append(PREFIX_TAG + s.tagName + " ")
         );
+        return sb.toString();
+    }
+
+    /**
+     * Returns the part of command string for the given {@code EditPersonDescriptor}'s details.
+     */
+    public static String getEditPersonDescriptorDetails(EditPersonDescriptor descriptor) {
+        StringBuilder sb = new StringBuilder();
+        descriptor.getName().ifPresent(name -> sb.append(PREFIX_NAME).append(name.fullName).append(" "));
+        descriptor.getPhone().ifPresent(phone -> sb.append(PREFIX_PHONE).append(phone.value).append(" "));
+        descriptor.getEmail().ifPresent(email -> sb.append(PREFIX_EMAIL).append(email.value).append(" "));
+        descriptor.getAddress().ifPresent(address -> sb.append(PREFIX_ADDRESS).append(address.value).append(" "));
+        if (descriptor.getTags().isPresent()) {
+            Set<Tag> tags = descriptor.getTags().get();
+            if (tags.isEmpty()) {
+                sb.append(PREFIX_TAG);
+            } else {
+                tags.forEach(s -> sb.append(PREFIX_TAG).append(s.tagName).append(" "));
+            }
+        }
         return sb.toString();
     }
 }
--- a/src/test/java/seedu/address/testutil/TypicalPersons.java
+++ b/src/test/java/seedu/address/testutil/TypicalPersons.java
@@ -35,7 +35,7 @@ public class TypicalPersons {
     public static final Person CARL = new PersonBuilder().withName("Carl Kurz").withPhone("95352563")
             .withEmail("heinz@example.com").withAddress("wall street").build();
     public static final Person DANIEL = new PersonBuilder().withName("Daniel Meier").withPhone("87652533")
-            .withEmail("cornelia@example.com").withAddress("10th street").build();
+            .withEmail("cornelia@example.com").withAddress("10th street").withTags("friends").build();
     public static final Person ELLE = new PersonBuilder().withName("Elle Meyer").withPhone("9482224")
             .withEmail("werner@example.com").withAddress("michegan ave").build();
     public static final Person FIONA = new PersonBuilder().withName("Fiona Kunz").withPhone("9482427")
--- a/src/test/java/seedu/address/testutil/PersonBuilder.java
+++ b/src/test/java/seedu/address/testutil/PersonBuilder.java
@@ -20,7 +20,6 @@ public class PersonBuilder {
     public static final String DEFAULT_PHONE = "85355255";
     public static final String DEFAULT_EMAIL = "alice@gmail.com";
     public static final String DEFAULT_ADDRESS = "123, Jurong West Ave 6, #08-111";
-    public static final String DEFAULT_TAGS = "friends";
 
     private Name name;
     private Phone phone;
@@ -33,7 +32,7 @@ public class PersonBuilder {
         phone = new Phone(DEFAULT_PHONE);
         email = new Email(DEFAULT_EMAIL);
         address = new Address(DEFAULT_ADDRESS);
-        tags = SampleDataUtil.getTagSet(DEFAULT_TAGS);
+        tags = new HashSet<>();
     }
 
     /**


Rinder5:
I wanted to update the tests in a different commit, but I think each commit is supposed to be passing. So I have a bunch of code which doesn't make sense, which I'll mention here.

Two tests break after removing the default "friends" tag:
- FindCommandSystemTest: This one uses the "friends" from to Daniel, so I added it to him in TypicalPersons
- AddressBookParserTest: This one breaks either parseCommand_edit or parseCommand_add (and others) depending on how empty tags are handled in PersonBuilder. If tags are set as null, many tests break as the field is not allowed to be empty. If tags are set as an empty hashset (commit implementation), parseCommand_edit fails, as EditPersonDescriptor adds "t/" to the command, which should not exist. This issue does not occur if there is a tag assigned to the person. I decided to do this by using a person from TypicalPersons, because I thought it might look weird to use the standard personbuilder except that it added a tag.

Rinder5:
While the PR does fix that issue, I felt it didn't really describe what I was doing with the PR, since I barely even touch TypicalPersons. If you think it should still be named that though, I can change it.

vivekscl:
>While the PR does fix that issue, I felt it didn't really describe what I was doing with the PR, since I barely even touch TypicalPersons.

Oh, but the reasoning behind the PR title is as stated here:
deepinscreenshot_select-area_20180313161517

So it doesn't mean that `TypicalPersons` should have a major change right? In this case, the PR title should be changed to the issue title? What are your thoughts @yamidark?

yamidark:
>In this case, the PR title should be changed to the issue title?

I think we should keep with the format, the issue title as the PR's title since there's a rationale behind it (easy tracing). We also can't really tell from the current PR title what we are fixing (just what we are doing to fix it).

Rinder5:
Ok then! Thanks for the input, I'll change the title.

src/test/java/seedu/address/testutil/TypicalPersons.java
```diff
@@ -35,7 +35,7 @@
     public static final Person CARL = new PersonBuilder().withName("Carl Kurz").withPhone("95352563")
             .withEmail("heinz@example.com").withAddress("wall street").build();
     public static final Person DANIEL = new PersonBuilder().withName("Daniel Meier").withPhone("87652533")
-            .withEmail("cornelia@example.com").withAddress("10th street").build();
+            .withEmail("cornelia@example.com").withAddress("10th street").withTags("friends").build();
```
Zhiyuan-Amos: This change isn't clearly explained in the commit message; why does `DANIEL` require a change here but not the other person constants?

src/test/java/seedu/address/logic/parser/AddressBookParserTest.java
```diff
@@ -63,7 +64,7 @@ public void parseCommand_delete() throws Exception {
 
     @Test
     public void parseCommand_edit() throws Exception {
-        Person person = new PersonBuilder().build();
+        Person person = TypicalPersons.ALICE;
```
Zhiyuan-Amos: The reason why this has to change is unclear.

Zhiyuan-Amos:
I'm thinking that EditPersonDescriptorBuilder shouldn't take in a Person for its constructor because it can be quite misleading. @vivekscl @yamidark what do you think?

Zhiyuan-Amos:
Also, I'm of the opinion that we shouldn't use PersonUtil#getPersonDetails(Person) for EditCommand. We should create a new method PersonUtil#getPersonDetails(EditPersonDescriptor).

vivekscl:
>I'm thinking that EditPersonDescriptorBuilder shouldn't take in a Person for its constructor because it can be quite misleading as you aren't sure which fields you actually are editing

Sorry for the late reply @Zhiyuan-Amos! Regarding this, it can be misleading sometimes. However, in some cases, this would prove quite useful right? Like in the case where we are building a duplicate person. In cases where it can be misleading, we could change those tests such that they only use the empty constructor and then the other builder methods like `withName` right?

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +36,29 @@ public static String getPersonDetails(Person person) {
         );
         return sb.toString();
     }
+
```
vivekscl: There should be header comments for this method.

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +36,29 @@ public static String getPersonDetails(Person person) {
...
+            sb.append(PREFIX_EMAIL + descriptor.getEmail().get().value + " ");
+        }
+        if (descriptor.getAddress().isPresent()) {
+            sb.append(PREFIX_ADDRESS + descriptor.getAddress().get().value + " ");
+        }
+        if (descriptor.getTags().isPresent()) {
+            if (descriptor.getTags().get().isEmpty()) {
+                sb.append(PREFIX_TAG);
+            }
+            descriptor.getTags().get().stream().forEach(
```
vivekscl: You don't need a stream for this right? You could just use the `forEach` without the stream.

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +36,29 @@ public static String getPersonDetails(Person person) {
...
+                s -> sb.append(PREFIX_TAG + s.tagName + " ")
```
vivekscl: I think you should chain the `append` function rather than using a `+` to concatenate. Combined with the above comment to not use streams, the whole thing might look something like this:
`descriptor.getTags().get().forEach(s -> sb.append(PREFIX_TAG).append(s.tagName).append(" "));`

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +36,29 @@ public static String getPersonDetails(Person person) {
         );
         return sb.toString();
     }
+
+    public static String getPersonDetails(EditCommand.EditPersonDescriptor descriptor) {
```
vivekscl: Hmm, the argument type looks a bit off. Maybe you should import `EditCommand.EditPersonDescriptor` then just use `EditPersonDescriptor` alone? Like so:
`import seedu.address.logic.commands.EditCommand.EditPersonDescriptor;`

Zhiyuan-Amos:
> Regarding this, it can be misleading sometimes. However, in some cases, this would prove quite useful right? Like in the case where we are building a duplicate person. In cases where it can be misleading, we could change those tests such that they only use the empty constructor and then the other builder methods like withName right?

@vivekscl Yup you are right, thanks :)

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +36,29 @@ public static String getPersonDetails(Person person) {
...
+            sb.append(PREFIX_PHONE + descriptor.getPhone().get().value + " ");
+        }
+        if (descriptor.getEmail().isPresent()) {
+            sb.append(PREFIX_EMAIL + descriptor.getEmail().get().value + " ");
+        }
+        if (descriptor.getAddress().isPresent()) {
+            sb.append(PREFIX_ADDRESS + descriptor.getAddress().get().value + " ");
+        }
+        if (descriptor.getTags().isPresent()) {
+            if (descriptor.getTags().get().isEmpty()) {
```
Zhiyuan-Amos: We can extract out `descriptor.getTags().get()` as a variable.

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +39,33 @@ public static String getPersonDetails(Person person) {
...
         return sb.toString();
     }
+
+    /**
+     * Returns the part of command string for the given {@code EditPersonDescriptor}'s details.
+     */
+    public static String getPersonDetails(EditPersonDescriptor descriptor) {
+        StringBuilder sb = new StringBuilder();
+        descriptor.getName().ifPresent(
+            name -> sb.append(PREFIX_NAME).append(name.fullName).append(" ")
```
Zhiyuan-Amos: This fits in a single line doesn't it :P

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +39,33 @@ public static String getPersonDetails(Person person) {
...
+        );
+        descriptor.getAddress().ifPresent(
+            address -> sb.append(PREFIX_ADDRESS).append(address.value).append(" ")
+        );
+        if (descriptor.getTags().isPresent()) {
+            Set<Tag> tags = descriptor.getTags().get();
+            if (tags.isEmpty()) {
+                sb.append(PREFIX_TAG);
+            }
+            tags.forEach(
```
Zhiyuan-Amos: we should do an `if-else` to make things clearer.

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +39,26 @@ public static String getPersonDetails(Person person) {
...
+    /**
+     * Returns the part of command string for the given {@code EditPersonDescriptor}'s details.
+     */
+    public static String getPersonDetails(EditPersonDescriptor descriptor) {
+        StringBuilder sb = new StringBuilder();
+        descriptor.getName().ifPresent(name -> sb.append(PREFIX_NAME).append(name.fullName).append(" "));
+        descriptor.getPhone().ifPresent(phone -> sb.append(PREFIX_PHONE).append(phone.value).append(" "));
+        descriptor.getEmail().ifPresent(email -> sb.append(PREFIX_EMAIL).append(email.value).append(" "));
+        descriptor.getAddress().ifPresent(
+            address -> sb.append(PREFIX_ADDRESS).append(address.value).append(" ")
```
Zhiyuan-Amos: Why did you not edit this :(

Rinder5: oops >.>

yamidark:
[1/2] Could add in commit message that:
AddressBookParserTest#parseCommand_edit() **may** fail if we want to test for case 2 but PersonUtil#getPersonDetails(Person) will only give us attributes for case 1 instead.

Also, commit title should not start with capital after **:**
PersonUtil: **a**dd method...

src/test/java/seedu/address/testutil/PersonUtil.java
```diff
@@ -35,4 +39,24 @@ public static String getPersonDetails(Person person) {
         );
         return sb.toString();
     }
+
+    /**
+     * Returns the part of command string for the given {@code EditPersonDescriptor}'s details.
+     */
+    public static String getPersonDetails(EditPersonDescriptor descriptor) {
```
yamgent: Is it still appropriate to call this method `getPersonDetails`, since the details here in this case refers to the changes we are making to a person, rather than the an actual person itself (as with the case with `getPersonDetails(Person)`)?

Zhiyuan-Amos:
@damithc gentle reminder for your review prof.

src/test/java/seedu/address/testutil/PersonUtil.java
Zhiyuan-Amos: @Rinder5 do rmb to make this change :)

Zhiyuan-Amos:
@yamgent ok with you? :)

Zhiyuan-Amos:
> Commit title:
More than 72 chars :O.
Maybe we can rename this to:
PersonUtil: add method getEditPersonDescriptorDetails?

Doesn't this look better: `PersonUtil: add getEditPersonDescriptorDetails(EditPersonDescriptor)` :p there's no need to specify `method` if you're specifying the parameter. Also, it's our convention to specify the parameter when we are talking about a method :p

PR #850, opened 2018-03-15
HelpWindow: multiple help windows can be opened #819

### LINKED ISSUE #819: HelpWindow: multiple help windows can be opened
We should disallow multiple help windows to be opened.

To reproduce:
1. Click on `Help` button.
2. Execute "help" in the command box.

yamidark:
Fixes #819.
```
MainWindow creates a new HelpWindow instance and shows it whenever
MainWindow#handleHelp() is called.

This allows multiple help windows to be opened at the same time by
executing HelpCommand, pressing Help menu button or F1 multiple times.

Multiple help windows should be disallowed since each help window
are exactly the same thus having extra copies is unnecessary.

Let's update MainWindow to hold and show a single HelpWindow instance.
```

### DIFF
--- a/src/main/java/seedu/address/ui/HelpWindow.java
+++ b/src/main/java/seedu/address/ui/HelpWindow.java
@@ -61,4 +61,18 @@ public class HelpWindow extends UiPart<Stage> {
         logger.fine("Showing help page about the application.");
         getRoot().show();
     }
+
+    /**
+     * Returns true if the help window is currently being shown.
+     */
+    public boolean isShowing() {
+        return getRoot().isShowing();
+    }
+
+    /**
+     * Focuses on the help window.
+     */
+    public void focus() {
+        getRoot().requestFocus();
+    }
 }
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -38,6 +38,7 @@ public class MainWindow extends UiPart<Stage> {
     private PersonListPanel personListPanel;
     private Config config;
     private UserPrefs prefs;
+    private HelpWindow helpWindow;
 
     @FXML
     private StackPane browserPlaceholder;
@@ -72,6 +73,8 @@ public class MainWindow extends UiPart<Stage> {
 
         setAccelerators();
         registerAsAnEventHandler(this);
+
+        helpWindow = new HelpWindow();
     }
 
     public Stage getPrimaryStage() {
@@ -165,7 +168,6 @@ public class MainWindow extends UiPart<Stage> {
      */
     @FXML
     public void handleHelp() {
-        HelpWindow helpWindow = new HelpWindow();
         helpWindow.show();
     }
 
--- a/src/main/java/seedu/address/ui/MainWindow.java
+++ b/src/main/java/seedu/address/ui/MainWindow.java
@@ -164,11 +164,15 @@ public class MainWindow extends UiPart<Stage> {
     }
 
     /**
-     * Opens the help window.
+     * Opens the help window or focuses on it if it's already opened.
      */
     @FXML
     public void handleHelp() {
-        helpWindow.show();
+        if (!helpWindow.isShowing()) {
+            helpWindow.show();
+        } else {
+            helpWindow.focus();
+        }
     }
 
     void show() {
--- a/src/test/java/guitests/GuiRobot.java
+++ b/src/test/java/guitests/GuiRobot.java
@@ -75,9 +75,16 @@ public class GuiRobot extends FxRobot {
      * Returns true if the window with {@code stageTitle} is currently open.
      */
     public boolean isWindowShown(String stageTitle) {
-        return listTargetWindows().stream()
+        return getNumberOfWindowsShown(stageTitle) >= 1;
+    }
+
+    /**
+     * Returns the number of windows with {@code stageTitle} that are currently open.
+     */
+    public int getNumberOfWindowsShown(String stageTitle) {
+        return (int) listTargetWindows().stream()
                 .filter(window -> window instanceof Stage && ((Stage) window).getTitle().equals(stageTitle))
-                .count() >= 1;
+                .count();
     }
 
     /**
--- a/src/test/java/guitests/guihandles/StageHandle.java
+++ b/src/test/java/guitests/guihandles/StageHandle.java
@@ -44,6 +44,13 @@ public abstract class StageHandle {
     }
 
     /**
+     * Returns true if currently focusing on this stage.
+     */
+    public boolean isFocused() {
+        return stage.isFocused();
+    }
+
+    /**
      * Retrieves the {@code query} node owned by the {@code stage}.
      *
      * @param query name of the CSS selector for the node to retrieve.
--- a/src/test/java/seedu/address/ui/HelpWindowTest.java
+++ b/src/test/java/seedu/address/ui/HelpWindowTest.java
@@ -1,6 +1,8 @@
 package seedu.address.ui;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static seedu.address.ui.HelpWindow.USERGUIDE_FILE_PATH;
 
 import java.net.URL;
@@ -29,5 +31,26 @@ public class HelpWindowTest extends GuiUnitTest {
     public void display() {
         URL expectedHelpPage = HelpWindow.class.getResource(USERGUIDE_FILE_PATH);
         assertEquals(expectedHelpPage, helpWindowHandle.getLoadedUrl());
+    }
+
+    @Test
+    public void isShowing_helpWindowIsShowing_returnsTrue() {
+        guiRobot.interact(helpWindow::show);
+        assertTrue(helpWindow.isShowing());
+    }
+
+    @Test
+    public void isShowing_helpWindowIsHiding_returnsFalse() {
+        guiRobot.interact(helpWindow.getRoot()::hide);
+        assertFalse(helpWindow.isShowing());
+    }
+
+    @Test
+    public void focus_helpWindowNotFocused_focused() {
+        guiRobot.interact(helpWindow::show);
+        assertFalse(helpWindow.getRoot().isFocused());
+
+        guiRobot.interact(helpWindow::focus);
+        assertTrue(helpWindow.getRoot().isFocused());
     }
 }
--- a/src/test/java/systemtests/HelpCommandSystemTest.java
+++ b/src/test/java/systemtests/HelpCommandSystemTest.java
@@ -73,6 +73,19 @@ public class HelpCommandSystemTest extends AddressBookSystemTest {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void help_multipleCommands_onlyOneHelpWindowOpen() {
+        getMainMenu().openHelpWindowUsingMenu();
+
+        getMainWindowHandle().focus();
+        getMainMenu().openHelpWindowUsingAccelerator();
+
+        getMainWindowHandle().focus();
+        executeCommand(HelpCommand.COMMAND_WORD);
+
+        assertEquals(1, guiRobot.getNumberOfWindowsShown(HelpWindowHandle.HELP_WINDOW_TITLE));
+    }
+
     /**
      * Asserts that the help window is open, and closes it after checking.
      */


src/main/java/seedu/address/ui/MainWindow.java
```diff
@@ -164,11 +164,15 @@ GuiSettings getCurrentGuiSetting() {
     }
 
     /**
-     * Opens the help window.
+     * Opens the help window or focus on it if already opened.
```
Zhiyuan-Amos: focus**es** on it if **it's** already opened.

src/main/java/seedu/address/ui/HelpWindow.java
```diff
@@ -61,4 +61,18 @@ public void show() {
...
+
+    /**
+     * Returns true if the help window is currently being shown.
+     */
+    public boolean isShowing() {
+        return getRoot().isShowing();
+    }
+
+    /**
+     * Focus on the help window.
```
Zhiyuan-Amos: Focus**es**

yamidark:
>Can we do tests for this?

Will look into adding these tests in.

src/test/java/guitests/GuiRobot.java
```diff
@@ -80,6 +80,13 @@ public boolean isWindowShown(String stageTitle) {
                 .count() >= 1;
     }
 
+    /**
+     * Returns the number of windows with {@code stageTitle} currently open.
+     */
+    public int getNumberOfWindowShown(String stageTitle) {
```
Zhiyuan-Amos: Window**s**

src/test/java/guitests/GuiRobot.java
```diff
@@ -80,6 +80,13 @@ public boolean isWindowShown(String stageTitle) {
                 .count() >= 1;
     }
 
+    /**
+     * Returns the number of windows with {@code stageTitle} currently open.
```
Zhiyuan-Amos: ... **that are** currently open.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -55,6 +56,14 @@ public void openHelpWindow() {
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
 
+        // check that only 1 help window will be opened even with multiple commands
```
Zhiyuan-Amos: asserts that executing multiple commands to open help window will result in only 1 help window opened.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -55,6 +56,14 @@ public void openHelpWindow() {
...
+        getMainMenu().openHelpWindowUsingMenu();
+        getMainWindowHandle().focus();
```
Zhiyuan-Amos: we should comment here on why there's a need to call `focus()`.

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +40,14 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Remove focus from the {@code Stage}.
```
Zhiyuan-Amos: Removes focus from the current stage that is in focus,

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,23 @@ public void display() {
         URL expectedHelpPage = HelpWindow.class.getResource(USERGUIDE_FILE_PATH);
         assertEquals(expectedHelpPage, helpWindowHandle.getLoadedUrl());
     }
+
+    @Test
+    public void isShowing() {
```
Zhiyuan-Amos: can we try to follow the 3-part structure for test naming? I can't quite think of an idea on how to name it though.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -82,6 +86,12 @@ public void openHelpWindow() {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void helpWindowIsShowing() {
+        getMainMenu().openHelpWindowUsingAccelerator();
+
```
Zhiyuan-Amos: extra new line

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -82,6 +86,12 @@ public void openHelpWindow() {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void helpWindowIsShowing() {
```
Zhiyuan-Amos: what is this test for?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -56,12 +56,16 @@ public void openHelpWindow() {
         executeCommand(HelpCommand.COMMAND_WORD);
         assertHelpWindowOpen();
 
-        // check that only 1 help window will be opened even with multiple commands
+        // check that if help window is open when command is given, we will focus on the opened help window
```
Zhiyuan-Amos: we shouldn't mix these tests together. Also, you didn't test for the case of `not focused -> handleHelp() -> focused`.

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: Also I think we should separate this into a different test, because we aren't specifically testing for the different ways of opening up help windows.

src/test/java/systemtests/HelpCommandSystemTest.java
yamidark: Test for `HelpWindow#isShowing()`?

Zhiyuan-Amos: But there's no assertion below to verify that it's showing isn't it :P

yamidark: >But there's no assertion below to verify that it's showing isn't it :P

Whoops, I think this test was brought over the `HelpWindowTest` but I forgot to remove it here.

src/test/java/guitests/GuiRobot.java
```diff
@@ -80,6 +80,13 @@ public boolean isWindowShown(String stageTitle) {
                 .count() >= 1;
     }
 
+    /**
+     * Returns the number of windows with {@code stageTitle} that are currently open.
+     */
+    public int getNumberOfWindowsShown(String stageTitle) {
+        return (int) listTargetWindows().stream()
+                .filter(window -> window instanceof Stage && ((Stage) window).getTitle().equals(stageTitle)).count();
```
vivekscl: I think the `.count();` should be at the bottom. Like each stream operation show start on a new line right?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -73,14 +74,27 @@ public void openHelpWindow() {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void help_multipleCommands_oneHelpWindowOpen() {
+        // execute multiple commands to open help window
+        getMainMenu().openHelpWindowUsingMenu();
+        // focus on main window so that command to open help window will trigger
```
vivekscl: Hmm, do we need this comment? :thinking:

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -73,14 +74,27 @@ public void openHelpWindow() {
...
+        getMainWindowHandle().focus();
+        getMainMenu().openHelpWindowUsingAccelerator();
+        getMainWindowHandle().focus();
+        executeCommand(HelpCommand.COMMAND_WORD);
+        // assert that only 1 help window is open even with multiple commands to open the help window
```
vivekscl: I think this should be:
assert that only 1 help window is open **after using** multiple commands to open the help window

src/test/java/guitests/GuiRobot.java
```diff
@@ -80,6 +80,13 @@ public boolean isWindowShown(String stageTitle) {
                 .count() >= 1;
```
vivekscl: On a side note, now that we have `GuiRobot#getNumberOfWindowsShown(String)` we can use this method in `GuiRobot#isWindowShown(String)` right? Like, `getgetNumberOfWindowsShown(stageTitle) >= 1`.

src/test/java/guitests/guihandles/StageHandle.java
```diff
@@ -43,6 +43,13 @@ public void focus() {
         logger.info("Finishing focus on" + windowTitle);
     }
 
+    /**
+     * Returns true if we are currently focusing on the this stage.
+     */
+    public boolean isFocused() {
```
vivekscl: This method is never used right? When I checked the usages of this method in other places it says that the method comes from `Window#isFocused()`.

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +40,14 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Removes focus from the current stage that is in focus.
```
vivekscl: I think `Removes focus from the current stage.` suffices since it is implied that the current stage would be in focus.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
         assertEquals(1, guiRobot.getNumberOfWindowsShown(HELP_WINDOW_TITLE));
     }
 
+    @Test
+    public void help_helpWindowNotFocused_helpWindowFocused() {
```
vivekscl: Maybe a better name for this method would be `help_helpWindowOutOfFocus_focusOnHelpWindow`?

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
...
+        getMainMenu().openHelpWindowUsingAccelerator();
+        Stage helpWindowStage = guiRobot.getStage(HELP_WINDOW_TITLE);
+
+        // assert command to open help window focuses on the open help window
```
vivekscl: assert **that** the command  to open help window focuses on the **currently opened** help window

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
         URL expectedHelpPage = HelpWindow.class.getResource(USERGUIDE_FILE_PATH);
         assertEquals(expectedHelpPage, helpWindowHandle.getLoadedUrl());
     }
+
+    @Test
+    public void isShowing_helpWindowIsShowing_true() {
+        guiRobot.interact(() -> helpWindow.show());
```
vivekscl: You could use method references here right? like, `guiRobot.interact(helpWindow::show)`

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
...
+        assertTrue(helpWindow.isShowing());
+    }
+
+    @Test
+    public void isShowing_helpWindowIsHiding_false() {
+        guiRobot.interact(() -> helpWindow.getRoot().hide());
```
vivekscl: Could use method references here as well: `guiRobot.interact(helpWindowgetRoot()::hide)`

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
...
+        assertFalse(helpWindow.isShowing());
+    }
+
+    @Test
+    public void focus_helpWindowNotFocused_focused() {
+        guiRobot.interact(() -> helpWindow.show());
```
vivekscl: Same here. `guiRobot.interact(helpWindow::show)`

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
...
+        guiRobot.removeFocus();
+        assertFalse(helpWindow.getRoot().isFocused());
+
+        guiRobot.interact(() -> helpWindow.focus());
```
vivekscl: Same here. `guiRobot.interact(helpWindow::focus)`

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: Initially, I told @yamidark to write this comment because I thought it would be unclear to people who aren't familiar with ui tests. 

However, yes I agree that this comment doesn't look too nice. Let's leave a new line instead of writing comments to show the demarcation between each opening of HelpWindow.

```java
getMainMenu().openHelpWindowUsingMenu();

getMainWindowHandle().focus();
getMainMenu().openHelpWindowUsingAccelerator();

getMainWindowHandle().focus();
executeCommand(HelpCommand.COMMAND_WORD);

assertEquals(1, guiRobot.getNumberOfWindowsShown(HELP_WINDOW_TITLE));
```

src/test/java/systemtests/HelpCommandSystemTest.java
Zhiyuan-Amos: There's no need for this comment; this is intuitive following the test method name and the assert method.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -73,14 +74,27 @@ public void openHelpWindow() {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void help_multipleCommands_oneHelpWindowOpen() {
+        // execute multiple commands to open help window
```
Zhiyuan-Amos: This comment is unnecessary since the test method name is rather descriptive.

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +40,14 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Removes focus from the current stage.
+     */
+    public void removeFocus() {
+        press(KeyCode.ALT, KeyCode.TAB);
+        release(KeyCode.ALT, KeyCode.TAB);
```
Zhiyuan-Amos: Does this work? Cos `Alt-Tab` will be pressed until `release(...)` is called. Therefore, there may be some scenarios when the `Alt-Tab` causes it to revert back to the current stage.

We can take a look at this [SO solution](https://stackoverflow.com/a/25467653/8828382).

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
         URL expectedHelpPage = HelpWindow.class.getResource(USERGUIDE_FILE_PATH);
         assertEquals(expectedHelpPage, helpWindowHandle.getLoadedUrl());
     }
+
+    @Test
+    public void isShowing_helpWindowIsShowing_true() {
```
Zhiyuan-Amos: `_returnsTrue`

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -30,4 +32,26 @@ public void display() {
...
+        guiRobot.interact(helpWindow::show);
+        assertTrue(helpWindow.isShowing());
+    }
+
+    @Test
+    public void isShowing_helpWindowIsHiding_false() {
```
Zhiyuan-Amos: `_returnsFalse`

src/test/java/guitests/GuiRobot.java
```diff
@@ -75,11 +75,17 @@ public void waitForEvent(BooleanSupplier event, int timeOut) {
...
     }
 
+    /**
+     * Returns the number of windows with {@code stageTitle} that are currently open.
+     */
+    public int getNumberOfWindowsShown(String stageTitle) {
+        return (int) listTargetWindows().stream()
+                .filter(window -> window instanceof Stage && ((Stage) window).getTitle().equals(stageTitle))
+                .count();
+    }
```
Zhiyuan-Amos: missing new line

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -1,5 +1,6 @@
 package systemtests;
 
+import static guitests.guihandles.HelpWindowHandle.HELP_WINDOW_TITLE;
```
vivekscl: There is no need to import this. Since it's a class level constant, you can just call it using the class name. Refer to this [SO](https://stackoverflow.com/questions/824011/a-simple-way-to-create-static-imports-for-constants-in-existing-code) post for more information on why it's better not to use static imports.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
         assertEquals(1, guiRobot.getNumberOfWindowsShown(HELP_WINDOW_TITLE));
     }
 
+    @Test
+    public void help_helpWindowOutOfFocus_focusOnHelpWindow() {
+        getMainMenu().openHelpWindowUsingAccelerator();
+        Stage helpWindowStage = guiRobot.getStage(HELP_WINDOW_TITLE);
```
Zhiyuan-Amos: You should not access the `Stage` directly, but access them through `Handle`s :P Which is why you need this [method](https://github.com/se-edu/addressbook-level4/compare/c94261e57a5ee4a79f983cd5c601f588d7e87d8c~1...c94261e57a5ee4a79f983cd5c601f588d7e87d8c#diff-5600101e1e39f84f3280c83b9f3d20edL49). But since the method is only used in `HelpWindowHandle`, maybe we should put there instead of `StageHandle`.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -73,14 +74,27 @@ public void openHelpWindow() {
...
+    }
+
     /**
      * Asserts that the help window is open, and closes it after checking.
      */
     private void assertHelpWindowOpen() {
         assertTrue(ERROR_MESSAGE, HelpWindowHandle.isWindowPresent());
         guiRobot.pauseForHuman();
 
-        new HelpWindowHandle(guiRobot.getStage(HelpWindowHandle.HELP_WINDOW_TITLE)).close();
```
vivekscl: You can bring back this line after removing the static import as mentioned previously.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
...
+        Stage helpWindowStage = guiRobot.getStage(HELP_WINDOW_TITLE);
+
+        // assert command to open help window focuses on the open help window
+        getMainWindowHandle().focus();
+        getMainMenu().openHelpWindowUsingAccelerator();
+        assertTrue(helpWindowStage.isFocused());
+
+        getMainWindowHandle().focus();
+        executeCommand(HelpCommand.COMMAND_WORD);
+        assertTrue(helpWindowStage.isFocused());
```
Zhiyuan-Amos: I wonder whether there's a need for lines 101-103.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
         assertEquals(1, guiRobot.getNumberOfWindowsShown(HELP_WINDOW_TITLE));
     }
 
+    @Test
+    public void help_helpWindowOutOfFocus_focusOnHelpWindow() {
+        getMainMenu().openHelpWindowUsingAccelerator();
+        Stage helpWindowStage = guiRobot.getStage(HELP_WINDOW_TITLE);
+
+        // assert command to open help window focuses on the open help window
```
Zhiyuan-Amos: I think this comment is unnecessary because the test method name is self-explanatory.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -87,6 +88,21 @@ public void help_multipleCommands_oneHelpWindowOpen() {
...
+        getMainWindowHandle().focus();
+        getMainMenu().openHelpWindowUsingAccelerator();
+        assertTrue(helpWindowStage.isFocused());
```
Zhiyuan-Amos: after line 97 we should `assertFalse(helpWindowStage.isFocused());`

yamidark: >after line 97 we should assertFalse(helpWindowStage.isFocused());

For some reason, when the test is done in `headless` mode, this `assertFalse` line would always fail, which is why I removed it.

Zhiyuan-Amos: I tried, it seems that when the test is executed on `headless` mode and `getMainWindowHandle().focus()` is called, weirdly both `MainWindow` and `HelpWindow` are in focus.

Zhiyuan-Amos: As such, this actually means that the test isn't executing as per intended, since `MainWindow` was already in focus from the start.

yamidark: Tried using some other `JavaFx` object stages, but couldn't really get them to work, so I used the help windows to try this test:

```Java
    private HelpWindow hw1;
    private HelpWindow hw2;

    @Test
    public void test() throws Exception {
        guiRobot.interact(() -> hw1 = new HelpWindow());
        guiRobot.interact(() -> hw2 = new HelpWindow());

        FxToolkit.setupStage((stage) -> stage.setScene(hw1.getRoot().getScene()));
        FxToolkit.setupStage((stage) -> stage.setScene(hw2.getRoot().getScene()));
        FxToolkit.showStage();

        guiRobot.interact(hw1::show);
        guiRobot.interact(hw2::show);

        assertFalse(hw1.getRoot().isFocused());
        assertTrue(hw2.getRoot().isFocused());
        guiRobot.interact(hw1::focus);
        assertTrue(hw1.getRoot().isFocused());
        assertFalse(hw2.getRoot().isFocused());
    }
```

Test passes normally when run in `Intellji`, but when run on `gradlew` `headless` mode test fails as the 2 help window aren't in focus even after showing.
We can also focus on both of them in `headless` mode too when only 1 of them should be in focus at a time.

yamidark: Should we leave this test like this for now, and add comments that in `headless` mode any of the windows that were once in focus will never leave focus?

Zhiyuan-Amos: Good job investigating 👍 Okay, the issue now is: We are updating the main code base but we aren't able to write tests to verify the updated behavior.

As such, we should remove this test because this test doesn't do anything. `getMainWindowHandle().focus()` is supposed to remove the focus from `HelpWindow`, then we verify that executing help command will cause the `HelpWindow` to go back in focus. But now, `HelpWindow` doesn't go out of focus, so we don't get to test the functionality that executing help command will cause the `HelpWindow` to go back in focus.

However, prior to removing this test, I think it's good that we make a sample class so that we can raise an issue to `TestFX` for verification. This class shouldn't be tied to the rest of our code base; you should create a new project e.g.:

```java
public class Foo {
    public static void main(String args[]) {
        // open 2 stages
        // perform assertions here. It doesn't matter that this isn't a test class, we can still do assertFalse etc here.
    }
}
```

Once we have raised an issue in TestFX and it is verified, then we can leave a comment in our code base that we aren't writing a test for this method because of the issue in TestFX. :)

yamidark: Ok, managed to reproduce this using 2 `VBox`:

```Java
    Stage stage1;
    Stage stage2;

    @Test
    public void test() throws Exception {
        FxRobot fxRobot = new FxRobot();
        FxToolkit.registerPrimaryStage();

        VBox vBox = new VBox();
        Scene scene = FxToolkit.setupScene(() -> new Scene(vBox,300,250));
        fxRobot.interact(() -> {
            stage1 = new Stage();
            stage1.setScene(scene);
        });

        VBox vBox2 = new VBox();
        Scene scene2 = FxToolkit.setupScene(() -> new Scene(vBox2,300,250));
        fxRobot.interact(() -> {
            stage2 = new Stage();
            stage2.setScene(scene2);
        });

        fxRobot.interact(stage1::show);
        assertTrue(stage1.isFocused());
        assertFalse(stage2.isFocused());

        fxRobot.interact(stage2::show);
        assertFalse(stage1.isFocused());
        assertTrue(stage2.isFocused());
    }
```
Test passes when run normally, but fails on `headless` since `assertFalse(stage1.isFocused())` fails.
So I should post this issue on TestFX?

Zhiyuan-Amos: Great work. Just wondering whether there's a need for the VBox to be present I.e. Can we simplify our code further? Otherwise, I think we are good to go. :+1

yamidark: Tried doing this with just 2 `stage` alone, but the test doesn't work, can't seem to even focus on them, and the test hangs in `gradle` `headless` mode. So, I believe there's a need for the VBox here.

Zhiyuan-Amos: Ok. Yup, do post this issue on TestFX, include the whole class (not just the test method) ya :)

yamidark:
Issue posted on [TestFX](https://github.com/TestFX/TestFX/issues/548).

Zhiyuan-Amos:
This [test](https://github.com/se-edu/addressbook-level4/compare/1f16b1c764d337e10402329e5cd6ad40345ebfc3~1...1f16b1c764d337e10402329e5cd6ad40345ebfc3#diff-3558a92e4dc4346714c9e81d44c8ebfcL49) shouldn't be deleted as we still need to test that `HelpWindow#focus()` works.

src/test/java/systemtests/HelpCommandSystemTest.java
```diff
@@ -73,6 +73,19 @@ public void openHelpWindow() {
         assertNotEquals(StatusBarFooter.SYNC_STATUS_INITIAL, getStatusBarFooter().getSyncStatus());
     }
 
+    @Test
+    public void help_multipleCommands_oneHelpWindowOpen() {
```
Zhiyuan-Amos: maybe we should be clearer and specify **only**OneHelpWindowOpen.

Zhiyuan-Amos:
@yamidark you can continue on this PR since [Monocle](https://github.com/javafxports/openjdk-jfx/issues/50) has responded by placing a bug tag. :)

yamidark:
>Create an issue in our code base to note that we need to write test for this method, but it cannot be written 

Noted. Will do that after this PR is merged.

Zhiyuan-Amos:
@yamidark can do last call for reviews on Gitter? :)

Zhiyuan-Amos:
@yamidark great work! 👍 :)

PR #860, opened 2018-03-25
Tidy up build.gradle #706

### LINKED ISSUE #706: Gradle: update to latest version

Zhiyuan-Amos:
Fixes #706

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -164,7 +164,6 @@ task headless {
         println "Setting headless mode properties."
         test {
             systemProperties = [
-                'java.awt.robot': 'true',
                 'testfx.robot': 'glass',
                 'testfx.headless': 'true',
                 'prism.order': 'sw',
--- a/build.gradle
+++ b/build.gradle
@@ -9,8 +9,8 @@ plugins {
     id 'java'
     id 'jacoco'
     id 'checkstyle'
-    id "com.github.kt3k.coveralls" version "2.4.0"
-    id "com.github.johnrengelman.shadow" version '2.0.3'
+    id 'com.github.kt3k.coveralls' version '2.4.0'
+    id 'com.github.johnrengelman.shadow' version '2.0.3'
     id 'org.asciidoctor.convert' version '1.5.6'
     id 'application'
 }
@@ -23,7 +23,7 @@ targetCompatibility = JavaVersion.VERSION_1_9
 
 repositories {
     mavenCentral()
-    maven { url "https://oss.sonatype.org/content/repositories/snapshots/" }
+    maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
 }
 
 checkstyle {
@@ -70,7 +70,7 @@ dependencies {
 }
 
 shadowJar {
-    archiveName = "addressbook.jar"
+    archiveName = 'addressbook.jar'
 
     destinationDir = file("${buildDir}/jar/")
 }
@@ -161,7 +161,7 @@ test {
 
 task headless {
     doLast {
-        println "Setting headless mode properties."
+        println 'Setting headless mode properties.'
         test {
             systemProperties = [
                 'testfx.robot': 'glass',
--- a/build.gradle
+++ b/build.gradle
@@ -2,7 +2,7 @@
  * Gradle Configuration File
  *
  * For more details take a look at the Java Quickstart chapter in the Gradle
- * user guide available at http://gradle.org/docs/2.2.1/userguide/tutorial_java_projects.html
+ * user guide available at http://gradle.org/docs/4.6/userguide/tutorial_java_projects.html
  */
 
 plugins {
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
     id 'jacoco'
     id 'checkstyle'
     id "com.github.kt3k.coveralls" version "2.4.0"
-    id "com.github.johnrengelman.shadow" version '1.2.3'
+    id "com.github.johnrengelman.shadow" version '2.0.3'
     id 'org.asciidoctor.convert' version '1.5.6'
     id 'application'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -60,9 +60,6 @@ dependencies {
         exclude group: 'org.testfx', module: 'testfx-internal-java8'
     }
     testImplementation group: 'org.testfx', name: 'testfx-junit', version: testFxVersion
-    testImplementation group: 'org.testfx', name: 'testfx-legacy', version: '4.0.8-alpha', {
-        exclude group: 'junit', module: 'junit'
-    }
 
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: jUnitVersion
 
--- a/build.gradle
+++ b/build.gradle
@@ -143,8 +143,8 @@ test {
         boolean runNonGuiTests = gradle.taskGraph.hasTask(nonGuiTests)
 
         if (!runGuiTests && !runNonGuiTests) {
-            runGuiTests = true;
-            runNonGuiTests = true;
+            runGuiTests = true
+            runNonGuiTests = true
         }
 
         if (runNonGuiTests) {
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,6 @@
-/*
- * Gradle Configuration File
- *
- * For more details take a look at the Java Quickstart chapter in the Gradle
- * user guide available at http://gradle.org/docs/4.6/userguide/tutorial_java_projects.html
- */
+// Gradle Configuration File
+// For more details take a look at the Java Quickstart chapter in the Gradle
+// user guide available at http://gradle.org/docs/4.6/userguide/tutorial_java_projects.html
 
 plugins {
     id 'java'
@@ -120,11 +117,9 @@ allTests.dependsOn nonGuiTests
 test {
     systemProperty 'testfx.setup.timeout', '60000'
 
-    /*
-     * Prints the currently running test's name in the CI's build log,
-     * so that we can check if tests are being silently skipped or
-     * stalling the build.
-     */
+    // Prints the currently running test's name in the CI's build log,
+    // so that we can check if tests are being silently skipped or
+    // stalling the build.
     if (System.env.'CI') {
         beforeTest { descriptor ->
             logger.lifecycle("Running test: ${descriptor}")
@@ -193,11 +188,9 @@ asciidoctor {
     ]
 }
 
-/*
- * Copies stylesheets into the directory containing generated HTML files as
- * Asciidoctor does not copy linked CSS files to the output directory when rendering.
- * This is needed for linked stylesheets and embedded stylesheets which import other files.
- */
+// Copies stylesheets into the directory containing generated HTML files as
+// Asciidoctor does not copy linked CSS files to the output directory when rendering.
+// This is needed for linked stylesheets and embedded stylesheets which import other files.
 task copyStylesheets(type: Copy) {
     from "${asciidoctor.sourceDir}/stylesheets"
     into "${asciidoctor.outputDir}/html5/stylesheets"
--- a/build.gradle
+++ b/build.gradle
@@ -181,14 +181,16 @@ asciidoctor {
     sourceDir 'docs'
     outputDir "${buildDir}/docs"
 
-    attributes linkcss: true,
-               stylesheet: 'gh-pages.css',
-               'source-highlighter': 'coderay',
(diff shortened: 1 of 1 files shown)


Zhiyuan-Amos:
@yamidark @vivekscl wanna have a shot at reviewing this PR? :p

vivekscl:
Just to clarify, IntellJ complains about these as well:
```
task(guiTests)
task(nonGuiTests)
task(allTests)
```
and
```
sourceDirectories = files(allprojects.sourceSets.main.allSource.srcDirs)
classDirectories =  files(allprojects.sourceSets.main.output)
executionData = files(allprojects.jacocoTestReport.executionData)
```

Do these need to be cleaned up as well?

Zhiyuan-Amos:
> Is the `` around descriptor needed? The formatting is not reflected in the commit message right?

@vivekscl I used the `` around descriptor to denote that it's code, as 'descriptor' can seem to be like an English word (it isn't clear that I'm using 'descriptor' as code or as an English word). What do you think? :)

> Just to clarify, IntellJ complains about these as well:

> task(guiTests)
task(nonGuiTests)
task(allTests)

> and

> sourceDirectories = files(allprojects.sourceSets.main.allSource.srcDirs)
classDirectories =  files(allprojects.sourceSets.main.output)
executionData = files(allprojects.jacocoTestReport.executionData)
Do these need to be cleaned up as well?

Yes, I think they should be cleaned up. However, I don't know how to do so :p 

The issue with `allprojects.sourceSets.main.allSource.srcDirs` is that Gradle can't infer the argument i.e. it doesn't know what's the type of `allprojects.sourceSets.main.allSource.srcDirs`. Since AB-Level4 isn't a multiple project, so I think there's no need for `allprojects`. So, we can actually remove that. But then again I'm not too sure since coveralls use that too (see line 92). I guess I should try uh?

I have spent some time trying to figure out the `task` thing but I don't know how to fix it... ><

Thanks for your reviews @vivekscl & @yamidark!

vivekscl:
>I have spent some time trying to figure out the task thing but I don't know how to fix it. Edit: I followed the Gradle user guide and added task myTask, but that itself receives a warning from IntelliJ as well. As such, I think that IntelliJ is giving a false warning?

I did some searching and it does seem like it might be a bug with IntelliJ.

build.gradle
```diff
@@ -172,14 +172,14 @@ asciidoctor {
...
+    attributes = [linkcss: true,
                stylesheet: 'gh-pages.css',
                'source-highlighter': 'coderay',
                icons: 'font',
                experimental: true,
                sectlinks: true,
                idprefix: '',  // for compatibility with GitHub preview
-               idseparator: '-'
+               idseparator: '-']
 }
```
vivekscl: Just wondering, should these lines be indented to be in line with `linkcss`?

vivekscl:
Everything else LGTM :+1: .

build.gradle
Zhiyuan-Amos: Good catch!

Zhiyuan-Amos:
Had to close and re-open the PR to trigger the Travis-CI build. Weird.

Zhiyuan-Amos:
@vivekscl @yamidark I updated IntelliJ to a newer version, and it prompted me another warning in `build.gradle`...... As such, I updated the 6th commit :p sorry and thanks for the review guys!

Zhiyuan-Amos:
> The link on the commit message isn't working because of the "at the back. Could add a whitespace before the at the back to point to the correct link.

Actually I think it's fine, because I'm not trying to direct readers to click on the link (unlike the cases where we have footnotes to direct readers to click on the link for more information). I'm just trying to denote the code as it is. @vivekscl what do you think? :)

pyokagan:
[4/8]

> The wrong methods are called as these abbreviations point to getter
> methods Report#isEnabled() and MavenArtifactRepository#getUrl().

This seems strange, `isEnabled()` and `getUrl()` don't take any arguments, so we should be getting at least an error at runtime. Furthermore, many of the code examples in the Internet don't use the `=`.

A little digging online brings up http://mrhaki.blogspot.sg/2016/02/gradle-goodness-methods-generated.html , which states that Gradle generates those additional methods and proxies them to their respective setters.

vivekscl:
>Actually I think it's fine, because I'm not trying to direct readers to click on the link (unlike the cases where we have footnotes to direct readers to click on the link for more information). I'm just trying to denote the code as it is. 

Yup, I agree. Since the intent was to denote the code, I think it's fine.

Zhiyuan-Amos:
@pyokagan thanks for your comments! :)

build.gradle
```diff
@@ -156,11 +156,11 @@ test {
...
-        systemProperty 'java.awt.robot', 'true'
-        systemProperty 'testfx.robot', 'glass'
-        systemProperty 'testfx.headless', 'true'
-        systemProperty 'prism.order', 'sw'
-        systemProperty 'prism.text', 't2k'
+        systemProperties = ['java.awt.robot': 'true',
+                            'testfx.robot': 'glass',
+                            'testfx.headless': 'true',
+                            'prism.order': 'sw',
+                            'prism.text': 't2k']
```
pyokagan: Maybe we could format it like this?
```groovy
systemProperties = [
    'java.awt.robot': 'true',
    'testfx.robot': 'glass',
    'testfx.headless': 'true',
    'prism.order': 'sw',
    'prism.text': 't2k',
]
```

Reasons:

* No custom indentation to be consistent with our Java code style
* Trailing comma on the last item and closing bracket on a separate line so we have a cleaner diff when adding/removing items at the end.

Same for the `attributes` below.

Zhiyuan-Amos: Looks better, thanks Paul!

Zhiyuan-Amos:
Added new commits to replace deprecated stuffs.

Zhiyuan-Amos:
Remove unused stuffs.

build.gradle
```diff
@@ -121,7 +121,7 @@ test {
      */
     if (System.env.'CI') {
         beforeTest { descriptor ->
-            logger.lifecycle('Running test: ' + descriptor)
+            logger.lifecycle('Running test: ' + descriptor.toString())
```
pyokagan: Actually, should this line be using string interpolation instead? i.e. `"Running test: ${descriptor}"`

Zhiyuan-Amos:
> How do you know its unused? I see references to AWT in the JavaFX code https://github.com/javafxports/openjdk-jfx/search?q=awt

I thought that it's unused in the sense that the tests still pass now, and also because we aren't likely to use classes that depend on `java.awt.robot`. Correct me if I'm not wrong, the reason why we have to set `java.awt.robot` to be `true` is because some classes in `openjdk` uses it? Sorry for the newb question, still not too familiar with Gradle. I tried googling up on this, and the closest that I can find it [this](https://stackoverflow.com/questions/27403410/headless-testing-with-javafx-and-testfx).

pyokagan:
@Zhiyuan-Amos 
> I thought that it's unused in the sense that the tests still pass now, and also because we aren't likely to use classes that depend on java.awt.robot.

"All tests still pass" isn't a sufficient enough reason for anything, though, since tests don't catch all bugs. There needs to be a more detailed explanation rather than saying it is "unused".

Note that this situation is very different from the case of "a class/method in our code base is unused", since we can easily check that in the IDE. System properties can affect the way our dependencies behave, so we need to be more careful with it.

Also, you need to consider the possibility that the author put it there for a reason.

---

So, let's do some investigation:

The original commit that introduced it is not in our repo, but in the old hubturbo one. Specially, https://github.com/HubTurbo/addressbook/commit/2eef6947984d0991a5e48d2a39759d34fa85b229 , which shows that it was indeed added specifically to fix uh... something. Maybe @m133225 could shed some light on that :-P

pyokagan:
@Zhiyuan-Amos If @m133225 does not get back to us, I think a pretty fool-proof way of checking if `java.awt.robot` does anything at all is to grep for it inside the openjdk code base (where the AWT source code should be). If there are hits, it means it does something. If not, then it should be safe to remove it.

Zhiyuan-Amos:
@pyokagan I downloaded the [`openjdk` code base](https://github.com/javafxports/openjdk-jfx) and performed `grep -r 'java.awt.robot' .` in the directory and it returned no results. I'm not too familiar with grep, but from what I searched, it seems to mean `search recursively (in all sub directories also) for files that contain "java.awt.robot" at this . (current) directory` [(here's the source I referred to)](https://stackoverflow.com/questions/9217185/how-do-i-use-grep-to-search-the-current-directory-for-all-files-having-the-a-str).

I didn't receive any results, and as such, if my grep command is right, I suppose it's safe to delete it?

pyokagan:
@Zhiyuan-Amos 
>  I downloaded the openjdk code base 

That's the openjfx (JavaFX) code base, not the openjdk one. The correct one is at http://hg.openjdk.java.net/jdk/jdk10/

Anyway here are my results:
```
$ grep -r 'java\.awt\.robot'
test/jdk/lib/testlibrary/ExtendedRobot.java: * lifecycle. Constructor reads vm integer property {@code java.awt.robotdelay} and sets the delay value
test/jdk/lib/testlibrary/ExtendedRobot.java:    //    syncDelay = AccessController.doPrivileged(new GetIntegerAction("java.awt.robotdelay", DEFAULT_SYNC_DELAY));
```

I also verified that the Java AWT source code is in `src/java.desktop` (can grep for `package java.awt` inside that directory)

So I think we can quite safely say that setting the system property `java.awt.robot` will not do anything.

pyokagan:
> http://hg.openjdk.java.net/jdk/jdk10/

Well, maybe I should have used the jdk9 one instead :-P

Zhiyuan-Amos:
> Well, maybe I should have used the jdk9 one instead :-P

I used the [jdk9](http://jdk.java.net/9/) one and there's no references to `java.awt.robot`.

> That's the openjfx (JavaFX) code base, not the openjdk one.

oh right, sorry for the mistake >< If I get your point correctly, you're saying that there's no need to set `java.awt.robot` to `true since `java.awt` package doesn't contain `java.awt.robot`? Sorry for the newb question.

pyokagan:
@Zhiyuan-Amos 
> If I get your point correctly, you're saying that there's no need to set java.awt.robot to true since java.awt package doesn't contain java.awt.robot? Sorry for the newb question. 

If `java.awt.robot` did something, I would expect that there is a `System.getProperty("java.awt.robot")` call inside the AWT source code in openjdk.

However, as the string `java.awt.robot` does not exist inside the source code at all, it is likely that setting `java.awt.robot` to `true` will not do anything at all.

If we want to do a more fool-proof check, we can also check all our dependencies as well :-P I expect library developers to be sane enough to not introduce a `java.awt` system property in a non-AWT library though.

Zhiyuan-Amos:
> If java.awt.robot did something, I would expect that there is a System.getProperty("java.awt.robot") call inside the AWT source code in openjdk.

Thanks for the explanation Paul! :) Appreciate it. :+1:

pyokagan:
@Zhiyuan-Amos What happened to travis?

pyokagan:
Since this PR is titled "tidy up build.gradle" and it's so thorough, here's some other cleanups to make it even more thorough :D :D

* Replace all string literals which do not use string interpolation with single quotes. (double-quotes creates a GString which is heavier than a regular String).
* Replace all usages of: `/* comments */` with `// comments`. These kind of `/* comments */` can then be used for commenting out blocks of code during debugging.

eugenepeh:
Btw I found the [Travis build request](https://travis-ci.org/se-edu/addressbook-level4/builds/362455312) that ended up in error. Apparently travis was waiting for report and hence didn't trigger any new build request.

I've restarted the build. Hopefully it will solve the problem.

yamgent:
> I noticed that running gradlew shadowJar runs the asciidoctor as well (producing full html files in build directory)

`HelpWindow` is using `/docs/UserGuide.html`. We need this document to be generated so that it can be packed in the `jar` file.

I don't think we implement it to only generate `/docs/UserGuide.html` for `shadowJar`, so it currently generates the entire documentation using `asciidoctor`.

Zhiyuan-Amos:
@eugenepeh sorry how did you find the build report? :O

eugenepeh:
@Zhiyuan-Amos go to https://travis-ci.org/se-edu/addressbook-level4/, you should be able to see the list of Branches, Build History, Pull Requests and etc. You can then slowly dig out the build report.... :D

PR #863, opened 2018-03-26
PersonListPanelTest: fix flaky test

Zhiyuan-Amos:
```
LIST_CREATION_TIMEOUT has a value of 5500ms, which causes
PersonListPanelTest#performanceTest() to fail if the list creation takes
longer than the set amount of time.

This value is too small and this test executed on Travis CI sometimes
fails. [1]

Let's increase the value of LIST_CREATION_TIMEOUT by 4000ms so that
this test becomes very unlikely to fail.

[1]: https://travis-ci.org/se-edu/addressbook-level4/builds/358246241
```

### DIFF
--- a/src/test/java/seedu/address/ui/PersonListPanelTest.java
+++ b/src/test/java/seedu/address/ui/PersonListPanelTest.java
@@ -2,7 +2,6 @@ package seedu.address.ui;
 
 import static java.time.Duration.ofMillis;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
 import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
 import static seedu.address.testutil.EventsUtil.postNow;
 import static seedu.address.testutil.TypicalIndexes.INDEX_SECOND_PERSON;
@@ -11,18 +10,13 @@ import static seedu.address.ui.testutil.GuiTestAssert.assertCardDisplaysPerson;
 import static seedu.address.ui.testutil.GuiTestAssert.assertCardEquals;
 
 import java.io.File;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Logger;
 
 import org.junit.Test;
-
-import com.google.common.base.Stopwatch;
 
 import guitests.guihandles.PersonCardHandle;
 import guitests.guihandles.PersonListPanelHandle;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
-import seedu.address.commons.core.LogsCenter;
 import seedu.address.commons.events.ui.JumpToListRequestEvent;
 import seedu.address.commons.util.FileUtil;
 import seedu.address.commons.util.XmlUtil;
@@ -37,11 +31,7 @@ public class PersonListPanelTest extends GuiUnitTest {
 
     private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/sandbox/");
 
-    private static final Logger logger = LogsCenter.getLogger(PersonListPanelTest.class);
-
-    private static final int TEST_TIMEOUT = 8000;
     private static final long CARD_CREATION_AND_DELETION_TIMEOUT = 2500;
-    private static final long LIST_CREATION_TIMEOUT = TEST_TIMEOUT - CARD_CREATION_AND_DELETION_TIMEOUT;
 
     private PersonListPanelHandle personListPanelHandle;
 
@@ -73,12 +63,6 @@ public class PersonListPanelTest extends GuiUnitTest {
     /**
      * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
      * to execute.
-     * <ol>
-     *     <li>If {@code AssertionError} is thrown, then the failure occurs during the preparation of the test
-     *     (creating the xml file containing the large number of persons).</li>
-     *     <li>If {@code AssertionFailedError} is thrown, then the failure occurs during the creation and deletion of
-     *     the person cards.</li>
-     * </ol>
      */
     @Test
     public void performanceTest() throws Exception {
@@ -92,29 +76,13 @@ public class PersonListPanelTest extends GuiUnitTest {
 
     /**
      * Returns a list of persons containing {@code personCount} persons that is used to populate the
-     * {@code PersonListPanel}. Logs the time taken for method execution if it finishes before
-     * {@code LIST_CREATION_TIMEOUT}.
-     *
-     * @throws AssertionError if this method takes too long to execute.
+     * {@code PersonListPanel}.
      */
     private ObservableList<Person> createBackingList(int personCount) throws Exception {
-        Stopwatch stopwatch = Stopwatch.createStarted();
-
         File xmlFile = createXmlFileWithPersons(personCount);
         XmlSerializableAddressBook xmlAddressBook =
                 XmlUtil.getDataFromFile(xmlFile, XmlSerializableAddressBook.class);
-        ObservableList<Person> personList =
-                FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
-
-        long createListTime = stopwatch.elapsed(TimeUnit.MILLISECONDS);
-        String createListMessage = "List creation took: " + createListTime + "ms. ";
-        if (createListTime >= LIST_CREATION_TIMEOUT) {
-            fail(createListMessage + "Time limit exceeded.");
-        } else {
-            logger.info(createListMessage + "Continuing test.");
-        }
-
-        return personList;
+        return FXCollections.observableArrayList(xmlAddressBook.toModelType().getPersonList());
     }
 
     /**
--- a/src/test/java/seedu/address/ui/PersonListPanelTest.java
+++ b/src/test/java/seedu/address/ui/PersonListPanelTest.java
@@ -61,8 +61,8 @@ public class PersonListPanelTest extends GuiUnitTest {
     }
 
     /**
-     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
-     * to execute.
+     * Verifies that creating and deleting large number of persons in {@code PersonListPanel} requires lesser than
+     * {@code CARD_CREATION_AND_DELETION_TIMEOUT} milliseconds to execute.
      */
     @Test
     public void performanceTest() throws Exception {
--- a/src/test/java/seedu/address/ui/PersonListPanelTest.java
+++ b/src/test/java/seedu/address/ui/PersonListPanelTest.java
@@ -1,7 +1,9 @@
 package seedu.address.ui;
 
+import static java.time.Duration.ofMillis;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertTimeoutPreemptively;
 import static seedu.address.testutil.EventsUtil.postNow;
 import static seedu.address.testutil.TypicalIndexes.INDEX_SECOND_PERSON;
 import static seedu.address.testutil.TypicalPersons.getTypicalPersons;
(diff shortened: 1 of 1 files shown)


src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -37,7 +37,7 @@
 
     private static final Logger logger = LogsCenter.getLogger(PersonListPanelTest.class);
 
-    private static final int TEST_TIMEOUT = 8000;
+    private static final int TEST_TIMEOUT = 12000;
     private static final int CARD_CREATION_AND_DELETION_TIMEOUT = 2500;
     private static final int LIST_CREATION_TIMEOUT = TEST_TIMEOUT - CARD_CREATION_AND_DELETION_TIMEOUT;
```
yamidark: Would it be clearer if we set individual time outs for both these tests, then set the total timeout in line 84 as `@Test(timeout = LIST_CREATION_TIMEOUT + CARD_CREATION_AND_DELETION_TIMEOUT)` instead? Then in future if we wish to change the timeout for an individual test we won't need to adjust the other.

Zhiyuan-Amos:
Thanks for your prompt reviews @eugenepeh and @yamidark :D

> LIST_CREATION_TIMEOUT seems to be tested before CARD_CREATION_AND_DELETION_TIMEOUT. Wouldn't it have a similar chance of failing as well if LIST_CREATION_TIMEOUT can fail? Thus shouldn't CARD_CREATION_AND_DELETION_TIMEOUT receive an equivalent increase in ratio?

From the failing Travis builds that I've inspected, the tests are only failing because of `LIST_CREATION_TIMEOUT` being exceeded. These are the only 2 cases that I've observed:
1. `LIST_CREATION_TIMEOUT` and `CARD_CREATION_AND_DELETION_TIMEOUT` aren't being exceeded -> test passes.
1. `LIST_CREATION_TIMEOUT` exceeded -> test fails.

I have not observed that `LIST_CREATION_TIMEOUT` not exceeded while `CARD_CREATION_AND_DELETION_TIMEOUT` exceeded (I've rebuilt my PR about 4 times without any test failure), so it seems like this particular case doesn't happen. :)

src/test/java/seedu/address/ui/PersonListPanelTest.java
Zhiyuan-Amos: LOL. You are right.

eugenepeh:
>I have not observed that LIST_CREATION_TIMEOUT not exceeded while CARD_CREATION_AND_DELETION_TIMEOUT exceeded (I've rebuilt my PR about 4 times without any test failure), so it seems like this particular case doesn't happen. :)

Hmm? `LIST_CREATION_TIMEOUT` runs before `CARD_CREATION_AND_DELETION_TIMEOUT`. If it already fails in `LIST_CREATION_TIMEOUT`, why would it run `CARD_CREATION_AND_DELETION_TIMEOUT`? :P

Zhiyuan-Amos:
> Hmm? LIST_CREATION_TIMEOUT runs before CARD_CREATION_AND_DELETION_TIMEOUT. If it has already failed in LIST_CREATION_TIMEOUT, why would it run CARD_CREATION_AND_DELETION_TIMEOUT? :P

Yup, I get what you're saying. I'm saying that CARD_CREATION_AND_DELETION_TIMEOUT seems to be sufficient because: 
1. I've not observed that `LIST_CREATION_TIMEOUT` doesn't fail, but `CARD_CREATION_AND_DELETION_TIMEOUT` fails before this PR.
1. In this PR, after increasing `LIST_CREATION_TIMEOUT`, I've not observed `CARD_CREATION_AND_DELETION_TIMEOUT` failing; the test always pass.

I get what you're saying that if `LIST_CREATION_TIMEOUT` fails, we can't be sure whether `CARD_CREATION_AND_DELETION_TIMEOUT` fails or not. You're right to point out that, but for the reasons mentioned above, it seems that `CARD_CREATION_AND_DELETION_TIMEOUT` is within a safe value.

eugenepeh:
> In this PR, after increasing LIST_CREATION_TIMEOUT, I've not observed CARD_CREATION_AND_DELETION_TIMEOUT failing; the test always pass.

I see, how often has `LIST_CREATION_TIMEOUT` been failing since 4ec524489670b3a424c1b15aa3776bda84badacb though?

Zhiyuan-Amos:
@eugenepeh you can see the PRs with red crosses, generally Travis build fails about a third of the time. 

I do think that it's safer to increase the timing of the timeout, but it isn't broken now (I ran the build about 8 times) , so I'm not sure whether we should fix it or not ><

eugenepeh:
> I do think that it's safer to increase the timing of the timeout, but it isn't broken now (I ran the build about 8 times) , so I'm not sure whether we should increase it or not ><

I do agree that it might be safer to increase the timing of the timeout. But at the same time, it would be nice to know when or why it occurs too, e.g. Travis downtime / maintenance etc https://mobile.twitter.com/traviscistatus. If the frequency and timing of the PR failures coincide with the Travis downtime, I don't think the increase would help >_<

How about we log the output of the performance test timing as well for now to see if its caused by boundary increase?

Zhiyuan-Amos:
I don't think that it's caused by Travis downtime because sometimes when you restart a build, it fails, then sometimes when you restart, it passes.

As such, when I ran the tests locally, I noticed that the test execution timings fluctuate quite a bit. I'm guessing that Travis Ci takes a longer time to execute the method that creates an xml file, and the fluctuation causes the test to sometimes pass and to sometimes fail.

Zhiyuan-Amos:
We can log the timing, do you have any idea how to display it in the build log? ><

eugenepeh:
Hmm... 

https://github.com/se-edu/addressbook-level4/blob/a4ef5229e29bb1da91d462a650ee82c243ab1b7c/build.gradle#L114-L126

https://stackoverflow.com/questions/3963708/gradle-how-to-display-test-results-in-the-console-in-real-time

Think modifying gradle helps but it might take some work to display PerformanceTest only >_<

Zhiyuan-Amos:
Hmm ok I think there's a better solution. I'll try to be clear in my explanation:
1. If we upgrade our tests to JUnit 5, we have a simpler way of [writing the code for this test](https://github.com/se-edu/addressbook-level4/pull/811#pullrequestreview-106489082). We no longer need `@Timeout`.
1. Without the need for `@Timeout`, we don't need to have a timeout for the .xml file, since the purpose of this test is not to test how long it takes to create an .xml file with 10000 persons. See the [discussion](https://github.com/se-edu/addressbook-level4/pull/818#pullrequestreview-104077673) and the threads that follow after.
1. Without having to time the creation of the .xml file, then we don't need to think of how to log the timing in Gradle. Yay.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -71,53 +63,26 @@ public void handleJumpToListRequestEvent() {
...
-
-        Stopwatch stopwatch = Stopwatch.createStarted();
         initUi(backingList);
-        guiRobot.interact(backingList::clear);
-        if (stopwatch.elapsed(TimeUnit.MILLISECONDS) >= CARD_CREATION_AND_DELETION_TIMEOUT) {
-            fail("Creation and deletion of person cards exceeded time limit");
-        }
+
+        assertTimeoutPreemptively(ofMillis(CARD_CREATION_AND_DELETION_TIMEOUT),
+                () -> guiRobot.interact(backingList::clear),
```
Zhiyuan-Amos: @eugenepeh can I ask you a qn since you've been working on Checkstyle? 

I received an error message on Checkstyle `'(' should be on the previous line. [SeparatorWrap]`. I get the intention of this check, but it causes this line to fail the Checkstyle test :(

I can do it this way, but it looks really ugly.
```java
assertTimeoutPreemptively(ofMillis(CARD_CREATION_AND_DELETION_TIMEOUT), () ->
                guiRobot.interact(backingList::clear),
        "Creation and deletion of person cards exceeded time limit");
```

Zhiyuan-Amos:
@yamgent tagging you to take a look at this small PR since we discussed on it previously :)

src/test/java/seedu/address/ui/PersonListPanelTest.java
eugenepeh: I see. Checkstyle is not perfect in some sense, I not sure if they have implement new options to counter such scenario; 
I can probably let them know if they don't and see if they can fixed it.
Meanwhile perhaps can do something like this?

<pre>
assertTimeoutPreemptively(
    ofMillis(CARD_CREATION_AND_DELETION_TIMEOUT), () -> guiRobot.interact(backingList::clear),
    "Creation and deletion of person cards exceeded time limit"
    );
</pre>

Zhiyuan-Amos: Thanks Eugene for your suggestion :)

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -73,12 +63,6 @@ public void handleJumpToListRequestEvent() {
     /**
      * Verifies that creating and deleting large number of persons in {@code PersonListPanel} does not take too long
```
yamgent: Btw, since you are cleaning this up... I think we should use `{@code CARD_CREATION_AND_DELETION_TIMEOUT}` instead of `take too long`.

Zhiyuan-Amos: Thanks for your prompt review :)

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -72,25 +74,20 @@ public void handleJumpToListRequestEvent() {
...
-    // The timeout makes the test fail fast. Without it, the test can stall for a long time if
-    // guiRobot#interact(Runnable) takes very long to complete execution because guiRobot#interact(Runnable) only
-    // returns after the Runnable completes execution.
-    @Test(timeout = TEST_TIMEOUT)
+    @Test
     public void performanceTest() throws Exception {
         ObservableList<Person> backingList = createBackingList(10000);
-
-        Stopwatch stopwatch = Stopwatch.createStarted();
         initUi(backingList);
```
yamidark: Shouldn't this also be inside `assertTimeoutPreemptively` too, since we are also checking for the time taken for creation of person cards?

Zhiyuan-Amos: You are right!

yamgent: Was that taking a long time as well? I was under the impression that it was the clearing of the `PersonCard`s that was the main bottleneck....

Zhiyuan-Amos: It depends on the implementation of the `MappedList` that we were using (see my [comment in the issue](https://github.com/se-edu/addressbook-level4/issues/796#issuecomment-357432776)). :)

In short, the bottleneck can be either in the clearing or creating of the `PersonCard`s, and so we should time both.

src/test/java/seedu/address/ui/PersonListPanelTest.java
```diff
@@ -72,25 +74,20 @@ public void handleJumpToListRequestEvent() {
...
-        Stopwatch stopwatch = Stopwatch.createStarted();
-        initUi(backingList);
-        guiRobot.interact(backingList::clear);
-        if (stopwatch.elapsed(TimeUnit.MILLISECONDS) >= CARD_CREATION_AND_DELETION_TIMEOUT) {
-            fail("Creation and deletion of person cards exceeded time limit");
-        }
+        assertTimeoutPreemptively(ofMillis(CARD_CREATION_AND_DELETION_TIMEOUT), () -> {
+            initUi(backingList);
+            guiRobot.interact(backingList::clear);
+        }, "Creation and deletion of person cards exceeded time limit");
```
Zhiyuan-Amos: @eugenepeh your previous suggestion on resolving this checkstyle issue doesn't work now because I need to execute more than 1 line of code :(

Zhiyuan-Amos:
@damithc appveyor is failing because Help Window tests sometimes stall. The tests are introduced in #850 but Appveyor didn't fail during then (which was why it was left undetected), and I've notified @yamidark to look into it. The test is flaky; sometimes passes sometimes fails.

src/test/java/seedu/address/ui/PersonListPanelTest.java
eugenepeh: I see.

In this case I think need 1 more indentation because its a definition within parameter.
So maybe something like this?
<pre>
assertTimeoutPreemptively(ofMillis(CARD_CREATION_AND_DELETION_TIMEOUT), () -> {
            initUi(backingList);
            guiRobot.interact(backingList::clear);
        }, "Creation and deletion of person cards exceeded time limit");
</pre>

Sorry for the late reply >_< rushing fyp report.

Zhiyuan-Amos: No problem, thanks for still reviewing this PR! :)

I can't do your suggestion though, because:

image

eugenepeh: I see, in that case, I think its okay to leave it as it is for now.

PR #866, opened 2018-03-28
Use java.nio.file for file/path handling #750

### LINKED ISSUE #554: StorageManagerTest#getTempFilePath(String): Fix missing file separator
`StorageManagerTest#getTempFilePath(String)` uses a temporary folder to store temporary files created and used by the test.

However, the file path getter method does not use a file separator between the folder path and the file name:

```java
private String getTempFilePath(String fileName) {
    return testFolder.getRoot().getPath() + fileName;
}
```

So the temporary folder can be `~/temp/junit123/`, but the file is created outside `junit123` temporary folder (e.g. `~/temp/junit123file.xml`. While the tests can still work as normal, it is not intuitive. 

The way to fix this is simply to add a `File.separator` in between:

```java
private String getTempFilePath(String fileName) {
    return testFolder.getRoot().getPath() + File.separator + fileName;
}
```

As this is merely an inconvenience that happens in tests (just have a harder time finding the file), this is not a very important issue to fix now.

### LINKED ISSUE #750: Use java.nio.file for file/path handling
Constructing paths using string operations is error prone, but there's lots of areas in the code base where this is done, which sets a bad example.

Lots of APIs in the code base also take in Strings as file paths, which further encourages this kind of code.

vivekscl:
Fixes #750 
Fixes #554 

### Things to take note of
* The performance test in `PersonListPanelTest` is rather fickle. It fails sometimes, while passes in others. As such, `TEST_TIMEOUT` will be set to 12000 as a temporary measure until #863 is merged.
* Even though many classes in the UI component use `String` arguments for file names for `Class#getResource(String)`, the file names were changed to be of type `Path` and then converted to type `String` when needed. Not entirely sure if this is needed, but since using `Strings` for file names is still error prone I refactored it anyway. Do let me know if this shouldn't be the case. 

### Issues to raise
* Many parts of the codebase have methods that declare that they throw an `IOException` even though it is never thrown in the first place. One example is `JsonUserPrefsStorage#readUserPrefs(Path)`.
* A few methods have return values that are never used. One such example is `FileUtil#createFile(File)`.
* Many methods can be made to have a more restrictive access modifier. One such case would be the methods in `AddressBookSystemTest` which could be made to be private. Not sure if this was intended for future use though.
* `UserPrefs#getAddressBookName` is unused.

### DIFF
--- a/src/main/java/seedu/address/AppParameters.java
+++ b/src/main/java/seedu/address/AppParameters.java
@@ -1,22 +1,28 @@
 package seedu.address;
 
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Map;
 import java.util.Objects;
+import java.util.logging.Logger;
 
 import javafx.application.Application;
+import seedu.address.commons.core.LogsCenter;
+import seedu.address.commons.util.FileUtil;
 
 /**
  * Represents the parsed command-line parameters given to the application.
  */
 public class AppParameters {
+    private static final Logger logger = LogsCenter.getLogger(AppParameters.class);
 
-    private String configPath;
+    private Path configPath;
 
-    public String getConfigPath() {
+    public Path getConfigPath() {
         return configPath;
     }
 
-    public void setConfigPath(String configPath) {
+    public void setConfigPath(Path configPath) {
         this.configPath = configPath;
     }
 
@@ -28,7 +34,11 @@ public class AppParameters {
         Map<String, String> namedParameters = parameters.getNamed();
 
         String configPathParameter = namedParameters.get("config");
-        appParameters.setConfigPath(configPathParameter);
+        if (configPathParameter != null && !FileUtil.isValidPath(configPathParameter)) {
+            logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
+            configPathParameter = null;
+        }
+        appParameters.setConfigPath(configPathParameter != null ? Paths.get(configPathParameter) : null);
 
         return appParameters;
     }
--- a/src/main/java/seedu/address/AppParameters.java
+++ b/src/main/java/seedu/address/AppParameters.java
@@ -1,0 +1,54 @@
+package seedu.address;
+
+import java.util.Map;
+import java.util.Objects;
+
+import javafx.application.Application;
+
+/**
+ * Represents the parsed command-line parameters given to the application.
+ */
+public class AppParameters {
+
+    private String configPath;
+
+    public String getConfigPath() {
+        return configPath;
+    }
+
+    public void setConfigPath(String configPath) {
+        this.configPath = configPath;
+    }
+
+    /**
+     * Parses the application command-line parameters.
+     */
+    public static AppParameters parse(Application.Parameters parameters) {
+        AppParameters appParameters = new AppParameters();
+        Map<String, String> namedParameters = parameters.getNamed();
+
+        String configPathParameter = namedParameters.get("config");
+        appParameters.setConfigPath(configPathParameter);
+
+        return appParameters;
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (other == this) {
+            return true;
+        }
+
+        if (!(other instanceof AppParameters)) {
+            return false;
+        }
+
+        AppParameters otherAppParameters = (AppParameters) other;
+        return Objects.equals(getConfigPath(), otherAppParameters.getConfigPath());
+    }
+
+    @Override
+    public int hashCode() {
+        return configPath.hashCode();
+    }
+}
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -1,7 +1,6 @@
 package seedu.address;
 
 import java.io.IOException;
-import java.util.Map;
 import java.util.Optional;
 import java.util.logging.Logger;
 
@@ -57,7 +56,8 @@ public class MainApp extends Application {
         logger.info("=============================[ Initializing AddressBook ]===========================");
         super.init();
 
-        config = initConfig(getApplicationParameter("config"));
+        AppParameters appParameters = AppParameters.parse(getParameters());
+        config = initConfig(appParameters.getConfigPath());
 
         UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
         userPrefs = initPrefs(userPrefsStorage);
@@ -73,11 +73,6 @@ public class MainApp extends Application {
         ui = new UiManager(logic, config, userPrefs);
 
         initEventsCenter();
-    }
-
-    private String getApplicationParameter(String parameterName) {
-        Map<String, String> applicationParameters = getParameters().getNamed();
-        return applicationParameters.get(parameterName);
     }
 
     /**
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -1,6 +1,7 @@
 package seedu.address;
 
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Optional;
 import java.util.logging.Logger;
 
@@ -109,9 +110,9 @@ public class MainApp extends Application {
      * The default file path {@code Config#DEFAULT_CONFIG_FILE} will be used instead
      * if {@code configFilePath} is null.
      */
-    protected Config initConfig(String configFilePath) {
+    protected Config initConfig(Path configFilePath) {
         Config initializedConfig;
-        String configFilePathUsed;
+        Path configFilePathUsed;
 
         configFilePathUsed = Config.DEFAULT_CONFIG_FILE;
 
@@ -146,7 +147,7 @@ public class MainApp extends Application {
      * reading from the file.
      */
     protected UserPrefs initPrefs(UserPrefsStorage storage) {
-        String prefsFilePath = storage.getUserPrefsFilePath();
+        Path prefsFilePath = storage.getUserPrefsFilePath();
         logger.info("Using prefs file : " + prefsFilePath);
 
         UserPrefs initializedPrefs;
--- a/src/main/java/seedu/address/commons/util/FileUtil.java
+++ b/src/main/java/seedu/address/commons/util/FileUtil.java
@@ -45,19 +45,16 @@ public class FileUtil {
     }
 
     /**
-     * Creates a file if it does not exist along with its missing parent directories
-     *
-     * @return true if file is created, false if file already exists
+     * Creates a file if it does not exist along with its missing parent directories.
      */
-    public static boolean createFile(Path file) throws IOException {
+    public static void createFile(Path file) throws IOException {
         if (Files.exists(file)) {
-            return false;
+            return;
         }
 
         createParentDirsOfFile(file);
 
         Files.createFile(file);
-        return true;
     }
 
     /**
--- a/src/main/java/seedu/address/commons/util/FileUtil.java
+++ b/src/main/java/seedu/address/commons/util/FileUtil.java
@@ -5,7 +5,9 @@ import static seedu.address.commons.util.AppUtil.checkArgument;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 
 /**
  * Writes and reads files
@@ -16,6 +18,20 @@ public class FileUtil {
 
     public static boolean isFileExists(Path file) {
         return Files.exists(file) && Files.isRegularFile(file);
+    }
+
+    /**
+     * Returns true if {@code path} can be converted into a {@code Path} via {@link Paths#get(String)},
+     * otherwise returns false.
+     * @param path A string representing the file path. Cannot be null.
+     */
+    public static boolean isValidPath(String path) {
+        try {
+            Paths.get(path);
+        } catch (InvalidPathException ipe) {
+            return false;
+        }
+        return true;
     }
 
     /**
--- a/src/main/java/seedu/address/commons/util/FileUtil.java
+++ b/src/main/java/seedu/address/commons/util/FileUtil.java
@@ -5,6 +5,7 @@ import static seedu.address.commons.util.AppUtil.checkArgument;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 
 /**
  * Writes and reads files
@@ -13,15 +14,15 @@ public class FileUtil {
 
     private static final String CHARSET = "UTF-8";
 
-    public static boolean isFileExists(File file) {
-        return file.exists() && file.isFile();
+    public static boolean isFileExists(Path file) {
+        return Files.exists(file) && Files.isRegularFile(file);
     }
 
     /**
      * Creates a file if it does not exist along with its missing parent directories.
      * @throws IOException if the file or directory cannot be created.
      */
-    public static void createIfMissing(File file) throws IOException {
+    public static void createIfMissing(Path file) throws IOException {
         if (!isFileExists(file)) {
             createFile(file);
         }
@@ -32,14 +33,15 @@ public class FileUtil {
      *
      * @return true if file is created, false if file already exists
      */
-    public static boolean createFile(File file) throws IOException {
-        if (file.exists()) {
+    public static boolean createFile(Path file) throws IOException {
+        if (Files.exists(file)) {
             return false;
         }
 
         createParentDirsOfFile(file);
 
-        return file.createNewFile();
+        Files.createFile(file);
+        return true;
     }
 
     /**
@@ -48,36 +50,37 @@ public class FileUtil {
      * @param dir the directory to be created; assumed not null
      * @throws IOException if the directory or a parent directory cannot be created
      */
-    public static void createDirs(File dir) throws IOException {
-        if (!dir.exists() && !dir.mkdirs()) {
-            throw new IOException("Failed to make directories of " + dir.getName());
+    public static void createDirs(Path dir) throws IOException {
+        if (Files.exists(dir)) {
+            return;
         }
+        Files.createDirectories(dir);
     }
 
     /**
      * Creates parent directories of file if it has a parent directory
      */
-    public static void createParentDirsOfFile(File file) throws IOException {
-        File parentDir = file.getParentFile();
+    public static void createParentDirsOfFile(Path file) throws IOException {
+        Path parentDir = file.getParent();
 
         if (parentDir != null) {
-            createDirs(parentDir);
+            Files.createDirectories(parentDir);
         }
     }
 
     /**
      * Assumes file exists
      */
-    public static String readFromFile(File file) throws IOException {
-        return new String(Files.readAllBytes(file.toPath()), CHARSET);
+    public static String readFromFile(Path file) throws IOException {
+        return new String(Files.readAllBytes(file), CHARSET);
     }
 
     /**
      * Writes given string to a file.
      * Will create the file if it does not exist yet.
      */
-    public static void writeToFile(File file, String content) throws IOException {
-        Files.write(file.toPath(), content.getBytes(CHARSET));
+    public static void writeToFile(Path file, String content) throws IOException {
+        Files.write(file, content.getBytes(CHARSET));
     }
 
     /**
--- a/src/main/java/seedu/address/commons/util/FileUtil.java
+++ b/src/main/java/seedu/address/commons/util/FileUtil.java
@@ -1,8 +1,5 @@
 package seedu.address.commons.util;
 
-import static seedu.address.commons.util.AppUtil.checkArgument;
-
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
@@ -58,19 +55,6 @@ public class FileUtil {
     }
 
     /**
-     * Creates the given directory along with its parent directories
-     *
-     * @param dir the directory to be created; assumed not null
-     * @throws IOException if the directory or a parent directory cannot be created
-     */
-    public static void createDirs(Path dir) throws IOException {
-        if (Files.exists(dir)) {
-            return;
-        }
-        Files.createDirectories(dir);
-    }
-
-    /**
      * Creates parent directories of file if it has a parent directory
      */
     public static void createParentDirsOfFile(Path file) throws IOException {
@@ -94,16 +78,6 @@ public class FileUtil {
      */
     public static void writeToFile(Path file, String content) throws IOException {
         Files.write(file, content.getBytes(CHARSET));
-    }
-
-    /**
-     * Converts a string to a platform-specific file path
-     * @param pathWithForwardSlash A String representing a file path but using '/' as the separator
-     * @return {@code pathWithForwardSlash} but '/' replaced with {@code File.separator}
-     */
-    public static String getPath(String pathWithForwardSlash) {
-        checkArgument(pathWithForwardSlash.contains("/"));
-        return pathWithForwardSlash.replace("/", File.separator);
     }
 
 }
--- a/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
+++ b/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
@@ -2,9 +2,11 @@ package seedu.address.storage;
 
 import static java.util.Objects.requireNonNull;
 
-import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.logging.Logger;
 
@@ -45,14 +47,14 @@ public class XmlAddressBookStorage implements AddressBookStorage {
                                                                                  FileNotFoundException {
         requireNonNull(filePath);
 
-        File addressBookFile = new File(filePath);
+        Path addressBookFile = Paths.get(filePath);
 
-        if (!addressBookFile.exists()) {
+        if (!Files.exists(addressBookFile)) {
             logger.info("AddressBook file "  + addressBookFile + " not found");
             return Optional.empty();
         }
 
-        XmlSerializableAddressBook xmlAddressBook = XmlFileStorage.loadDataFromSaveFile(new File(filePath));
+        XmlSerializableAddressBook xmlAddressBook = XmlFileStorage.loadDataFromSaveFile(Paths.get(filePath));
         try {
             return Optional.of(xmlAddressBook.toModelType());
         } catch (IllegalValueException ive) {
@@ -74,7 +76,7 @@ public class XmlAddressBookStorage implements AddressBookStorage {
         requireNonNull(addressBook);
         requireNonNull(filePath);
 
-        File file = new File(filePath);
+        Path file = Paths.get(filePath);
         FileUtil.createIfMissing(file);
         XmlFileStorage.saveDataToFile(file, new XmlSerializableAddressBook(addressBook));
     }
--- a/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
+++ b/src/main/java/seedu/address/storage/XmlAddressBookStorage.java
@@ -6,7 +6,6 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.logging.Logger;
 
@@ -23,13 +22,13 @@ public class XmlAddressBookStorage implements AddressBookStorage {
 
     private static final Logger logger = LogsCenter.getLogger(XmlAddressBookStorage.class);
 
-    private String filePath;
+    private Path filePath;
 
-    public XmlAddressBookStorage(String filePath) {
+    public XmlAddressBookStorage(Path filePath) {
         this.filePath = filePath;
     }
 
-    public String getAddressBookFilePath() {
+    public Path getAddressBookFilePath() {
         return filePath;
     }
 
@@ -43,22 +42,20 @@ public class XmlAddressBookStorage implements AddressBookStorage {
      * @param filePath location of the data. Cannot be null
      * @throws DataConversionException if the file is not in the correct format.
      */
-    public Optional<ReadOnlyAddressBook> readAddressBook(String filePath) throws DataConversionException,
+    public Optional<ReadOnlyAddressBook> readAddressBook(Path filePath) throws DataConversionException,
                                                                                  FileNotFoundException {
         requireNonNull(filePath);
 
-        Path addressBookFile = Paths.get(filePath);
-
-        if (!Files.exists(addressBookFile)) {
-            logger.info("AddressBook file "  + addressBookFile + " not found");
+        if (!Files.exists(filePath)) {
+            logger.info("AddressBook file "  + filePath + " not found");
             return Optional.empty();
         }
 
-        XmlSerializableAddressBook xmlAddressBook = XmlFileStorage.loadDataFromSaveFile(Paths.get(filePath));
+        XmlSerializableAddressBook xmlAddressBook = XmlFileStorage.loadDataFromSaveFile(filePath);
         try {
             return Optional.of(xmlAddressBook.toModelType());
         } catch (IllegalValueException ive) {
-            logger.info("Illegal values found in " + addressBookFile + ": " + ive.getMessage());
+            logger.info("Illegal values found in " + filePath + ": " + ive.getMessage());
             throw new DataConversionException(ive);
         }
     }
@@ -72,13 +69,12 @@ public class XmlAddressBookStorage implements AddressBookStorage {
      * Similar to {@link #saveAddressBook(ReadOnlyAddressBook)}
      * @param filePath location of the data. Cannot be null
      */
-    public void saveAddressBook(ReadOnlyAddressBook addressBook, String filePath) throws IOException {
+    public void saveAddressBook(ReadOnlyAddressBook addressBook, Path filePath) throws IOException {
         requireNonNull(addressBook);
         requireNonNull(filePath);
 
-        Path file = Paths.get(filePath);
-        FileUtil.createIfMissing(file);
-        XmlFileStorage.saveDataToFile(file, new XmlSerializableAddressBook(addressBook));
+        FileUtil.createIfMissing(filePath);
+        XmlFileStorage.saveDataToFile(filePath, new XmlSerializableAddressBook(addressBook));
     }
 
 }
--- a/src/test/java/seedu/address/AppParametersTest.java
+++ b/src/test/java/seedu/address/AppParametersTest.java
@@ -2,6 +2,7 @@ package seedu.address;
 
 import static org.junit.Assert.assertEquals;
 
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -19,13 +20,20 @@ public class AppParametersTest {
     @Test
     public void parse_validConfigPath_success() {
         parametersStub.namedParameters.put("config", "config.json");
-        expected.setConfigPath("config.json");
+        expected.setConfigPath(Paths.get("config.json"));
         assertEquals(expected, AppParameters.parse(parametersStub));
     }
 
     @Test
     public void parse_nullConfigPath_success() {
         parametersStub.namedParameters.put("config", null);
+        assertEquals(expected, AppParameters.parse(parametersStub));
+    }
+
+    @Test
+    public void parse_invalidConfigPath_success() {
+        parametersStub.namedParameters.put("config", "a\0");
+        expected.setConfigPath(null);
         assertEquals(expected, AppParameters.parse(parametersStub));
     }
 
--- a/src/test/java/seedu/address/AppParametersTest.java
+++ b/src/test/java/seedu/address/AppParametersTest.java
@@ -1,0 +1,50 @@
+package seedu.address;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Test;
+
+import javafx.application.Application;
+
+public class AppParametersTest {
+
+    private final ParametersStub parametersStub = new ParametersStub();
+    private final AppParameters expected = new AppParameters();
+
+    @Test
+    public void parse_validConfigPath_success() {
+        parametersStub.namedParameters.put("config", "config.json");
+        expected.setConfigPath("config.json");
+        assertEquals(expected, AppParameters.parse(parametersStub));
+    }
+
+    @Test
+    public void parse_nullConfigPath_success() {
+        parametersStub.namedParameters.put("config", null);
+        assertEquals(expected, AppParameters.parse(parametersStub));
+    }
+
+    private static class ParametersStub extends Application.Parameters {
+        private Map<String, String> namedParameters = new HashMap<>();
+
+        @Override
+        public List<String> getRaw() {
+            throw new AssertionError("should not be called");
+        }
+
+        @Override
+        public List<String> getUnnamed() {
+            throw new AssertionError("should not be called");
+        }
+
+        @Override
+        public Map<String, String> getNamed() {
+            return Collections.unmodifiableMap(namedParameters);
+        }
+    }
+}
--- a/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
@@ -3,8 +3,9 @@ package seedu.address.commons.util;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 
-import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.logging.Level;
 
@@ -18,7 +19,7 @@ import seedu.address.commons.exceptions.DataConversionException;
 
 public class ConfigUtilTest {
 
-    private static final String TEST_DATA_FOLDER = FileUtil.getPath("./src/test/data/ConfigUtilTest/");
+    private static final Path TEST_DATA_FOLDER = Paths.get("src", "test", "data", "ConfigUtilTest");
 
     @Rule
     public ExpectedException thrown = ExpectedException.none();
@@ -75,12 +76,12 @@ public class ConfigUtilTest {
         Config config = new Config();
         config.setAppTitle("Typical App Title");
         config.setLogLevel(Level.INFO);
-        config.setUserPrefsFilePath("C:\\preferences.json");
+        config.setUserPrefsFilePath(Paths.get("C:\\preferences.json"));
         return config;
     }
 
     private Optional<Config> read(String configFileInTestDataFolder) throws DataConversionException {
-        String configFilePath = addToTestDataPathIfNotNull(configFileInTestDataFolder);
+        Path configFilePath = addToTestDataPathIfNotNull(configFileInTestDataFolder);
         return ConfigUtil.readConfig(configFilePath);
     }
 
@@ -100,7 +101,7 @@ public class ConfigUtilTest {
     public void saveConfig_allInOrder_success() throws DataConversionException, IOException {
         Config original = getTypicalConfig();
 
-        String configFilePath = testFolder.getRoot() + File.separator + "TempConfig.json";
+        Path configFilePath = testFolder.getRoot().toPath().resolve("TempConfig.json");
 
         //Try writing when the file doesn't exist
         ConfigUtil.saveConfig(original, configFilePath);
@@ -116,13 +117,13 @@ public class ConfigUtilTest {
     }
 
     private void save(Config config, String configFileInTestDataFolder) throws IOException {
-        String configFilePath = addToTestDataPathIfNotNull(configFileInTestDataFolder);
+        Path configFilePath = addToTestDataPathIfNotNull(configFileInTestDataFolder);
         ConfigUtil.saveConfig(config, configFilePath);
     }
 
-    private String addToTestDataPathIfNotNull(String configFileInTestDataFolder) {
+    private Path addToTestDataPathIfNotNull(String configFileInTestDataFolder) {
         return configFileInTestDataFolder != null
-                                  ? TEST_DATA_FOLDER + configFileInTestDataFolder
+                                  ? TEST_DATA_FOLDER.resolve(configFileInTestDataFolder)
                                   : null;
     }
 
--- a/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/ConfigUtilTest.java
@@ -76,7 +76,7 @@ public class ConfigUtilTest {
         Config config = new Config();
         config.setAppTitle("Typical App Title");
         config.setLogLevel(Level.INFO);
-        config.setUserPrefsFilePath(Paths.get("C:\\preferences.json"));
+        config.setUserPrefsFilePath(Paths.get("preferences.json"));
         return config;
     }
 
--- a/src/test/java/seedu/address/commons/util/FileUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/FileUtilTest.java
@@ -1,12 +1,16 @@
 package seedu.address.commons.util;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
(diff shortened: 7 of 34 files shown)


src/test/java/seedu/address/commons/util/ConfigUtilTest.java
```diff
@@ -75,19 +76,19 @@ private Config getTypicalConfig() {
...
         config.setAppTitle("Typical App Title");
         config.setLogLevel(Level.INFO);
-        config.setUserPrefsFilePath("C:\\preferences.json");
+        config.setUserPrefsFilePath(Paths.get("C:\\preferences.json"));
         return config;
     }
 
-    private Optional<Config> read(String configFileInTestDataFolder) throws DataConversionException {
-        String configFilePath = addToTestDataPathIfNotNull(configFileInTestDataFolder);
+    private Optional<Config> read(Path configFileInTestDataFolder) throws DataConversionException {
```
pyokagan: Since this is a test method, it should be fine to either leave it alone taking in a `String`, or add an overload that takes in a `String`, so that we don't need to spam `Paths.get()` in all the tests. Being a test method, it is fine to get a `InvalidPathException` since it will fail the test anyway.

Same goes for all the other test methods.

src/main/java/seedu/address/MainApp.java
```diff
@@ -116,13 +118,13 @@ private void initLogging(Config config) {
      */
     protected Config initConfig(String configFilePath) {
```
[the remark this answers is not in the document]
vivekscl: In this method, there's a possibility that the `MainApp#getApplicationParameter(String)` returns a null String and causes the `Paths.get(String)` method to throw a `NullPointerException`. So I didn't add it in this case. A workaround would be to add a null check before returning the path but that would be more complicated than just returning a String right?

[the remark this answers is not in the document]
pyokagan: Well, putting aside the null check for the moment...

Would it be right for the application to crash with an `InvalidPathException` if the user inputs an invalid path?

At the boundaries between user input and the application you need to do a proper validation check, so yes you do need the complexity.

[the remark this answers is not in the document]
vivekscl: Oh what I meant is that, isn't there already a check where if the String is null, the default config file path will be used instead? So in this case, it's redundant to add an extra null check right?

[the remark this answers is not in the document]
pyokagan: Yes, it is redundant. But you'll need to balance it against the "handle errors properly" requirement that I mentioned up thread.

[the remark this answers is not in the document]
vivekscl: Ahh, I understand now. Will make the changes accordingly. Thanks for clearing up my doubts! :)

vivekscl:
On a side note, I tried to place the conversion of all usages of `File` in the code base to `Path` as a separate commit but it almost always results in that commit failing tests since many of the method signatures were changed because of the conversion. So I placed it together with the first commit instead.

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -48,17 +48,17 @@ public static boolean createFile(File file) throws IOException {
      * @param dir the directory to be created; assumed not null
      * @throws IOException if the directory or a parent directory cannot be created
      */
-    public static void createDirs(File dir) throws IOException {
-        if (!dir.exists() && !dir.mkdirs()) {
-            throw new IOException("Failed to make directories of " + dir.getName());
+    public static void createDirs(Path dir) throws IOException {
```
pyokagan: Seems like this whole function can be removed and replaced with `Files.createDirectories()`

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -32,14 +32,14 @@ public static void createIfMissing(File file) throws IOException {
...
+    public static boolean createFile(Path filePath) throws IOException {
+        if (Files.exists(filePath)) {
             return false;
         }
 
-        createParentDirsOfFile(file);
+        createParentDirsOfFile(filePath);
 
-        return file.createNewFile();
+        return filePath.toFile().createNewFile();
```
pyokagan: Seems like it can be replaced with `Files.createFile()`

src/main/java/seedu/address/MainApp.java
```diff
@@ -75,9 +77,11 @@ public void init() throws Exception {
         initEventsCenter();
     }
 
-    private String getApplicationParameter(String parameterName) {
+    private Optional<Path> getApplicationParameter(String parameterName) {
```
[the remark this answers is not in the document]
vivekscl: Ah, my bad. Should have studied the method more carefully :sweat_smile: .

src/main/java/seedu/address/MainApp.java
```diff
@@ -75,9 +77,11 @@ public void init() throws Exception {
...
         Map<String, String> applicationParameters = getParameters().getNamed();
-        return applicationParameters.get(parameterName);
+        return applicationParameters.get(parameterName) == null
+                ? Optional.empty()
+                : Optional.of(Paths.get(applicationParameters.get(parameterName)));
```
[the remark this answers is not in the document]
vivekscl: It doesn't crash, but it does throw an exception and then starts the app like so:
deepinscreenshot_select-area_20180402120431

But that's still bad practice, so I'll make the changes accordingly.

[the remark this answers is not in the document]
vivekscl: After trying to pass an invalid path to the master branch of this repo, the same error pops up and then the app starts, so it seems like this behaviour was already there from the beginning.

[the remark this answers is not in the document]
pyokagan: That's a `NoSuchFileException`. `Paths.get` only throws an `InvalidPathException`.

src/main/java/seedu/address/MainApp.java
```diff
@@ -57,7 +59,9 @@ public void init() throws Exception {
         logger.info("=============================[ Initializing AddressBook ]===========================");
         super.init();
 
-        config = initConfig(getApplicationParameter("config"));
+        config = getApplicationParameter("config") != null
+                ? initConfig(Paths.get(getApplicationParameter("config")))
+                : initConfig(null);
```
pyokagan: Eh, this is smushing too many things into one line.

At least, I'd expect something like:
```java
Path configPath = getApplicationParameter("config") != null ? Path.get(getApplicationParameter("config")) : null;
config = initConfig(configPath);
```
(Operations on the `configPath` is separated from the `config`)

*However*, this code still does not handle the possible `InvalidPathException` from `Paths.get()`. You need something like:

```java
String configPathParameter = getApplicationParameter("config");
if (configPathParameter != null && !isValidPath(configPathParameter)) {
    // Handle error
    // Considering the the rest of MainApp's code thinks its okay to not report errors to users,
    // I'm inclined to using something like:
    logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
    configPathParameter = null;
}
Path configPath = configPathParameter != null ? Path.get(configPathParameter) : null;
config = initConfig(configPath);
```

Yeah, Java sucks.

src/test/java/seedu/address/commons/util/XmlUtilTest.java
```diff
@@ -122,7 +123,7 @@ public void saveDataToFile_missingFile_fileNotFoundException() throws Exception
 
     @Test
     public void saveDataToFile_validFile_dataSaved() throws Exception {
-        TEMP_FILE.createNewFile();
+        TEMP_FILE.toFile().createNewFile();
```
pyokagan: `Files.createFile()` looked like it could work here, except that it throws an exception if the file already exists.

Maybe use `FileUtil.createFile()`?

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -88,6 +89,8 @@ public static void writeToFile(File file, String content) throws IOException {
     public static String getPath(String pathWithForwardSlash) {
         checkArgument(pathWithForwardSlash.contains("/"));
         return pathWithForwardSlash.replace("/", File.separator);
+    public static void writeToFile(Path filePath, String content) throws IOException {
```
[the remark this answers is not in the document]
pyokagan: So, this compiles on the last commit because the problem is fixed in [2/2]. Do ensure that the project compiles + passes all tests on *every commit*.

src/main/java/seedu/address/MainApp.java
vivekscl: For this, I tried testing if the `InvalidPathException` would be thrown using invalid paths as arguments, but it always returns true the `isValidPath` check on Linux. This [SO post](https://stackoverflow.com/questions/468789/is-there-a-way-in-java-to-determine-if-a-path-is-valid-without-attempting-to-cre) describes the same issue with Linux return true for any kind of path. Is it ok if I proceed with this check since it could work on Windows and MacOS?

pyokagan: @vivekscl 
> This SO post describes the same issue with Linux return true for any kind of path.

Seems like it has been fixed already.
https://ideone.com/4DONru

> Is it ok if I proceed with this check since it could work on Windows and MacOS?

Yeah, of course. It's all about the API contract. If `Paths.get` claims that it *could* throw an `InvalidPathException`, then that needs to be handled, whether it does actually throw one or not.

src/main/java/seedu/address/MainApp.java
```diff
@@ -80,6 +89,21 @@ private String getApplicationParameter(String parameterName) {
         return applicationParameters.get(parameterName);
     }
 
+    /**
+     * Checks if the path given is valid.
+     * @param path Points to the config file entered by the user.
+     *             Cannot be null.
+     * @return     true if path is valid, false if invalid.
+     */
+    private static boolean isValidPath(String path) {
```
pyokagan: This method has general utility, and so should be in `FileUtil` instead?

src/main/java/seedu/address/MainApp.java
```diff
@@ -57,7 +60,13 @@ public void init() throws Exception {
         logger.info("=============================[ Initializing AddressBook ]===========================");
         super.init();
 
-        config = initConfig(getApplicationParameter("config"));
+        String configPathParameter = getApplicationParameter("config");
+        if (configPathParameter != null && !isValidPath(configPathParameter)) {
+            logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
+            configPathParameter = null;
+        }
+        Path configPath = configPathParameter != null ? Paths.get(configPathParameter) : null;
```
pyokagan: Alright, but now we have a SLAP problem since the nitty-gritty of command-line parsing is at a much lower level compared to the high level setup of the application architecture (model/storage/logic/ui).

So, maybe extract out application parameter parsing into a separate method?

```java
private static class AppParameters {
    Path configPath;
}

/**
 * Parses application command-line parameters.
 */
private AppParameters getAppParameters() {
    AppParameters appParameters = new AppParameters();
    Map<String, String> namedParameters = getParameters().getNamed();

    String configPathParameter = namedParameters.get("config");
    if (configPathParameter != null && !isValidPath(configPathParameter)) {
        logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
        configPathParameter = null;
    }
    appParameters.configPath = configPathParameter != null ? Paths.get(configPathParameter) : null;

    // If future devs want to add additional command line arguments, they can add it here

    return appParameters;
}

...

AppParameters appParameters = getAppParameters();
config = initConfig(appParameters.configPath);
```

Note that this means that `getApplicationParameter(String)` is not needed anymore, and I prefer it that way since it forces developers to put all the application parameter parsing into one place for greater cohesiveness.

src/test/java/seedu/address/commons/util/ConfigUtilTest.java
```diff
@@ -75,12 +76,12 @@ private Config getTypicalConfig() {
         Config config = new Config();
         config.setAppTitle("Typical App Title");
         config.setLogLevel(Level.INFO);
-        config.setUserPrefsFilePath("C:\\preferences.json");
+        config.setUserPrefsFilePath(Paths.get("preferences.json"));
```
pyokagan: Argh, I realised I forgot to mention: this should be in a separate commit, since its a different problem from what this commit is solving.

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -13,35 +14,20 @@
...
-    public static void createIfMissing(File file) throws IOException {
+    public static void createIfMissing(Path file) throws IOException {
         if (!isFileExists(file)) {
-            createFile(file);
+            Files.createFile(file);
         }
     }
 
-    /**
-     * Creates a file if it does not exist along with its missing parent directories
```
pyokagan: Why should the "create missing parent directories" functionality be removed?

Let's try to preserve the existing behavior of the code base as much as possible.

src/main/java/seedu/address/MainApp.java
vivekscl: This means that we would have to write tests for this method and as such, `FileUtilTest` shouldn't be removed?

src/main/java/seedu/address/MainApp.java
vivekscl: On a side note, I would like to clarify why we are using a static class. Here is my reasoning:

You need a static class `AppParameters`, to encapsulate all the parameters that will be added in the future, rather than creating variables as and when you need them in the MainApp class which violates SLAP. And since it doesn't need to access anything in the outer class it was declared in (aka MainApp) it is declared static.

The `getAppParameters()` method is declared as a method inside MainApp but outside the static class because you have to create a new instance of the AppParameters class every time you get the app parameters. This is because every time you start the app, you would be using new parameters so you need a new instance of `AppParameters`.

src/main/java/seedu/address/MainApp.java
pyokagan: Yes

src/main/java/seedu/address/MainApp.java
pyokagan: > You need a static class AppParameters, to encapsulate all the parameters that will be added in the future, rather than creating variables as and when you need them in the MainApp class which makes it less cohesive. And since it doesn't need to access anything in the outer class it was declared in (aka MainApp) it is declared static.

Yes.

> The getAppParameters() method is declared as a method inside MainApp but outside the static class because you have to create a new instance of the AppParameters class every time you get the app parameters. 

Not exactly, it's because I need access to  `javafx.application.Application#getParameters()`.

src/main/java/seedu/address/MainApp.java
```diff
@@ -206,6 +205,29 @@ public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
...
+     * Static class encapsulating the app parameters.
+     */
+    private static class AppParameters {
+        private Path configPath;
+    }
+
+    /**
+     * Parses application command-line parameters.
+     */
+    private AppParameters getAppParameters() {
```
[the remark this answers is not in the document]
pyokagan: @vivekscl codacy seems to be correct here.

src/test/java/seedu/address/storage/XmlSerializableAddressBookTest.java
```diff
@@ -2,30 +2,31 @@
...
 
 public class XmlSerializableAddressBookTest {
 
-    private static final String TEST_DATA_FOLDER = FileUtil.getPath("src/test/data/XmlSerializableAddressBookTest/");
-    private static final File TYPICAL_PERSONS_FILE = new File(TEST_DATA_FOLDER + "typicalPersonsAddressBook.xml");
-    private static final File INVALID_PERSON_FILE = new File(TEST_DATA_FOLDER + "invalidPersonAddressBook.xml");
+    private static final Path TEST_DATA_FOLDER = Paths.get("src", "test", "data", "XmlSerializableAddressBookTest");
+    private static final Path TYPICAL_PERSONS_FILE_PATH = TEST_DATA_FOLDER.resolve("typicalPersonsAddressBook.xml");
+    private static final Path INVALID_PERSON_FILE_PATH = TEST_DATA_FOLDER.resolve("invalidPersonAddressBook.xml");
+    private static final Path INVALID_TAG_FILE_PATH = TEST_DATA_FOLDER.resolve("invalidTagAddressBook.xml");
```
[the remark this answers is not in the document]
pyokagan: @vivekscl codacy seems to be correct here.

Zhiyuan-Amos:
@vivekscl 

> The performance test in PersonListPanelTest is rather fickle. It fails sometimes, while passes in others. As such, TEST_TIMEOUT will be set to 12000 as a temporary measure until #863 is merged.

Resolved :)

src/main/java/seedu/address/MainApp.java
```diff
@@ -206,6 +205,29 @@ public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
         this.stop();
     }
 
+    /**
+     * Static class encapsulating the app parameters.
```
pyokagan: `Represents the command-line parameters given to the application` or something.

src/test/java/seedu/address/commons/util/FileUtilTest.java
```diff
@@ -28,4 +30,17 @@ public void getPath() {
...
+    public void isValidPath() {
+
+        // valid path
+        assertTrue(FileUtil.isValidPath("valid/file/path"));
+
+        // invalid path
+        assertFalse(FileUtil.isValidPath("a\0"));
+
+        // null path
+        assertFalse(FileUtil.isValidPath(null));
```
pyokagan: Is passing `null` to `isValidPath` a valid use case, or does it indicate that there is a programming error somewhere?

vivekscl: Although `isValidPath` is only used in `MainApp`, it is now a general utility method in `FileUtil` it has to guard against null arguments in the future to indicate an error right?

pyokagan: If a caller somehow passes a `null` to `isValidPath`, it means the caller messed up somewhere and so a `NullPointerException` should be thrown instead.

Passing a `null` to `isValidPath` is not meaningful at all since a `null` means different things in different contexts -- only the caller will know what the `null` means, and it is not `isValidPath`'s job to decide.

(You can draw parallels with how `Paths.get()` throws `InvalidPathException` for invalid paths, but throws `NullPointerException` if you pass in a null)

vivekscl: Ohh, I understand now! So in this case, I don't handle the `NullPointerException` so that it'll be thrown in order to let the caller know that he messed up?

pyokagan: Yes

vivekscl: On a side note, is the change of the parameter name `file` to `filePath` throughout the codebase necessary? From the commit you gave me as a reference, it seems like it's unnecessary :sweat_smile: .

pyokagan: Up to you, for that commit I just did the bare minimum.

vivekscl: >Up to you, for that commit I just did the bare minimum.

Now that the first commit is such that all usages of `File` were converted to `Path`, it looks like I'm randomly changing the parameter names in the next commit. Additionally, as you mentioned last time, a `Path` still points to a file so it seems like it's fine to leave the parameter names alone.

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -18,6 +20,21 @@ public static boolean isFileExists(Path file) {
         return Files.exists(file) && Files.isRegularFile(file);
     }
 
+    /**
+     * Checks if the path given is valid.
```
pyokagan: I wonder if `Returns true if the path name is valid, false otherwise.` would provide the same information while allowing you to drop the `@return` below.

Also, seems that some definition of a "valid path" would be needed, which is a pain since it is platform-dependent. I also note that the [javadocs](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html#get(java.lang.String,%20java.lang.String...)) for `Paths.get()` claims that it throws an `InvalidPathException`  if "the path string cannot be converted to a `Path`".

So, Maybe we can side-step this issue by giving a definition like this:

```
Returns true if {@code path} can be converted into a {@code Path} via {@link Paths#get(String)}, otherwise returns false
```

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -27,21 +44,14 @@ public static void createIfMissing(Path file) throws IOException {
...
     /**
-     * Creates a file if it does not exist along with its missing parent directories
-     *
-     * @return true if file is created, false if file already exists
+     * Creates a file if it does not exist along with its missing parent directories.
      */
-    public static boolean createFile(Path file) throws IOException {
-        if (Files.exists(file)) {
-            return false;
+    public static void createFile(Path file) throws IOException {
```
pyokagan: Unnecessary change? Why is there a need to change the return type?

src/main/java/seedu/address/storage/XmlAddressBookStorage.java
```diff
@@ -43,22 +42,20 @@ public String getAddressBookFilePath() {
      * @param filePath location of the data. Cannot be null
      * @throws DataConversionException if the file is not in the correct format.
      */
-    public Optional<ReadOnlyAddressBook> readAddressBook(String filePath) throws DataConversionException,
-            FileNotFoundException {
+    public Optional<ReadOnlyAddressBook> readAddressBook(Path filePath) throws DataConversionException,
+                                                                                 FileNotFoundException {
```
pyokagan: Unnecessary change.

src/test/java/seedu/address/commons/util/FileUtilTest.java
```diff
@@ -28,4 +30,18 @@ public void getPath() {
...
+    public void isValidPath() {
+
+        // valid path
+        assertTrue(FileUtil.isValidPath("valid/file/path"));
+
+        // invalid path
+        assertFalse(FileUtil.isValidPath("a\0"));
+
+        // null path -> throws NullPointerException
+        thrown.expect(NullPointerException.class);
```
pyokagan: Use `Assert.assertThrows()` instead. Using `thrown.expect` here means we can't add any other assertions after this.

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -13,81 +13,68 @@
...
+        return Files.exists(file) && Files.isRegularFile(file);
     }
 
     /**
-     * Creates a file if it does not exist along with its missing parent directories.
-     * @throws IOException if the file or directory cannot be created.
+     * Checks if the path given is valid.
+     * @param path A String representing the file path.
+     *             Cannot be null.
+     * @return     true if path is valid, false if invalid.
```
pyokagan: Addendum: let's not do manual alignment here, otherwise if the argument name `path` is changed to something that is longer or shorter we will need to re-align everything.

src/main/java/seedu/address/commons/util/FileUtil.java
vivekscl: Oh, that's because `Files.createFile(Path)` doesn't return boolean but a Path instead, and the return value of `createFile` wasn't used either.

pyokagan: Should be done in a separate commit.

src/main/java/seedu/address/MainApp.java
```diff
@@ -206,6 +205,30 @@ public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
         this.stop();
     }
 
+
```
pyokagan: No need for this extra newline, one blank line of spacing is sufficient.

src/main/java/seedu/address/MainApp.java
```diff
@@ -206,6 +205,30 @@ public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
...
+     * Parses application command-line parameters.
+     */
+    private AppParameters getAppParameters() {
+        AppParameters appParameters = new AppParameters();
+        Map<String, String> namedParameters = getParameters().getNamed();
+
+        String configPathParameter = namedParameters.get("config");
+        if (configPathParameter != null && !FileUtil.isValidPath(configPathParameter)) {
+            logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
+            configPathParameter = null;
```
pyokagan: I wonder if it is possible to test this code path.

Given that this is in the `MainApp`, the only way this could be done is by implementing a system test. System tests have a whole assortment of problems, though.

So, maybe we can do something like this: Move `AppParameters` into a separate top-level class of its own, like this:

<details>
<summary>Big Diff</summary>

```diff
diff --git a/src/main/java/seedu/address/AppParameters.java b/src/main/java/seedu/address/AppParameters.java
index e69de29..cecc6a4 100644
--- a/src/main/java/seedu/address/AppParameters.java
+++ b/src/main/java/seedu/address/AppParameters.java
@@ -0,0 +1,44 @@
+package seedu.address;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.logging.Logger;
+
+import javafx.application.Application;
+import seedu.address.commons.core.LogsCenter;
+import seedu.address.commons.util.FileUtil;
+
+/**
+ * Represents the parsed command-line parameters given to the application.
+ */
+public class AppParameters {
+    private static final Logger logger = LogsCenter.getLogger(AppParameters.class);
+
+    private Path configPath;
+
+    public Path getConfigPath() {
+        return configPath;
+    }
+
+    public void setConfigPath(Path configPath) {
+        this.configPath = configPath;
+    }
+
+    /**
+     * Parses the application command-line parameters.
+     */
+    public static AppParameters parse(Application.Parameters parameters) {
+        AppParameters appParameters = new AppParameters();
+        Map<String, String> namedParameters = parameters.getNamed();
+
+        String configPathParameter = namedParameters.get("config");
+        if (configPathParameter != null && !FileUtil.isValidPath(configPathParameter)) {
+            logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
+            configPathParameter = null;
+        }
+        appParameters.setConfigPath(configPathParameter != null ? Paths.get(configPathParameter) : null);
+
+        return appParameters;
+    }
+}
diff --git a/src/main/java/seedu/address/MainApp.java b/src/main/java/seedu/address/MainApp.java
index ce5ce34..134b9f3 100644
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -60,8 +60,8 @@ public class MainApp extends Application {
         logger.info("=============================[ Initializing AddressBook ]===========================");
         super.init();
 
-        AppParameters appParameters = getAppParameters();
-        config = initConfig(appParameters.configPath);
+        AppParameters appParameters = AppParameters.parse(getParameters());
+        config = initConfig(appParameters.getConfigPath());
 
         UserPrefsStorage userPrefsStorage = new JsonUserPrefsStorage(config.getUserPrefsFilePath());
         userPrefs = initPrefs(userPrefsStorage);
@@ -205,30 +205,6 @@ public class MainApp extends Application {
         this.stop();
     }
 
-
-    /**
-     * Parses application command-line parameters.
-     */
-    private AppParameters getAppParameters() {
-        AppParameters appParameters = new AppParameters();
-        Map<String, String> namedParameters = getParameters().getNamed();
-
-        String configPathParameter = namedParameters.get("config");
-        if (configPathParameter != null && !FileUtil.isValidPath(configPathParameter)) {
-            logger.warning("Invalid config path " + configPathParameter + ". Using default config path.");
-            configPathParameter = null;
-        }
-        appParameters.configPath = configPathParameter != null ? Paths.get(configPathParameter) : null;
-        return appParameters;
-    }
-
-    /**
-     * Represents the command-line parameters given to the application.
-     */
-    private static class AppParameters {
-        private Path configPath;
-    }
-
     public static void main(String[] args) {
         launch(args);
     }

```

</details>

And you can then create a `AppParametersTest` which can do all the testing we need without requiring a system test.

By the way, if you are feeling creative, you could move this `AppParameters` thing into a preparatory commit :-)

src/main/java/seedu/address/MainApp.java
```diff
@@ -206,6 +205,30 @@ public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
...
+        }
+        appParameters.configPath = configPathParameter != null ? Paths.get(configPathParameter) : null;
+        return appParameters;
```
pyokagan: Add a blank line above this line to separate it from the "configPath parsing" code block.

src/main/java/seedu/address/MainApp.java
vivekscl: Preparatory commit as in, a separate commit after [2/5] for the moving of the `AppParameters` class?

vivekscl:
>Eh, not sure if the whole method needed to be re-written this way.
Some people might complain about "arrowhead code".

Arrowhead code means that there is many nested "if" statements right? But in this case, there's only one "if" statement. So doesn't that mean it's ok to re-write the method this way?

>I don't see File#createNewFile() anywhere in the diff.

In this case, should I mention that it was done in the previous commit instead?

pyokagan:
@vivekscl 
> Arrowhead code means that there is many nested "if" statements right? But in this case, there's only one "if" statement. So doesn't that mean it's ok to re-write the method this way?

You can look up "Make the happy path prominent" in https://nus-cs2103-ay1718s2.github.io/website/book/codeQuality/

> In this case, is it better to mention that it was done in the previous commit instead?

Actually, I don't see how "`Additionally, after the replacement of  File#createNewFile() with Files#createFile(Path), FileUtil#createFile(Path) can be simplified further.`" is even relevant to "changing the return type to `void`".

The commit message claims that it somehow makes the code "simpler", but I don't see how the final result is any simpler at all.

vivekscl:
>Actually, I don't see how "Additionally, after the replacement of File#createNewFile() with Files#createFile(Path), FileUtil#createFile(Path) can be simplified further." is even relevant to "changing the return type to void".

My thought process was that `File#createNewFile()` returned a boolean value, but now that it has been replaced, we have to separately `return true` which isn't required since the return type isn't used. And so that is part of the reason for changing the return type to `void` right? Furthermore, once the return type is changed to `void`, the `if` statement would not have any body and so inverting the condition and placing the file creation process within the `if` statement simplifies the method right?

pyokagan:
@vivekscl 
> My thought process was that File#createNewFile() returned a boolean value, but now that it has been replaced, we have to separately return true which isn't required since the return type isn't used. And so that is part of the reason for changing the return type to void right? 

Yes, that part is fine. I'm not objecting to the change of return type.

>  Furthermore, once the return type is changed to void, the if statement would not have any body and so inverting the condition and placing the file creation process within the if statement simplifies the method right?

In terms of complexity, they are the same.

Note that number of lines of code is just as illusion. If our code style did not require brackets, then it just becomes

```java
if (Files.exists(file))
    return;
createParentDirsOfFile(file);
Files.createFile(file);
```

vs

```java
if (!Files.exists(file)) {
    createParentDirsOfFile(file);
    Files.createFile(file);
}
```

i.e. the same number of lines

And if our code style allows guard clauses to be a single line, then it becomes:

```java
if (Files.exists(file)) return;
createParentDirsOfFile(file);
Files.createFile(file);
```

i.e. 3 lines vs 4

Plus, how about the link I gave you on the "happy path principle"?

vivekscl:
>In terms of complexity, they are the same.
Note that number of lines of code is just as illusion.

Oh, now it's completely clear! Didn't realised that I misunderstood lesser lines of code as more simple :sweat_smile: . And the "happy path principle" really does make it clearer than trying to reduce the number of lines of code for superficial simplicity. Thanks for clearing that up! :smile: 

On a side note, the commit message doesn't require this portion now right?
`Additionally, after the replacement of File#createNewFile() with Files#createFile(Path), FileUtil#createFile(Path) can be simplified further.`

vivekscl:
On a side note, I tried adding more tests for `parse` but I can't pass in a `null` path since the app parameters will not accept any null characters in the input `String`. And since Unix operating systems consider `null` characters as invalid paths, I couldn't test for that an invalid path either.

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,44 @@
...
+import java.nio.file.Paths;
+import java.util.Map;
+import java.util.logging.Logger;
+
+import javafx.application.Application;
+import seedu.address.commons.core.LogsCenter;
+import seedu.address.commons.util.FileUtil;
+
+/**
+* Represents the parsed command-line parameters given to the application.
```
pyokagan: You need a space in front of the `*`

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,44 @@
...
+*/
+public class AppParameters {
+    private static final Logger logger = LogsCenter.getLogger(AppParameters.class);
+
+    private Path configPath;
```
pyokagan: Should be a `String` instead. This commit should just be a simple extraction of code from `MainApp`, without any other changes.

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -18,6 +20,20 @@ public static boolean isFileExists(Path file) {
         return Files.exists(file) && Files.isRegularFile(file);
     }
 
+    /**
+     * Returns true if {@code path} can be converted into a {@code Path} via {@link Paths#get(String)}, otherwise
+     * returns false.
```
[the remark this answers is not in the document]
pyokagan: https://github.com/se-edu/addressbook-level4/pull/866#discussion_r181410642

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,38 @@
...
+    private Application testApp;
+
+    @BeforeClass
+    public static void setupBeforeClass() {
+        SystemTestSetupHelper.initialize();
+    }
+
+    private void setApplicationParameters(String parameter) {
+        try {
+            testApp = FxToolkit.setupApplication(TestApp.class, "--config=" + parameter);
```
pyokagan: There's no need to use TestFX at all. (And preferably we should not use it as much as possible otherwise we get all kinds of issues with GUI testing).

Something like this:

```java

public class AppParametersTest {

    @Test
    public void parse() {
        // valid path
        ParametersStub parametersStub = new ParametersStub();
        parametersStub.namedParameters.put("config", "config.json");

        AppParameters expected = new AppParameters();
        expected.setConfigPath(Paths.get("config.json"));

        assertEquals(expected, AppParameters.parse(parametersStub));
    }

    private static class ParametersStub extends Application.Parameters {
        private Map<String, String> namedParameters = new HashMap<>();

        @Override
        public List<String> getRaw() {
            throw new AssertionError("should not be called");
        }

        @Override
        public List<String> getUnnamed() {
            throw new AssertionError("should not be called");
        }

        @Override
        public Map<String, String> getNamed() {
            return Collections.unmodifiableMap(namedParameters);
        }
    }
}
```

Then you can test for other corner cases like `null` path and such.

src/main/java/seedu/address/AppParameters.java
vivekscl: A simple extraction meaning, I would have to remove the `getAppParameters` from the `MainApp` in this commit as well?

vivekscl: > Not sure  what you mean by that, I don't see any getAppParameters in this PR.

Oh I meant , since it's an extraction, should we remove `MainApp#getApplicationParameters(String)` in this commit?

pyokagan: > Oh I meant , since it's an extraction, should we remove MainApp#getApplicationParameters(String) in this commit?

Assuming you meant `MainApp#getApplicationParameter(String)` (singular), then yes.

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Map;
+
+import org.junit.Test;
+
+import javafx.application.Application;
+
+public class AppParametersTest {
+
+    @Test
+    public void parse() {
```
pyokagan: Try to split this up into separate tests. Otherwise you will have the same problem as in the system tests where a change in a test at the beginning leads to a "snowball" effect, requiring changes in later lines.

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,56 @@
...
+        // valid path
+        ParametersStub parametersStub = new ParametersStub();
+        parametersStub.namedParameters.put("config", "config.json");
+
+        AppParameters expected = new AppParameters();
+        expected.setConfigPath("config.json");
+
+        assertEquals(expected.getConfigPath(), AppParameters.parse(parametersStub).getConfigPath());
```
pyokagan: In the future, besides checking if `getConfigPath()` returns the correct value, we also need to check if the other properties (when added in the future) of the `AppParameters` are unchanged as well.

To prepare for the future, should just be an
```java
assertEquals(expected, AppParameters.parse(parametersStub));
```

You will need to implement a suitable `equals` as well.

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Parses the application command-line parameters.
+     */
+    public static AppParameters parse(Application.Parameters parameters) {
+        AppParameters appParameters = new AppParameters();
+        Map<String, String> namedParameters = parameters.getNamed();
+
+        String configPathParameter = namedParameters.get("config");
+        if (configPathParameter != null && !FileUtil.isValidPath(configPathParameter)) {
```
pyokagan: Since this is pure refactoring, there should not be any behavioral changes. So, there should not be any test for a "valid path" here, to be consistent with the old behavior of just allowing any string.

vivekscl: So you mean that the behavior should be like this:
```java
    public static AppParameters parse(Application.Parameters parameters) {
        AppParameters appParameters = new AppParameters();
        Map<String, String> namedParameters = parameters.getNamed();

        String configPathParameter = namedParameters.get("config");
        appParameters.setConfigPath(configPathParameter);

        return appParameters;
    }
```

or just exactly the same as `MainApp#getApplicationParameter(String)`?
```java
    private String getApplicationParameter(String parameterName) { 
        Map<String, String> applicationParameters = getParameters().getNamed(); 
        return applicationParameters.get(parameterName); 
    } 
```

pyokagan: The first one.

src/test/java/seedu/address/AppParametersTest.java
vivekscl: In this case, there won't be an invalid path since any string is accepted. So we can only test for null path and a valid path right?

pyokagan: Yes, but you can always add it in [3/6] though.

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.logging.Logger;
+
+import javafx.application.Application;
+import seedu.address.commons.core.LogsCenter;
+
+/**
+ * Represents the parsed command-line parameters given to the application.
+ */
+public class AppParameters {
+    private static final Logger logger = LogsCenter.getLogger(AppParameters.class);
```
pyokagan: `logger` not used in this commit, should be added in [3/6] instead.

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,52 @@
...
+            return true;
+        }
+
+        if (!(other instanceof AppParameters)) {
+            return false;
+        }
+
+        AppParameters otherAppParameters = (AppParameters) other;
+        return (otherAppParameters.getConfigPath() == null && this.getConfigPath() == null)
+                || otherAppParameters.getConfigPath().equals(this.getConfigPath());
```
pyokagan: `Objects.equals(otherAppParameters.getConfigPath(), getConfigPath())`

`AppParameters#hashCode()` also needs to be implemented.

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import javafx.application.Application;
+
+public class AppParametersTest {
+
+    private ParametersStub parametersStub = new ParametersStub();
+    private AppParameters expected = new AppParameters();
+
+    @Test
+    public void parse_validParameters_success() {
```
pyokagan: `parse_validConfigPath_success`

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,51 @@
...
+        parametersStub.namedParameters.put("config", "config.json");
+        expected.setConfigPath("config.json");
+        assertEquals(expected, AppParameters.parse(parametersStub));
+    }
+
+    @Test
+    public void parse_nullPath_success() {
```
pyokagan: `parse_nullConfigPath_success`

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,51 @@
...
+        parametersStub.namedParameters.put("config", null);
+        expected.setConfigPath(null);
```
pyokagan: Not sure if we need to explicitly set the configPath, or we can just leave it at the default? :thinking: 

I think we should leave it as the default, since a value of `null` means "not present", which means the `AppParameters` should be left untouched?

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -30,6 +31,13 @@ public void parse_nullPath_success() {
         assertEquals(expected, AppParameters.parse(parametersStub));
     }
 
+    @Test
+    public void parse_invalidPath_success() {
```
pyokagan: `parse_invalidConfigPath_success`

vivekscl:
@pyokagan a gentle reminder to review :smile:

pyokagan:
@vivekscl Sorry, you'll need to wait until the weekend. In the meantime you can do a self-review or ask others to review.

vivekscl:
>Sorry, you'll need to wait until the weekend. In the meantime you can do a self-review or ask others to review.

Alright, I understand. Is it alright if ask Prof for a review, or is that allowed only after approval from others?

pyokagan:
@vivekscl Any of the senior devs or @damithc can merge it if they are confident enough in the PR :-) (approval == reviewer vouches for the PR)

vivekscl:
>Any of the senior devs or @damithc can merge it if they are confident enough in the PR :-) (approval == reviewer vouches for the PR)

Alright, understood! @Zhiyuan-Amos @yamgent @damithc, if anyone is free, maybe take a look? :smile:

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,54 @@
...
+            return false;
+        }
+
+        AppParameters otherAppParameters = (AppParameters) other;
+        return Objects.equals(otherAppParameters.getConfigPath(), getConfigPath());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(configPath);
```
Zhiyuan-Amos: Since there's only 1 field, should we do `configPath.hashCode()` instead?

src/test/java/seedu/address/AppParametersTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+
+import org.junit.Test;
+
+import javafx.application.Application;
+
+public class AppParametersTest {
+
+    private ParametersStub parametersStub = new ParametersStub();
+    private AppParameters expected = new AppParameters();
```
Zhiyuan-Amos: we can have `final` for these fields

src/main/java/seedu/address/commons/util/FileUtil.java
```diff
@@ -18,6 +20,20 @@ public static boolean isFileExists(Path file) {
         return Files.exists(file) && Files.isRegularFile(file);
     }
 
+    /**
+     * Returns true if {@code path} can be converted into a {@code Path} via {@link Paths#get(String)},
+     * otherwise returns false.
+     * @param path A String representing the file path. Cannot be null.
```
Zhiyuan-Amos: `String` should either be `string` or `{@code String}`? :P

yamgent:
> [2/6] Should we set AppParameters constructor to be private since this class is only instantiated through parse(Application.Parameters)?
> 
> Also, I'm wondering whether this class should be immutable (i.e. no need for setter method) since there's no reason for us to set the config path right?

The setter method is used in the test, and there's no other way to create this object with the current implementation. So perhaps there can be a `AppParameters(string)` constructor as well?

src/main/java/seedu/address/AppParameters.java
```diff
@@ -0,0 +1,54 @@
...
+        if (other == this) {
+            return true;
+        }
+
+        if (!(other instanceof AppParameters)) {
+            return false;
+        }
+
+        AppParameters otherAppParameters = (AppParameters) other;
+        return Objects.equals(otherAppParameters.getConfigPath(), getConfigPath());
```
pyokagan: For most of the `equals()` in our code base, the order is "this object", then the "other object", so this should be:
```java
return Objects.equals(getConfigPath(), otherAppParameters.getConfigPath());
```

vivekscl:
>[2/6] Should we set AppParameters constructor to be private since this class is only instantiated through parse(Application.Parameters)?
Also, I'm wondering whether this class should be immutable (i.e. no need for setter method) since there's no reason for us to set the config path right?

>The setter method is used in the test, and there's no other way to create this object with the current implementation. So perhaps there can be a AppParameters(string) constructor as well?

So essentially, there's no need for a setter method and instead we can replace it with a public constructor that takes in a string?

pyokagan:
@yamgent 
> The setter method is used in the test, and there's no other way to create this object with the current implementation. So perhaps there can be a AppParameters(string) constructor as well?

This will make tests harder to write in the future when we start introducing additional optional configuration properties, since you can't easily create an `AppParameters` object with all default property values EXCEPT the value that is being tested.

Zhiyuan-Amos:
@pyokagan 

> Should we set AppParameters constructor to be private since this class is only instantiated through parse(Application.Parameters)?

> Also, I'm wondering whether this class should be immutable (i.e. no need for setter method) since there's no reason for us to set the config path right?

Do you have any thoughts on this? Implementing `AppParameters` this way will definitely make tests way harder to write as mentioned by @yamgent. 2 alternatives that I've thought of:
1. Reflection.
2. Instead of comparing equality between 2 `AppParameters` objects, we compare the equality of the `configPath` only. This is a weaker test though.

vivekscl:
>Instead of comparing equality between 2 AppParameters objects, we compare the equality of the configPath only. This is a weaker test though.

Just my thoughts on this:
Is there a need to make `AppParameters` immutable? In cases where `AppParameters#parse` is called multiple times, the values get modified so it's essentially supposed to be mutable? :thinking:

pyokagan:
> This will make tests harder to write in the future when we start introducing additional optional configuration properties, since you can't easily create an AppParameters object with all default property values EXCEPT the value that is being tested.

Sorry, I realized that @yamgent was actually saying something like my message.

To make it clear: I'm -1 on making it immutable. It provides no benefit, and just makes tests harder to write.

pyokagan:
@Zhiyuan-Amos What problem are you trying to solve by making it immutable?

pyokagan:
> Sorry, I realized that @yamgent was actually saying something like my message.

Whoops, don't think this statement is right either.

Anyway:

* Making `AppParameters` immutable -- Disagree. No benefit since only one `AppParameters` will be kept around, makes tests harder to write (you'll need to write a `AppParametersBuilder` to restore sanity, but this is going into [FizzBuzzEnterpriseEdition](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) territory. Just keep it simple)

* No default constructor for `AppParameters`, require callers to use `AppParameters(String configPath)` constructor. If we have more parameters in the future the constructor will be extended to have more arguments. -- Disagree. This will make tests much more verbose when we start introducing additional optional configuration properties.

vivekscl:
Merge commit message:
```
String operations are used for the construction of file paths. Also,
many APIs in the codebase take in String arguments as file paths.

This method of file/path handling is error prone as creating paths using
string concatenation can lead to the creation of invalid paths.
Furthermore, checking if two paths resolve to the same file using string
equality is incorrect as it only tests if the strings are equal.

Let's refactor the codebase to use java.nio.file for file/path
handling as it provides a safe and platform-independent way to
perform operations on paths and properly separate the concerns of
parsing/stringifying paths and manipulating paths.

More information on why String operations are error-prone:
http://twistedoakstudios.com/blog/Post4872_dont-treat-paths-like-strings
```

pyokagan:
@yamgent Oh that's bad, thanks for the sharp eyes.

vivekscl:
>Commit message contains the messages for both [2/6] and [3/6].

@yamgent thanks for catching that! Really bad mistake on my part :sweat_smile: .

yamgent:
@vivekscl you can last call review this.

vivekscl:
@yamgent I guess this is ready to merge? :smile:

PR #868, opened 2018-03-29
build.gradle: remove unused testfx-legacy dependency

Rinder5:
`Testfx-legacy is depreciated and no longer supported. We use
testfx-core version 4.0.12-alpha, which is past the supported
testfx-legacy version of 4.0.8-alpha, and breaks it, according to
https://github.com/TestFX/TestFX#testfx-legacy-deprecated.`

`Let's remove the unused testfx-legacy dependency.`


yamidark:
This is already being handled in #860 [11/12], so should not need to repeat this.

Rinder5:
yup, looks good

PR #869, opened 2018-03-29
build.gradle: remove field with default value #857

### LINKED ISSUE #857: build.gradle: remove field with default value
https://github.com/se-edu/addressbook-level4/blob/a4ef5229e29bb1da91d462a650ee82c243ab1b7c/gradle.properties#L1

The above line is a default value as seen in [Gradle documentation](https://docs.gradle.org/4.6/userguide/gradle_daemon.html).

It is unnecessary to specify default values.

Let's remove it.

Rinder5:
fixes #857

The line `org.gradle.daemon=true` in gradle.properties is a default
value as seen in
https://docs.gradle.org/4.6/userguide/gradle_daemon.html

It is unnecessary to specify default values.

Let's remove it.


Rinder5:
will do a larger PR with more insignificant changes, as per discussion in https://github.com/se-edu/addressbook-level4/pull/870

PR #870, opened 2018-03-29
Clean up "d.FirstTimers" issues

### LINKED ISSUE #245: Rename references to the "last shown list"
The code base (namely, the code in the `logic.commands` package) contains references to the `lastShownList` and `last person listing`.

This is a relic of the AddressBook-Level3 code, which did have a "last shown list". However, in Level 4, we use a "filtered person list" instead.

All references to `lastShownList` and `last person listing` should be renamed into their level 4 equivalents names (`filteredPersonList` and `displayed person list` ?)

Documentation as well.

### LINKED ISSUE #397: Person's field classes: Remove the starting word "Person" from the error messages
The invalid values error message for each person's field starts with a redundant "Person" word. Remove that word in the error messages.

---

* Address: `"Person addresses can take any values, and it should not be blank"`
  * Should be: `"Addresses can take any values, and it should not be blank"`
* Email: `"Person emails should be 2 alphanumeric/period strings separated by '@'"`
  * Should be: `"Emails should be 2 alphanumeric/period strings separated by '@"`
* Name: `"Person names should only contain alphanumeric characters and spaces, and it should not be blank"`
  * Should be: `"Names should only contain alphanumeric characters and spaces, and it should not be blank"`
* Phone: `"Person phone numbers should only contain numbers"`
  * Should be: `"Phone numbers should only contain numbers"`

### LINKED ISSUE #450: docs/TestDocument.adoc is no longer needed
`docs/TestDocument.adoc` was used to test whether the asciidoctor workflow works, at a time when all our documentations were still in markdown.

Now that we have two files in asciidoctor (namely `DeveloperGuide.adoc` and `UserGuide.adoc`), `docs/TestDocument.adoc` is now redundant.

Let's remove it.

### LINKED ISSUE #484: FindCommand's help message should specify that it is case-insensitive
Find command allows searching for keywords without case-sensitivity.

Fix the header comment and MESSAGE_USAGE to reflect that behavior.

```diff
-            + "the specified keywords (case-sensitive) and displays them as a list with index numbers.\n"
+            + "the specified keywords (case-insensitive) and displays them as a list with index numbers.\n"
```

### LINKED ISSUE #521: StatusBarFooter & CommandBox: Change styleClass
Both fxml files use `StackPane` as the root node. However, both use 'anchor-pane' as their `styleClass`. 

As 'anchor-pane' style class is not used elsewhere in the code base, let's rename it to 'stack-pane', and update the naming accordingly in StatusBarFooter & CommandBox.

### LINKED ISSUE #565: CommandBoxHandle#run() should not return any values
Because no one is using return value (YAGNI)

### LINKED ISSUE #629: CommandBox: rename method handleCommandInputChanged()
`handleCommandInputChanged()` method name sounds misleading since this event only triggers when user press the "Enter" button and not every time when user types something. We should rename it to something that match what it is doing. E.g. `handleCommandEntered()`.

### LINKED ISSUE #637: Command: Remove getMessageForPersonListShownSummary(int)
Command: Move getMessageForPersonListShownSummary(int)

This method is only called by FindCommand.

There is no need for this method to be placed in the super class
Command as other commands do not use this method.

Let's move getMessageForPersonListShownSummary(int) to FindCommand as a
private method.

### LINKED ISSUE #663: UniquePersonList: update asObservableList() method name
`asObservableList()` returns an unmodifiable `ObservableList`.

It is not very clear from the method name that we are returning an
unmodifiable `ObservableList`.

Let's update the method name to `asUnmodifiableObservableList()`.

### LINKED ISSUE #822: typicalPersonsAddressBook.xml: remove @code tag for header comment

### LINKED ISSUE #857: build.gradle: remove field with default value

### LINKED ISSUE #858: CollectionUtilTest: remove usage of deprecated API
> `CollectionUtilTest.java uses or overrides a deprecated API.`

Upon compiling the test code, the above message is displayed. Upon inspection, the `new Integer(x)` used by the test is causing this warning to occur.

Let's replace all instances of `new Integer(x)` with `x`.

Rinder5:
fixes #857
fixes #822
fixes #245
fixes #397
fixes #450
fixes #858
fixes #637
fixes #629
fixes #565
fixes #663
fixes #484
fixes #521

```
We have 37 open issues labeled "d.FirstTimers". This is 37% of our 
open issues.

Many of these issues do not serve as helpful issues for new 
contributors to start on. They usually involve renaming something or 
deleting something small. Also, the way to do it is completely stated 
in the issue itself, meaning that the new contributor would only need 
to follow the instructions given to complete it.

The issue is that many of these issues are unattractive. As a result, 
many of these issues are unresolved, and results in many small 
problems with the code base which can easily be corrected (as 
we end up waiting for a new first timer to fix it).

Let's close some.
```


yamgent:
@Rinder5 this PR is fixing a `FirstTimer` issue, did you get permission from Prof to fix them? :P

Rinder5:
We have too many first timer issues! =0

pyokagan:
Yes, I do think some of the first timer issues are long overdue to be resolved: e.g. #450 

However, do group all of them into one PR, and resolve at least 6 first-timer issues in one single PR. Thanks.

Rinder5:
kk, will do that then.

any preference? if not I'll try to remove the most braindead ones, to let new developers at least have to think a bit

pyokagan:
Yeah you can start with the "braindead" ones first. After that, you can move on to solving the rest if you want as well. I can always generate more first timer issues if needed.

src/test/data/XmlSerializableAddressBookTest/typicalPersonsAddressBook.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <!-- AddressBook save file which contains the same Person values as in
-{@code TypicalPersons#getTypicalAddressBook()}-->
+{TypicalPersons#getTypicalAddressBook()}-->
```
pyokagan: Maybe the braces `{}` need to be removed as well?

Actually, could this line be joined with the previous line?

src/test/java/seedu/address/commons/util/CollectionUtilTest.java
```diff
@@ -101,8 +101,8 @@ public void elementsAreUnique() {
         assertNotUnique("abc", "abc");
         assertNotUnique("abc", "", "abc", "ABC");
         assertNotUnique("", "abc", "a", "abc");
-        assertNotUnique(1, new Integer(1));
```
pyokagan: Eh, I actually think the author wanted to test whether `CollectionUtil#elementsAreUnique()` works when primitive values and object values are mixed together.

I think the correct fix would be to use `Integer.valueOf()` instead.

pyokagan:
@Rinder5 
> They usually involve renaming something or deleting something small. Also, the way to do it is completely stated in the issue itself, meaning that the new contributor would only need to follow the instructions given to complete it. 

Actually, that is precisely the point of First Timer issues. They are meant to get contributors acquainted with our contribution practices, not to test their competency. It's not a competition.

The only problem is that some first timer issues are not "attractive", and so no one does them. For those issues, I think it's fine for someone to clear the backlog once in awhile.

src/main/java/seedu/address/model/person/Phone.java
```diff
@@ -11,7 +11,7 @@
 
 
     public static final String MESSAGE_PHONE_CONSTRAINTS =
-            "Phone numbers can only contain numbers, and should be at least 3 digits long";
+            "Phone numbers should only contain numbers, and it should be at least 3 digits long";
```
vivekscl: Not sure if this part of #397. Seems like an issue on its own, so I don't think this should be part of this commit.

vivekscl:
>Also, since FindCommand#getMessageForPersonListShownSummary(int) is used only within its own class and instance, maybe we canr emove the static and get the method to call model.getFilteredPersonList().size() to get the size by itself?

Sounds good :+1:

Rinder5:
@vivekscl I removed the commit for removing file.exists. Thanks!

src/main/java/seedu/address/logic/commands/FindCommand.java
```diff
@@ -21,10 +22,19 @@ public FindCommand(NameContainsKeywordsPredicate predicate) {
         this.predicate = predicate;
     }
 
+    /**
+     * Constructs a feedback message to summarise an operation that displayed a listing of persons.
+     *
+     * @return summary message for persons displayed
```
Zhiyuan-Amos: `@return` isn't necessary, we can join both sentences together: `Returns the summary message...`

Zhiyuan-Amos: I wonder why not let's just inline this method though, since it's just 1 line :P

Zhiyuan-Amos:
@Rinder5 

image

Can you update the README file as well `JUnit 4 -> JUnit 5`? :) Thanks!

Zhiyuan-Amos:
@vivekscl having another look at this?

@Rinder5 
> [v7 9/15] ~~Command: inline getMessageForPersonListShownSummary(int)~~
Inline Command#getMessageForPersonListShownSummary(int)
[v7 10/15] ~~CommandBox: rename method handleCommandInputChanged()~~
Rename CommandBox#handleCommandInputChanged()

PR #875, opened 2018-04-04
Dev guide: checkstyle setup: require to apply version change #715

### LINKED ISSUE #715: Dev guide: checkstyle setup: require to apply version change
image
It turns out one needs to click `Apply` after selecting the checkstyle version, before choosing the checkstyle file. Otherwise the plugin will use the previous version to validate the config file and throw an error.

Rinder5:
fixes #715

```
UsingCheckstyle.adoc: update documentation to include clicking `Apply`

We do not specify that clicking `Apply` is required after selecting the
Checkstyle version before choosing the checkstyle file. If not clicked,
Checkstyle will use the previous version to validate the config file
and throw an error.

Let's update UsingCheckstyle.adoc to include clicking `Apply`
```


yamidark:
We should update the troubleshooting section with the note too, so that students that somehow missed the pressing `Apply` note earlier can find the solution more easily.

untitled

PR #880, opened 2018-04-16
HelpWindowTest: fix failing test in non-headless mode

yamidark:
```
HelpWindowTest#focus_helpWindowNotFocused_focused() test checks that
the HelpWindow#focus() method will cause the HelpWindow to be in focus
when called. The test first asserts that the HelpWindow is not in focus
first after it is shown, being calling this method and asserting that
the HelpWindow is now in focus.

When tests are run in non-headless mode, the HelpWindow will be in
focus immediately after it is shown, thus our first assertion is
incorrect, causing the test to consistently fail in non-headless mode.

Let's add a method GuiRobot#removeFocus(), and update the test to call
this method to remove focus from the HelpWindow after it is shown
to ensure the first assertion is correct in non-headless mode.
```

### DIFF
--- a/src/test/java/guitests/GuiRobot.java
+++ b/src/test/java/guitests/GuiRobot.java
@@ -40,6 +40,13 @@ public class GuiRobot extends FxRobot {
     }
 
     /**
+     * Returns true if tests are run in headless mode.
+     */
+    public boolean isHeadlessMode() {
+        return isHeadlessMode;
+    }
+
+    /**
      * Waits for {@code event} to be true by {@code DEFAULT_WAIT_FOR_EVENT_TIMEOUT_MILLISECONDS} milliseconds.
      *
      * @throws EventTimeoutException if the time taken exceeds {@code DEFAULT_WAIT_FOR_EVENT_TIMEOUT_MILLISECONDS}
--- a/src/test/java/seedu/address/ui/HelpWindowTest.java
+++ b/src/test/java/seedu/address/ui/HelpWindowTest.java
@@ -46,8 +46,11 @@ public class HelpWindowTest extends GuiUnitTest {
     }
 
     @Test
-    public void focus_helpWindowNotFocused_focused() {
+    public void focus_helpWindowNotFocused_focused() throws Exception {
         guiRobot.interact(helpWindow::show);
+
+        // Focus on another stage to remove focus from the helpWindow
+        FxToolkit.setupStage(Stage::requestFocus);
         assertFalse(helpWindow.getRoot().isFocused());
 
         guiRobot.interact(helpWindow::focus);
--- a/src/test/java/seedu/address/ui/HelpWindowTest.java
+++ b/src/test/java/seedu/address/ui/HelpWindowTest.java
@@ -3,6 +3,7 @@ package seedu.address.ui;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeFalse;
 import static seedu.address.ui.HelpWindow.USERGUIDE_FILE_PATH;
 
 import java.net.URL;
@@ -47,6 +48,7 @@ public class HelpWindowTest extends GuiUnitTest {
 
     @Test
     public void focus_helpWindowNotFocused_focused() throws Exception {
+        assumeFalse("Test skipped in headless mode: Window focus behavior is buggy.", guiRobot.isHeadlessMode());
         guiRobot.interact(helpWindow::show);
 
         // Focus on another stage to remove focus from the helpWindow
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,8 @@
 // For more details take a look at the Java Quickstart chapter in the Gradle
 // user guide available at http://gradle.org/docs/4.6/userguide/tutorial_java_projects.html
 
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
 plugins {
     id 'java'
     id 'jacoco'
@@ -117,12 +119,14 @@ allTests.dependsOn nonGuiTests
 test {
     systemProperty 'testfx.setup.timeout', '60000'
 
-    // Prints the currently running test's name in the CI's build log,
-    // so that we can check if tests are being silently skipped or
-    // stalling the build.
-    if (System.env.'CI') {
-        beforeTest { descriptor ->
-            logger.lifecycle("Running test: ${descriptor}")
+    testLogging {
+        events TestLogEvent.FAILED, TestLogEvent.SKIPPED
+
+        // Prints the currently running test's name in the CI's build log,
+        // so that we can check if tests are being silently skipped or
+        // stalling the build.
+        if (System.env.'CI') {
+            events << TestLogEvent.STARTED
         }
     }
 


yamidark:
@vivekscl @Rinder5 for your review.

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +40,13 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Removes focus from the current stage.
+     */
+    public void removeFocus() throws Exception {
+        FxToolkit.setupStage(dummyStage -> dummyStage.requestFocus());
```
vivekscl: Should there be a try-catch block instead of simply throwing the exception?

yamidark: Hmm, do you mean something like what `UiPartRule` does:

```java
catch (TimeoutException te) {
       throw new AssertionError("Timeout should not happen.", te);
}
```

yamidark:
Actually, should I add in the comments that the `removeFocus()` does not work in `headless` mode due to an issue with [Monocle](https://github.com/javafxports/openjdk-jfx/issues/50), and make the test `return` immediately if in `headless` mode?

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +40,13 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Removes focus from the current stage.
```
pyokagan: A more accurate description would be "focus on the registered TestFX stage". If the "current stage" is the registered TestFX stage, then this won't "remove focus" from it.

With that said, I think this is too oddly specific to be inside GuiRobot. We don't usually "remove focus", we just focus on another window instead.

Just inline this directly in the test?

pyokagan:
@yamidark 
> Actually, should I add in the comments that the removeFocus() does not work in headless mode due to an issue with Monocle, and make the test method return immediately if in headless mode?

Actually, I'm not sure why the test is even run in headless mode? For headless mode to be useful its results need to match the real-world head-full mode, otherwise its test results are meaningless since the fact that a test passes/fails in headless mode does not mean that it would pass/fail in the real-world.

If Monocle's window focusing behavior does not match the "real world" then there is no point trying to run focus tests with it. Better to be explicit about that by printing a gigantic "TEST SKIPPED IN HEADLESS MODE, RUN IT IN HEAD-FULL MODE INSTEAD" message.

pyokagan:
> If Monocle's window focusing behavior does not match the "real world" then there is no point trying to run focus tests with it.

Actually I take that back. There's still some value in the test if it fails if `requestFocus` is not called for whatever reason.

src/test/java/guitests/GuiRobot.java
vivekscl: Yup!

yamidark: Since I will be inling the method directly into the test, should we still do a `try-catch` block in the test?

pyokagan:
@yamidark 

Sorry, I re-assert again that the test should not be run in headless mode :-). Its relying on undefined behavior, and I think that is part of the reason why `HelpWindowTest` is stalling our builds.

Consider this: (Make sure to run it in **headless** mode)

```java
package seedu.address.ui;

import static org.junit.Assert.assertTrue;

import org.junit.Test;
import org.testfx.api.FxRobot;
import org.testfx.api.FxToolkit;

import javafx.scene.Scene;
import javafx.scene.layout.Pane;
import javafx.stage.Stage;

public class HelpWindowTest {
    Stage stage1;
    Stage stage2;

    @Test
    public void test() throws Exception {
        FxRobot fxRobot = new FxRobot();
        FxToolkit.registerPrimaryStage();

        fxRobot.interact(() -> {
            stage1 = new Stage();
            stage1.setScene(new Scene(new Pane()));

            stage2 = new Stage();
            stage2.setScene(new Scene(new Pane()));
        });

        fxRobot.interact(stage1::show);
        assertTrue(stage1.isFocused()); // NOTE: stage1 is focused once it is shown!!!
    }
}
```

Note that upon calling `stage1::show`, `stage1` is immediately focused.

Now, consider the following case where **`stage1` and `stage2` share the same scene**:

```java
package seedu.address.ui;

import static org.junit.Assert.assertTrue;

import org.junit.Test;
import org.testfx.api.FxRobot;
import org.testfx.api.FxToolkit;

import javafx.scene.Scene;
import javafx.scene.layout.Pane;
import javafx.stage.Stage;

public class HelpWindowTest {
    Stage stage1;
    Stage stage2;

    @Test
    public void test() throws Exception {
        FxRobot fxRobot = new FxRobot();
        FxToolkit.registerPrimaryStage();

        fxRobot.interact(() -> {
            Scene scene = new Scene(new Pane());

            stage1 = new Stage();
            stage1.setScene(scene);

            stage2 = new Stage();
            stage2.setScene(scene);
        });

        fxRobot.interact(stage1::show);
        assertTrue(stage1.isFocused()); // ASSERTION FAILURE!!!!
    }
}
```

`stage1` is now not automatically focused when it is shown.

So, on normal operation in headless mode stages are supposed to be focused automatically when they are shown. But when stages share a Scene (which the JavaFX docs specifically say not to), then the focus does not occur.

I think this is exactly what is happening in `HelpWindowTest`. Specifically this:

```java
Stage helpWindowStage = FxToolkit.setupStage((stage) -> stage.setScene(helpWindow.getRoot().getScene()));
```

I also believe that this is tied to `HelpWindowTest` randomly stalling our builds, since from my investigation it stalls due to a deadlock between the Application thread and the renderer thread because the Application thread miscounts the number of scenes.

yamidark:
@pyokagan Should I also make the other tests in `HelpWindowTest` also skip in headless mode, since they also seem to have undefined behaviour as well?

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -46,8 +49,16 @@ public void isShowing_helpWindowIsHiding_returnsFalse() {
     }
 
     @Test
-    public void focus_helpWindowNotFocused_focused() {
+    public void focus_helpWindowNotFocused_focused() throws Exception {
+        if (guiRobot.isHeadlessMode()) {
```
pyokagan: Any way of properly telling JUnit to skip the test? (So that it is reported as a skipped test rather than a test that succeeded)

yamidark: Will look up on this.

Zhiyuan-Amos: You can use [Assumption](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/Assumptions.html#assumingThat-boolean-org.junit.jupiter.api.function.Executable-)

Zhiyuan-Amos: Hmm that's weird, cos in the docs:

> In direct contrast to failed assertions, failed assumptions do not result in a test failure; rather, a failed assumption results in a test being aborted.

Zhiyuan-Amos: I'm referring to the org.junit.jupiter.api.Assumptions though, see the docs I linked above :P

yamidark: >I'm referring to the org.junit.jupiter.api.Assumptions though, see the docs I linked above :P

@_@, ok maybe I confuse with the other links I had xD.
Anyway, I can't seem to get conditional execution to work on my side:
```
@DisabledIf("guiRobot.isHeadlessMode()")
```
Should I just stick with `org.junit.Assume`? It doesn't seem to inform us when we skip the test though.

pyokagan: You can do something like this to print the names of skipped tests:

```diff
diff --git a/build.gradle b/build.gradle
index 75a518e..5c0329e 100644
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,8 @@
 // For more details take a look at the Java Quickstart chapter in the Gradle
 // user guide available at http://gradle.org/docs/4.6/userguide/tutorial_java_projects.html
 
+import org.gradle.api.tasks.testing.logging.TestLogEvent
+
 plugins {
     id 'java'
     id 'jacoco'
@@ -117,12 +119,14 @@ allTests.dependsOn nonGuiTests
 test {
     systemProperty 'testfx.setup.timeout', '60000'
 
-    // Prints the currently running test's name in the CI's build log,
-    // so that we can check if tests are being silently skipped or
-    // stalling the build.
-    if (System.env.'CI') {
-        beforeTest { descriptor ->
-            logger.lifecycle("Running test: ${descriptor}")
+    testLogging {
+        events TestLogEvent.FAILED, TestLogEvent.SKIPPED
+
+        // Prints the currently running test's name in the CI's build log,
+        // so that we can check if tests are being silently skipped or
+        // stalling the build.
+        if (System.env.'CI') {
+            events << TestLogEvent.STARTED
         }
     }
 

```

Still haven't figured out how to print the reason for test abortion though.

src/test/java/seedu/address/ui/HelpWindowTest.java
```diff
@@ -46,8 +47,12 @@ public void isShowing_helpWindowIsHiding_returnsFalse() {
     }
 
     @Test
-    public void focus_helpWindowNotFocused_focused() {
+    public void focus_helpWindowNotFocused_focused() throws Exception {
+        assumeFalse(guiRobot.isHeadlessMode());
```
pyokagan: Should use `assumeFalse(String message, boolean b)` to provide a reason for why the test is aborted.

src/test/java/guitests/GuiRobot.java
```diff
@@ -39,6 +39,13 @@ public void pauseForHuman() {
         sleep(PAUSE_FOR_HUMAN_DELAY_MILLISECONDS);
     }
 
+    /**
+     * Returns true if tests are run in headless mode.
+     */
+    public boolean isHeadlessMode() {
```
pyokagan: This is not used in the commit at all.

pyokagan:
@yamidark Please submit another version because your PR has been modified.

PR #893, opened 2018-08-06
parsers: throw ParseException consistently

pyokagan:
```
Some parsers in the `parsers` package throw ParseException on a parse
error, others throw IllegalValueException.

Given that they are *all* parsers, let's make them all throw
ParseException for consistency.
```

### DIFF
--- a/src/main/java/seedu/address/logic/commands/AddCommand.java
+++ b/src/main/java/seedu/address/logic/commands/AddCommand.java
@@ -54,7 +54,7 @@ public class AddCommand extends Command {
             model.commitAddressBook();
             return new CommandResult(String.format(MESSAGE_SUCCESS, toAdd));
         } catch (DuplicatePersonException e) {
-            throw new CommandException(MESSAGE_DUPLICATE_PERSON);
+            throw new CommandException(MESSAGE_DUPLICATE_PERSON, e);
         }
 
     }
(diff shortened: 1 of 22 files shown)


pyokagan:
Coverage down due to the `ParseException(String message, Throwable cause)` not being used any more, which is strange because there are definitely places in the code base where we catch an exception and then throw a `ParseException` in response.

pyokagan:
> which is strange because there are definitely places in the code base where we catch an exception and then throw a ParseException in response.

Basically a long-winded way of saying that the code base is not using chained exceptions consistently, which needs to be fixed.

pyokagan:
@Zhiyuan-Amos 
> Just one question for [3/3], in Assert#assertThrows(Class<? extends Throwable>, String, VoidCallable) should AssertionFailedError be thrown with the cause actualException as well?

Yes, you are right.

`AssertionFailedError` does not have a constructor with a `cause` argument though, so I guess I'll need to find a reason to make it use `AssertionError` instead.

pyokagan:
From the range-diff (new feature cooking in upstream Git):

```
1:  2ae8dddb4 = 1:  2ae8dddb4 ParserUtil methods: throw ParseExceptions on invalid input
2:  c348aa969 = 2:  c348aa969 parsers: remove unnecessary catch-throws of ParseExceptions
-:  --------- > 3:  d7b3f1cc0 Remove use of `junit.framework` package
3:  90291f105 ! 4:  69728c1f2 Use chained exceptions consistently throughout the code base
```

Tells me that [v1 1/3] corresponds with [v2 1/4] with no changes, [v1 2/3] corresponds with [v2 2/4] with no changes, [v2 3/4] is a new commit, and [v1 3/3] corresponds with [v2 4/4] with changes.

PR #894, opened 2018-08-06
Replace uses of `org.junit.Assert.fail` with `throw new AssertionError(...)`

pyokagan:


src/test/java/seedu/address/commons/util/CollectionUtilTest.java
```diff
@@ -102,7 +101,7 @@ private void assertNullPointerExceptionThrown(Object... objects) {
     private void assertNullPointerExceptionThrown(Collection<?> collection) {
         try {
             requireAllNonNull(collection);
-            fail("The expected NullPointerException was not thrown");
+            throw new AssertionError("The expected NullPointerException was not thrown");
```
eugenepeh: Missing period. Though its not there in the first place but might as well add it back in, since going to modify this line, for consistency sake.

PR #895, opened 2018-08-07
ParserUtil: remove `Optional` variants of parse methods

pyokagan:


pyokagan:
What empty check?

eugenepeh:
>What empty check?

We have null check, invalid check, valid check but missing empty string check.

pyokagan:
> We have null check, invalid check, valid check but missing empty string check.

That's a probable deficiency with `ParserUtil` (although I think it is because all of the corner cases are already tested in `NameTest`, `EmailTest` etc.), but it's not the focus of this PR.

src/main/java/seedu/address/logic/parser/EditCommandParser.java
```diff
@@ -43,10 +43,18 @@ public EditCommand parse(String args) throws ParseException {
         }
 
         EditPersonDescriptor editPersonDescriptor = new EditPersonDescriptor();
-        ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME)).ifPresent(editPersonDescriptor::setName);
-        ParserUtil.parsePhone(argMultimap.getValue(PREFIX_PHONE)).ifPresent(editPersonDescriptor::setPhone);
-        ParserUtil.parseEmail(argMultimap.getValue(PREFIX_EMAIL)).ifPresent(editPersonDescriptor::setEmail);
-        ParserUtil.parseAddress(argMultimap.getValue(PREFIX_ADDRESS)).ifPresent(editPersonDescriptor::setAddress);
+        if (argMultimap.getValue(PREFIX_NAME).isPresent()) {
+            editPersonDescriptor.setName(ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME).get()));
```
Zhiyuan-Amos: Do you think the code will be more readable if we instead do:
`editPersonDescriptor.setName(ParserUtil.parseName(argMultimap.getValue(PREFIX_NAME).orElse(null)));`?

In this case, `editPersonDescriptor` allows us to set null values (to indicate that the particular field won't be edited).

Issue #871, opened 2018-03-31
Java 9: use Factory Methods to Create Immutable Collections


Zhiyuan-Amos:
We can convert method calls like `Arrays#asList(...)` to use `List#of(...)`. https://dzone.com/articles/java-9-factory-methods-to-create-immutable-collect

As this is likely to be a mechanical change, do take note to list down the steps that you've taken to resolve this issue. Refer to our [coding conventions](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-organization) for more details.

PR #896, opened 2018-08-07
model: convert checked exceptions to runtime exceptions

pyokagan:
```
DuplicatePersonException and PersonNotFoundException are checked
exceptions. They are thrown by methods to signify that their relevant
preconditions were violated.

Using checked exceptions for such a purpose is slightly useful in that
it will force callers to handle the case where the above preconditions
are not met, such as when the methods are called with invalid user
input.

However, it imposes a HUGE cost on callers where the preconditions are
already known to be met (e.g. in test code, or when the user input has
already been validated before hand). In such a case, callers are forced
to add try-catch blocks around the method call even if they know that
the exception will never be thrown, bloating up the code. It is also
impossible to test the catch blocks as well since correct code will
ensure that the precondition holds and thus the exception will never be
thrown, leading to reduced code coverage.

Checked exceptions also don't work very well with the Java Streams API,
since the API doesn't accept lambdas which could throw checked
exceptions.

In AB-4, the amount of code which benefits from DuplicatePersonException
and PersonNotFoundException being checked exceptions is much smaller
than the amount of code which is negatively impacted.

As such, let's make the tradeoff in the other direction, by making
DuplicatePersonException and PersonNotFoundException runtime exceptions.
New callers _could_ forget to check that the preconditions hold before
calling the methods in question (although test cases should catch that),
but this is balanced out by the huge benefit of having more concise and
testable code.
```


src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -110,13 +110,12 @@ public ReadOnlyAddressBook getAddressBook() {
         }
 
         @Override
-        public void deletePerson(Person target) throws PersonNotFoundException {
```
yamidark: [1/10] `PersonNotFoundException` import not removed

src/test/java/seedu/address/logic/commands/AddCommandTest.java
```diff
@@ -181,6 +182,11 @@ public ReadOnlyAddressBook getAddressBook() {
     private class ModelStubAcceptingPersonAdded extends ModelStub {
         final ArrayList<Person> personsAdded = new ArrayList<>();
 
+        @Override
+        public boolean hasPerson(Person person) {
+            return personsAdded.stream().anyMatch(addedPerson -> addedPerson.isSamePerson(person));
```
Zhiyuan-Amos: lambda expression can be simplified to : `person::isSamePerson`

src/test/java/seedu/address/model/UniquePersonListTest.java
```diff
@@ -1,18 +1,185 @@
...
+    @Test
+    public void setPerson_targetPersonNotInList_throwsPersonNotFoundException() {
+        thrown.expect(PersonNotFoundException.class);
+        uniquePersonList.setPerson(ALICE, ALICE);
+    }
+
+    @Test
+    public void setPerson_editedPersonIsSamePerson_success() {
+        uniquePersonList.add(ALICE);
+        uniquePersonList.setPerson(ALICE, ALICE);
```
Zhiyuan-Amos: Should we do `uniquePersonList.setPerson(ALICE, new PersonBuilder(ALICE).build());` instead?
It's unclear here why we are using the same object references.

src/test/java/seedu/address/logic/commands/AddCommandTest.java
pyokagan: Sorry, I'll remember to run tests on every commit next time ;_;

pyokagan:
@yamidark Sorry, please do concentrate on https://github.com/se-edu/addressbook-level2/pull/388 , I think @damithc would prefer that. Sorry for requesting your review and taking your time, this PR is not so important.

src/main/java/seedu/address/model/util/SampleDataUtil.java
```diff
@@ -42,15 +41,11 @@
...
-            AddressBook sampleAb = new AddressBook();
-            for (Person samplePerson : getSamplePersons()) {
-                sampleAb.addPerson(samplePerson);
-            }
-            return sampleAb;
-        } catch (DuplicatePersonException e) {
-            throw new AssertionError("sample data cannot contain duplicate persons", e);
+        AddressBook sampleAb = new AddressBook();
+        for (Person samplePerson : getSamplePersons()) {
+            sampleAb.addPerson(samplePerson);
```
Zhiyuan-Amos: will this PR be simplifying some of the loops such as this one?
`Arrays.stream(getSamplePersons()).forEach(sampleAb::addPerson);`

Since you mentioned in your commit message that:
```
Checked exceptions also don't work very well with the Java Streams API,
since the API doesn't accept lambas which could throw checked
exceptions.
```

pyokagan: Eh let's keep that for another PR, to avoid delaying this one.

src/test/java/seedu/address/model/UniquePersonListTest.java
pyokagan: If they refer to the same object (for whatever reason), the API should still allow it, right?

pyokagan:
@eugenepeh 
> [10/10] Since UniquePersonList is in the package seedu.address.model.person;, UniquePersonListTest should be correspondingly the same.

fyi, this is not a new file, but I'll do it anyway in this PR since it's easy enough to do.

src/test/java/seedu/address/model/UniquePersonListTest.java
pyokagan: Yeah, after some more thought, I think I'll keep it. The "different object reference case" is already tested in `setPerson_editedPersonHasSameIdentity_success`

src/main/java/seedu/address/model/util/SampleDataUtil.java
Zhiyuan-Amos: Ok! Oh I spotted a spelling mistake for your commit msg: `lambas` :p

pyokagan:
@Zhiyuan-Amos 
> Since we are on Java 9, should we use List.of(...) instead of Arrays.asList(...)?

I considered that, but we don't use `List.of(...)` at all in our code base, and this PR may not be a good time to "break new ground" in that sense. Putting aside immutability, `Arrays.asList(...)` is the same as `List.of(...)` and not that many more characters to type.

I'll wait for #871

pyokagan:
[3/11] Also added some `requireNonNull()`s, missed that out in v1.

src/test/java/seedu/address/model/person/UniquePersonListTest.java
```diff
@@ -0,0 +1,184 @@
...
+    @Test
+    public void setPersons_nullList_throwsNullPointerException() {
+        thrown.expect(NullPointerException.class);
+        uniquePersonList.setPersons((List<Person>) null);
+    }
+
+    @Test
+    public void setPersons_list_replacesOwnListWithProvidedList() {
+        uniquePersonList.add(ALICE);
+        List<Person> personList = Arrays.asList(BOB);
```
Zhiyuan-Amos: IntelliJ would prefer you to use `Collections.singletonList()` here. :p

pyokagan: Foiled again.

Does it still prefer `Collections.singletonList()` over `List.of()`, I wonder.

PR #897, opened 2018-08-08
ClearCommandTest: fix tautological tests

Zhiyuan-Amos:
```
Tests in ClearCommandTest always pass because the actual and expected
model refer to the same instance;
`assertEquals(expectedModel, actualModel)` is always true.

As a result, the model's state isn't correctly verified.

Let's update these tests to correctly verify the model's state.
```

### DIFF
--- a/src/test/java/seedu/address/logic/commands/ClearCommandTest.java
+++ b/src/test/java/seedu/address/logic/commands/ClearCommandTest.java
@@ -6,6 +6,7 @@ import static seedu.address.testutil.TypicalPersons.getTypicalAddressBook;
 import org.junit.Test;
 
 import seedu.address.logic.CommandHistory;
+import seedu.address.model.AddressBook;
 import seedu.address.model.Model;
 import seedu.address.model.ModelManager;
 import seedu.address.model.UserPrefs;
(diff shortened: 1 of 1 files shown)


pyokagan:
Just one nit:

> ```
> ClearCommandTest: fix tautological tests
> 
> Tests in ClearCommandTest always pass because the actual and expected
> model refer to the same instance;
> ```

OK, I understand what this is saying.

> ```
> `assertEquals(expectedModel, actualModel)` is always true.
> ```

Can remove this line imo, it confuses more than it clarifies.

I guess it should be something like:

```
`assertEquals(expectedModel, actualModel)` will always return true if
expectedModel and actualModel reference the same object.
```

> ```
> As a result, the model's state isn't correctly verified.
> 
> Let's update these tests to correctly verify the model's state.
> ```

Fix looks okay.

pyokagan:
Sorry, I take that back, can you add back this to the commit message:

```
Tests in ClearCommandTest always pass because the actual and expected
model refer to the same instance;
```

Just having "`assertEquals(expectedModel, actualModel)` is always true because
expectedModel and actualModel reference the same object." says nothing about the problem.

Issue #451, opened 2017-05-26
HelpCommand#setData and ExitCommand#setData should be overwritten


Zhiyuan-Amos:
`HelpCommand#setData` and `ExitCommand#setData` are currently assigning `model` to `HelpCommand` and `ExitCommand` respectively.

There is no need for `HelpCommand` and `ExitCommand` to have access to `model`. 

Let's overwrite `HelpCommand#setData` and `ExitCommand#setData` to not assign `model` to their respective classes.

pyokagan:
Incidentally fixed by #898

PR #898, opened 2018-08-08
Combine Command#setData(...) and Command#execute() together

pyokagan:
```
Ensure that all command tests verify their model state and command
history.

Our commands forgot that their bound model and commandHistory are part
of their state, and did not check them in their `#equals(...)`
implementations. Combine Command#setData(...) and Command#execute()
together to fix this, and allow us to streamline our command testing
code by removing all of the `prepareCommand(...)` helper methods.
```

### DIFF
--- a/src/main/java/seedu/address/logic/CommandHistory.java
+++ b/src/main/java/seedu/address/logic/CommandHistory.java
@@ -15,6 +15,10 @@ public class CommandHistory {
         userInputHistory = new LinkedList<>();
     }
 
+    public CommandHistory(CommandHistory commandHistory) {
+        userInputHistory = new LinkedList<>(commandHistory.userInputHistory);
+    }
+
     /**
      * Appends {@code userInput} to the list of user input entered.
      */
@@ -28,5 +32,27 @@ public class CommandHistory {
      */
     public List<String> getHistory() {
         return new LinkedList<>(userInputHistory);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        // short circuit if same object
+        if (obj == this) {
+            return true;
+        }
+
+        // instanceof handles nulls
+        if (!(obj instanceof CommandHistory)) {
+            return false;
+        }
+
+        // state check
+        CommandHistory other = (CommandHistory) obj;
+        return userInputHistory.equals(other.userInputHistory);
+    }
+
+    @Override
+    public int hashCode() {
+        return userInputHistory.hashCode();
     }
 }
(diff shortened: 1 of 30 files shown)


pyokagan:
The PR total diffstat is overall +279 −279, but that's because some setup was needed to teach the command tests to do stricter verification of the model and command history.

The big payoff in the test code can be seen in the main dish of the PR, [5/5], with a total diffstat of +128 −220 due to the test helper code it eliminates.

Zhiyuan-Amos:
This PR somewhat resolves #451 too 🎉

src/test/java/seedu/address/logic/commands/CommandTestUtil.java
```diff
@@ -89,21 +92,26 @@ public static void assertCommandSuccess(Command command, Model actualModel, Stri
      * Executes the given {@code command}, confirms that <br>
      * - a {@code CommandException} is thrown <br>
      * - the CommandException message matches {@code expectedMessage} <br>
-     * - the address book and the filtered person list in the {@code actualModel} remain unchanged
+     * - the address book and the filtered person list in the {@code actualModel} remain unchanged <br>
+     * - {@code actualCommandHistory} remains unchanged
```
Zhiyuan-Amos: Nit: missing period

src/main/java/seedu/address/logic/CommandHistory.java
```diff
@@ -15,6 +15,10 @@ public CommandHistory() {
         userInputHistory = new LinkedList<>();
     }
 
+    public CommandHistory(CommandHistory commandHistory) {
+        userInputHistory = new LinkedList<>(commandHistory.userInputHistory);
```
Zhiyuan-Amos: Since this constructor is only used by tests, should we create a helper class in the test code that creates a copy of `CommandHistory` instead?

src/test/java/seedu/address/logic/CommandHistoryTest.java
```diff
@@ -24,4 +35,45 @@ public void add() {
...
+
+        // different types -> returns false
+        assertFalse(commandHistoryWithA.equals(5.0f));
+
+        // different values -> returns false
+        assertFalse(commandHistoryWithA.equals(commandHistoryWithB));
+    }
+
+    @Test
+    public void hashcode() {
```
Zhiyuan-Amos: Capital 'C'?

pyokagan: `hashCode()` will conflict with `Object#hashCode()` (iow it will need to return an int and all that).

Granted, this is subtle and I'm open to other suggestions.

PR #903, opened 2018-08-14
Remove unnecessary uses of `this`

### LINKED ISSUE #856: Remove unnecessary use of `this`
Refer to #729.
There are still places in the codebase where the use of `this` goes against our coding standard.
One example is at:
https://github.com/se-edu/addressbook-level4/blob/2f4405c75cd21111952565a9706a9793b475c41e/src/main/java/seedu/address/commons/core/GuiSettings.java#L19-L23

pyokagan:
Fixes #856

### DIFF
--- a/src/main/java/seedu/address/MainApp.java
+++ b/src/main/java/seedu/address/MainApp.java
@@ -199,7 +199,7 @@ public class MainApp extends Application {
     @Subscribe
     public void handleExitAppRequestEvent(ExitAppRequestEvent event) {
         logger.info(LogsCenter.getEventHandlingLogMessage(event));
-        this.stop();
+        stop();
     }
 
     public static void main(String[] args) {
--- a/src/main/java/seedu/address/commons/core/GuiSettings.java
+++ b/src/main/java/seedu/address/commons/core/GuiSettings.java
@@ -17,15 +17,15 @@ public class GuiSettings implements Serializable {
     private Point windowCoordinates;
 
     public GuiSettings() {
-        this.windowWidth = DEFAULT_WIDTH;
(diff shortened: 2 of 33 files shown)


pyokagan:
> XmlAdaptedPerson still contains unnecessary uses of this.

Yeah, but that is shadowed later on in the method body. Confirm you want?

pyokagan:
Actually, now that I look at the result:

```java
        if (name == null) {
            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Name.class.getSimpleName()));
        }
        if (!Name.isValidName(name)) {
            throw new IllegalValueException(Name.MESSAGE_NAME_CONSTRAINTS);
        }
        final Name name = new Name(this.name);

        if (phone == null) {
            throw new IllegalValueException(String.format(MISSING_FIELD_MESSAGE_FORMAT, Phone.class.getSimpleName()));
        }
        if (!Phone.isValidPhone(phone)) {
            throw new IllegalValueException(Phone.MESSAGE_PHONE_CONSTRAINTS);
        }
        final Phone phone = new Phone(this.phone);
```

it doesn't look *too* bad :-P

But there is likely a reason why @yamidark chose to use `this.name` etc. and such (because he found it more readable from writing the code), and so the proper solution would be to rename the variables.

yamidark:
>But there is likely a reason why @yamidark chose to use this.name etc.

IIRC, I placed the 2 checks together at 1st (`this.name != null && !Name.isValidName(this.name)`), which is why I copied using `this.name`, before I separated them later in order to give separate exception messages (didn't really consider readability at that time).

But looking at the renamed variables, I agree that it looks more readable (to me) now compared to before.

PR #904, opened 2018-08-15
checkstyle: add SingleSpaceSeparator

### LINKED ISSUE #876: Explore using checkstyle SingleSpaceSeparator
e.g.

https://github.com/se-edu/addressbook-level4/blob/ae383756eaef6702531f5258fe501b7c0e2c831d/src/main/java/seedu/address/model/person/Email.java#L12

Code accidentally having two many spaces has come up too many times during code review, so It'll be good to have an automatic check for that.

Also, it seems that instances of that has slipped into the code base, so it'll be good to clean that up using checkstyle for detection.

http://checkstyle.sourceforge.net/config_whitespace.html#SingleSpaceSeparator

pyokagan:
Fixes #876

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -356,6 +356,25 @@
       <property name="severity" value="warning"/>
     </module>
 
+    <!-- Checks that non-whitespace characters are separated by no more than one whitespace character.
+         a = 1; // Allowed
+         a  = 1; // Not allowed (more than one space before =)
+    -->
+    <module name="SingleSpaceSeparator">
+      <!-- Validate whitespace surrounding comments as well.
+
+           a = 1; // Allowed (single space before start of comment)
+           a = 1; /* Allowed (single space before start of comment) */
+           /* Allowed (single space after end of comment) */ a = 1;
+           a = 1;  // Not allowed (more than one space before start of comment)
+           a = 1;  /* Not allowed (more than one space before start of comment) */
+           /* Not allowed (more than one space after end of comment) */  a = 1;
+
+           This doesn't validate whitespace within comments so a comment /* like  this */ is allowed.
+      -->
+      <property name="validateComments" value="true"/>
+    </module>
+
     <!--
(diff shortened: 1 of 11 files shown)


config/checkstyle/checkstyle.xml
```diff
@@ -356,6 +356,22 @@
...
+    <module name="SingleSpaceSeparator">
+      <!-- Validate whitespace surrounding comments as well.
+           a = 1; //  Allowed (single space before start of comment)
+           a = 1; /* Allowed (single space before start of comment) */
+           /* Allowed (single space after end of comment) */ a = 1;
+           a = 1;  // Not allowed (more than one space before start of comment)
+           a = 1;  /* Not allowed (more than one space before start of comment) */
+           /* Not allowed (more than one space after end of comment) */  a = 1;
+      -->
+      <property name="validateComments" value="true" />
```
eugenepeh: Maybe can remove the space before `/>`

config/checkstyle/checkstyle.xml
```diff
@@ -356,6 +356,22 @@
       <property name="severity" value="warning"/>
     </module>
 
+    <!-- Checks that non-whitespace characters are separated by no more than one whitespace character.
+         a = 1; // Allowed
+         a  = 1; // Not allowed (more than one space before =)
+    -->
+    <module name="SingleSpaceSeparator">
+      <!-- Validate whitespace surrounding comments as well.
+           a = 1; //  Allowed (single space before start of comment)
```
eugenepeh: Seems like you're trying to demo white space within comments are allowed but not very obvious.

Maybe can change the comments to `//   Allowed (multiple whitespace characters within comments)` since above already has a similar example.

pyokagan: I decided to do something like this:

```diff
            /* Not allowed (more than one space after end of comment) */  a = 1;
+
+           This doesn't validate whitespace within comments so something /* like  this */ is allowed.
```

so we won't have redundant info in all the comments.

PR #910, opened 2018-09-07
HelpWindowTest: fix bogus "focus on another stage" code

pyokagan:


pyokagan:
@damithc This fix should probably be pulled into the semester repos as well.

damithc:
> @damithc This fix should probably be pulled into the semester repos as well.

Will do.

PR #911, opened 2018-09-07
DeveloperGuide: update "Setting up" section

yamidark:
```
Due to an ongoing issue[1] with some of the newer versions of IntelliJ,
the IDE may report code errors for JAXB APIs, although the code can be
build and run successfully since we have already imported the
dependencies through gradle.

This may cause confusion amongst users.

Let's update the "Setting up" section to address this issue, and how
to resolve it.

[1]: https://youtrack.jetbrains.com/issue/IDEA-189060
```

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -11,6 +11,7 @@ ifdef::env-github[]
 :tip-caption: :bulb:
 :note-caption: :information_source:
 :warning-caption: :warning:
+:experimental:
 endif::[]
 :repoURL: https://github.com/se-edu/addressbook-level4/tree/master
 
@@ -46,6 +47,10 @@ Do not disable them. If you have disabled them, go to `File` > `Settings` > `Plu
 . Click `OK` to accept the default settings
 . Open a console and run the command `gradlew processResources` (Mac/Linux: `./gradlew processResources`). It should finish with the `BUILD SUCCESSFUL` message. +
 This will generate all resources required by the application and tests.
+. Open link:{repoURL}/src/main/java/seedu/address/storage/XmlAdaptedPerson.java[`XmlAdaptedPerson.java`] and link:{repoURL}/src/main/java/seedu/address/ui/MainWindow.java[`MainWindow.java`] and check for any code errors
+.. Due to an ongoing https://youtrack.jetbrains.com/issue/IDEA-189060[issue] with some of the newer versions of IntelliJ, code errors may be detected even if the project can be built and run successfully
+.. To resolve this, place your cursor over any of the code section highlighted in red. Press kbd:[ALT + ENTER], and select `Add '--add-modules=...' to module compiler options` for each error
+. Repeat this for the test folder as well (e.g. check link:{repoURL}/src/test/java/seedu/address/commons/util/XmlUtilTest.java[`XmlUtilTest.java`] and link:{repoURL}/src/test/java/seedu/address/ui/HelpWindowTest.java[`HelpWindowTest.java`] for code errors, and if so, resolve it the same way)
 
 === Verifying the setup
 


pyokagan:
Strangely, I don't get errors for jaxb imports on both AB-3 and AB-4. I do get errors for JavaFX imports on both AB-3 and AB-4 though.

pyokagan:
Does anyone share the same results as me?

yamidark:
>Does anyone share the same results as me?

Tried with the Community Edition, seems like we would get errors with Javafx there, while if we are using Ultimate Edition, we would get errors with `java.xml.bind`. I believe @yamgent faced the same problem with Javafx imports  in se-edu/se-book#62.

Since different users may face different import issues, will update the DeveloperGuide to mention checking both sides.

docs/DeveloperGuide.adoc
```diff
@@ -46,6 +46,10 @@ Do not disable them. If you have disabled them, go to `File` > `Settings` > `Plu
 . Click `OK` to accept the default settings
 . Open a console and run the command `gradlew processResources` (Mac/Linux: `./gradlew processResources`). It should finish with the `BUILD SUCCESSFUL` message. +
 This will generate all resources required by the application and tests.
+. Open the `XmlAdaptedPerson` and `MainWindow` file and check for any code errors
+.. Due to an ongoing https://youtrack.jetbrains.com/issue/IDEA-189060[issue] with some of the newer versions of IntelliJ, code errors may be detected even if the project can be built and run successfully
+.. To resolve this, place your cursor over any of the code section highlighted in red. Press kbd:[ALT + ENTER], and select `Add '--add-modules=...' to module compiler options` for each error
```
pyokagan: The `kbd:[ALT + ENTER]` isn't rendered properly. I think the `:experimental:` attribute isn't set in this file.

docs/DeveloperGuide.adoc
```diff
@@ -46,6 +46,10 @@ Do not disable them. If you have disabled them, go to `File` > `Settings` > `Plu
...
+. Repeat this for the test folder as well (check `XmlUtilTest` and `HelpWindowTest` file for code errors, and if so, resolve it the same way)
```
pyokagan: Since you'll need to submit a new iteration for the `kbd:[ALT + ENTER]` issue, some more suggestions:

The `XmlUtilTest` and `HelpWindowTest` are just examples, right? So,

```
(e.g. check `XmlUtilTest` and `HelpWindowTest` for code errors, and if so, ...
```

docs/DeveloperGuide.adoc
```diff
@@ -46,6 +46,10 @@ Do not disable them. If you have disabled them, go to `File` > `Settings` > `Plu
 . Click `OK` to accept the default settings
 . Open a console and run the command `gradlew processResources` (Mac/Linux: `./gradlew processResources`). It should finish with the `BUILD SUCCESSFUL` message. +
 This will generate all resources required by the application and tests.
+. Open the `XmlAdaptedPerson` and `MainWindow` file and check for any code errors
```
pyokagan: Should be:

```
Open `XmlAdaptedPerson` and `MainWindow`, and check for any code errors
```

(Since `XmlAdaptedPerson` and `MainWindow` are names. Similar example on why the original sounds wrong: `Talk to the Bob and Jill person`.)

If you want to make it clear they are files, then can write:

```
Open `XmlAdaptedPerson.java` and `MainWindow.java` and check for any code errors.
```

We also have `repoURL` defined here as well, so you can use:

```
Open link:{repoURL}/src/main/java/seedu/address/storage/XmlAdaptedPerson.java[`XmlAdaptedPerson.java`] and link:{repoURL}/src/main/java/seedu/address/ui/MainWindow.java[`MainWindow.java`] and check for any code errors.
```

docs/DeveloperGuide.adoc
yamgent: It renders fine on Netlify, are you trying it locally?

pyokagan: We need the docs to render the same on both GitHub and on GitHub Pages.

yamidark:
@damithc This should probably be pulled into this semester's repos as well.

damithc:
> @damithc This should probably be pulled into this semester's repos as well.

done. Thanks for the fix @yamidark and for the review @pyokagan

PR #916, opened 2018-09-20
Shorten names of constants of Person's attributes #399

### LINKED ISSUE #399: Shorten names of constants of Person's attributes
Each of these classes: `Name`, `Phone`, `Email`, `Address` and `Tag` has a `MESSAGE_X_CONSTRAINTS` and `X_VALIDATION_REGEX` (X referring to the name of the class).

It is unnecessary to have `X` as part of the names of these constants. For example, `Name.MESSAGE_NAME_CONSTRAINTS` doesn't seem to provide more information than `Name.MESSAGE_CONSTRAINTS`.

apoorva17:
Fixes #399 
 ```
Each of these classes: Name, Phone, Email, Address and Tag has a 
MESSAGE_X_CONSTRAINTS and X_VALIDATION_REGEX (X referring to the name of 
the class).

It is unnecessary to have X as part of the names of these constants. For 
example, Name.MESSAGE_NAME_CONSTRAINTS doesn't seem to provide more 
information than Name.MESSAGE_CONSTRAINTS.

As such, let's remove X as part of the names of these constants.
```

### DIFF
--- a/src/main/java/seedu/address/model/person/Address.java
+++ b/src/main/java/seedu/address/model/person/Address.java
@@ -9,14 +9,13 @@ import static seedu.address.commons.util.AppUtil.checkArgument;
  */
 public class Address {
 
-    public static final String MESSAGE_ADDRESS_CONSTRAINTS =
-            "Addresses can take any values, and it should not be blank";
+    public static final String MESSAGE_CONSTRAINTS = "Addresses can take any values, and it should not be blank";
 
     /*
      * The first character of the address must not be a whitespace,
      * otherwise " " (a blank string) becomes a valid input.
      */
-    public static final String ADDRESS_VALIDATION_REGEX = "[^\\s].*";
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by pyokagan** [comment]
@apoorva17 I'll take a closer look later, but in the meantime can you wrap the commit message to a width of 72 characters per line?

File: src/main/java/seedu/address/model/person/Address.java [hunk 1]
```diff
@@ -9,14 +9,14 @@
  */
 public class Address {
 
-    public static final String MESSAGE_ADDRESS_CONSTRAINTS =
+    public static final String MESSAGE_CONSTRAINTS =
             "Addresses can take any values, and it should not be blank";
```
pyokagan: You can combine these lines into one now:

```java
    public static final String MESSAGE_CONSTRAINTS = "Addresses can take any values, and it should not be blank";
```

PR #932, opened 2018-11-27
Make UserPrefs truly a part of Model

pyokagan:
```
Fix the codebase such that UserPrefs is truly a model component, and any
other code accessing user preferences will need to go through the Model
interface.
```


src/test/java/seedu/address/model/ModelManagerTest.java
```diff
@@ -22,6 +25,59 @@
...
+    }
+
+    @Test
+    public void setGuiSettings_nullGuiSettings_throwsNullPointerException() {
+        thrown.expect(NullPointerException.class);
+        modelManager.setGuiSettings(null);
+    }
+
+    @Test
+    public void setGuiSettings() {
```
yamgent: [v1 12/13] 
`setGuiSettings_validGuiSettings_setsGuiSettings()`

pyokagan:
> Line 191 in stop() still uses the userPrefs directly without consulting Model, which is not correct because Model now holds its own copy of userPrefs, and its content may change:

Good catch :+1:

src/main/java/seedu/address/commons/core/GuiSettings.java
```diff
@@ -6,14 +6,15 @@
 
 /**
  * A Serializable class that contains the GUI settings.
+ * Guarantees: immutable.
```
yamgent: [v2 3/15]

Actually, still not quite immutable yet. :P

The caller of `getWindowCoordinates()` can modify the window coordinates setting:

```java
    @Test
    public void immutable() {
        Point originalCoordinates = new Point(5, 5);
        GuiSettings settings = new GuiSettings(640.0, 480.0, originalCoordinates.x, originalCoordinates.y);

        // Modification of returned value should not result in modification of the actual object's coordinates
        Point returnedCoordinates = settings.getWindowCoordinates();
        returnedCoordinates.x = 10;
        returnedCoordinates.y = 10;
        assertEquals(originalCoordinates, settings.getWindowCoordinates());
    }
```

Maybe `getWindowCoordinates()` should return a copy of the coordinates, rather than the original. The `null` case will also need to be handled then if done this way.

pyokagan: Good catch again :+1: <sup>(Its actually still immutable by definition though :stuck_out_tongue: )</sup>

pyokagan: Stepping back, the point coordinates should be stored as doubles since they are represented as doubles in JavaFX.

I tried seeing if I could replace it with a `javafx.geom.Point2D`, but it can't be serialized by Jackson >:( Bah. Anyway, small issue and nearly inconsequential.

PR #934, opened 2018-12-03
checkstyle: allow missing javadoc for private methods

pyokagan:


fzdy1914:
This can be done using the suppressions.xml file.

pyokagan:
@fzdy1914 
> This can be done using the suppressions.xml file.

What do you mean?

fzdy1914:
The goal of this PR can be done by adding a new section in here [suppressions.xml](https://github.com/se-edu/addressbook-level4/blob/master/config/checkstyle/suppressions.xml)

pyokagan:
@fzdy1914 
>  I thought the pr is to allow missing javadoc for private methods only in tests.

Yes, I'm thinking of removing the check entirely for all private methods, since I think it's more trouble than it's worth. I still haven't decided on whether I should go forward with it or not though.

Seems like RepoSense has the [same rule enabled](https://github.com/reposense/RepoSense/blob/a9ac842f3e3f8ff9d1fd7fbdb00aa520256ffda8/config/checkstyle/checkstyle.xml#L384-L401), although it seems like it only causes a warning, not an error. (As opposed to AB-4 where it is an error.) Maybe you could try adding javadocs to all private methods in the code base and see if its worth it or not ;-)

damithc:
@fzdy1914 shall we merge this?

fzdy1914:
@damithc Instead of removing the whole inspection, I am referring to set the severity to a warning first. What's your opinion on the javadoc of the private methods issue?

damithc:
> @damithc Instead of removing the whole inspection, I am referring to set the severity to a warning first. What's your opinion on the javadoc of the private methods issue?

The downside of that is we end up with a bunch of warning that are always there; over time, we stop reading the warnings at all (because we _expect_ them to be there anyway) which increases the risk of missing other unrelated warnings. 

If we remove it entirely, we can still expect reviewers to ask for javadoc comments for non-trivial private methods. So, I don't mind removing it altogether.

fzdy1914:
@damithc  That's true, let's get this remove.

PR #939, opened 2018-12-09
ci: do not retry on test failures

pyokagan:


docs/UsingTravis.adoc
```diff
@@ -10,7 +10,6 @@ Travis CI can run the projects' tests automatically whenever new code is pushed
 The current Travis CI set up performs the following things whenever someone push code to the repo:
 
 * Runs the `./gradlew clean headless allTests coverage coveralls -i` command (see <<UsingGradle#, UsingGradle>> for more details on what this command means).
```
pyokagan: I'm aware of the error here but it's left for a later commit.

PR #944, opened 2018-12-13
MainApp: don't forcefully terminate the application on exit

pyokagan:
No merge commit


pyokagan:
Seems like Travis didn't like the `copyDummySearchPage` task disappearing from master. Will submit a new version.

PR #945, opened 2018-12-13
Logic: expose command history as an ObservableList

pyokagan:


pyokagan:
Seems like Travis didn't like the `copyDummySearchPage` task disappearing from master. Will submit a new version.

src/main/java/seedu/address/logic/commands/HistoryCommand.java
```diff
@@ -20,7 +20,7 @@
     @Override
     public CommandResult execute(Model model, CommandHistory history) {
         requireNonNull(history);
-        List<String> previousCommands = history.getHistory();
+        ArrayList<String> previousCommands = new ArrayList<>(history.getHistory());
```
Zhiyuan-Amos: Are we doing this instead of:

`**List**<String> previousCommands = new ArrayList<>(history.getHistory());`

to specifically show that `previousCommands` is a modifiable list?

pyokagan: It's done this way because there's nothing to be gained by using an interface over the original type of the object.

PR #955, opened 2019-01-25
AddressBook 3.5

pyokagan:
Will be merged into https://github.com/se-edu/addressbook-level35

Note that this is just a stop-gap measure that should only be used for the next few semesters.


pyokagan:
@damithc Anything else you want to remove?

damithc:
> @damithc Anything else you want to remove?

Can't think of anything. Maybe cleaner to keep this as _AB-4  minus GUI Tests_?

@yamgent @Zhiyuan-Amos @eugenepeh @yamidark anything?

While we are on this topic, is it possible to also have a _AB-3.5 minus GUI plus CLI_ (to replace AB-3)? Or the Model/Logic is too optimized for the GUI? Of course it's not something to do in the near future. Just a thought only.

damithc:
> This is sort of like a trial run for the creation of an AB-3 that will replace the old AB-3. From the perspective, what should be removed or changed? (So that we can test them out early and get feedback from the current semester)

My previous approach was for each levels to introduce breadth-first enhancements e.g., introduce polymorphism across the code base. But if we want to unify multiple AB levels into one repo, I guess depth-first increments are more appropriate? For example, AB3 + GUI = AB4 ?

pyokagan:
@damithc 

Sorry, just realized I missed this:

> While we are on this topic, is it possible to also have a AB-3.5 minus GUI plus CLI (to replace AB-3)? Or the Model/Logic is too optimized for the GUI? 

Do you mean in the same style as the current AB3? (a "CLI in a GUI") If so, two concerns:

* The way person numbering works in AB-4 is slightly different from AB-3. When a person is deleted in AB-4, the next person would take its index and so on. When a person is deleted in AB-3, the indices do not change. It's a bit of effort to re-implement AB-3's behavior, so I recommend that AB-3 just take on AB-4's behavior. 

    Of course, it's highly likely that AB-3 was coded to behave that way since there is no "persons list" displayed in the GUI. I recommend that AB-3 have a "persons list" in its GUI as well.

* I'm not sure why a "CLI in a GUI" is considered to be simpler than the GUI in AB-4. Code-wise they are around the same level of complexity. (Except for the indices thing as mentioned earlier)

damithc:
> Do you mean in the same style as the current AB3? (a "CLI in a GUI") If so, two concerns:

AB3 can be a pure CLI in this new depth-first approach. 'CLI in a GUI' was the breadth-first approach.

> * so I recommend that AB-3 just take on AB-4's behavior.

No objections from me.

pyokagan:
If we believe that students have more trouble when the sloc count is higher, then a simple way forward would be to remove all "nice-to-have" features. The easy gains are:

* Remove undo/redo (-800 lines)
* Remove history command (-750 lines)
* Remove select command (-500 lines, at which point the browser panel becomes useless and can be removed as well, -80 lines)
* Remove use of invalidation listeners (-200 lines)
* Remove GuiSettings (-200 lines)

damithc:
> If we believe that students have more trouble when the sloc count is higher, then a simple way forward would be to remove all "nice-to-have" features. The easy gains are:

Yup, we can do that as well. What do we remove from AB-3 to get AB-2?

pyokagan:
@damithc 
> Yup, we can do that as well. What do we remove from AB-3 to get AB-2?

I'm not thinking that far ahead yet :-P When all the low hanging fruit (removing features) have all been picked, the only remaining things we can do I think is to simplify the architecture, which is a bit more work.

pyokagan:
@Zhiyuan-Amos 
> Looks good! +1

What looks good? Me? :stuck_out_tongue:

okkhoy:
@pyokagan @damithc and others.

One of the teams on CS2113 did something [like the above comment](https://github.com/se-edu/addressbook-level4/pull/955#issuecomment-457615762). They did not want to set up Gradle/Travis/Autopublishing/Checkstyle etc., missing in AB-3 but didn't want to do GUI of AB-4.
So in effect they had a AB-3.5 in their project (with undo/redo and history still being there).

I am thinking this could be AB-3.5?

pyokagan:
@okkhoy 
> One of the teams on CS2113 did something [like the above comment](https://github.com/se-edu/addressbook-level4/pull/955#issuecomment-457615762).

Which team? It is hard to draw a comparison without the actual source code.

okkhoy:
Here you go: https://github.com/CS2113-AY1819S1-T13-2/main

pyokagan:
@okkhoy Alright, it's merged into https://github.com/se-edu/addressbook-level35 . Have fun.

@damithc Please enable Travis + AppVeyor for that repo. Thanks!

damithc:
> @damithc Please enable Travis + AppVeyor for that repo. Thanks!

Done

Issue #874, opened 2018-04-01
Upgrade JUnit4 tests to JUnit5


Rinder5:
Our tests use JUnit4. Let's upgrade them to use JUnit5 instead.

- [ ] Change imports to JUnit5 version. (e.g. `org.junit.jupiter.api.Test` instead of
`org.junit.Test`)
- [ ] Remove `Test` being used for specifying expectations. We use this to
specify timeout in `PersonListPanelTest`. Instead we should now use
`assertTimeoutPreemptively`
- [ ] Rename `@Before` to `@BeforeEach`
- [ ] Rename `@After` to `@AfterEach`
- [ ] Rename `@BeforeClass` to `@BeforeAll`
- [ ] Rename `@AfterClass` to `@AfterAll`
- [ ] Group assertions together using `assertAll` for related assertions
- [ ] Remove `@Rule` and `@ClassRule`. Instead, we need to
`@ExtendWith`, which generally requires creating an extension class.

Mentor: @Zhiyuan-Amos 
Assigned to: @sijie123

sijie123:
While migrating from JUnit 4 to 5, I ran into a small issue...
In JUnit 5, assertThrows only checks that it is of the correct type.
If we want to assert the error message, we have to take the exception re-thrown by assertThrows and do comparisons ourselves. As this routine is commonly used, I feel that a custom method within our custom Assert class is the best way to go.
However, this means that I have to import static both Junit's assertThrows (the plain one), as well as our Assert.assertThrows (the enhanced one) in test classes that use both types of assertions (e.g. AppUtilTest).
I'm not sure if this is a good idea, since it will be confusing for readers -- why are we importing assertThrows from 2 places?

The alternative is that I add a wrapper to the plain assertThrows in our custom Assert. However, we will still import things like AssertEquals or AssertTrue from JUnit, which becomes the same problem as above.

Any suggestions?

sijie123:
Looking through git blame for existing Assert class,
I think the reason why the current Assert class exists with 2 methods is because earlier versions of JUnit didn't have good support for assertThrown, hence we implemented our own versions.

Zhiyuan-Amos:
> In JUnit 5, assertThrows only checks that it is of the correct type. If we want to assert the error message, we have to take the exception re-thrown by assertThrows and do comparisons ourselves.

Does the following `assertThrows` API work?

```
public static <T extends Throwable> T assertThrows(Class<T> expectedType,
                                                   Executable executable,
[...]

PR #956, opened 2019-01-27
[WIP] Migrate from JUnit 4 to JUnit 5

sijie123:
This is still a work in progress.

Following #874, this PR will contain a series of commits that will migrate all JUnit 4 tests to JUnit 5.

### DIFF
--- a/src/test/java/seedu/address/testutil/Assert.java
+++ b/src/test/java/seedu/address/testutil/Assert.java
@@ -1,23 +1,43 @@
 package seedu.address.testutil;
 
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.function.Executable;
+
 /**
  * A set of assertion methods useful for writing tests.
  */
 public class Assert {
 
     /**
-     * Asserts that the {@code callable} throws the {@code expected} Exception.
+     * Asserts that the {@code executable} throws the {@code expectedType} Exception.
+     * This is a wrapper method that invokes
+     * {@code Assertions.assertThrows(Class<? extends Throwable>, Executable)}, to maintain
+     * consistency with our custom
+     * {@see assertThrowsWithMessage(Class<? extends Throwable>, String, Executable)} method.
      */
-    public static void assertThrows(Class<? extends Throwable> expected, VoidCallable callable) {
-        assertThrows(expected, null, callable);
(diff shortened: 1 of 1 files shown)


sijie123:
At this juncture, I wanted to get @Zhiyuan-Amos to help me take a quick look at my code before I continue making more significant changes.
Particularly,
1. Is this what you guys are looking for? I changed the tests to JUnit 5 style - do they look right?
2. I also changed everything from AssertTrue(a.equals(b)) into AssertEquals(a,b). Would this be a better way moving forward?

Zhiyuan-Amos:
Hi @sijie123, what you've done looks good so far! 👍 

> I also changed everything from AssertTrue(a.equals(b)) into AssertEquals(a,b). Would this be a better way moving forward?

What you've suggested is right according to this [SO post](https://stackoverflow.com/questions/24821983/junit-which-is-recommended-asserttrue-or-assertequals-for-string). Do include the comments from this post in your commit message as the justification for making this change.

src/test/java/seedu/address/commons/core/index/IndexTest.java
```diff
@@ -57,36 +57,33 @@ private void assertCreateOneBasedFailure(int invalidOneBasedIndex) {
      * (if it is one based), and confirms that an {@code IndexOutOfBoundsException} is thrown.
      */
     private void assertCreateFailure(int invalidIndex, boolean isZeroBased) {
```
Zhiyuan-Amos: On a side note, if you have the time, maybe you can look into which methods can be removed. For example, in the past, I think we created this method because replicating the following in 2 places (`assertCreateOneBasedFailure` and `assertCreateZeroBasedFailure`) is verbose:

```
try {
  ...
  throw new AssertionError("...");
} catch (IOOBE ie) {
  // expected behaviour
}
```

Since we can use `assertThrows` now, we can tweak `assertCreateOneBasedFailure` and `assertCreateZeroBasedFailure` to look something like:

`assertThrows(IndexOutOfBoundsException.class, Index::fromZeroBased);`

This removes the need for this method.

sijie123:
Closing this PR as this was mainly used to get feedback on the scope and style of the migration.
I intend to migrate in small PRs, each focusing on a certain group of tests.

PR #460, opened 2017-06-01
StringUtil#isUnsignedInteger(String): Fix unhandled NFE #436

Zhiyuan-Amos:
Fixes #436

PR #975, opened 2019-02-24
StringUtilTest: Rename test method name

sijie123:
Currently, StringUtilTest has a method named
isUnsignedPositiveInteger. However, the method tested is actually
named isNonZeroUnsignedInteger. This can cause confusion in the
naming.

It appears that the tested method was renamed to
isNonZeroUnsignedInteger in #460, however the test was not renamed
then.

To maintain consistent naming convention, lets rename the test method
to isNonZeroUnsignedInteger too.

### DIFF
--- a/src/test/java/seedu/address/commons/util/StringUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/StringUtilTest.java
@@ -17,10 +17,10 @@ public class StringUtilTest {
     @Rule
     public ExpectedException thrown = ExpectedException.none();
 
-    //---------------- Tests for isUnsignedPositiveInteger --------------------------------------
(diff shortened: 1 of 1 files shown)


pyokagan:
Side note, but what happened to the commit message in 0352e6b6e3f51910edbff7a786f6ae24f453d38f  :-/ ?

> ```
> StringUtil#isUnsignedInteger(String) attempts to convert any given
> number to an integer (as a step in checking if it is an unsigned integer).
>
> When the input is a number higher than Integer.MAX_VALUE,
> the method fails with an exception instead of returning false.
> ```

The original code of `isUnsignedInteger()` was

```java
    return s != null && s.matches("^0*[1-9]\\d*$");
```

which doesn't do any integer conversion, and can't possibly throw a NFE.

PR #976, opened 2019-02-24
FileUtilTest: Add test for Windows path

sijie123:
Currently, FileUtilTest#isValidPath() only tests for isValidPath for a linux path. However, our users can also develop and use AB4 on a Windows PC. To provide assurance that our FileUtil#isValidPath(String) method works on Windows computers as well, we should add coverage to test for paths on Windows.

Hence, let's add another test to ensure that FileUtil#isValidPath returns the correct result for a valid path in the Windows naming convention.

### DIFF
--- a/src/test/java/seedu/address/commons/util/FileUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/FileUtilTest.java
@@ -12,7 +12,8 @@ public class FileUtilTest {
     @Test
     public void isValidPath() {
         // valid path
-        assertTrue(FileUtil.isValidPath("valid/file/path"));
+        assertTrue(FileUtil.isValidPath("valid/file/path")); //Linux
+        assertTrue(FileUtil.isValidPath("valid\\file\\path")); //Windows
 
         // invalid path
         assertFalse(FileUtil.isValidPath("a\0"));


src/test/java/seedu/address/commons/util/FileUtilTest.java
```diff
@@ -12,7 +12,8 @@
     @Test
     public void isValidPath() {
         // valid path
-        assertTrue(FileUtil.isValidPath("valid/file/path"));
+        assertTrue(FileUtil.isValidPath("valid/file/path")); //Linux
+        assertTrue(FileUtil.isValidPath("valid\\file\\path")); //Windows
```
pyokagan: I don't like the idea of the test since it is not a "Windows path", It is a valid Linux path as well. So, it's not correct to annotate them with the `// Linux` and `// Windows` comments.

Also, compared to `valid/file/path`, `valid\file\path` has different meanings on Windows and Linux. On Linux it just means "a file with the name `valid\file\path`", which is probably not the intention here.

Anyway, as the very sparse "invalid path" test case below hints, these are only simple smoke tests to check to see if `isValidPath()` is properly wired up to `Paths.get()`. So I don't really see the value in adding more tests.

sijie123:
I think I was kinda loose on my Windows/Linux terminology use. 
According to this [Stack Overflow](https://stackoverflow.com/questions/1589930/so-what-is-the-right-direction-of-the-paths-slash-or-under-windows) page, actually "a/b/c" is not a valid filesystem path on Windows, although it is valid as a URI.

In any case, I agree with you in that the method is essentially a hook to `Paths.get()`, so if we can assume that `Paths.get()` is implemented correctly, then this extra test certainly doesn't add value.

I will close this issue.

PR #980, opened 2019-02-28
Upgrade to Gradle 5.2.1

fzdy1914:
We are currently using Gradle 4.8.1 as our default wrapper, which is a
version before JDK 11 comes.

Gradle 5.2.1 has been announced for a while and it gives better support
for JDK11.[1]

As we are going to use JDK 11+ as our supported JDK, let’s upgrade to
Gradle 5.2.1 to better suit JDK 11.

Also, let's update all the relevant plugins to fit JDK 11+ better.

[1] https://docs.gradle.org/5.2.1/release-notes.html


fzdy1914:
I will push all the commits one by one to see if the CI fails.

pyokagan:
> I will push all the commits one by one to see if the CI fails.

You can do something like `git rebase -i -x './gradlew headless check'` locally on your machine as well, since CI is pretty slow.

fzdy1914:
> You can do something like `git rebase -i -x './gradlew headless check'` locally on your machine as well, since CI is pretty slow.

Thanks for your suggestion.

pyokagan:
> plugin to its latest stable version 4.0.4 to solve the problem.

Seems like the shadow plugin version **5.0.0** was released 12 hours ago :-P https://github.com/johnrengelman/shadow/releases

build.gradle
```diff
@@ -66,13 +66,13 @@ shadowJar {
 }
 
 task coverage(type: JacocoReport) {
-    sourceDirectories = files(sourceSets.main.allSource.srcDirs)
-    classDirectories =  files(sourceSets.main.output)
-    executionData = files(jacocoTestReport.executionData)
+    getSourceDirectories().from(files(sourceSets.main.allSource.srcDirs))
+    getClassDirectories().from(files(sourceSets.main.output))
+    getExecutionData().from(files(jacocoTestReport.executionData))
```
pyokagan: We use Groovy's [getter shortcut notation](http://groovy-lang.org/style-guide.html#_getters_and_setters) throughout the `build.gradle` file, so it would be good to be consistent. E.g.:

```groovy
executionData.from(files(jacocoTestReport.executionData))
```

Same for the rest.

pyokagan:
But really though, for [2/4] it is sufficient to say that we are upgrading the "shadow" plugin just to be up to date with its development.

And with that same reason, we can also justify updating the other third-party plugins that we use as well.

fzdy1914:
> But really though, for [2/4] it is sufficient to say that we are upgrading the "shadow" plugin just to be up to date with its development.
> 
> And with that same reason, we can also justify updating the other third-party plugins that we use as well.

So, I do not need to write the errors in the commit message?

fzdy1914:
> Okay, but I still don't understand where you got the "suggested version" from.
> That's pretty important information that would let me know if the fix is
> correct and reasonable or not.

See the warning message, it tells us to use getSourceDirectories().from(...).

pyokagan:
@fzdy1914 
> So, I do not need to write the errors in the commit message?

Nevermind, ignore my comment in https://github.com/se-edu/addressbook-level4/pull/980#issuecomment-468650780 , since from your perspective the errors are important it is better to keep it that way.

pyokagan:
> See the warning message, it tells us to use getSourceDirectories().from(...).

I see, that's fine then. :-)

pyokagan:
> > See the warning message, it tells us to use getSourceDirectories().from(...).
> 
> I see, that's fine then. :-)

But the log messages need to be in the commit message.

fzdy1914:
> is incompatible with

fzdy1914:
> Hmm, the error doesn't occur when I use shadow plugin version 4.0.0? 🤔

Version 4.0.0 is enough when using Java 8. However, the support for java 11 comes in version 4.0.2. 
See the release [here](https://github.com/johnrengelman/shadow/releases/tag/4.0.2)

> I'm not quite sure how the linked issue relates to the error message?
> Can double check?

This issue is related to JDK11. Since we have reorder the commit, it is not related now anymore. 

> Seems like [johnrengelman/shadow#447](https://github.com/johnrengelman/shadow/issues/447) would be more relevant.

Correct, this is more related.

fzdy1914:
> ./gradlew wrapper --gradle-version 5.2.1
> ```
> Running it yields the following changes:
> 
> ```diff
> diff --git a/gradle/wrapper/gradle-wrapper.jar b/gradle/wrapper/gradle-wrapper.jar
> index f6b961fd..87b738cb 100644
> Binary files a/gradle/wrapper/gradle-wrapper.jar and b/gradle/wrapper/gradle-wrapper.jar differ
> diff --git a/gradlew b/gradlew
> index cccdd3d5..af6708ff 100755
> --- a/gradlew
> +++ b/gradlew
> @@ -28,7 +28,7 @@ APP_NAME="Gradle"
>  APP_BASE_NAME=`basename "$0"`
>  
>  # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
> -DEFAULT_JVM_OPTS=""
> +DEFAULT_JVM_OPTS='"-Xmx64m"'
>  
>  # Use the maximum available, or set MAX_FD != -1 to use that value.
>  MAX_FD="maximum"
> diff --git a/gradlew.bat b/gradlew.bat
> index f9553162..6d57edc7 100644
> --- a/gradlew.bat
> +++ b/gradlew.bat
> @@ -14,7 +14,7 @@ set APP_BASE_NAME=%~n0
>  set APP_HOME=%DIRNAME%
>  
>  @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
> -set DEFAULT_JVM_OPTS=
> +set DEFAULT_JVM_OPTS="-Xmx64m"
>  
>  @rem Find java.exe
>  if defined JAVA_HOME goto findJavaFromJavaHome
> ```
> 
> which needs to be in the commit.

Very weird, when I run this command on my side, it does not yield this change. I will try to investigate it.

build.gradle
```diff
@@ -66,13 +66,13 @@ shadowJar {
 }
 
 task coverage(type: JacocoReport) {
-    sourceDirectories = files(sourceSets.main.allSource.srcDirs)
-    classDirectories =  files(sourceSets.main.output)
-    executionData = files(jacocoTestReport.executionData)
+    sourceDirectories.from(files(sourceSets.main.allSource.srcDirs))
+    classDirectories.from(files(sourceSets.main.output))
+    executionData.from(files(jacocoTestReport.executionData))
```
pyokagan: Sorry, I didn't notice this last time, but it seems that our `build.gradle` file omits parentheses when allowed to. (See http://groovy-lang.org/style-guide.html#_omitting_parentheses)

So, this should be:

```groovy
sourceDirectories.from files(sourceSets.main.allSource.srcDirs)
```
?

As an aside, looking at https://docs.gradle.org/current/javadoc/org/gradle/api/file/ConfigurableFileCollection.html#from-java.lang.Object...- , we may not actually need to use `files()` here, but even if that's true *that is a topic for another commit*.

pyokagan:
@fzdy1914 
> Very weird, when I run this command on my side, it does not yield this change. I will try to investigate it.

The changes seem consistent with https://github.com/gradle/gradle/commit/58e5c9f696d74127d42540e00514881e7491fc9b , which was released in Gradle 5.0.0, so I think it's correct.

I think the command may need to be run twice. Once to upgrade Gradle to 5.2.1 (while running on Gradle 4.8.1), and another time to upgrade the wrapper while running on Gradle 5.2.1.

fzdy1914:
> @fzdy1914
> 
> > Very weird, when I run this command on my side, it does not yield this change. I will try to investigate it.
> 
> The changes seem consistent with [gradle/gradle@58e5c9f](https://github.com/gradle/gradle/commit/58e5c9f696d74127d42540e00514881e7491fc9b) , which was released in Gradle 5.0.0, so I think it's correct.
> 
> I think the command may need to be run twice. Once to upgrade Gradle to 5.2.1 (while running on Gradle 4.8.1), and another time to upgrade the wrapper while running on Gradle 5.2.1.

Today, I have tried the command again and the change happens. Sorry for the inconvenience caused.

fzdy1914:
> [1/4]
> 
> Very importantly, placement of citations need to be precise to prevent miscommunication with future readers.

Sorry, I parsed the sentence incorrect previously, I will fix it now.

fzdy1914:
> The [Gradle wrapper](https://docs.gradle.org/5.2.1/userguide/gradle_wrapper.html) document
> is _extremely_ long, so it would be good to quote the parts that you are
> referring to in the citation, so that future readers can easily find it by searching the document with Ctrl+F:

@pyokagan Can I know if the long quotation line needed to be separated or just remains in single line?

pyokagan:
@fzdy1914 
> @pyokagan Can I know if the long quotation line needed to be separated or just remain in a single line?

For text wrapping in commit messages we follow what [Linus says](https://github.com/torvalds/linux/pull/17#issuecomment-5660604):

> Yes, commit messages should have proper word-wrap, with empty lines in
between paragraphs, and at the same time sometimes you need a long
line without word-wrap (compiler error messages or other "non-prose"
explanation).

In other words, prose should be word-wrapped while non-prose should not be word-wrapped.

The Gradle wrapper document is prose so it should be word-wrapped, I think, but you can just follow your own judgement based on the rules that I outlined above :-)

pyokagan:
> > Yes, commit messages should have proper word-wrap, with empty lines in
> > between paragraphs, and at the same time sometimes you need a long
> > line without word-wrap (compiler error messages or other "non-prose"
> > explanation).

Addendum: this [message](https://github.com/torvalds/linux/pull/17#issuecomment-5661185) is more instructive:

> Word-wrapping is a property of the text. And the tool you use to
visualize things cannot know. End result: you do word-wrapping at the
only stage where you can do it, namely when writing it. Not when
showing it.

> Some things should not be word-wrapped. They may be some kind of
quoted text - long compiler error messages, oops reports, whatever.
Things that have a certain specific format.

fzdy1914:
> [1/4]
> 
> Very importantly, placement of citations need to be precise to prevent miscommunication with future readers.
> 
> > ```
> > build.gradle: remove wrapper task
> > 
> > Since Gradle 4.8 overriding built-in tasks, such as the `wrapper` task[1],
> > ```
> 
> `Since Gradle 4.8` -> `Since Gradle 4.8,`
> 
> > ```
> > has been deprecated[2] and since Gradle 5.0 would actually produce an
> > ```
> 
> `` such as the `wrapper` task[1] `` -> `` such as the `wrapper` task ``
> 
> `has been deprecated[2]` -> `has been deprecated[1]`
> 
> Footnote [1] is the one that talks about how "defining a custom `wrapper` or `init` task is
> deprecated", and so should be the only source cited here.
> 
> > ```
> > error:
> > ```
> 
> `error:` -> `error[2]:`
> 
> Footnote [2] is the one that talks about Gradle 5.0 producing an error,
> and so should only be cited here.
> 
> > ```
> > > Cannot add task 'wrapper' as a task with that name already exists
> > 
> > Specifying the gradle version used for executing the build via creating
> > a "wrapper" task is not required since Gradle will use the version
> > specified in the `gradle/wrapper/gradle-wrapper.properties` file, which
> > we already have[3].
> > ```
> 
> `` specified in the `gradle/wrapper/gradle-wrapper.properties` file `` -> `` specified in the `gradle/wrapper/gradle-wrapper.properties` file[3] ``
> 
> `we already have[3]` -> `we already have.`
> 
> Footnote [3] talks about the `gradle/wrapper/gradle-wrapper.properties` file,
> _not_ about whether we have it or not.
> 
> > ```
> > As we are planning to upgrade to Gradle 5.2.1, let's remove the wrapper
> > task.
> > 
> > [1] https://docs.gradle.org/4.8/release-notes.html#overwriting-gradle's-
> > built-in-tasks
> > ```
> 
> URLs need to be on a single line.
> 
> > ```
> > [2]https://docs.gradle.org/5.0/userguide/upgrading_version_4.html
> > [3]https://docs.gradle.org/5.2.1/userguide/gradle_wrapper.html
> > ```
> 
> The [Gradle wrapper](https://docs.gradle.org/5.2.1/userguide/gradle_wrapper.html) document
> is _extremely_ long, so it would be good to quote the parts that you are
> referring to in the citation, so that future readers can easily find it by searching the document with Ctrl+F:
> 
> > The generated Wrapper properties file, gradle/wrapper/gradle-wrapper.properties, stores the information about the Gradle distribution.
> > 
> > * The server hosting the Gradle distribution...
> > * The type of Gradle distribution...
> > * The Gradle version used for executing the build.

build.gradle
fzdy1914: Yes, it seems that it will call `file()` automatically

fzdy1914:
> @fzdy1914 One last thing: can you mention in [3/4] that you ran `./gradlew wrapper --gradle-version 5.2.1` to upgrade the version? That information is very important I think, otherwise readers may wonder why the wrapper was modified.
> 
> Also, did you need to run `./gradlew wrapper --gradle-version 5.2.1` once or twice in the end? That might be worth mentioning as well.

Once is enough, I found that I accidentally ran this command in another repo yesterday, that's the reason why my git did not catch the change.

pyokagan:
> ```
> build.gradle: upgrade shadow plugin to 5.0.0
> 
> The current shadow plugin we are using is 2.0.3, which is incompatible
> with Gradle 5.2.1 and gives the error below when running shadowJar
> task:
> 
>     Method com/github/jengelman/gradle/plugins/shadow/internal/DependencyFileCollection.getBuildDependencies()Lorg/gradle/api/tasks/TaskDependency; is abstract
> 
> As we are planning to upgrade to Gradle 5.2.1, let's upgrade the shadow
> plugin to its latest stable version 5.0.0 to solve the problem[1].
> ```

In this case it would be the latest version of the shadow plugin that still supports Gradle 4.x, which is shadow 4.0.4.

pyokagan:
[2/4]

> ```
> The current shadow plugin we are using is 2.0.3, which is incompatible
> with Gradle 5.2.1 and gives the error below when running shadowJar
> task:
> 
>     Method com/github/jengelman/gradle/plugins/shadow/internal/DependencyFileCollection.getBuildDependencies()Lorg/gradle/api/tasks/TaskDependency; is abstract
> 
> As we are planning to upgrade to Gradle 5.2.1, let's upgrade the shadow
> plugin to its latest version that still supports Gradle 4.x, version 4.0.4, to
> ```

This line exceeds 72 characters :-/

> ```
> solve the problem[1].
> ```

fzdy1914:
> This line exceeds 72 characters :-/

So I need to change it right?

pyokagan:
@fzdy1914 
> So I need to change it right?

No need, It's a minor issue. Just be careful.

PR #986, opened 2019-03-14
StringUtilTest: Remove Hamcrest String matcher

sijie123:
Currently, in StringUtilTest#getDetails_exceptionGiven, we use
Hamcrest's string matching and assert libraries.
This library was introduced in [1], however no documentation was
logged on why we specifically used these Hamcrest libraries. In fact,
this is the only instance in the entire project where we use this
dependency.

We are only using the contains API of the Hamcrest library, whose
corresponding functionality is already implemented in Java standard
String library. Hence, the Hamcrest library serves little extra
benefit, and introuduces extra dependencies for our tests.

Hence, let's remove the Hamcrest libraries and replace them with the
equivalent Java and JUnit APIs.

[1] https://github.com/se-edu/addressbook-level4/commit/75942c90c04ff0469d5bb26873b85cd094f37797

### DIFF
--- a/src/test/java/seedu/address/commons/util/StringUtilTest.java
+++ b/src/test/java/seedu/address/commons/util/StringUtilTest.java
@@ -1,7 +1,5 @@
 package seedu.address.commons.util;
 
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
@@ -145,8 +143,8 @@ public class StringUtilTest {
 
     @Test
     public void getDetails_exceptionGiven() {
-        assertThat(StringUtil.getDetails(new FileNotFoundException("file not found")),
-                   containsString("java.io.FileNotFoundException: file not found"));
+        assertTrue(StringUtil.getDetails(new FileNotFoundException("file not found"))
+            .contains("java.io.FileNotFoundException: file not found"));
     }
 
     @Test


src/test/java/seedu/address/commons/util/StringUtilTest.java
```diff
@@ -145,8 +143,8 @@ public void containsWordIgnoreCase_validInputs_correctResult() {
 
     @Test
     public void getDetails_exceptionGiven() {
-        assertThat(StringUtil.getDetails(new FileNotFoundException("file not found")),
-                   containsString("java.io.FileNotFoundException: file not found"));
+        assertTrue(StringUtil.getDetails(new FileNotFoundException("file not found"))
+            .contains("java.io.FileNotFoundException: file not found"));
```
pyokagan: According to https://github.com/junit-team/junit4/wiki/matchers-and-assertthat , using `assertThat()` gives us a more detailed error message if the assertion fails, compared to using `assertTrue()` which would just give us a generic "expected true, but got false".

Should we be concerned about that?

pyokagan:
> Eh? Why did we even have access to this library in this first place?
> I don't see it being listed as a dependency in `build.gradle`.

To clarify: I'm confused at the use of "dependency" being used in the commit message, since JUnit4 guarantees that the Hamcrest library is available for use with JUnit4's assertions. While https://github.com/se-edu/addressbook-level4/commit/75942c90c04ff0469d5bb26873b85cd094f37797 introduced the first use of the Hamcrest matchers API, it doesn't really "introduce the library", since it was always available.

Now, I do have a hunch why the word "dependency" is being used here, but I guess it's better for me not to carelessly assume :-)

src/test/java/seedu/address/commons/util/StringUtilTest.java
sijie123: That is certainly a concern - I will add it into the commit message to reflect this. Thanks!

However, I feel that seeing a more detailed error message if the assertion fails is not absolutely critical. If the assertion had failed, a developer can see the cause by simply adding a debugger to log the state.

I see value in having more informative error message, but I would think that consistency is more valuable here. It's odd to use assertThat for everything except for this test, and there's nothing too special about this test to justify using an entirely different library.

As a very-very-future extension, it might make sense to apply assertThat to the entire AB4 test suite if it helps developers code with clarity. Right now, I think it would make sense to remove this lone assertThat statement to bring it in line with the rest of the code.

pyokagan:
@sijie123 
> I might have used the term dependency a bit loosely here - I'll correct it in the commit log.

This didn't really seem to be done? Well, not a really a big deal since anyone with some background knowledge will still understand what the commit message is talking about. Anyway, the important parts are in.

> ^ This paragraph won't be in the commit message, because right now we assume that we're still working with JUnit 4.

I see you understand how we do things around here :stuck_out_tongue:

PR #959, opened 2019-02-01
Migrate from JUnit 4 to JUnit 5

sijie123:
Our tests use JUnit4. 

JUnit 5 provides many APIs that simplify our tests. For example, instead of a lengthy try/catch procedure to test for correct exception handling behaviour, we can now use assertThrows. This makes it easier for readers to understand our code, and to implement more tests.

Let's upgrade to use JUnit 5.

Resolves #874 

Progress/To-do:
[x] Change imports to JUnit5 version. (e.g. org.junit.jupiter.api.Test instead of org.junit.Test)
[x] Remove Test being used for specifying expectations. We use this to specify timeout in PersonListPanelTest. Instead we should now use assertTimeoutPreemptively. (Done in #863)
[x] Rename @Before to @BeforeEach
[x] Rename @After to @AfterEach
[x] Rename @BeforeClass to @BeforeAll
[x] Rename @AfterClass to @AfterAll
[-] Group assertions together using assertAll for related assertions (Not sure if we want to do this in this PR? Leaving it as separate asserts right now as feedback to user is clearer when tests fail)
[x] Remove @Rule and @ClassRule. Instead, we need to @ExtendWith, which generally requires creating an extension class.

PR #1009, opened 2019-04-25
TemporaryFolder: Remove unnecessary call to TemporaryFolder#newFile

sijie123:
LogicManagerTest and MainWindowCloseTest calls newFile() on a
temporary folder to initialize an empty file for the AddressBoook and
UserPrefs storage.

This is unnecessary, as the file is automatically created the first
time it is written to.

Instead of explicitly creating a new file, let's simply reference a
file path with Path#resolve instead.

Reference: https://github.com/se-edu/addressbook-level4/pull/959#discussion_r276946591


sijie123:
Closed with changes made in #959 instead.

PR #157, opened 2016-10-15
Document checkstyle use #156

m133225:
Fixes #156.

<s>Still need to verify if reimporting works smoothly.</s>

PR #1011, opened 2019-05-11
.gitignore: standardize and specify entries

### LINKED ISSUE #926: .gitignore: remove config.json
Can we remove `config.json` from the `.gitignore` file?

Reason: students need to add a `_reposense/config.json` for their forks.

fzdy1914:
Fixes #926.

We ignore config.json and preferences.json by adding entries in
.gitignore.

The entries added are too general so git will also ignore other files
with same name, which is an unwanted behavior.

Let's specific the entries in .gitignore to solve the problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,20 +1,21 @@
-build/
-.gradle/
-.checkstyle
-.settings/
-.idea/
-lib/*
-*.iml
-*.log
-*.log.*
-*.csv
-config.json
-src/test/data/sandbox/
-preferences.json
-.DS_Store
-./screenshot*.png
-classes/
-/data/
-/bin/
+# Gradle build files
+/.gradle/
+/build/
 src/main/resources/docs/
-out/
+
+# IDEA files
+/.idea/
+/out/
+/*.iml
+
+# Storage/log files
+/data/
+/config.json
+/preferences.json
+/*.log.*
+
+# Test sandbox files
+src/test/data/sandbox/
+
+# MacOS custom attributes files created by Finder
+.DS_Store


.gitignore
```diff
@@ -1,20 +1,15 @@
...
+/.gradle/
+/.idea/
+/.settings/
+src/main/resources/docs/
+src/test/data/sandbox/
+/config.json
+/preferences.json
+.DS_Store
 *.iml
 *.log
```
pyokagan: I believe this is meant to ignore AB log files.  However, AB log files are always of the format `addressbook.log.0`, `addressbook.log.1` etc. So, this line can be removed?

.gitignore
```diff
@@ -1,20 +1,15 @@
...
 *.log.*
```
pyokagan: I believe this is meant to ignore AB log files. However, we only support running AB from the repo root. So, this should be `/*.log.*`?

A more specific pattern would be `/addresbook.log.*`, however given that it is likely that students would change the name of the application (and the log filename, since they would likely be doing a mass find-and-replace), `/*.log.*` would be acceptable as well I think. (Then again, that find-and-replace would also replace `/addressbook.log.*` as well? :thinking: )

fzdy1914: If they choose the scope of mass find-and-replace to be `src` to avoid changing the docs, this file name won't be replaced. It may be better to just leave it to be  `/*.log.*`

.gitignore
```diff
@@ -1,20 +1,15 @@
-build/
-.gradle/
-.checkstyle
-.settings/
-.idea/
-lib/*
+/build/
+/data/
+/out/
+/.checkstyle/
```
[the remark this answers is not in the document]
fzdy1914: I have tried using IntelliJ's checkstyle plugin and this folder is not created. After blaming this file, I found it was added in https://github.com/se-edu/addressbook-level4/pull/157, which I strongly suspect, this is generated by eclipse's checkstyle plugin. So I will just remove it.

PR #1012, opened 2019-05-11
Fix improper usage of `an`

### LINKED ISSUE #357: Typo in some command parser parse(String)'s header comments
For example, in `SelectCommandParser`:

```java
    /**
     * Parses the given {@code String} of arguments in the context of the SelectCommand
     * and returns an SelectCommand object for execution.
     */
```

It should be `returns a`, not `returns an`.

This happens in other command parsers as well.

fzdy1914:
Fixes #357. 

There are some typos in our codebase related to the usage of `an`.

Let's fix the improper usage of `an` to solve the problem.

### DIFF
--- a/src/main/java/seedu/address/logic/parser/DeleteCommandParser.java
+++ b/src/main/java/seedu/address/logic/parser/DeleteCommandParser.java
@@ -13,7 +13,7 @@ public class DeleteCommandParser implements Parser<DeleteCommand> {
 
     /**
      * Parses the given {@code String} of arguments in the context of the DeleteCommand
-     * and returns an DeleteCommand object for execution.
+     * and returns a DeleteCommand object for execution.
      * @throws ParseException if the user input does not conform the expected format
      */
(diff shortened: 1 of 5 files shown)


fzdy1914:
> Still, I just happened to spot this instance that was missed out:
> 
> https://github.com/se-edu/addressbook-level4/blob/10586cb7fa6af793bf274fa58f5d667ab4d0caa8/src/test/java/seedu/address/ui/TestFxmlObject.java#L6
> 
> Perhaps we could fix this as well in the same commit?

From my understanding, it is not a misuse of `an`. The beginning sound of `FXML` actually needs `an`. 

Attached image:
{EADCDC52-36FF-DE50-1695-B8E1278838E4}

pyokagan:
> {EADCDC52-36FF-DE50-1695-B8E1278838E4}

What application is that? :-O

Okay, lets go with this then.

fzdy1914:
> What application is that? :-O

It is Grammarly.

PR #1013, opened 2019-05-11
Rename Model#getFilteredPersonList() and Logic#getFilteredPersonList()

### LINKED ISSUE #827: Rename Logic#getFilteredPersonList()
`Logic` doesn't need to expose to the world that it uses a filtered list internally.
Furthermore, the `UI` doesn't need to know what kind of list to bind itself too, as long as it is an `ObservableList`.

fzdy1914:
Fixes #827.

We use method with name `getFilteredPersonList()` to get the list to
display in UI.

The method name should follow the terminology used by the application
domain. And our User Guide (which should be using terms from the
application domain since its meant for users) actually uses the term
"displayed person list".

Let's rename `getFilteredPersonList()` to `getDisplayPersonList()` to
follow the terminology.

### DIFF
--- a/src/main/java/seedu/address/logic/Logic.java
+++ b/src/main/java/seedu/address/logic/Logic.java
@@ -31,8 +31,8 @@ public interface Logic {
      */
     ReadOnlyAddressBook getAddressBook();
 
-    /** Returns an unmodifiable view of the filtered list of persons */
-    ObservableList<Person> getFilteredPersonList();
+    /** Returns an unmodifiable view of the filtered list of persons to be displayed. */
+    ObservableList<Person> getDisplayPersonList();
 
     /**
      * Returns an unmodifiable view of the list of commands entered by the user.
--- a/src/main/java/seedu/address/logic/LogicManager.java
+++ b/src/main/java/seedu/address/logic/LogicManager.java
@@ -72,8 +72,8 @@ public class LogicManager implements Logic {
     }
 
     @Override
-    public ObservableList<Person> getFilteredPersonList() {
-        return model.getFilteredPersonList();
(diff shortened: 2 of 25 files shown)


pyokagan:
Ehh, I'm still on the fence about #827 . Let me think... 

@fzdy1914 Do you personally think this is something that should be done?

fzdy1914:
@pyokagan Personally, I do not have a strong opinion of the name of this method. I think the previous name is also acceptable. So, it is not a must-need-done change.

Zhiyuan-Amos:
Comments from [issue](https://github.com/se-edu/addressbook-level4/issues/827#issuecomment-412888304):

> FilteredPersonList simply means that it is returning a list that can be indirectly manipulated by the users of the Logic API (through another method updateFilteredPersonList(Predicate<Person>).

> The UI does need to know which list to bind itself to. The application won't be very useful if the UI bound itself to the model.getAddressBook().getPersonList() list instead

I agree with these comments. I'm wondering whether `getFilteredPersonList()` suggests that the list of persons can be filtered, which couples the `Logic` component to the existence of `FindCommand` (iirc, only `FindCommand` performs the filtering), thus justifying the renaming of this method to a more generic name `getDisplayPersonList()`.

On a side note, I think we could write clearer header comments in `Logic` to suggest how this `list` can be modified.

pyokagan:
@Zhiyuan-Amos I replied in https://github.com/se-edu/addressbook-level4/issues/827 (so we can keep the discussion in one place). I actually think the name "displayed person list" might be *worse* than just keeping it at the current status quo, despite the name "filtered person list" not being ideal.

PR #1015, opened 2019-05-28
diagrams: rename .pptx files to match their .png files

pyokagan:
Some of the pptx files in the `docs/diagrams/` directory do not have the
same base filename as their PNG file counterparts in the `docs/images/`
directory. For example, the PNG file counterpart for
`docs/diagrams/HighLevelSequenceDiagrams.pptx` is
`docs/images/SDforDeletePerson.png`.

Such inconsistent naming can confuse developers when updating the
diagrams. This unfortunately already happened before in 5509159e
(diagrams: restore original LogicComponentSequenceDiagram.pptx,
2019-02-27).

As such, let's rename the pptx files so that their filenames match their
corresponding PNG files. For example, we rename
`docs/diagrams/HighLevelSequenceDiagrams.pptx` to
`docs/diagrams/SDforDeletePerson.pptx` so that it matches
`docs/images/SDforDeletePerson.png`.


pyokagan:
Verified using:
```sh
cd docs/diagrams && for x in *.pptx; do test ! -f "../images/${x%*.pptx}.png" && echo "NOT FOUND: $x"; done
```

PR #1021, opened 2019-07-22
Migrate from PowerPoint to PlantUml

j-lum:


docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+==== Link lengths
+By default, a short link (`->`) points to right and a long link (`-->`)
+points downwards. you can extend any link to make it longer (`-->`).
+
+.Length of arrows and its effects
+image::ArrowLength.png[]
+
+==== Link directions
+Clever usage of arrow directions will resolve most layout issues.
```
damithc: The sentence is rather 'mysterious'. :-p Add a follow up sentence? 
> For example, the table below shows how the way in which you specify arrows can results in drastically different layouts for the same diagram.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+package "Rule Of Thumb"{
+    Class C
+    A --> B
+    C --> D
+}
+----
+|image::PackagesAndConsistency.png[]
+|===
+
+We strongly recommend that you explicitly define all symbols to avoid any
```
damithc: ~~strongly~~

damithc: Or, format as a tip.
> :bulb: Explicitly define all symbols to avoid any ...

docs/Documentation.adoc
```diff
@@ -26,7 +26,11 @@ We chose asciidoc over Markdown because asciidoc, although a bit more complex th
 See <<UsingGradle#rendering-asciidoc-files, UsingGradle.adoc>> to learn how to render `.adoc` files locally to preview the end result of your edits.
```
damithc: Commit message: ~~acceptable~~ diagrams.

docs/DeveloperGuide.adoc
```diff
@@ -26,12 +26,13 @@ Refer to the guide <<SettingUp#, here>>.
 === Architecture
 
 .Architecture Diagram
-image::Architecture.png[width="600"]
+image::ArchitectureDiagram.png[]
 
 The *_Architecture Diagram_* given above explains the high-level design of the App. Given below is a quick overview of each component.
```
damithc: Diagrams are missing in https://deploy-preview-1021--seedu-ab4.netlify.com/developerguide

docs/UsingPlantUml.adoc
```diff
@@ -47,6 +47,15 @@ Any changes you make in editor pane on the left will be reflected in the preview
...
 //TODO: Discussion about why we're not using asciidoctor-diagram
 
+== Export UML diagrams
+
+A helpful Gradle task `:plantUml` is provided to generate all diagrams in the `docs/diagrams` directory and output them in the `.png` format under `docs/images` directory.
+
+NOTE: You will have to `git add` any new diagrams generated!
+
+When editing your documentation, you might find it useful to have Gradle monitor changes to your PlantUML diagrams and and automatically export them.
+You can do this by running `gradlew -t plantUml` in a terminal.
```
damithc: Can also mention the meaning of `-t`

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+ifdef::env-github[]
+:tip-caption: :bulb:
+:note-caption: :information_source:
+endif::[]
+
+== Introduction to PlantUML
+
+PlantUML is a popular tool in this project to create UML diagrams.
+For more information about the basics of PlantUML, head over to
+link:http://plantuml.com/[its official website].
```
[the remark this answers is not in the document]
j-lum: Changed it to an shorthand link like the rest.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+Graphviz is a dependency that PlantUML requires to generate more advanced diagrams.
+Head over to the https://www.graphviz.org/download/[downloads page] on the offcial Graphviz website and follow instructions to install Graphviz.
+
+=== Installing the `PlantUML integration` plugin for IntelliJ IDEA
+
+While this is not compulsory, the `PlantUML integration` provides a live preview of the PlantUML file you are working on.
+
+.Settings - Plugins - Marketplace: install the `PlantUML integration` plugin
+image::PlantUmlPlugin.png[]
```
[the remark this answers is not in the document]
j-lum: Removed.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+=== Configuring `PlantUML integration`
```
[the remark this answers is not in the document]
j-lum: Combined with previous section.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+Go to `Settings` > `Other Settings` > `PlantUML` or search for PlantUML.
+Configure the path to the `dot` executable.
```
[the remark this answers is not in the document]
j-lum: Specified that it comes with GraphViz and its location.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,209 @@
...
+
+.Settings - Other Settings - PlantUML: input the path to your dot executable
+image::ConfiguringGraphviz.png[]
+
+== Create/Edit PlantUML diagrams
+
+After installing the `PlantUML integration` plugin, simply create or open any `.puml` file to start editing it.
+
+.Editing `DeleteSequenceDiagram.puml`
+image::OpenPumlFile.png[]
```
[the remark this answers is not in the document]
j-lum: Refactored.

docs/UsingPlantUml.adoc
j-lum: Added the line in.

docs/UsingPlantUml.adoc
j-lum: Converted to a tip.

docs/UsingPlantUml.adoc
j-lum: Explained continuous build mode.

docs/Documentation.adoc
j-lum: Amended.

docs/DeveloperGuide.adoc
j-lum: Something's wrong with netlify, it keeps pulling version 8059 instead of version 1.2019.8.
Think it's ignoring the buildscript or something. Will remove the dependsOn for now.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+You can fine-tune the formatting of PlantUML diagrams with the `skinparam` command.
+For example, `skinparam backgroundColor transparent` turns the background of the diagram transparent.
+
+For a comprehensive list of ``skinparam``s head over to the link:https://plantuml-documentation.readthedocs.io/en/latest/[unofficial PlantUML skinparam documentation].
+
+***
+
+=== Repositioning elements in PlantUML diagrams
+
+While PlantUML's automatic layout engine usually produces satisfactory results, it is not uncommon to get a lemon, especially on larger diagrams.
```
damithc: add `Here is an example where the default layout generated by PlantUML has a lot of overlapping lines that are hard to decipher`:`.

damithc: `it is not uncommon to get a lemon` -> `at times the result can be not less than ideal`

damithc: Remove Logic and Model (they don't contribute to the point being discussed)?

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+.The UI class diagram without additional formatting
+image::RawUiDiagram.png[]
+
+NOTE: In most cases, you should consider decomposing the diagram into smaller ones or focusing on a more specific portion of the diagram.
+
+Here are some of the techniques we used in this project to obtain a more palatable diagram.
+
+==== Link lengths
+By default, a short link (`->`) points to right and a long link (`-->`)
+points downwards. you can extend any link to make it longer (`-->`).
```
damithc: Unnecessary line breaks?
`a short link` -> `a short arrow`?
image
Looks like asciidoc converts it to a proper arrow, which may not be what we want here.

docs/diagrams/ArchitectureDiagram.puml
```diff
@@ -0,0 +1,40 @@
...
+
+Logs -up- Commons
+Hidden .down.> Commons
+Hidden .down.> Commons
+Hidden .down.> Commons
+
+Storage .right.>File
+User --> UI
+hide circle
+@enduml
```
damithc: In the original diagram, Main is not connected to any other component. There is a bunch of outgoing arrows to indicate that it is _connected to everything_.

docs/diagrams/LogicClassDiagram.puml
```diff
@@ -0,0 +1,65 @@
+@startuml
```
damithc: Some differences noted by comparing original to new diagram. I may have missed some. 

image
* `CommandResult` missing
* It's easier to understand if `Parser` is above `XYZCommandParser`
* `creates` should be a dashed arrow
* `Prefix` should have three incoming arrows
* `<<abstract>>` -> `{abstract}`
* `XYZCommand` should inherit, not depend on, `Command`

docs/diagrams/UiClassDiagram.puml
```diff
@@ -0,0 +1,67 @@
+@startuml
```
damithc: image

* Missing `--->` to `Ui`
* `<<abstract>>` -> `{abstract}`

docs/diagrams/DeleteSequenceDiagram.puml
```diff
@@ -0,0 +1,69 @@
+@startuml
```
damithc: image

* Background colors can be much lighter.
* `DeleteCommandParser` lifeline should not go beyond the `X`. Furthermore, `X` should not be attached to the activation box.
* Some return arrows are given as solid lines (should be dashed)

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+.Editing `DeleteSequenceDiagram.puml`
+image::EditingDeleteSequenceDiagram.png[]
+Any changes you make in editor pane on the left will be reflected in the preview pane on the right.
+However, do take note that these changes _will not_ be reflected in the developers guide until you export the diagram.
+//TODO: Discussion about why we're not using asciidoctor-diagram
+
+== Export UML diagrams
+
+A helpful Gradle task `:plantUml` is provided to generate all diagrams in the `docs/diagrams` directory and output them in the `.png` format under `docs/images` directory.
```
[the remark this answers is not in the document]
j-lum: The colon was workaround some issues on older versions of Gradle. Since we're going to bump the version to v5.5 we can remove it.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+After installing the `PlantUML integration` plugin, simply create or open any `.puml` file to start editing it.
+
+.Editing `DeleteSequenceDiagram.puml`
+image::EditingDeleteSequenceDiagram.png[]
+Any changes you make in editor pane on the left will be reflected in the preview pane on the right.
+However, do take note that these changes _will not_ be reflected in the developers guide until you export the diagram.
+//TODO: Discussion about why we're not using asciidoctor-diagram
+
+== Export UML diagrams
```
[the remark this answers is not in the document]
j-lum: Changed.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+...
+!define LOGIC_COLOR #3333C4
+!define LOGIC_COLOR_T1 #7777DB
+!define LOGIC_COLOR_T2 #5252CE
+!define LOGIC_COLOR_T3 #1616B0
+!define LOGIC_COLOR_T4 #101086
+...
+----
+
+and use it in another PlantUML file like this:
```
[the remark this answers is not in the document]
j-lum: Split them into two sentences.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+You can do this by running `gradlew -t plantUml` in a terminal.
+The `-t` flag enables _continuous build_ which will watch for changes to files and execute specified tasks.
+In this case, gradle will run the task `plantUml` when it detects changes to your PlantUML diagrams.
+
+== Common tasks
+
+=== Applying consistent formatting to PlantUML diagrams
+
+It is highly recommended to consistently color your UML diagrams as an visual aid.
+You can achieve this by creating a dictionary of colors and import it à la CSS
```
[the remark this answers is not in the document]
j-lum: Changed to `like.`

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+== Export UML diagrams
+
+A helpful Gradle task `:plantUml` is provided to generate all diagrams in the `docs/diagrams` directory and output them in the `.png` format under `docs/images` directory.
+
+NOTE: You will have to `git add` any new diagrams generated!
+
+When editing your documentation, you might find it useful to have Gradle monitor changes to your PlantUML diagrams and and automatically export them.
+You can do this by running `gradlew -t plantUml` in a terminal.
+The `-t` flag enables _continuous build_ which will watch for changes to files and execute specified tasks.
+In this case, gradle will run the task `plantUml` when it detects changes to your PlantUML diagrams.
```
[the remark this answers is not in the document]
j-lum: Tip-ified.

docs/diagrams/UiClassDiagram.puml
j-lum: The abstract part is a bit hacky since puml doesn't support the {abstract} syntax.
`"{abstract}\nUiPart" as UiPart` seems render correctly.

docs/diagrams/DeleteSequenceDiagram.puml
j-lum: There's actually no way to end a lifeline early. 
 
We can hack around the X positioning by issuing a `-[hidden]->` from DeleteCommandParser before the destroy though. Used the hack and indicated it in the comments.

Changed style.puml to feature really light colors for T1 and T2.

docs/diagrams/LogicClassDiagram.puml
j-lum: - Added `CommandResult`
- I understand where you're coming from but the diagrams is pretty volatile now, we should probably get this split up soon.
- Dashed create arrow
- Added missing CliSyntax ..> Prefix
- Fix relationship.

This diagram is always one hidden arrow away from collapsing into an unreadable arrow singularity, I'll plaster over it for now but we should really go back to the old plan of breaking it up into 3 separate diagrams. I'll go do the write up soon.

j-lum:
- Change the arrow diagram to feature the 4 arrow lengths side by side.
- Color code all the examples to make it easier to grok.
- Since we're no longer in-lining diagrams, extracted them out into their own folder.

docs/diagrams/ArchitectureDiagram.puml
damithc: Not addressed?

j-lum: I missed this for some reason. 

In the original diagram main has 3 stubby arrows pointing to, but not reaching UI, Model and Commons. Do we just leave main dangling without arrows in the new version or do we want dummy arrows pointing to invisible classes?

docs/DeveloperGuide.adoc
```diff
@@ -168,29 +169,29 @@ Given below is an example usage scenario and how the undo/redo mechanism behaves
 
 Step 1. The user launches the application for the first time. The `VersionedAddressBook` will be initialized with the initial address book state, and the `currentStatePointer` pointing to that single address book state.
 
-image::UndoRedoStartingStateListDiagram.png[width="800"]
+image::UndoRedoState0.png[]
 
 Step 2. The user executes `delete 5` command to delete the 5th person in the address book. The `delete` command calls `Model#commitAddressBook()`, causing the modified state of the address book after the `delete 5` command executes to be saved in the `addressBookStateList`, and the `currentStatePointer` is shifted to the newly inserted address book state.
 
-image::UndoRedoNewCommand1StateListDiagram.png[width="800"]
+image::UndoRedoState1.png[]
```
damithc: image

The object box needs a border; otherwise it looks like highlighted text.

This placement of `delete 5` is not meaningful. Remove it altogether and add above the diagram 
> After executing command `delete 5`:

Also, remove `=0`, `=1`.
Same for the other diagrams in this sequence.

docs/diagrams/UndoSequenceDiagram.puml
```diff
@@ -0,0 +1,56 @@
+@startuml
+!include style.puml
+
+box Logic LOGIC_COLOR_T1
+participant ":LogicManager" as LogicManager LOGIC_COLOR
+participant ":AddressBookParser" as AddressBookParser LOGIC_COLOR
+participant ":UndoCommand" as UndoCommand LOGIC_COLOR
```
damithc: should be `u:UndoCommand`

docs/diagrams/ArchitectureDiagram.puml
damithc: Ideally, have some dummy arrows.

build.gradle
```diff
@@ -227,6 +237,13 @@ task deployOfflineDocs(type: Copy) {
     }
 }
 
+apply plugin: 'com.cosminpolifronie.gradle.plantuml'
```
pyokagan: Is this needed? I seem to remember that the `plugins` block does the same thing as `apply plugin`.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,214 @@
...
+participant ":LogicManager" as LogicManager LOGIC_COLOR
+participant ":AddressBookParser" as AddressBookParser LOGIC_COLOR
+participant ":UndoCommand" as UndoCommand LOGIC_COLOR
+end box
+----
+
+You can fine-tune the formatting of PlantUML diagrams with the `skinparam` command.
+For example, `skinparam backgroundColor transparent` turns the background of the diagram transparent.
+
+For a comprehensive list of ``skinparam``s head over to the link:https://plantuml-documentation.readthedocs.io/en/latest/[unofficial PlantUML skinparam documentation].
```
pyokagan: We can just use `https://plantuml-documentation.../[unofficial PlantUML...]` without the `link:` here.

docs/diagrams/LogicClassDiagram.puml
```diff
@@ -0,0 +1,70 @@
...
```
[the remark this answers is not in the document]
j-lum: `LogicManager` instantiates an `AddressBookParser` in its constructor and uses it in `execute()`. Then the `AddressBookParser`  instance is used to create a `XZYCommandParser`. 

But my UML is a bit rusty so I could be wrong here.

docs/diagrams/LogicClassDiagram.puml
```diff
@@ -0,0 +1,70 @@
...
+!include style.puml
```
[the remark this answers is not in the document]
j-lum: Removed the [width=800] on it.

docs/diagrams/DeleteSequenceDiagram.puml
```diff
@@ -0,0 +1,71 @@
...
+!include style.puml
```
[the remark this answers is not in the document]
j-lum: Added a note to both of the sequence diagrams with destroys

docs/diagrams/BetterModelClassDiagram.puml
```diff
@@ -0,0 +1,24 @@
+@startuml
```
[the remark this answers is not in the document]
j-lum: There's some limitations to arrow lengths when directions are specified so I rearranged the entire diagram.

docs/DeveloperGuide.adoc
j-lum: Added titles to all states, removed the relation text, changed CurrentState = x to "Current state".

docs/diagrams/UndoSequenceDiagram.puml
j-lum: Named u.

build.gradle
j-lum: You're right, it's not. It was a left over from when I was figuring out why builds broke on netlify.

docs/UsingPlantUml.adoc
j-lum: Removed the link:

pyokagan:
@j-lum Please clean up the commits. No "implemented requested changes" commits are allowed.

j-lum:
Squashed!

[Developer Guide](https://deploy-preview-1021--seedu-ab4.netlify.com/documentation)

[Using PlantUML](https://deploy-preview-1021--seedu-ab4.netlify.com/usingplantuml)

Will squash every iteration from now on, quite painful to do it all at once I'm afraid.

docs/Documentation.adoc
```diff
@@ -26,7 +26,11 @@ We chose asciidoc over Markdown because asciidoc, although a bit more complex th
...
 Alternatively, you can download the AsciiDoc plugin for IntelliJ, which allows you to preview the changes you have made to your `.adoc` files in real-time.
 
-== Publishing Documentation
+=== Editing Diagrams
+
+See <<UsingPlantUml#, UsingPlantUml.adoc>> to find out how to create and update the UML diagrams in the developer guide.
+
+=== Publishing Documentation
```
pyokagan: We are making this a level 3 header here, when it was previously level 2. Most of the headings in this document are level 2. Intended?

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+.Editing `DeleteSequenceDiagram.puml`
+image::EditingDeleteSequenceDiagram.png[]
+Any changes you make in editor pane on the left will be reflected in the preview pane on the right.
+However, do take note that these changes _will not_ be reflected in the developers guide until you export the diagram.
+//TODO: Discussion about why we're not using asciidoctor-diagram
+
+== Export PlantUML diagrams
+
+A Gradle task `plantUml` is provided to generate all diagrams in the `docs/diagrams` directory and output them in the `.png` format under `docs/images` directory.
```
pyokagan: Doesn't make sense to have this piece of text in this commit if the `plantUml` task does not even exist yet.

I do believe `UsingGradle.adoc` also needs to be updated with this new task.

j-lum: Oops must have squashed it into the wrong commit by accident.

docs/Documentation.adoc
j-lum: I was thinking that the editing-publish loop should stay in a section by itself. Thoughts?
An alternative would be to have all of them at level 2 but that would be quite visually noisy.

build.gradle
```diff
@@ -227,6 +237,11 @@ task deployOfflineDocs(type: Copy) {
     }
 }
 
+plantUml {
+    render input: 'docs/diagrams/*.puml', output: 'docs/images', format: 'png'
+    render input: 'docs/diagrams/plantuml/*.puml', output: 'docs/images/plantuml', format: 'png'
```
pyokagan: There doesn't seem to be any files in `docs/diagrams/plantuml/`. What do we need this line for?

build.gradle
```diff
@@ -185,22 +195,22 @@ asciidoctor {
...
+            'source-highlighter': 'coderay',
+            icons: 'font',
+            experimental: true,
+            sectlinks: true,
+            idprefix: '',  // for compatibility with GitHub preview
+            idseparator: '-',
+            'site-root': "${sourceDir}",  // must be the same as sourceDir, do not modify
+            'site-name': 'AddressBook-Level4',
+            'site-githuburl': 'https://github.com/se-edu/addressbook-level4',
+            'site-seedu': true,  // delete this line if your project is not a fork (not a SE-EDU project)
```
pyokagan: Unnecessary change.

build.gradle
```diff
@@ -185,22 +195,22 @@ asciidoctor {
     outputDir "${buildDir}/docs"
 
     options = [
-        template_dirs: [file("${sourceDir}/templates")],
+            template_dirs: [file("${sourceDir}/templates")],
```
pyokagan: Unnecessary change.

build.gradle
```diff
@@ -167,10 +177,10 @@ task headless {
...
         test {
             systemProperties = [
-                'testfx.robot': 'glass',
-                'testfx.headless': 'true',
-                'prism.order': 'sw',
-                'prism.text': 't2k',
+                    'testfx.robot': 'glass',
+                    'testfx.headless': 'true',
+                    'prism.order': 'sw',
+                    'prism.text': 't2k',
```
pyokagan: Unnecessary change.

build.gradle
```diff
@@ -4,10 +4,20 @@
...
 
+buildscript {
+    dependencies {
+        classpath('com.cosminpolifronie.gradle:gradle-plantuml-plugin:1.6.0') {
+            exclude group: 'net.sourceforge.plantuml', module: 'plantuml'
+        }
+        classpath 'net.sourceforge.plantuml:plantuml:1.2019.7'
+    }
+}
+
```
pyokagan: I don't think we need this block if we are using the `plugins` block below.

pyokagan:
Please hold, I haven't taken a look at [2/3] yet and I need to go off now.

The `plantUml` task is currently giving me output like this, haven't had time to investigate yet:

ArchitectureDiagram

j-lum:
The plantuml plugin seems a bit flaky on netlify as well. The whole `buildscript` block was to pin down a version of `plantuml` which appeared to resolve the issues at least a while. 

Let's just remove the task and go back to using the plantuml jar and manual copy-pasting until I can figure out what's going wrong with the plugin. I'll just purge the last commit and rewrite some stuff.

fzdy1914:
@j-lum Bumping Gradle version is quite a dangerous operation. I suggest do it in a separate PR / or at least separate commit.  Do verify all Gradle functionalities works in 5.5. One high-risk task is `shadowJar`

j-lum:
@fzdy1914 Yeah I've been seeing a lot of depreciated stuff in our gradle build. Since we're not using the plantuml plugin anymore I won't be bumping it.

docs/UsingPlantUml.adoc
j-lum: Never mind, nuking the commit that adds the task.

build.gradle
j-lum: Oops, added the diagrams in together with the rest. Missed it out during the squash.

build.gradle
j-lum: Nuked the commit that made this change.

build.gradle
j-lum: Explained below but no point having a convenience function that breaks the build randomly for some people. We should probably consider rolling our own if we have the time and manpower.

docs/Documentation.adoc
j-lum: Actually, let's just go with convention here and promote everything to level 2. Not much time left so let's squeeze this out before the sem starts.

pyokagan:
> The `plantUml` task is currently giving me output like this, haven't had time to investigate yet:
> 
> ArchitectureDiagram

So I guess it's because plantuml tries to resolve includes relative to the current working directory, which is whatever current working directory the Gradle daemon happens to have.

One way this could be fixed is to somehow explicitly tell plantuml to search in the `docs/diagrams` directory instead. This configuration option is something that the plantuml gradle plugin really should be supporting, but doesn't, so we would need to do something icky like this:

```diff
diff --git a/build.gradle b/build.gradle
index b56b9453..6c0e6412 100644
--- a/build.gradle
+++ b/build.gradle
@@ -237,6 +237,8 @@ task deployOfflineDocs(type: Copy) {
     }
 }
 
+System.setProperty("plantuml.include.path", file("docs/diagrams").toString())
+
 plantUml {
     render input: 'docs/diagrams/*.puml', output: 'docs/images', format: 'png'
     render input: 'docs/diagrams/plantuml/*.puml', output: 'docs/images/plantuml', format: 'png'
```

This only works because the gradle plugin (fortunately) sets the isolation mode to none, so Gradle uses the same Java process that runs the build script:

https://github.com/cosminpolifronie/gradle-plantuml-plugin/blob/3b5d5762597a9eccb1874f6854528ff0a5fd8c83/src/main/groovy/com/cosminpolifronie/gradle/plantuml/tasks/PlantUmlTask.groovy#L79

docs/diagrams/UiClassDiagram.puml
```diff
@@ -0,0 +1,70 @@
...
+CommandBox -d--|> UiPart
+BrowserPanel -d--|> UiPart
+PersonListPanel -d--|> UiPart
+PersonCard -d-|> UiPart
+StatusBarFooter -d-|> UiPart
+HelpWindow -d-|> UiPart
+
+PersonCard ..> Model
+UiManager -right-> Logic
+MainWindow -left-> Logic
```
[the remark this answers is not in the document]
pyokagan: My LogicClassDiagram looks much, much different:

LogicClassDiagram

pyokagan:
> @pyokagan Your idea?

Well, keeping our dependencies up-to-date is always a good thing, and I'm confident that @j-lum would be able to handle any trouble and make it work :+1: :+1: :+1: :+1:

docs/diagrams/CommitActivityDiagram.puml
```diff
@@ -0,0 +1,13 @@
+@startuml
+start
+:User executes command;
+'New
```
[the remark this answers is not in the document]
j-lum: That's a workaround to the old diamond notation back. Should have been clearer with comment to indicate that we're using the new notation.

docs/diagrams/UiClassDiagram.puml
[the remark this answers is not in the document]
j-lum: Yeah let me figure out what fonts work.

j-lum:
@pyokagan I would really prefer to not have to muck around to get the plugin working for now. I'll take a look at what I can do to improve upon it in my free time though. 

As for bumping gradle, I'm sure we can postpone it to another PR.
On Duke, I've actually had success with v5.5 running the latest version of Shadow and importing JavaFX stuff through the openjfx plugin which seems to clean up the dependencies block. When Duke is stable we can consider porting the changes over to AB to make things a bit more consistent.

j-lum:
Also the default font happens to be Java's sans-serif which is system dependent. 
Changing the font to Arial doesn't seem to break anything on my side (some boxes got smaller) and it seems to be cross platform so let's so stick with that for now?

docs/DeveloperGuide.adoc
```diff
@@ -199,15 +204,15 @@ If the `currentStatePointer` is at index `addressBookStateList.size() - 1`, poin
 
 Step 5. The user then decides to execute the command `list`. Commands that do not modify the address book, such as `list`, will usually not call `Model#commitAddressBook()`, `Model#undoAddressBook()` or `Model#redoAddressBook()`. Thus, the `addressBookStateList` remains unchanged.
 
-image::UndoRedoNewCommand3StateListDiagram.png[width="800"]
+image::UndoRedoState4.png[].png[width="800"]
```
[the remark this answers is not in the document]
j-lum: I'm actually surprised that it still rendered properly.

docs/diagrams/UiClassDiagram.puml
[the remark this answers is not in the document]
j-lum: Changing it to left instead of down causes all the arrows pointing to UiPart to clump together so I'm just taking the one bad arrow.
image

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,215 @@
...
+== Export PlantUML diagrams
+
+The `PlantUML integration` plugin allows you to export individual diagrams to a location of your choosing.
+
+However, the plugin does not support diagrams with multiple pages.
+If you have diagrams using the `newpage` feature, you will need to download the `plantuml` jar from the http://plantuml.com/download[official website].
+
+NOTE: You will have to `git add` any new diagrams generated!
+
+TIP: `plantuml.jar` will monitor changes and automatically export diagrams that have been modified!
```
pyokagan: Is this true? I can't find any info on the web about this.

docs/diagrams/style.puml
```diff
@@ -0,0 +1,75 @@
...
+skinparam BackgroundColor #FFFFFFF
+
+skinparam Shadowing false
+
+skinparam Class {
+    FontColor #FFFFFF
+    BorderThickness 1
+    BorderColor #FFFFFF
+    StereotypeFontColor #FFFFFF
+    FontName Arial
```
pyokagan: If we are changing the font, then all of the diagrams will need to be re-generated as well.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,215 @@
...
+== Set up PlantUML
+
+=== Installing Graphviz
+
+Graphviz is a dependency that PlantUML requires to generate more advanced diagrams.
+Head over to the https://www.graphviz.org/download/[downloads page] on the offcial Graphviz website and follow instructions to install Graphviz.
+
+=== Installing the `PlantUML integration` plugin for IntelliJ IDEA
+
+While this is not compulsory, the `PlantUML integration` provides a live preview of the PlantUML file you are working on.
```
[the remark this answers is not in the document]
j-lum: There's the plantuml.jar method but we should just remove this and get everyone on the plugin.

docs/UsingPlantUml.adoc
j-lum: Yeah the jar does that which is quite nice. The downside is that it always exports everything and it only exports to the directory that the puml files reside in.

docs/diagrams/style.puml
j-lum: Let me just remove the last bits of hides and regenerate everything.

pyokagan:
Okay, I managed to regenerate mostly the same diagram outputs on my Windows system, so that's good enough for me I guess, since it's not much different from Powerpoint where a specific piece of software needs to be used to generate the diagrams.

docs/UsingPlantUml.adoc
```diff
@@ -0,0 +1,215 @@
...
+
+PlantUML is a popular tool in this project to create UML diagrams.
+For more information about the basics of PlantUML, head over to http://plantuml.com/[its official website].
+
+== Set up PlantUML
+
+=== Installing Graphviz
+
+Graphviz is a dependency that PlantUML requires to generate more advanced diagrams.
+Head over to the https://www.graphviz.org/download/[downloads page] on the offcial Graphviz website and follow instructions to install Graphviz.
```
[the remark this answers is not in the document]
j-lum: Fixed the typos and the weird sentence.

docs/UsingPlantUml.adoc
j-lum: Also the reason why I didn't add the puml files this commit was because it's kind of icky to have the pptx files sit right with puml ones. 

Also the part on editing documentation comes in the next commit when we update the DG so I think it's ok to have them in the same commit?

j-lum:
Actually now that I think of it. The reason why I put the down plantuml as popular was because it was one of the things we considered when looking for alternative options. I'll go rewrite the commit to reflect that more clearly.

j-lum:
Whoops forgot to export the plantuml directory.

docs/UsingPlantUml.adoc
pyokagan: Well just put everything into one commit then.

pyokagan: > I didn't add the puml files this commit was because it's kind of icky to have the pptx files sit right with puml ones.

If foo is generated from bar, then foo and bar *must* be in the same commit.

j-lum: Hmm never considered making it one large commit but that works. Let me push again real quick.

docs/UsingPlantUml.adoc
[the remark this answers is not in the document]
pyokagan: > There's the plantuml.jar method but we should just remove this and get everyone on the plugin.

@j-lum Was this missed out?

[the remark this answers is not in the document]
pyokagan: Anyway, I've re-read through `UsingPlantUml.adoc`. One problem is that it does not make it clear that there are two ways of using PlantUML: via downloading plantuml.jar and using it directly or by using the PlantUML integration plugin. On one hand, it says that the PlantUML integration is "not compulsory", yet on the other hand most of the document (e.g. the "Create/Edit PlantUML diagrams" section and the "Export PlantUML diagrams" section) only explain how to perform the operations using the PlantUML integration plugin.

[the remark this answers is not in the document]
pyokagan: The document needs to make it clear what method (using plantuml.jar XOR using the PlantUML integration plugin) students are expected to use.

j-lum:
@pyokagan I think we should endorse one clear path to do things to minimise confusion. I'm leaning towards just the plugin. Students lose some QOL improvements like auto publishing and multipage rendering but it shouldn't be too bad. Thoughts before I edit the document?

pyokagan:
> @pyokagan I think we should endorse one clear path to do things to minimise confusion. I'm leaning towards just the plugin.

> Students lose some QOL improvements like auto publishing

I guess you meant the "watch the filesystem for changes and rebuild automatically when source files change" functionality? Shouldn't be a big loss.

> and multipage rendering but it shouldn't be too bad.

This too, since we are supposed to only generate one image per puml file.

> Thoughts before I edit the document?

Sounds like a good approach.

j-lum:
@pyokagan Surprisingly, not having the daemon watching for changes and rebuilding was actually quite  a huge pain point for me. But until the gradle plugin I was using starts to play nice with !include we'll have to default to using the intellij plugin.

pyokagan:
Alright, I've checked the images and they seem good. Let's try not to touch them anymore :+1:

j-lum:
@pyokagan Glad you're still around, I'm pushing the changes to the adoc. Just changing the optional line to making them install it and removing the tip on the plantuml jar.

pyokagan:
>  I'll merge.

By tonight.

pyokagan:
@j-lum Can you help me merge AB-4 master into AB-3?

You can use my (trial) merge resolution as a reference: (Note the removal/renaming of some classes that don't exist in AB-3):

```diff
diff --cc docs/diagrams/ArchitectureDiagram.puml
index 00000000,ff6d7b52..a80a6624
mode 000000,100644..100644
--- a/docs/diagrams/ArchitectureDiagram.puml
+++ b/docs/diagrams/ArchitectureDiagram.puml
@@@ -1,0 -1,45 +1,41 @@@
+ @startuml
 -!include <office/Concepts/globe_internet>
+ !include <office/Concepts/documents>
+ !include <office/Users/user>
+ !include style.puml
+ 
+ Package " "<<Rectangle>>{
+     Class UI UI_COLOR
+     Class Logic LOGIC_COLOR
+     Class Storage STORAGE_COLOR
+     Class Model MODEL_COLOR
+     Class Main MODEL_COLOR_T1
+     Class Commons LOGIC_COLOR_T2
+     Class "Log Center" as Logs UI_COLOR_T2
+     Class Hidden #FFFFFF
+     Class HiddenUI #FFFFFF
+     Class HiddenModel #FFFFFF
+ 
+ 
+ }
+ Class "<$user>" as User MODEL_COLOR_T2
 -Class "<$globe_internet> \nWeb" as Web LOGIC_COLOR_T1
+ Class "<$documents>" as File UI_COLOR_T1
+ 
+ 
+ HiddenUI -up[hidden]-> UI
+ HiddenModel -left[hidden]-> Model
+ Main -up-> HiddenUI
+ Main -left-> HiddenModel
+ UI -> Logic
+ UI -right-> Model
+ Logic -> Storage
+ Logic -down-> Model
+ 
 -UI .left.> Web
 -
+ Logs -right- Commons
+ Hidden .down.> Commons
+ Hidden .down.> Commons
+ Hidden .down.> Commons
+ 
+ Storage .right.>File
+ User --> UI
+ Main --> Hidden
+ @enduml
diff --cc docs/diagrams/BetterModelClassDiagram.puml
index 00000000,82b410d7..7790472d
mode 000000,100644..100644
--- a/docs/diagrams/BetterModelClassDiagram.puml
+++ b/docs/diagrams/BetterModelClassDiagram.puml
@@@ -1,0 -1,21 +1,21 @@@
+ @startuml
+ !include style.puml
+ skinparam arrowThickness 1.1
+ skinparam arrowColor MODEL_COLOR
+ skinparam classBackgroundColor MODEL_COLOR
+ 
 -VersionedAddressBook *-right-> "1" UniquePersonList
 -VersionedAddressBook *-right-> "1" UniqueTagList
++AddressBook *-right-> "1" UniquePersonList
++AddressBook *-right-> "1" UniqueTagList
+ UniqueTagList -[hidden]down- UniquePersonList
+ UniqueTagList -[hidden]down- UniquePersonList
+ 
+ UniqueTagList *-right-> "*" Tag
+ UniquePersonList o-right-> Person
+ 
+ Person o-up-> "*" Tag
+ 
+ Person *--> Name
+ Person *--> Phone
+ Person *--> Email
+ Person *--> Address
+ @enduml
diff --cc docs/diagrams/LogicClassDiagram.puml
index 00000000,88519178..64103968
mode 000000,100644..100644
--- a/docs/diagrams/LogicClassDiagram.puml
+++ b/docs/diagrams/LogicClassDiagram.puml
@@@ -1,0 -1,64 +1,62 @@@
+ @startuml
+ !include style.puml
+ skinparam arrowThickness 1.1
+ skinparam arrowColor LOGIC_COLOR_T4
+ skinparam classBackgroundColor LOGIC_COLOR
+ 
+ package Logic {
+ 
+ package Parser {
+ Interface Parser <<Interface>>
+ Class AddressBookParser
+ Class XYZCommandParser
+ Class CliSyntax
+ Class ParserUtil
+ Class ArgumentMultimap
+ Class ArgumentTokenizer
+ Class Prefix
+ }
+ 
+ package Command {
+ Class XYZCommand
+ Class CommandResult
+ Class "{abstract}\nCommand" as Command
+ }
+ 
+ Interface Logic <<Interface>>
+ Class LogicManager
 -Class CommandHistory
+ }
+ 
+ package Model{
+ Class HiddenModel #FFFFFF
+ }
+ 
+ Class HiddenOutside #FFFFFF
+ HiddenOutside ..> Logic
+ 
+ LogicManager .up.|> Logic
 -LogicManager -left->"1" CommandHistory
+ LogicManager -->"1" AddressBookParser
+ AddressBookParser .left.> XYZCommandParser: creates >
+ 
+ XYZCommandParser ..> XYZCommand : creates >
+ XYZCommandParser ..|> Parser
+ XYZCommandParser ..> ArgumentMultimap
+ XYZCommandParser ..> ArgumentTokenizer
+ ArgumentTokenizer .left.> ArgumentMultimap
+ XYZCommandParser ..> CliSyntax
+ CliSyntax ..> Prefix
+ XYZCommandParser ..> ParserUtil
+ ParserUtil .down.> Prefix
+ ArgumentTokenizer .down.> Prefix
+ XYZCommand -up-|> Command
+ LogicManager ..> Command : executes >
+ 
+ LogicManager --> Model
+ Command .right.> Model
+ note right of XYZCommand: XYZCommand = AddCommand, \nFindCommand, etc
+ 
+ Logic ..> CommandResult
+ LogicManager .down.> CommandResult
+ Command .left.> CommandResult
+ CommandResult -[hidden]-> Parser
+ @enduml
diff --cc docs/diagrams/ModelClassDiagram.puml
index 00000000,1e1ce927..e85a00d4
mode 000000,100644..100644
--- a/docs/diagrams/ModelClassDiagram.puml
+++ b/docs/diagrams/ModelClassDiagram.puml
@@@ -1,0 -1,58 +1,56 @@@
+ @startuml
+ !include style.puml
+ skinparam arrowThickness 1.1
+ skinparam arrowColor MODEL_COLOR
+ skinparam classBackgroundColor MODEL_COLOR
+ 
+ Package Model <<Rectangle>>{
+ Interface ReadOnlyAddressBook <<Interface>>
+ Interface Model <<Interface>>
+ Interface ObservableList <<Interface>>
+ Class AddressBook
+ Class ReadOnlyAddressBook
+ Class Model
+ Class ModelManager
+ Class UserPrefs
+ Class ReadOnlyUserPrefs
 -Class VersionedAddressBook
+ 
+ Package Person {
+ Class Person
+ Class Address
+ Class Email
+ Class Name
+ Class Phone
+ Class UniquePersonList
+ }
+ 
+ Package Tag {
+ Class Tag
+ }
+ }
+ 
+ Class HiddenOutside #FFFFFF
+ HiddenOutside ..> Model
+ 
+ AddressBook .up.|> ReadOnlyAddressBook
 -VersionedAddressBook .up.|> AddressBook
+ 
+ ModelManager .up.|> Model
+ Model .right.> ObservableList
 -ModelManager o--> "1" VersionedAddressBook
++ModelManager o--> "1" AddressBook
+ ModelManager o-left-> "1" UserPrefs
+ UserPrefs .up.|> ReadOnlyUserPrefs
+ 
 -VersionedAddressBook *--> "1" UniquePersonList
++AddressBook *--> "1" UniquePersonList
+ UniquePersonList o--> "*" Person
+ Person *--> Name
+ Person *--> Phone
+ Person *--> Email
+ Person *--> Address
+ Person *--> "*" Tag
+ 
+ Name -[hidden]right-> Phone
+ Phone -[hidden]right-> Address
+ Address -[hidden]right-> Email
+ 
+ ModelManager -->"1" Person : filtered list
+ @enduml
diff --cc docs/diagrams/UiClassDiagram.puml
index 00000000,a727b2d7..92746f9f
mode 000000,100644..100644
--- a/docs/diagrams/UiClassDiagram.puml
+++ b/docs/diagrams/UiClassDiagram.puml
@@@ -1,0 -1,64 +1,60 @@@
+ @startuml
+ !include style.puml
+ skinparam arrowThickness 1.1
+ skinparam arrowColor UI_COLOR_T4
+ skinparam classBackgroundColor UI_COLOR
+ 
+ package UI <<Rectangle>>{
+ Interface Ui <<Interface>>
+ Class "{abstract}\nUiPart" as UiPart
+ Class UiManager
+ Class MainWindow
+ Class HelpWindow
+ Class ResultDisplay
 -Class BrowserPanel
+ Class PersonListPanel
+ Class PersonCard
+ Class StatusBarFooter
+ Class CommandBox
+ }
+ 
+ package Model <<Rectangle>> {
+ Class HiddenModel #FFFFFF
+ }
+ 
+ package Logic <<Rectangle>> {
+ Class HiddenLogic #FFFFFF
+ }
+ 
+ Class HiddenOutside #FFFFFF
+ HiddenOutside ..> Ui
+ 
+ UiManager .left.|> Ui
+ UiManager -down-> MainWindow
+ MainWindow --> HelpWindow
+ MainWindow *-down-> CommandBox
+ MainWindow *-down-> ResultDisplay
 -MainWindow *-down-> BrowserPanel
+ MainWindow *-down-> PersonListPanel
+ MainWindow *-down-> StatusBarFooter
+ 
+ PersonListPanel -down-> PersonCard
+ 
+ MainWindow -left-|> UiPart
+ 
+ ResultDisplay --|> UiPart
+ CommandBox --|> UiPart
 -BrowserPanel --|> UiPart
+ PersonListPanel --|> UiPart
+ PersonCard --|> UiPart
+ StatusBarFooter --|> UiPart
+ HelpWindow -down-|> UiPart
+ 
+ PersonCard ..> Model
+ UiManager -right-> Logic
+ MainWindow -left-> Logic
+ 
+ PersonListPanel -[hidden]left- HelpWindow
+ HelpWindow -[hidden]left- CommandBox
+ CommandBox -[hidden]left- ResultDisplay
 -ResultDisplay -[hidden]left- BrowserPanel
 -BrowserPanel -[hidden]left- StatusBarFooter
++ResultDisplay -[hidden]left- StatusBarFooter
+ 
+ MainWindow -[hidden]-|> UiPart
+ @enduml
diff --cc docs/images/ArchitectureDiagram.png
index 00000000,3ef3bf9f..4f7316a2
mode 000000,100644..100644
Binary files differ
diff --cc docs/images/BetterModelClassDiagram.png
index 00000000,f3099204..3aece69a
mode 000000,100644..100644
Binary files differ
diff --cc docs/images/LogicClassDiagram.png
index dce88c11,15f7fcd7..e8843a3c
Binary files differ
diff --cc docs/images/ModelClassDiagram.png
index 3a784055,bf2930ea..3ffded6d
Binary files differ
```

Issue #728, opened 2017-11-05
FileUtil: isFileExists(File) may be unnecessary


PierceAndy:
The current method `FileUtil#isFileExists(File)` is basically a combination of two boolean method calls:
```
public static boolean isFileExists(File file) {
        return file.exists() && file.isFile();
}
```

However, as per the [official docs for java.io.File#isFile()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#isFile), this method returns "true if and only if the file denoted by this abstract pathname exists and is a normal file; false otherwise".

Since ` java.io.File#isFile()` already checks for existence, `FileUtil#isFileExists(File)` appears unnecessary.

fzdy1914:
This issue can be closed, as `isFileExists` is now edited to be:
```
    public static boolean isFileExists(Path file) {
        return Files.exists(file) && Files.isRegularFile(file);
    }
```

pyokagan:
@fzdy1914 
> This issue can be closed, as `isFileExists` is now edited to be:
> 
> ```
>     public static boolean isFileExists(Path file) {
>         return Files.exists(file) && Files.isRegularFile(file);
>     }
> ```

Probably updated, but not closed though since conceptually the same issue exists, just with a different API. `Files#isRegularFile(Path)` already checks for file existence, so the call to `Files.exists(file)`, and in fact even the whole `isFileExists(Path)` helper method, can be removed.

This comment will serve as the update.

delislej:
i would like to fix this!

PR #1022, opened 2019-10-22
Remove redundant method

delislej:
fix issue #728

hopefully this is what you wanted!


damithc:
@delislej thanks for the PR. Our PR submission process is given in https://oss-generic.github.io/process/docs/QuestionsIssuesPrs.html#submitting-a-pr

PR #496, opened 2017-06-15
Documentation: Add hyperlink to return to top of file #430

Zhiyuan-Amos:
Fixes #430.

PR #1023, opened 2019-12-07
Documentation: Add return to top shortcuts

### LINKED ISSUE #430: Add Return to top shortcuts to our documentations
Currently all the shortcut links in our documentations are "one way trip".
We should make it further user-friendly by adding return to top link at the end of each section.

C-likethis123:
Fixes #430.

I'm pushing this PR up for early feedback.

### DIFF
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -149,6 +149,8 @@ The `Storage` component,
 
 Classes used by multiple components are in the `seedu.addressbook.commons` package.
 
+<<DeveloperGuide#, Return to Top>>
+
 == Implementation
 
 This section describes some noteworthy details on how certain features are implemented.
--- a/docs/DeveloperGuide.adoc
+++ b/docs/DeveloperGuide.adoc
@@ -71,6 +71,7 @@ image::ArchitectureSequenceDiagram.png[]
 
 The sections below give more details of each component.
 
+<<DeveloperGuide#, Return to Top>>
 [[Design-Ui]]
 === UI component
 
@@ -87,6 +88,8 @@ The `UI` component,
 
 * Executes user commands using the `Logic` component.
 * Listens for changes to `Model` data so that the UI can be updated with the modified data.
+
+<<DeveloperGuide#, Return to Top>>
 
 [[Design-Logic]]
 === Logic component
@@ -111,6 +114,8 @@ image::DeleteSequenceDiagram.png[]
 
 NOTE: The lifeline for `DeleteCommandParser` should end at the destroy marker (X) but due to a limitation of PlantUML, the lifeline reaches the end of diagram.
 
+<<DeveloperGuide#, Return to Top>>
+
 [[Design-Model]]
 === Model component
 
@@ -131,6 +136,8 @@ As a more OOP model, we can store a `Tag` list in `Address Book`, which `Person`
  +
 image:BetterModelClassDiagram.png[]
 
+<<DeveloperGuide#, Return to Top>>
+
 [[Design-Storage]]
 === Storage component
 
@@ -143,6 +150,8 @@ The `Storage` component,
 
 * can save `UserPref` objects in json format and read it back.
 * can save the Address Book data in json format and read it back.
+
+<<DeveloperGuide#, Return to Top>>
 
 [[Design-Commons]]
 === Common classes
@@ -216,6 +225,8 @@ The following activity diagram summarizes what happens when a user executes a ne
 
 image::CommitActivityDiagram.png[]
 
+<<DeveloperGuide#, Return to Top>>
+
 ==== Design Considerations
 
 ===== Aspect: How undo & redo executes
@@ -236,6 +247,8 @@ image::CommitActivityDiagram.png[]
 ** Pros: We do not need to maintain a separate list, and just reuse what is already in the codebase.
 ** Cons: Requires dealing with commands that have already been undone: We must remember to skip these commands. Violates Single Responsibility Principle and Separation of Concerns as `HistoryManager` now needs to do two different things.
 // end::undoredo[]
+
+<<DeveloperGuide#, Return to Top>>
 
 // tag::dataencryption[]
 === [Proposed] Data Encryption
@@ -259,6 +272,8 @@ We are using `java.util.logging` package for logging. The `LogsCenter` class is 
 * `INFO` : Information showing the noteworthy actions by the App
 * `FINE` : Details that is not usually noteworthy but may be useful in debugging e.g. print the actual list instead of just its size
 
+<<DeveloperGuide#, Return to Top>>
+
 [[Implementation-Configuration]]
 === Configuration
 
@@ -268,14 +283,19 @@ Certain properties of the application can be controlled (e.g user prefs file loc
 
 Refer to the guide <<Documentation#, here>>.
 
+<<DeveloperGuide#, Return to Top>>
+
 == Testing
 
 Refer to the guide <<Testing#, here>>.
+
+<<DeveloperGuide#, Return to Top>>
 
 == Dev Ops
 
 Refer to the guide <<DevOps#, here>>.
 
+<<DeveloperGuide#, Return to Top>>
 [appendix]
 == Product Scope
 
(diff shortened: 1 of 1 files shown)


Zhiyuan-Amos:
Hello @C-likethis123, thank you for contributing to this repository!

A couple of comments:

1. This PR fixes *part* of #430 as your PR only made changes for `DeveloperGuide.adoc` - we have `UserGuide.adoc` etc as well that requires to be updated. Therefore, the phrasing of your initial message should specify that accordingly. 

The reason being, using the phrase `Fixes #xxx` on the PR message (which you can observe if you hover over `Fixes`) causes the corresponding issue to be closed when this PR is merged. This isn't what we want to happen.

2. There are some sections with missing `return to top` shortcuts, e.g. 3.2 & 3.4. Is this intended? @eugenepeh, can you advise whether we should have the `return to top` shortcuts for each `section` (i.e. 1, 2, 3) or for each `sub-section` (i.e. 3.1, 3.1.1, 3.2)?

3. The code for returning to the top is `<<DeveloperGuide#, Return to Top>>`. Does the functionality break if `#` is removed?

4. Concerning your commit messages, you should have only 1 commit with the following commit message, following the pattern of our [guide](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message)

```
{current situation} -- use present tense

{why it needs to change}

{what is being done about it} -- use imperative mood
```

So, you can actually copy and paste the commit message I've written in my previous [PR](https://github.com/se-edu/addressbook-level4/pull/496/commits/d6fd15adf731680ca77364dfcdd52cb2939e23e5), which I've edited slightly to match what you're doing:

```
Developer Guide: Add hyperlink to return to top of file

The shortcut links in the Table of Contents of Developer Guide is a "one way trip” as there is no quick way to return back to the Table of Contents. {current situation}

The user has to use alternative slower methods such as scrolling to the top of the document. {why it needs to change}

Let’s make the Developer Guide more user-friendly by adding “return to top” link at the end of each section. {what is being done about it}
```

Do wait for @eugenepeh response before continuing. You should only have 1 commit. Since you already have 2 commits, you have to perform interactive rebase on Git to squash these commits. More info [here](https://stackoverflow.com/questions/1186535/how-to-modify-a-specified-commit/1186549#1186549).

eugenepeh:
Thanks for the tag, @Zhiyuan-Amos 

>There are some sections with missing return to top shortcuts, e.g. 3.2 & 3.4. Is this intended? @eugenepeh, can you advise whether we should have the return to top shortcuts for each section (i.e. 1, 2, 3) or for each sub-section (i.e. 3.1, 3.1.1, 3.2)?

The `return to top shortcuts` should only appear when the user have read all the necessary items under a topic.

In that case, minor sub-sections like 3.1.1, 3.1.2 etc should not have a `return to top shortcuts` as they are not independent of each other. Whereas, it is fine to have them individually for 3.1 and 3.2. If a `return to top shortcuts` is deemed necessary for a sub-section, the equivalents should also have the shortcuts. E.g. 3.1 has it, so 3.2 should also gets 1.

C-likethis123:
Thank you for the feedback! 

I was initially intending to make changes to the Developer Guide, and get feedback before making changes to the rest of the guides. In this case, should I change the other guides in this PR or should I make a separate PR?

Another question: should there be return to top shortcuts for very short sections? For example, section 1 is just a one-liner.

damithc:
We also need to consider the effect on generated PDF files as students need to submit PDF versions of DG. Will this add unnecessary clutter to the PDF files? Keep in mind that users can use the <kbd>Home</kbd> key to return to the top anyway. In other words, we should consider if this feature is worth adding after all.

Zhiyuan-Amos:
@eugenepeh any thoughts on the above?

eugenepeh:
>We also need to consider the effect on generated PDF files as students need to submit PDF versions of DG. Will this add unnecessary clutter to the PDF files? Keep in mind that users can use the Home key to return to the top anyway. In other words, we should consider if this feature is worth adding after all.

Yes, it does indeed adds unnecessary clutter as such link won't work when converted to PF. However, not sure if its still needed for the tutors to review the students' docs, I find this can help to save some pain for tutors who does their review using their mobile devices. That alone, I personally think is worth the unnecessary clutter.

damithc:
> Yes, it does indeed adds unnecessary clutter as such link won't work when converted to PF. However, not sure if its still needed for the tutors to review the students' docs, I find this can help to save some pain for tutors who does their review using their mobile devices. I used to do that occasionally when travelling / during lunch. That alone, I personally think is worth the unnecessary clutter.

This is a more general mobile UI issue, applicable to any online documentation. 'Go to top' used to be the solution but it is not optimal as that requires every web page on the Web to add such links all over the page. Recent mobile Browsers have a built-in alternative. For example, on my iPhone (several generations old), I can scroll to top by tapping at the top of the screen (i.e., the bar where battery level etc. are shown). Works for both Browser and PDF viewer. I guess Android has something similar?

damithc:
> I guess Android has something similar?

Looks like it doesn't? :-p  https://www.reddit.com/r/GooglePixel/comments/5gdnox/quick_scroll_to_top/

damithc:
So, I'm neutral on whether to add 'back to top' links. It's fine to add if it can be done without too much clutter. The benefit is mostly the convenience of viewers using non-iOS devices and viewing the HTML version. 
Docs review still happens but mostly at the end of the semester, using the PDF version. Mid-semester reviews are now done by students (i.e., peer review) using the HTML version but not using mobile devices as it is done during tutorials using the TV. Hence, the cases that can benefit from this enhancement has shrunk quite a lot since this issue was created.

eugenepeh:
>Looks like it doesn't? :-p https://www.reddit.com/r/GooglePixel/comments/5gdnox/quick_scroll_to_top/

Yup, for android case is really app specific I suppose. There's no universal way like Apple's. 

>So, I'm neutral on whether to add 'back to top' links. It's fine to add if it can be done without too much clutter. The benefit is mostly the convenience of viewers using non-iOS devices and viewing the HTML version.
Docs review still happens but mostly at the end of the semester, using the PDF version. Mid-semester reviews are now done by students (i.e., peer review) using the HTML version but not using mobile devices as it is done during tutorials using the TV. Hence, the cases that can benefit from this enhancement has shrunk quite a lot since this issue was created.

I see, in that case this issue isn't really necessary anymore. However, since @C-likethis123 have started working on this issue, I suppose we can continue with this feature to place some emphasis on user friendly, and also as a case study on who are your target users and benefits to them (is this feature worth adding in the first place etc). 

If documentations still have a hard requirement on pages, they can make decision on removing unnecessary clutters such as this :-P

damithc:
> I see, in that case this issue isn't really necessary anymore. However, since @C-likethis123 have started working on this issue, I suppose we can continue with this feature to place some emphasis on user friendly, and also as a case study on who are your target users and benefits to them (is this feature worth adding in the first place etc).

No objections from me.

Zhiyuan-Amos:
> I was initially intending to make changes to the Developer Guide, and get feedback before making changes to the rest of the guides. 

👍 

> In this case, should I change the other guides in this PR or should I make a separate PR?

You can change the other guides in this PR as the changes are trivial enough to be put into the same PR. :)

> Another question: should there be return to top shortcuts for very short sections? For example, section 1 is just a one-liner.

@eugenepeh any comments on this? I guess it makes sense for the purpose of consistency. Though I'm also open to the idea of closing this PR, since

> this issue isn't really necessary anymore.

Also, concerning:

> place some emphasis on user friendly, and also as a case study on who are your target users and benefits to them (is this feature worth adding in the first place etc).

We could update Developer Guide or something to direct them to our discussion here on why we've decided not to go ahead with having `return to top` feature.

C-likethis123:
Since this issue isn't necessary anymore, I'm okay with the idea of closing this PR and working on other issues.

damithc:
> Since this issue isn't necessary anymore, I'm okay with the idea of closing this PR and working on other issues.

@C-likethis123 Noted. Closing the PR. In fact, we might be archiving AB-4 (and AB1, AB2 too). You can contribute to AB3, or other [NUS-OSS projects](https://nus-oss.github.io/).

damithc:
@C-likethis123 Thanks again for your work on this PR (and thanks to the reviewers too).

